PR #23, opened 2024-03-21
ddp, fsdp : support passing model as argument

kshitij12345:
Previously failing script:

```python
import os
import thunder
import torch
import torch.distributed as torch_dist

world_size = int(os.environ.get("WORLD_SIZE", 1))
local_rank = int(os.environ.get("LOCAL_RANK", 0))
global_rank = int(os.environ.get("RANK", 0))
if world_size > 1:
    torch_dist.init_process_group(backend="nccl")
    pg = torch_dist.distributed_c10d._get_default_group()
device = torch.device("cuda", local_rank)
torch.cuda.set_device(device)

model = torch.nn.Linear(5, 10, bias=False, device=device)
x = torch.randn(2, 5, device=device)

def fwd_loss(m, x):
    return m(x).sum()

fwd_loss = thunder.jit(fwd_loss)

out = fwd_loss(model, x)

print(out)
if local_rank == 0:
    print("FN", thunder.last_traces(fwd_loss)[-1].python())
```


kshitij12345:
cc: @crcrpar @IvanYashchuk

IvanYashchuk:
What's the expected behavior if multiple models were passed as arguments?

kshitij12345:
Right now, `ddp` and `fsdp` both internally use the default process group . So even if we have multiple models they will all use the same process group for synchronization.

In future, if we add support for specifying the process group in `ddp` and `fsdp` then we may have to refactor the code as the GeneralJitContext also assumes there will be a single process group.

https://github.com/Lightning-AI/lightning-thunder/blob/a27bb65b1b9525c5b51d2b096ad122c10f5a2d6c/thunder/core/jit_ext.py#L520-L524
https://github.com/Lightning-AI/lightning-thunder/blob/a27bb65b1b9525c5b51d2b096ad122c10f5a2d6c/thunder/core/jit_ext.py#L1372-L1376

thunder/core/jit_ext.py
```diff
@@ -1347,6 +1347,18 @@ def bind_inputs(name, trace, input_vars, input_proxies):
...
+def _get_process_group_from(*fn_and_args):
+    # `ddp` and `fsdp` transforms add attribute `procses_group_for_ddp`
+    # on the Module that they wrap. This module could be passed to `thunder.jit`
+    # as the function to be jitted or as an argument of the function to be jitted.
+    # NOTE: `ddp` and `fsdp` use default process group. So finding the first
+    #       occurence is sufficient.
+    for fn_or_arg in fn_and_args:
+        pg = getattr(fn_or_arg, "process_group_for_ddp", None)
+        if pg is not None:
+            return pg
```
crcrpar, line 1359: nit

```suggestion
    for fn_or_arg in fn_and_args:
        if (pg := getattr(fn_or_arg, "process_group_for_ddp", None)) is not None:
            return pg
```

thunder/core/jit_ext.py
```diff
@@ -1347,6 +1347,17 @@ def bind_inputs(name, trace, input_vars, input_proxies):
...
 
+def _get_process_group_from(*fn_and_args):
+    # `ddp` and `fsdp` transforms add attribute `procses_group_for_ddp`
+    # on the Module that they wrap. This module could be passed to `thunder.jit`
+    # as the function to be jitted or as an argument of the function to be jitted.
+    # NOTE: `ddp` and `fsdp` use default process group. So finding the first
+    #       occurence is sufficient.
+    for fn_or_arg in fn_and_args:
+        if (pg := getattr(fn_or_arg, "process_group_for_ddp", None)) is not None:
+            return pg
```
crcrpar, line 1358: shouldn't we return `None` in case none of the args have the attribute?

crcrpar: uh rather for clarity?

crcrpar (approved):

thunder/core/jit_ext.py
```diff
@@ -1347,6 +1347,19 @@ def bind_inputs(name, trace, input_vars, input_proxies):
...
 
 
+def _get_process_group_from(*fn_and_args) -> Optional["ProcessGroup"]:
+    # `ddp` and `fsdp` transforms add attribute `procses_group_for_ddp`
+    # on the Module that they wrap. This module could be passed to `thunder.jit`
+    # as the function to be jitted or as an argument of the function to be jitted.
+    # NOTE: `ddp` and `fsdp` use default process group. So finding the first
+    #       occurence is sufficient.
+    for fn_or_arg in fn_and_args:
+        if (pg := getattr(fn_or_arg, "process_group_for_ddp", None)) is not None:
```
t-vi, line 1357: Nit: We're not starved for vertical space, so as a rule of thumb I would not use := if the number of non-whitespace characters get smaller from it.

t-vi: More broadly, I think we would have to check if it's all the same process group rather than assuming it.

Are we making fsdp/ddp "spill" over from one model to more here? Is that working as intended?

t-vi:
I must admit that it feels a bit hacky along the line of @IvanYashchuk comment, but we would also need to be very diligent about caching and "spilling" fsdp.
Maybe the intended use sheds some light on what's going on with that?

thunder/core/jit_ext.py
t-vi: Let's not do this. We're not short on space and it has more non-whitespace characters.

thunder/core/jit_ext.py
```diff
@@ -1347,6 +1348,26 @@ def bind_inputs(name, trace, input_vars, input_proxies):
...
+        pg = getattr(fn_or_arg, "process_group_for_ddp", None)
+        if pg is not None:
+            pgs.append(pg)
+
+    # attribute wasn't found, return None.
+    if pg == []:
+        return None
+
+    # check all pgs are equal
+    if not reduce(operator.eq, pgs):
```
t-vi, line 1366: Does this work though? operator.eq is mapping to bool.

kshitij12345: This doesn't work, thanks for catching this and help offline!

t-vi (approved):
Let's have this. Thank you @kshitij12345 @crcrpar @IvanYashchuk

PR #24, opened 2024-03-21
transformer_engine: wrap checker_fn in langctx and cleanup (PR2473)

kshitij12345:
This allows `checker_fn` to call TensorProxy methods.

Following snippet should work now:

```python
import thunder
from thunder.extend import OperatorExecutor, register_executor
import torch

my_exec = OperatorExecutor("my_exec")
register_executor(my_exec)

def _my_checker(x, y):
    # Helper function as input shape can be (*, Hin)
    def _view_input_as_2d(x):
        shape = x.shape
        return x.view((-1, shape[-1]))

    _view_input_as_2d(x)
    return True

my_exec.register_implementation(
    thunder.torch.mul,
    checker=_my_checker,
    execution_transform=lambda x, y: x * y,
)

def foo(x, y):
    return x * y

cfunc = thunder.jit(foo, executors=[my_exec])

# Works when requires_grad=False!
# x = torch.randn(16, 16, requires_grad=False)
x = torch.randn(16, 16, requires_grad=True)
cfunc(x, x)

```


thunder/executors/transformer_engineex.py
```diff
@@ -398,15 +403,6 @@ def linear_forwad_rule(a, w, bias):
     return primal, saved_for_backward
 
 
-def linear_forward_rule_checker(a: TensorProxy, w: TensorProxy, bias: None | TensorProxy) -> bool:
```
kshitij12345, line 401: unused.

carmocca (approved):
This is useful. I remember having to do

```python
    # cannot use `.view` so reduce the dimension manually
    # n_rows = X.view(-1, X.shape[-1]).shape[0]
    n_rows = reduce(int.__mul__, X.shape[:-1])
```

for the Unsloth RMSNorm kernel

PR #28, opened 2024-03-21
Minor Readme cosmetics

rasbt:
I was just reading through the README and noticed a few points and thought making a quick PR doesn't hurt :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ Thunder aims to be usable, understandable, and extensible.
 
 ## Performance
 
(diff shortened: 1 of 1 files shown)


t-vi (approved):
Thank you @rasbt

README.md
```diff
@@ -119,11 +119,11 @@ print(result)
 
 The compiled function `jfoo` takes and returns PyTorch tensors, just like the original function, so modules and functions compiled by Thunder can be used as part of larger PyTorch programs.
 
-## Train models
+## Model training with Thunder
```
lantiga, line 122: Let's keep `Train models` as the title

lantiga (approved):
Just one change

README.md
rasbt: I don't know where the "Train models" came from. In my copy it said "Running training" hence the change. But I like "Train models" even better than "Model training with models"

Borda (approved):

PR #29, opened 2024-03-21
triton_ce executor: don't error if triton is not installed (PR2466)

kshitij12345:
Similar to `nvfuser_ex`, we default `triton_ex` to be None and we register a `triton_ex` only if `triton` is available.


carmocca:
This is great. You should be able to removehttps://github.com/Lightning-AI/lightning-thunder/blob/f37ac34399c8f039637b527a79064d106abb4bc2/thunder/extend/__init__.py#L314 and update https://github.com/Lightning-AI/lightning-thunder/blob/f37ac34399c8f039637b527a79064d106abb4bc2/thunder/tests/test_extend.py#L130

t-vi (approved):
As Carlos says, this is great. Thank you @kshitij12345 , @carmocca .

PR #30, opened 2024-03-21
Remove deregister_augmented_forward_and_backward

IvanYashchuk:
OperatorExecutor.register_implementation(grad_transform=...) is now the preferred and only way to define backward autodiff rules with checker functions.

Check out the "So what is with the backward?" section of the following tutorial: https://github.com/Lightning-AI/lightning-thunder/blob/main/notebooks/adding_custom_operator_backward.ipynb

### DIFF
--- a/thunder/core/transforms.py
+++ b/thunder/core/transforms.py
@@ -2510,20 +2510,6 @@ def register_augmented_forward(op):
     return decorator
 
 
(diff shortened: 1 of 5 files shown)


t-vi (approved):
Thank you @IvanYashchuk

PR #37, opened 2024-03-21
Update LitGPT pin

carmocca:
## What does this PR do?

The main change is the removal of the underscore


Borda (approved):

carmocca:
@Borda Seems like the notebook jobs do not install the test dependencies (litgpt is missing)

Borda (approved):

Borda:
> @Borda Seems like the notebook jobs do not install the test dependencies (litgpt is missing)

it was because tests were installed in the base image but this need to be also explicitly listed in notebooks dependencies

PR #38, opened 2024-03-21
`Generator` as `skip_data_parallel_grad_sync`'s return type annotation (PR2475)

crcrpar:
ref:
- https://docs.python.org/3/library/typing.html#typing.Generator
- https://docs.python.org/3/library/collections.abc.html#collections.abc.Generator


carmocca (approved):
Thank you!

PR #40, opened 2024-03-21
source locations for initial traces

t-vi:
This implements source locations for initial traces.

Propagation will be separate, as discussed with @IvanYashchuk

### DIFF
--- a/thunder/core/interpreter.py
+++ b/thunder/core/interpreter.py
@@ -44,6 +44,7 @@ from types import (
 )
 
 from thunder.core.baseutils import Singleton, init_colors, extract_callable_name
+from thunder.core.codeutils import Positions
 
 
 #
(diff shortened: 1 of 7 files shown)


mruberry:
triage review — we should probably do a design review for this

thunder/core/interpreter.py
```diff
@@ -605,6 +606,13 @@ def push_frame_stack(self, frame: InterpreterFrame):
             pf = self._pop_frame_stack()
             assert pf is frame, "Frame stack inconsistency"
 
+    def get_current_user_source_location(self) -> tuple[str, Positions]:
```
IvanYashchuk, line 609: If I use `thunder.torch.softmax` why is the source location recorded as
```py
thunder/torch/__init__.py:4411: 	    return _softmax(a, dim=dim, dtype=dtype)
```
and with `torch.softmax` it points to the function being traced:
```py
<ipython-input-14-72fa81ca31be>:2: 	    return torch.softmax(a, 0)
```

IvanYashchuk (approved):
We should get this in and refine the information being stored later if needed. It doesn't affect any other systems in Thunder.

To me it wasn't obvious how to put the source locations into the initial trace in the first place, now I'll make the necessary changes to preserve this information.

Thank you!

lantiga (approved):
Great start, let’s get this in!

thunder/core/interpreter.py
t-vi: This is because the "symbol boundary" for softmax is `_softmax` apparently (the symbols are opaque functions to the JIT components, so they are the topmost frame). Note that softmax is special in how we define it in `thunder.torch` (to cater for the undocumented parameter). 
I see three options (for a follow-up PR):
- refuse to record locations in thunder.torch (deferring to the previous (next lower) stack frame if we are),
- provide a mechanism to fudge this (mark things "internal", might also be worthwile for "impl" functions if we see them),
- do nothing.

t-vi: Filed https://github.com/Lightning-AI/lightning-thunder/issues/475

PR #41, opened 2024-03-22
Add cat upcasting support for PyTorch compat

t-vi:


t-vi:
I don't think this will happen soon, so closing it. If anyone else wants to figure out what's going on in torch.compile , that's cool.

PR #42, opened 2024-03-22
only break long lines in unpacking

t-vi:


thunder/core/prims.py
```diff
@@ -812,12 +820,21 @@ def _unpack_tuple_printer(
     if len(bsym.output) == 0:
         return f"# {call_str} (empty tuple)"
 
+    parts = [f"{codeutils.prettyprint(out, literals_as_underscores=True)}, " for out in out_printables]
```
carmocca, line 823: Should you put the added code into a function that is called from both places?

thunder/core/prims.py
```diff
@@ -754,12 +754,20 @@ def unpack_sequence_printer(
     if len(bsym.output) == 0:
```
carmocca, line 754: Maybe you can remove the TODO in L733 now:
> TODO Review using multi-line unpacks more cleverly

thunder/core/prims.py
```diff
@@ -754,12 +754,20 @@ def unpack_sequence_printer(
...
+    parts = [f"{codeutils.prettyprint(out, literals_as_underscores=True)}, " for out in out_printables]
+    parts.append(f"= {call_str}")
 
-    lines.append(f"= {call_str}")
+    lines = []
+    line_parts = []
+    pos = 0
+    for p in parts:
+        if pos and pos + len(p) > 80:
+            lines.append("".join(line_parts) + " \\")
```
carmocca, line 765: It seems like you have an extra space

```suggestion
            lines.append("".join(line_parts) + "\\")
```

carmocca (commented):
I don't think you updated the backward trace. Here's what I used to check:

```python
from litgpt import GPT
import thunder
import torch

# Use only two layers to keep the traces shorter for the demonstration
model = GPT.from_name("Llama-2-7b-hf", n_layer=2).cuda()
model = thunder.jit(model)
x = torch.randint(model.max_seq_length, (2, 5), device="cuda")
y = model(x)  # forward, this may take a bit

forward_trace = thunder.last_traces(model)[-1]
print(forward_trace)
trace = thunder.last_backward_traces(model)[-1]
print(trace)
```

thunder/core/prims.py
t-vi: It was on the edge of what I'd put in it's own functions, but yeah.

t-vi:
Thank you @carmocca

carmocca (approved):
Amazing!

lantiga (approved):
Stamped!

carmocca:
I didn't rerun my snippet after your last commits :sweat_smile:. Opened #69

PR #45, opened 2024-03-22
Implement `no_sync` for `thunder.distributed.fsdp` (PR2457)

crcrpar:
## tldr

Enables `no_sync` for `thunder.jit(thunder.distributed.fsdp(model))`. The accompanied changes are:
- new argument of `return_none_instead_of_grads` of `ThunderFunction.forward`
  - This could be eliminated once a `TraceCtx`'s bound symbols are not deleted even if it just returns one or more `None`s
- removal of `no_sync` check before applying `dist_prims.synchronize` to args and kwargs
  - FSDP's forward needs this prim for its param AllGather
  - [ddp] `visitor_transform` removes `dist_prims.all_reduce`, `dist_prims.wait`, and preaveraging when `no_sync`
  - [fsdp] `visitor_transform` removes comms and puts `dist_prims.stash_grad_for_fsdp` and optional param `AllGather` when `no_sync`
    - The generated trace and its executable python code return unsynchronized unsharded gradients.
    - The prim's implementation accumulates the grads as `param._thunder_fsdp_unsharded_grad`.
    - `ThunderFunction`'s `backward` returns `None`s instead of such grads to avoid shape mismatch between params and unsharded grads.

---

as of https://github.com/Lightning-AI/lightning-thunder/pull/45/commits/fa61c49c480bfc16536f0575a81d76c873513f5d

- llama-2-7b-hf
- world size 8 H100s
- micro batch size 1
- global batch size 32
- gradient accumulation 4
- no bucketing (of AllGather and ReduceScatter)

### zero2

command: `torchrun --nproc-per-node=8 thunder/benchmarks/benchmark_litgpt.py --compile=thunder_inductor --distributed_mode=fsdp --nsys_enabled=False --micro_batch_size=1 --global_batch_size=32 --skip_data_sync <false|true> --model_name=Llama-2-7b-hf --shard_mode=zero2 --bucketing_mode=none --json_path "<filename>.json" --return_metrics_as_json=true`

|                         | w/ `no_sync` | w/o `no_sync` |
|-------------------------|--------------|---------------|
| tokens/sec              | 82713.0      | 80341.0       |
| memory consumption [GB] | 65.6         | 40.3          |

### zero3

command: `torchrun --nproc-per-node=8 thunder/benchmarks/benchmark_litgpt.py --compile=thunder_inductor --distributed_mode=fsdp --nsys_enabled=False --micro_batch_size=1 --global_batch_size=32 --skip_data_sync <false|true> --model_name=Llama-2-7b-hf --shard_mode=zero3 --bucketing_mode=none --json_path "<filename>.json" --return_metrics_as_json=true`

|                         | w/ `no_sync` | w/o `no_sync` |
|-------------------------|--------------|---------------|
| tokens/sec              | 77839.0      | 75511.9       |
| memory consumption [GB] | 52.5         | 27.1          |

### DIFF
--- a/thunder/benchmarks/benchmark_litgpt.py
+++ b/thunder/benchmarks/benchmark_litgpt.py
@@ -216,9 +216,11 @@ class Benchmark_litGPT:
                 from thunder.distributed import fsdp, FSDPType, FSDPBucketingStrategy
 
                 sharding_strategy = {"zero2": FSDPType.ZERO2, "zero3": FSDPType.ZERO3}[self.shard_mode]
-                bucketing_strategy = {"none": FSDPBucketingStrategy.NONE, "block": FSDPBucketingStrategy.BLOCK, "layer": FSDPBucketingStrategy.LAYER,}[
-                    self.bucketing_mode
-                ]
+                bucketing_strategy = {
+                    "none": FSDPBucketingStrategy.NONE,
+                    "block": FSDPBucketingStrategy.BLOCK,
+                    "layer": FSDPBucketingStrategy.LAYER,
+                }[self.bucketing_mode]
                 model = fsdp(
                     model,
                     broadcast_from=None,
--- a/thunder/benchmarks/benchmark_litgpt.py
+++ b/thunder/benchmarks/benchmark_litgpt.py
@@ -216,7 +216,7 @@ class Benchmark_litGPT:
                 from thunder.distributed import fsdp, FSDPType, FSDPBucketingStrategy
 
                 sharding_strategy = {"zero2": FSDPType.ZERO2, "zero3": FSDPType.ZERO3}[self.shard_mode]
-                bucketing_strategy = {"none": FSDPBucketingStrategy.NONE, "block": FSDPBucketingStrategy.BLOCK}[
+                bucketing_strategy = {"none": FSDPBucketingStrategy.NONE, "block": FSDPBucketingStrategy.BLOCK, "layer": FSDPBucketingStrategy.LAYER,}[
                     self.bucketing_mode
                 ]
                 model = fsdp(
--- a/thunder/benchmarks/benchmark_litgpt.py
+++ b/thunder/benchmarks/benchmark_litgpt.py
@@ -1,5 +1,6 @@
 import os
 import time
+from typing import Any
 
 import torch
 import functools
@@ -35,6 +36,22 @@ def configure_optimizers(model, weight_decay, learning_rate, betas, device_type)
     return optimizer
 
 
+def run_fwd_bwd_one_microbatch(
+    model: torch.nn.Module,
+    input_ids: torch.Tensor,
+    targets: torch.Tensor,
+    gradient_accumulation_steps: int,
+    te_ctx: Any,
+) -> torch.Tensor:
+    with te_ctx():
+        logits = model(input_ids)
+    logits = logits.reshape(-1, logits.size(-1))
+    targets = targets.reshape(-1)
+    loss = torch.nn.functional.cross_entropy(logits, targets, ignore_index=-1) / gradient_accumulation_steps
+    loss.backward()
+    return loss
+
+
 class Benchmark_litGPT:
     def __init__(
         self,
@@ -53,6 +70,7 @@ class Benchmark_litGPT:
         n_layers: int | None = None,
         profiler_start: int = 15,
         profiler_stop: int = 15,
+        skip_data_sync: bool = False,
     ):
         seed = 1337
         torch.manual_seed(seed)
@@ -132,11 +150,7 @@ class Benchmark_litGPT:
             assert (
                 self.global_batch_size % self.micro_batch_size * world_size == 0
             ), f"Global Batch Size {self.global_batch_size} should be a multiple Micro Batch Size {self.micro_batch_size} * World Size {world_size}."
-            # TODO: Remove when gradient accumulation is ready for benchmarking.
-            if self.gradient_accumulation_steps > 1:
-                print(
-                    f"[WARNING] Gradient Accumulation is not fully supported yet. Benchmarking results may not be accurate. Gradient Accumulation Steps = {self.gradient_accumulation_steps}"
-                )
+        self.skip_data_sync = skip_data_sync
 
         # Profiling Args
         self.nsys_enabled = nsys_enabled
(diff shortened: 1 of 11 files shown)


thunder/benchmarks/benchmark_litgpt.py
```diff
@@ -280,46 +294,43 @@ def train(self):
 
             te_ctx = nullcontext
 
+        if self.skip_data_sync:
+            data_sync_ctx = self.model.no_sync
+        else:
+            data_sync_ctx = nullcontext
```
crcrpar, line 300: cc @parthmannan

thunder/distributed/__init__.py
```diff
@@ -84,15 +84,45 @@ def _sync_grads(module: torch.nn.Module) -> None:
...
-        return
-    grads = [p.grad for p in params_with_grad]
-    torch._foreach_div_(grads, process_group.size())
-    with tdist.distributed_c10d._coalescing_manager(group=process_group, async_ops=True) as cm:
-        for g in grads:
-            tdist.distributed_c10d.all_reduce(g)
-    cm.wait()
+    process_group: ProcessGroup = thunder.compile_data(module).process_group_for_ddp
+    if getattr(module, "use_ddp", False):
+        params_with_grad = [p for p in module.parameters() if p.grad is not None]
```
crcrpar, line 89:
```suggestion
        params_with_grad = [p for p in module.parameters() if p.grad is not None]
        if not params_with_grad:
            return
```

thunder/executors/torch_autograd.py
```diff
@@ -19,8 +19,11 @@
 
 class ThunderFunction(torch.autograd.Function):
     @staticmethod
-    def forward(ctx, compiled_backward, saved_tensors, saved_other, flat_output, *flat_args):
+    def forward(
+        ctx, return_none_instead_of_grads, compiled_backward, saved_tensors, saved_other, flat_output, *flat_args
+    ):
         # Here we just propagate the tensors through the autograd graph
+        ctx.return_none_instead_of_grads = return_none_instead_of_grads
```
IvanYashchuk, line 26: How about naming this `is_fsdp_no_sync_enabled`? That way it will be easier to understand what this flag means outside of `ThunderFunction` code.

thunder/executors/torch_autograd.py
```diff
@@ -51,7 +54,12 @@ def backward(ctx, *args):
 
         # Inside the compiled backward we must clear the saved_tensors_list
         assert not saved_tensors_list, "saved_tensors_list must be empty after calling compiled_backward"
-        return (None, None, None, None, *grads)
+        # TODO(crcrpar): Remove if-else once `dist_prims.stash_grad_for_fsdp` starts to return `None`
```
IvanYashchuk, line 57: Could you please add a short reminding note here why all Nones are returned?

thunder/executors/torchex.py
```diff
@@ -1883,6 +1887,21 @@ def _update_bucket_view_prim_impl(
         views[index_of_dst_view].copy_(tensor)
         return tensor
 
+    # TODO(crcrpar): Update to return None instead
```
IvanYashchuk, line 1890: Why do you want to return None? If you return grad the dataflow in the trace is preserved.

thunder/executors/torchex.py
```diff
@@ -1883,6 +1887,21 @@ def _update_bucket_view_prim_impl(
...
+        grad: torch.Tensor,
+        param_fqn: str,
+        compile_data: CompileData,
+    ) -> None:
+        grad_name = "_thunder_fsdp_unsharded_grad"
+        param = compile_data.fn.get_parameter(param_fqn)
+        if torch.is_tensor(unsharded_grad := getattr(param, grad_name, None)):
+            unsharded_grad += grad
+        else:
+            setattr(param, grad_name, grad)
```
IvanYashchuk, line 1901: This `_stash_grad_for_fsdp_prim_impl` is great, I think we should do this accumulation in the general case as well. It gives us more control.
In https://github.com/Lightning-AI/lightning-thunder/pull/214 I'm experimenting with creating symbol that would be called in the forward trace and do `ThunderFunction.apply`.

thunder/executors/torch_autograd.py
IvanYashchuk: An alternative solution to returning Nones here would be modifying parameters in place to make them unsharded while inside the no_sync context and shard them again when exiting the context.
What do you think, would this work? We can access parameter tensors now within impl functions as done in `_stash_grad_for_fsdp_prim_impl`.

IvanYashchuk (commented):
Really cool work, I hope it will be useful despite higher memory usage. Here HF writes that for a 47B parameter model two gradient accumulation steps with PyTorch's FSDP already goes OOM: https://huggingface.co/docs/accelerate/concept_guides/gradient_synchronization#nosync-requires-additional-gpu-memory-when-using-fsdp.

Currently most likely due to previously reported memory usage issues in Thunder I can't run the example commands from the PR description, I get OOM.
I think we should fix memory problems first before proceeding with this PR.

thunder/executors/torchex.py
crcrpar: The unsharded gradients returned by the bound symbols and the trace calling this primitive will not be used thus I thought it'd be natural to return None

thunder/executors/torch_autograd.py
crcrpar: > modifying parameters in place to make them unsharded while inside the no_sync context and shard them again when exiting the context

iiuc it would further bloat up the memory footprint thus I'm not inclined to it.

thunder/executors/torchex.py
IvanYashchuk: Is there any safety against DCE removing this computation completely?

thunder/executors/torch_autograd.py
IvanYashchuk: For ZeRO2 I think it would even be less memory-consuming because today we hold both sharded and unsharded parameters after forward.
For ZeRO3 we could do this inplace unsharding only during backward.

IvanYashchuk (approved):
Nice feature! 

Are you aware of any `no_sync` usage with FSDP in the wild? Does DeepSpeed offer this functionality? How is HF's Accelerate or LIghtning's Fabric wrapping PyTorch's no_sync?

thunder/executors/torchex.py
crcrpar: I tried `DONT_DCE` with return of `None` but it didn't work so I left a comment

crcrpar:
@t-vi this ready for merge

crcrpar:
> How is HF's Accelerate or LIghtning's Fabric wrapping PyTorch's no_sync?

- fabric: https://lightning.ai/docs/fabric/stable/advanced/gradient_accumulation.html
- accelerate: 
    - no_wync wrapper: https://huggingface.co/docs/accelerate/en/concept_guides/gradient_synchronization
    - accumulate, abstraction of an optimizer step with gradient accumulation: https://huggingface.co/docs/accelerate/v0.29.3/en/package_reference/accelerator#accelerate.Accelerator.accumulate

t-vi (approved):
Thank you, @crcrpar  @IvanYashchuk

PR #47, opened 2024-03-22
`collections.namedtuple`: add lookaside

nikitaved:
General jit support for namedtuple types that are being created inside jitted callables.

### DIFF
--- a/thunder/core/interpreter.py
+++ b/thunder/core/interpreter.py
@@ -2600,7 +2600,8 @@ def _collections_namedtuple_lookaside(
     *,
     rename: bool = False,
     defaults: None | Iterable[Any] = None,
-    module: None | str = None):
+    module: None | str = None,
+):
     # Type checks {
     assert wrapped_isinstance(typename, str)
     assert wrapped_isinstance(field_names, Iterable)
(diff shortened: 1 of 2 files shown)


thunder/core/interpreter.py
```diff
@@ -2588,6 +2593,70 @@ def impl(self, other):
...
+        defaults = wrap_const(defaults)
+
+    if module is None:
+        module = wrap_const(module)
+    # }
+
+    # Run opaque namedtuple {
+    @interpreter_needs_wrap
+    def create_namedtuple(typename: str, field_names: str, **kwargs):
+        namedtuple_type = collections.namedtuple(typename, field_names, **kwargs)
```
t-vi, line 2628: If module is not in the kwargs it takes the module from the direct caller, i.e. it uses `thunder.core.interpreter`. Could you update providing the top stack frame's module if no module argument is given, please? 
```
In [1]:     from collections import namedtuple
   ...:     import thunder
   ...:     typename = "MyNamedTuple"
   ...:     field_names = ("a", "b", "c")
   ...: 
   ...:     # Test returnign just the type {
   ...:     def f():
   ...:         return namedtuple(typename, field_names)
   ...: 
   ...:     jf = thunder.jit(f)


In [2]: 

In [2]: jf()
Out[2]: thunder.core.interpreter.MyNamedTuple
```

Other than that, great work!

nikitaved: Addressed!

nikitaved:
@t-vi , I have addressed the module issue, thank you for pointing it out! Would you have other comments?

t-vi (approved):
OK, seems good to me now. I need to fix the `tuple` vs. `tuple.__new__` by having a new lookaside, but this is the next step.

PR #48, opened 2024-03-22
te executor: auto enable fp8_autocast ctx manager (PR2190)

kshitij12345:
**NOTE**: We will have to think about how to wrap only the forward computation with autocast context manager for grad transform. TransformerEngine [doc](https://docs.nvidia.com/deeplearning/transformer-engine/user-guide/examples/fp8_primer.html#Handling-backward-pass) mentions that the backward pass shouldn't be run under autocast context manager for multi-GPU settings.

TODO:

* [x] Add test with compile options of passing fp8_recipe.


lantiga:
Would it make sense to be more granular and explicitly identify the sections that need casting and insert enter and exit functions at the right places, instead of apply the context manager on the whole compute trace? I'm worried that this will hamper composability down the road.

kshitij12345:
Currently we have to annotate the complete trace. This is required because if we annotate each operator separately, in DDP setting, it will lead to sync points at all FP8 operators. By annotating the complete trace, we sync only at the last FP8 layer in the backward pass. 

We will bring this up with the TransformerEngine folks and have a better solution in future. Thanks for looking into this.

thunder/executors/transformer_engineex.py
```diff
@@ -332,6 +337,12 @@ def _te_functional_linear_backward_meta(
 
 LINEAR_CALLS_COUNTER = 0
 
+if TE_AVAILABLE:
+    _DEFAULT_RECIPE = recipe.DelayedScaling()
+
+IMPORT_CTX_TE_KEY = "transformer_engine"
+FP8_RECIPE_KEY = "fp8_recipe"
```
IvanYashchuk, line 344: What if Thunder gets its own fp8_recipe? Should we call it "te_fp8_recipe" already now?

IvanYashchuk (approved):
Looks great!

What happens if a jitted function with TE executor is called inside explicit `te.fp8_autocast` context?

thunder/executors/transformer_engineex.py
kshitij12345: Fair point, will update this. Thanks!

kshitij12345:
> What happens if a jitted function with TE executor is called inside explicit te.fp8_autocast context?

The inner most fp8_autocast is respected. So in this case, the one we wrapped the trace with will be respected. User will be able to pass their fp8_recipe via compile options (and we can add other customization options if required). This will also allow us to control the `process_group` for communication in future.

Example to show that the innermost context is respected:
```python
import torch
from transformer_engine.common.recipe import Format, DelayedScaling

fp8_small_amax_history_recipe = DelayedScaling(fp8_format=Format.HYBRID, amax_history_len=16, amax_compute_algo="max")

fp8_large_amax_history_recipe = DelayedScaling(fp8_format=Format.HYBRID, amax_history_len=100, amax_compute_algo="max")

my_linear = te.Linear(768, 768, bias=True)

inp = torch.rand((1024, 768)).cuda()

with te.fp8_autocast(enabled=True, fp8_recipe=fp8_small_amax_history_recipe):
    out_fp8 = my_linear(inp)

print(my_linear.fp8_meta["scaling_fwd"].amax_history.shape)  # 16, 3

my_linear = te.Linear(768, 768, bias=True)

with te.fp8_autocast(enabled=True, fp8_recipe=fp8_small_amax_history_recipe):
    with te.fp8_autocast(enabled=True, fp8_recipe=fp8_large_amax_history_recipe):
        out_fp8 = my_linear(inp)

print(my_linear.fp8_meta["scaling_fwd"].amax_history.shape)  # 100, 3
```

kshitij12345:
@t-vi please have a look and stamp this PR :)

t-vi (approved):
Look forward to the bright new future with more detailed building blocks. But for now LGTM.

Thank you @kshitij12345 @IvanYashchuk @lantiga

PR #50, opened 2024-03-22
Support nvfuser fd.add_output(output, alias_input)

kiya00:
Add a prim operator to support nvfuser `fd.add_output(output, alias_input)`, which allows in-place updating of the input tensor. Related context is we want to use it to update the running stats in batch norm.

This prim operator(currently named `prims.input_as_output`, happy to change it if you have a better name) should only be "used internally" when we are decomposing other operators. The torch implementation of this prim op is `copy_`, and nvfuser one is `fd.add_output(output, alias_input)`

TODOs:
use the inplace copy operator to implement the decomposed batch norm forward #139 

cc: @IvanYashchuk


thunder/tests/test_input_as_output_ops.py
```diff
@@ -0,0 +1,56 @@
...
+
+@instantiate(dtypes=(thunder.core.dtypes.inexact,))
+def test_input_as_output_prim(executor, device, dtype):
+    # Tests passing all arguments as function inputs
+    def torch_foo(x, y):
+        z = x * y
+        z = z * z
+        x.copy_(z)
+        # y = y*y
+        return y
```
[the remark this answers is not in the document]
kiya00: you mean the dce? for a internal decompose like this [func](https://github.com/Lightning-AI/lightning-thunder/blob/1857d1b74e3c98e9c0cf10c6685b0cbd22ce0ee3/thunder/tests/test_input_as_output_ops.py#L23) torch trace is 
```
@torch.no_grad()
@no_autocast()
def computation(x, y):
  # x: "cuda:0 f32[4, 4]"
  # y: "cuda:0 f32[4, 4]"
  z = torch.mul(x, y)  # z: "cuda:0 f32[4, 4]"
    # z = ltorch.mul(x, y)  # z: "cuda:0 f32[4, 4]"
      # z = prims.mul(x, y)  # z: "cuda:0 f32[4, 4]"
  t1 = torch.mul(z, z)  # t1: "cuda:0 f32[4, 4]"
    # t1 = ltorch.mul(z, z)  # t1: "cuda:0 f32[4, 4]"
      # t1 = prims.mul(z, z)  # t1: "cuda:0 f32[4, 4]"
  del z
  copy_(t1, x)
  del t1, x
  return y
```
but nvfuser currently fails because the return statement somehow gets before the other math, I'm still debugging it. When fixed I think we could expect a trace with `fd.add_output(t1, x)` replacing the `copy_(t1, x)`

thunder/core/prims.py
```diff
@@ -3492,3 +3493,20 @@ def embedding_backward_meta(grad, indices, num_weights, padding_idx, scale_grad_
 
 
 embedding_backward = make_prim(PrimIDs.EMBEDDING_BACKWARD, "embedding_backward", meta=embedding_backward_meta)
+
+
+def input_as_output_meta(
```
mruberry, line 3498: I'm not totally sure I understand the semantics of this primitive. 

Would it be easier to add a primitive, like `add_`, whose semantics were that it performed the addition inplace on the first operand?

kiya00: I think the motivation is for operator like batch_norm that has inplace update of the inputs(running mean/var), we want an internal prim op to mark it and let nvfuser update this input tensor in-place

kiya00:
The current test fails for nvfuser because it generates a trace with return statement appears before other lines:
```
@torch.no_grad()
@no_autocast()
def augmented_forward_fn(*args):
  # args: "Collection"
  t0, \
  t1, \
  = args
  return {'output': t1, 'flat_args': [t0, t1], 'flat_output': (t1,)}, ((), ())
  nvFusion0(t0, t1)
    # t2 = prims.mul(t0, t1)  # t2: "cuda:0 f32[2, 2]"
    # prims.input_as_output(t2, t0)
```
The reason is the [merge node function](https://github.com/Lightning-AI/lightning-thunder/blob/1857d1b74e3c98e9c0cf10c6685b0cbd22ce0ee3/thunder/executors/data_dependent_partition.py#L239), details see [here](https://gitlab-master.nvidia.com/-/snippets/6314)

t-vi:
I wonder if we should try to have a more thorough approach here. I'm not very creative, so I'd probably model some sort of "tensors sharing storage" and "versioning" approach and possibly make the tensors having the same storage outputs of the update op (or somesuch).

tfogal:
> I wonder if we should try to have a more thorough approach here. I'm not very creative, so I'd probably model some sort of "tensors sharing storage" and "versioning" approach and possibly make the tensors having the same storage outputs of the update op (or somesuch).

+1. What you're describing, TomV, is SSA with a deSSA pass that happens during/via allocation. Another implementation option for deSSA is to insert a copy along every update path, and then try to kill the copies in the allocator. The advantage of the latter is a simpler compiler (two dumber passes); the disadvantage is a slower compiler.

Anyway I'd encourage this approach. With just one source of potential copies (i.e. bnorm, here), it's pretty simple to deal with, though.

thunder/core/transforms.py
```diff
@@ -3121,7 +3123,9 @@ def deconstruct_forward_env_for_backward(trace, env):
     # arguments. See test_grad.py:test_torch_autograd_function for an example
     # where this is tested.
     bound_symbols = iter_bound_symbols(trace.bound_symbols)
-    saved_for_backward = tuple(env[sequencify(symbol.output)[0].name].residuals for symbol in bound_symbols)
+    saved_for_backward = tuple(
+        env[sequencify(symbol.output)[0].name].residuals for symbol in bound_symbols if symbol.output is not None
+    )
```
IvanYashchuk, line 3128: Could you please revert this change and the one in `reconstruct_forward_env_for_backward`?
Instead, we can modify the common function `iter_bound_symbols` to skip symbols with None output.

thunder/core/transforms.py
```diff
@@ -3466,7 +3472,8 @@ def put_grad(v: Variable, val: Any) -> None:
         # Having a single cotangent is a common case, so we flatten it
         # Otherwise, we will need to rewrite the pullback functions
         cotangents = tree_flatten(cotangents)[0]
-        residuals = forward_env[symbol_output[0].name].residuals
+        residuals = [] if symbol_output[0] is None else forward_env[symbol_output[0].name].residuals
+        # residuals = forward_env[symbol_output[0].name].residuals
```
IvanYashchuk, line 3476: Revert this change, please. Equivalent options are:
* Use `if symbol.output is None: continue` in the beginning of the loop
* Use `for symbol in reversed(list(iter_bound_symbols(trace.bound_symbols))):`

kiya00:
I'm afraid this is not a general solution to support in-place for Thunder, sorry for the misleading name of the PR. It's basically adding a prim operator to support the functionality provided by nvfuser `fd.add_output(output, alias_input)`, so that the in-place update of running stats in BN can apply  @t-vi @tfogal

thunder/executors/nvfuserex_impl.py
```diff
@@ -511,9 +511,6 @@ def get_fd(input_descriptors) -> FusionDefinition:
         return create_fd(bsyms, input_descriptors, sorted_unique_inputs, sorted_unique_outputs)
 
     fdw = FusionDefinitionWrapper(get_fd, name, get_fd.cache_info, get_fd.cache_clear)
-    # Avoid hitting nvFuser error when there is no output
-    if not sorted_unique_outputs:
-        return lambda *args: tuple()
```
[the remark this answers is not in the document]
IvanYashchuk: Is this needed to support a test case when the FusionDefinition consists of in place copy with no out-of-place outputs?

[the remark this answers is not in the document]
kiya00: yes, when the only 'output' of nvfusion is the in-place update of one input, the `sorted_unique_outputs` is empty and return a "empty" function

thunder/tests/test_copy_.py
IvanYashchuk, line 1: I suggest renaming this file to "test_inplace_copy.py", then it's more clear what it does. A trailing underscore in the filename might seem like a typo.

thunder/core/rematerialization.py
IvanYashchuk, line 1: Note to myself:
After undoing changes in this file the following test fails: `test_prim_copy__bwd_nvfuser_cuda_float16`.

thunder/core/rematerialization.py
```diff
@@ -87,7 +87,9 @@ def find_external_consumer_inputs(
         Tuple[ProxyInterface, ...]: Consumer's inputs that must be included in
         the input of the consumer.
     """
-    all_produced_vars = tuple(chain.from_iterable((y for y in x.flat_outs) for x in producer.subsymbols))
+    all_produced_vars = tuple(
+        chain.from_iterable((y for y in x.flat_outs if y is not None) for x in producer.subsymbols)
```
IvanYashchuk, line 91: Here's a better alternative because further code assumes we're iterating over proxies accessing its name.
```suggestion
        chain.from_iterable((y for y in x.flat_proxy_outs) for x in producer.subsymbols)
```

thunder/core/rematerialization.py
```diff
@@ -343,6 +345,8 @@ def add_edges(var):
             if user.sym.id in sym_skip_list:
                 continue
             for out in user.flat_outs:
+                if out is None:
+                    continue
```
IvanYashchuk, line 349:
```suggestion
            for out in user.flat_proxy_outs:
```

thunder/core/rematerialization.py
```diff
@@ -352,11 +356,14 @@ def add_edges(var):
...
-        add_edge("source", var.name + "_in", capacity=float("inf"))
-        add_edges(var)
+        if var is not None:
+            add_edge("source", var.name + "_in", capacity=float("inf"))
+            add_edges(var)
 
     for symbol in chain(producer.subsymbols, consumer.subsymbols):
         for var in symbol.flat_outs:
+            if var is None:
+                continue
```
IvanYashchuk, line 366:
```suggestion
        for var in symbol.flat_proxy_outs:
```

thunder/core/rematerialization.py
```diff
@@ -352,11 +356,14 @@ def add_edges(var):
         add_edge("source", "source", capacity=float("inf"))
 
     for var in required_producer_vars:
-        add_edge("source", var.name + "_in", capacity=float("inf"))
-        add_edges(var)
+        if var is not None:
+            add_edge("source", var.name + "_in", capacity=float("inf"))
+            add_edges(var)
```
IvanYashchuk, line 361: This change can be reverted when using `flat_proxy_outs`.

thunder/core/rematerialization.py
```diff
@@ -125,7 +127,7 @@ def apply_rematerialization_for_producer(
     )
     all_produced_vars = tuple(chain.from_iterable((y for y in x.flat_outs) for x in producer.subsymbols))
     # Choose the new producer's output from all the produced variables.
-    new_producer_output = tuple(x for x in all_produced_vars if x.name in new_producer_output_names)
+    new_producer_output = tuple(x for x in all_produced_vars if x is not None and x.name in new_producer_output_names)
```
IvanYashchuk, line 130: We can revert this change.
```suggestion
    new_producer_output = tuple(x for x in all_produced_vars if x.name in new_producer_output_names)
```

thunder/core/rematerialization.py
```diff
@@ -125,7 +127,7 @@ def apply_rematerialization_for_producer(
     )
     all_produced_vars = tuple(chain.from_iterable((y for y in x.flat_outs) for x in producer.subsymbols))
```
IvanYashchuk, line 128:
```suggestion
    all_produced_vars = tuple(chain.from_iterable((y for y in x.flat_proxy_outs) for x in producer.subsymbols))
```

thunder/core/rematerialization.py
```diff
@@ -153,7 +155,7 @@ def apply_rematerialization_for_consumer(
     external_inputs = find_external_consumer_inputs(producer, consumer)
     all_produced_vars = tuple(chain.from_iterable((y for y in x.flat_outs) for x in producer.subsymbols))
```
IvanYashchuk, line 156:
```suggestion
    all_produced_vars = tuple(chain.from_iterable((y for y in x.flat_proxy_outs) for x in producer.subsymbols))
```

thunder/core/rematerialization.py
```diff
@@ -153,7 +155,7 @@ def apply_rematerialization_for_consumer(
...
     cut_names = tuple(map(lambda x: x.name, cut)) if isinstance(cut[0], ProxyInterface) else tuple(cut)
-    cut_inputs = tuple(filter(lambda x: x.name in cut_names, (*all_produced_vars, *producer.args)))
+    cut_inputs = tuple(filter(lambda x: x is not None and x.name in cut_names, (*all_produced_vars, *producer.args)))
```
IvanYashchuk, line 158: We can revert this.
```suggestion
    cut_inputs = tuple(filter(lambda x: x.name in cut_names, (*all_produced_vars, *producer.args)))
```

IvanYashchuk (approved):
In `thunder/core/rematerialization.py` we can use `BoundSymbol.flat_proxy_outs` to remove "if not None" statements. Other than this I think this PR is good to merge.

thunder/core/rematerialization.py
kiya00, line 359:
```suggestion
        add_edge("source", var.name + "_in", capacity=float("inf"))
        add_edges(var)
```

kiya00:
Hi @t-vi @mruberry , I think it's ready to merge, could you take a look?

mruberry:
fyi @jjsjann123, too

mruberry (approved):

mruberry:
@Borda I'm not sure why this isn't merging?

kiya00:
Thanks @mruberry , probably because I didn't resolve some comments, let me do it now

Borda:
> Thanks @mruberry , probably because I didn't resolve some comments, let me do it now

Correct

PR #54, opened 2024-03-22
Make cudnnex tests deterministic.

wujingyue:
**If** we really want to cover multiple configs in the current randomization ranges, we should sweep the range instead of randomly picking one.


t-vi (approved):
Seems good to me. Thank you @wujingyue

PR #35, opened 2024-03-21
Enable cat for nvfuser >= 0.1.7

jacobhinkle:
Benchmarks are neutral. 

Before:
```
--------------------------------------------------------------------------------------------------------------------- benchmark: 24 tests ----------------------------------------------------------------------------------------------------------------------
Name (time in us)                                                 Min                       Max                      Mean                 StdDev                    Median                    IQR            Outliers          OPS            Rounds  Iterations
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_nanogpt_sdpa_fwd[thunder]                                85.9729 (1.0)            123.5039 (1.0)             90.1970 (1.0)           9.0070 (1.45)            87.2805 (1.0)           1.0572 (1.0)           2;3  11,086.8437 (1.0)          20           1
test_nanogpt_layer_norm_fwd[thunder]                          98.7879 (1.15)           236.1580 (1.91)           108.8600 (1.21)         30.0711 (4.83)           101.3121 (1.16)          3.2765 (3.10)          1;2   9,186.1104 (0.83)         20           1
test_nanogpt_gelu_fwd[thunder]                               110.0181 (1.28)           217.0510 (1.76)           115.9017 (1.28)         16.7485 (2.69)           112.1076 (1.28)          3.1011 (2.93)          1;3   8,628.0009 (0.78)         40           1
test_nanogpt_gelu_grad[thunder]                              293.2760 (3.41)           706.9011 (5.72)           324.0489 (3.59)         69.9223 (11.23)          299.2521 (3.43)         15.2696 (14.44)         2;9   3,085.9543 (0.28)         40           1
test_nanogpt_csa_fwd[thunder]                                312.5530 (3.64)           433.2711 (3.51)           320.6752 (3.56)         26.5578 (4.26)           314.2810 (3.60)          2.4339 (2.30)          1;1   3,118.4206 (0.28)         20           1
test_llama2_7b_rmsnorm_grad[thunder]                         413.3140 (4.81)           725.0749 (5.87)           494.8684 (5.49)         73.1271 (11.74)          511.1340 (5.86)        127.2761 (120.39)       20;0   2,020.7393 (0.18)         40           1
test_nanogpt_sdpa_grad[thunder]                              465.0408 (5.41)           629.3731 (5.10)           550.4548 (6.10)         57.3874 (9.22)           584.9096 (6.70)         97.4401 (92.17)         8;0   1,816.6797 (0.16)         20           1
[...]

PR #59, opened 2024-03-23
Allocate dQ, dK, and dV as a catted tensor to save a downstream cat in nvFuser.

wujingyue:
For https://github.com/NVIDIA/Fuser/issues/1768. 

The description on the added compile option explains what this optimization does.

This optimization is disabled by default for now. I'll try to enable it by default or even always after #35 is merged and bookend is [disabled by default](https://github.com/Lightning-AI/lit-thunder-LEGACY/issues/1562).


thunder/tests/test_cudnn_executor.py
```diff
@@ -209,25 +206,25 @@ def test_vjp_correctness_sdpa_cudnnex_manual(op, device, dtype, executor, comp):
...
-            executors_list=executor.executors_list() + [cudnn_ex],
+            executors_list=[cudnn_ex],
+            cudnn_sdpa_bwd_may_cat_grad_qkv=may_cat_grad_qkv,
         )
 
         actual_out, actual_grad = cfoo(filtered_args, (v,))
 
-        comp(actual_out, expect_out, atol=1e-2, rtol=1e-2)
-
+        torch.testing.assert_close(actual_out, expect_out, atol=1e-2, rtol=1e-2)
```
jjsjann123, line 227: wondering if tolerance should be part of the opinfo? Not really something that we should do in this PR.

jjsjann123 (approved):
:clap:

thunder/tests/test_cudnn_executor.py
```diff
@@ -186,15 +188,10 @@ def test_cudnn_vs_torch_consistency(op, device, dtype, *_):
             return result
 
 
-# NOTE Scaled_Dot_Product_Efficient_Attention_Backward does not support fp64 dtypes
-# RuntimeError: Only fp32, half & bf16 supported at the moment
-@ops(
```
wujingyue, line 191: I didn't manage to add an extra parameter (e.g. `may_cat_grad_qkv`) to the `ops` instantiation. So I reworked (and happened to simplify) the parameterization part of the test.

thunder/tests/test_cudnn_executor.py
wujingyue: It's already allowed to be, e.g., https://github.com/Lightning-AI/lightning-thunder/blob/c67c81428000da9fe26e5c62f750ede9a87ed361/thunder/tests/opinfos.py#L6409. Figuring out the right threshold for cudnn sdpa, however, is real art to me.

t-vi (approved):
Per @jjsjann123 review. Thank you @wujingyue @jjsjann123 .

wujingyue:
Hey @vedaanta-nvidia, are you interested in reviewing this before I merge?

thunder/tests/test_cudnn_executor.py
vedaanta-nvidia: True.

bf16 tolerances used for internal cudnn testing are also always higher than fp16.
Right now. sdpa tests in cudnn testing do:
`atol=2e-2 if input_type != torch.bfloat16 else 7e-2, rtol=2e-2`

Plus, it is known that there is slightly large mismatches between flash attention based kernels compared to pure math kernels. So depending on what reference is being run above, there are large swings in absolute difference.

wujingyue: Got it. Do you request a code change here? I thought I didn't change the behavior but I could be wrong if `comp` (what we used to use) is more complicated than just forwarding to `torch.testing.allclose`.

vedaanta-nvidia: I think its fine as it passes all tests. I can rework tolerances once it starts causing issues.

wujingyue:
> Hey @vedaanta-nvidia, are you interested in reviewing this before I merge?

Hey Vedaanta, do you have more comments before I merge?

thunder/executors/cudnnex.py
```diff
@@ -604,9 +668,20 @@ def _cudnn_sdpa_grad(
...
     )
 
-    g = get_grad(primal)
+    description = """\
+This flag is for enabling nvFuser's zipping optimization that seeks to avoid
+expensive concatenation. https://github.com/NVIDIA/Fuser/issues/1768 has more
+details. When this flag is true, cudnn_sdpa_bwd may cat dQ, dK and dV as one
+tensor and return them as slices of that tensor.
+"""
+    may_cat_grad_qkv: None | bool = get_compile_option("cudnn_sdpa_bwd_may_cat_grad_qkv", description)
```
vedaanta-nvidia, line 677: Nice, I did not know that executors can get hints/kwargs from thunder.compile.

thunder/executors/cudnnex.py
```diff
@@ -464,19 +480,53 @@ def cudnn_sdpa_backward_meta(
...
     *,
     scale: None | float = None,
-) -> (TensorProxy, TensorProxy, TensorProxy):
-    grad_query = TensorProxy(like=query)
-    grad_key = TensorProxy(like=key)
-    grad_value = TensorProxy(like=value)
+    cat_grad_qkv: bool,
+) -> tuple[TensorProxy, ...]:
+    if cat_grad_qkv:
+        grad_qkv = TensorProxy(
```
vedaanta-nvidia, line 486: The compile flag cudnn_sdpa_bwd_may_cat_grad_qkv suggests that cudnnex "may" output concatenated grads. Seems like its just assumed that cudnnex always be able to concatenate grads, as this code path only returns `grad_qkv`.
(Relying on cudnn to be able to do output grads in a strided fashion is a safe assumption for future cudnn versions too, as long as the embedding dim stride is 1.)

thunder/executors/cudnnex.py
```diff
@@ -207,6 +208,11 @@ def compute_NHWC_strides(shape):
 # And when registering for sdpa, cudnn assumes NHWC layout. (See _transform_sdpa_inputs())
 def _sdpa_enforce_input_tensor_contiguity(a: torch.Tensor) -> torch.Tensor:
     if a.stride(-1) == 1:
+        # TODO(vedaanta-nvidia): there's an inconsistency between
```
vedaanta-nvidia, line 211: nice catch. I think `_transform_sdpa_inputs` should not overwrite strides if already set.

So when cudnn checker function is called with TensorProxy:
- assume contiguous stride
- create cudnn execution plan (just fyi, this was removed for the time being in your previous PR)
- store in cache for potential performance boost in execution phase

And then at execution phase with actual torch tensors:
- Use tensor strides to redo cudnn graph/execution plan ([PPOPOSED] DONT OVERWRITE/ASSUME STRIDES)
- If tensor strides are already contiguous, execution benefits from cache hit
- Execute cudnn plan

I think above approach should be safe to do. What do you think?

vedaanta-nvidia (approved):

thunder/executors/cudnnex.py
wujingyue: > Use tensor strides to redo cudnn graph/execution plan

One potential problem is that the above step can fail even though the checker was happy. Still, with your proposal, cudnnex can at least throw a very clear runtime error when that happens, which is a strict improvement.

That being said, I think the fundamental problem here is: at this moment, a Thunder executor's checkers have no access to strides, but cudnnex needs strides to accurately check an sdpa op can be supported. Am I right? If so, we should probably discuss with Thunder folks on a better, maybe longer-term fix.

thunder/executors/cudnnex.py
wujingyue: > Seems like its just assumed that cudnnex always be able to concatenate grads

No, cat_grad_qkv is not the same as may_cat_grad_qkv. It combines `may_cat_grad_qkv` and [some size/stride checks](https://github.com/Lightning-AI/lightning-thunder/pull/59/files#diff-783705c9be7fba72357ae0678dac1da2961ddb64d37a40eb6e3bf03558f8e313R681) indicating "is able to cat".

PR #60, opened 2024-03-23
Fix error propagation from list init

t-vi:
@kiya00 With any bit of luck, this should also unblock your enumerate PR.

### DIFF
--- a/thunder/core/interpreter.py
+++ b/thunder/core/interpreter.py
@@ -2101,8 +2101,9 @@ class MutSequenceWrapperMethods(SequenceWrapperMethods):
         return wrap_const(cls.value())
 
     def __init__(self, iterable=()):
-        SequenceWrapperMethods.__init__(self, iterable)
-        return wrap_const(None)
+        # We need to propagate the return value because it could be JIT_SIGNALS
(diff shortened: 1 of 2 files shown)


thunder/core/interpreter.py
```diff
@@ -2101,8 +2101,9 @@ def __new__(cls, iterable=()):
         return wrap_const(cls.value())
 
     def __init__(self, iterable=()):
-        SequenceWrapperMethods.__init__(self, iterable)
-        return wrap_const(None)
+        # We need to propagate the return value because it could be JIT_SIGNALS
+        res = SequenceWrapperMethods.__init__(self, iterable)
+        return res
```
Borda, line 2106:
```suggestion
        return SequenceWrapperMethods.__init__(self, iterable)
```

nikitaved (approved):
LGTM! Thank you, @t-vi !

thunder/core/interpreter.py
t-vi: Yeah, we could, but I exect to add debug prints soonish.

t-vi:
Thank you for reviewing @nikitaved  @Borda

PR #75, opened 2024-03-25
Rename thunder-fwd-bwd->thunder for test_llama2_qkv_split_rope_7b_train

IvanYashchuk:
All other instances were already renamed to be called just "thunder".


t-vi (approved):
Thank you @IvanYashchuk

PR #78, opened 2024-03-26
Error on `torch` functions which are not part of the `ltorch`

nikitaved:
As per title. We error and recommending using `thunder.examine.examine` for checking support.

### DIFF
--- a/thunder/core/jit_ext.py
+++ b/thunder/core/jit_ext.py
@@ -912,11 +912,15 @@ def general_jit_lookaside(fn, *args, **kwargs) -> None | Callable:
             # Torch functions have __name__ defined
             fn_name = f"{fn.__module__}.{fn.__name__}"
 
-            # For now, only torch-like opaque functions are sharp edges
-            return _general_jit_sharp_edge(
-                f"Trying to call function {fn_name}, but it's unsupported. Please file an issue requesting support.",
-                None,
-            )
+            # Probably merge with sharp edges
+            calling_opaque_torch_msg = f"Trying to call function {fn_name}, but it is not yet supported. " \
+                                           "Please file an issue requesting support. " \
+                                           "To find out which operations are not yet recongnized by `thunder.jit`, " \
+                                           "please run `examine` as per:\n\n" \
+                                           "from thunder.examine import examine\n" \
+                                           "examine(<your thunder.jit callable argument>, ...)\n"
+
+            return do_raise(NotImplementedError(calling_opaque_torch_msg))
 
     return lookaside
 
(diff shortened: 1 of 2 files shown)


thunder/core/jit_ext.py
```diff
@@ -908,12 +908,16 @@ def is_from_torch(fn):
             # Torch functions have __name__ defined
             fn_name = f"{fn.__module__}.{fn.__name__}"
 
-            # For now, only torch-like opaque functions are sharp edges
-            return _general_jit_sharp_edge(
-                f"Trying to call function {fn_name}, but it's unsupported. Please file an issue requesting support.",
-                None,
+            # Probably merge with sharp edges
+            get_calling_opaque_torch_msg = (
+                lambda: f"Trying to call function {fn_name}, but it is not yet supported. "
```
carmocca, line 913: Why does this need to be a lambda?

thunder/core/jit_ext.py
```diff
@@ -908,12 +908,16 @@ def is_from_torch(fn):
...
+                "Please file an issue requesting support. "
+                "For getting a full list of unsupported functions we recommend "
+                "running `examine` from the `thunder.examine` module on your callable/PyTorch module."
```
carmocca, line 916: The current wording makes it sound like it will show all unsupported functions in the world

```suggestion
                "For getting a full list of unsupported functions in your function/module we recommend "
                "running `examine` from `thunder.examine`"
```

carmocca, line 916: Should we go the extra mile and include a tiny snippet?

```

examine(model_or_fn)(*inputs)
```

carmocca (approved):

thunder/core/jit_ext.py
nikitaved: Future proofing for sharp edges

thunder/core/jit_ext.py
nikitaved: Am I the only one who always gets confused about what the work `module` mean?

carmocca: In PyTorch? All of us. Each person has their own little convention.

My own is "module" when referring to code and implementations and "model" when talking about a specific instance, kind of like `model = MyModule()`. PyTorch Lightning also generally follows this.

nikitaved: Could you please elaborate on the wording? Or could you please provide an alternative that you believe is clearer?

nikitaved: I mean the confusion between a Python module and a PyTorch `nn.Module`.

thunder/core/jit_ext.py
nikitaved: Sorry, this is the terminal path, will fix that.

thunder/core/jit_ext.py
carmocca: Maybe let's just skip the word entirely and say

```
To find out which pieces failed to `thunder.jit`, we recommend running `examine`:

examine(what_you_passed_to_jit)(*inputs)
```

nikitaved:
@carmocca , I have addressed your comments.  Let me know if you have more to say.

carmocca:
Thank you @nikitaved!

PR #79, opened 2024-03-26
Preserve parameter names in the torch autograd path

IvanYashchuk:
Currently, FSDP bucketing relies on parameter names to determine parameters coming from the same nn.Module and group them together. If we don't preserve parameter names, then bucketing is broken. This PR fixes that and adds a test.


crcrpar (approved):
maybe it'd be nice to have a new test in `test_ddp.py` to make sure bucketing does pack multiple tensors into one so that future regressions can be detected a bit easier.

t-vi (commented):
Seems good do me. Thank you @IvanYashchuk @crcrpar

t-vi (approved):

PR #80, opened 2024-03-26
TE executor: DDP support (PR2408)

kshitij12345:
Whenever, the world size is greater than 1, the first TE module in the forward (i.e. the last to execute backward) takes care of syncing the fp8 meta-data state across all processes before the next iteration begins.

During forward, TE takes care of setting up for fp8 meta-data reduction if `world_size > 1` for each TE module.
https://github.com/NVIDIA/TransformerEngine/blob/a38b291b0d1b04847e8ab1df8550df642a03a27d/transformer_engine/pytorch/module/base.py#L552-L564

During backward of the first TE module in forward pass (/last in backward pass), it takes care of actually syncing the FP8 meta-data and this is by default synchronous/blocking. (It essentially does the torch.cat of all fp8 state, reduction, and torch.split on the reduced state back to individual buffers)
https://github.com/NVIDIA/TransformerEngine/blob/8255f87f3ee8076db21777795ce15b6ddf8754c0/transformer_engine/pytorch/module/base.py#L98-L100

This means there are constraints on re-ordering of the `te_linear`, see `NOTE: TransformerEngine Distributed Ordering Constraint` in `torch_autograd.py`.

Thanks @crcrpar for pointing me towards this.

This PR adds a DDP test for `thunder+TE executor` compared to `PyTorch Eager + TE`.

**Benchmark numbers on a real a model**:

Running `examples/llama2.c/train.py` on 2 Ada RTX6000
Cmd : `torchrun --nproc-per-node=2 train.py`

Without TE
```
95 | loss 8.2919 | lr 4.750000e-05 | 357.21ms | mfu 5.69%
96 | loss 8.3111 | lr 4.800000e-05 | 357.91ms | mfu 5.69%
97 | loss 8.2762 | lr 4.850000e-05 | 356.87ms | mfu 5.69%
98 | loss 8.2394 | lr 4.900000e-05 | 355.05ms | mfu 5.69%
99 | loss 8.2340 | lr 4.950000e-05 | 355.43ms | mfu 5.69%
100 | loss 8.1790 | lr 5.000000e-05 | 355.44ms | mfu 5.69%
```

With TE
```
95 | loss 8.3030 | lr 4.750000e-05 | 334.94ms | mfu 6.05%
96 | loss 8.3212 | lr 4.800000e-05 | 335.32ms | mfu 6.05%
97 | loss 8.2859 | lr 4.850000e-05 | 334.99ms | mfu 6.05%
98 | loss 8.2492 | lr 4.900000e-05 | 334.98ms | mfu 6.05%
99 | loss 8.2434 | lr 4.950000e-05 | 335.02ms | mfu 6.05%
100 | loss 8.1892 | lr 5.000000e-05 | 334.47ms | mfu 6.05%
```

<details>

<summary> Patch for Benchmark </summary>

```patch
diff --git a/examples/llama2.c/train.py b/examples/llama2.c/train.py
index 18290df0..dcb52561 100644
--- a/examples/llama2.c/train.py
+++ b/examples/llama2.c/train.py
@@ -28,6 +28,7 @@ from model import Transformer, ModelArgs
 from torch.distributed import destroy_process_group, init_process_group
 from torch.nn.parallel import DistributedDataParallel as DDP
+import transformer_engine.pytorch as te
 
 from tinystories import Task
 from export import model_export
@@ -60,7 +61,7 @@ dropout = 0.0
 # adamw optimizer
 gradient_accumulation_steps = 4  # used to simulate larger batch sizes
 learning_rate = 5e-4  # max learning rate
-max_iters = 100000  # total number of training iterations
+max_iters = 100  # total number of training iterations
 weight_decay = 1e-1
 beta1 = 0.9
 beta2 = 0.95
@@ -212,7 +213,8 @@ if compile == "thunder":
 
     import thunder
     from thunder.executors.sdpaex import sdpa_ex
-    executors = [sdpa_ex, thunder.nvfuser_executor, thunder.pytorch_executor]
+    from thunder.executors.transformer_engineex import transformer_engine_ex
+    executors = [transformer_engine_ex, sdpa_ex, thunder.nvfuser_executor, thunder.pytorch_executor]
 
     eval_model = thunder.compile(eval_model.eval(), disable_torch_autograd_support=True, executors_list=execu
tors)
     train_model = thunder.compile(train_model.train(), executors_list=executors)
@@ -316,7 +318,7 @@ while True:
             # I really dislike that this bloats the code and forces us to repeat code
             # looking at the source of that context manager, it just toggles this variable
             train_model.require_backward_grad_sync = micro_step == gradient_accumulation_steps - 1
-        with ctx:
+        with ctx, te.fp8_autocast():
             logits = train_model(X, Y)
             loss = F.cross_entropy(logits.view(-1, logits.size(-1)), Y.view(-1), ignore_index=-1)
             loss = loss / gradient_accumulation_steps
```
</details>


kshitij12345:
Blocked on https://github.com/Lightning-AI/lightning-thunder/issues/81

IvanYashchuk (approved):
Perfect!

kshitij12345:
Ping @t-vi for merging.

thunder/tests/distributed/test_ddp.py
```diff
@@ -28,6 +28,27 @@
...
 from thunder.tests.framework import instantiate
 
+# It is important to set this flag so that TE doesn't use
+# `torch.compile` to fuse a few operations. This is because
+# `torch.compile` creates a new process and that leads to
+# the error : daemonic processes are not allowed to have children
+# when running the tests.
+# With the setting below, we use `torch.jit` for this test suite
+# See: https://github.com/NVIDIA/TransformerEngine/blob/a38b291b0d1b04847e8ab1df8550df642a03a27d/transformer_engine/pytorch/jit.py#L11-L19
+os.environ["NVTE_TORCH_COMPILE"] = "0"
```
carmocca, line 38: It would be best to set this with pytest monkeypatch or unittest mock for the tests that run. This could otherwise become a source of future painful test interacitons

thunder/tests/distributed/test_ddp.py
```diff
@@ -1237,6 +1258,187 @@ def finalize_pg(pg):
...
+
+        return comparison_exceptions
+
+
+def _test_ddp_transformer_engine_llama_sanity(input_data):
+    # Test Description: We run a dummy training loop for a Transformer Model
+    # We run a few iterations to see that TransformerEngine doesn't throw internal assertion
+    # due to reordering of forward and backward operators.
+    # (This test will fail without `_rearrange_transformer_engine_linear` in `torch_autograd.py`)
+    # For more details, see docstring for `_rearrange_transformer_engine_linear` in transformer_engine_ex.py.
```
carmocca, line 1395: I don't love this test. It's generally better to check that something happens (linears were rearranged) than checking that something doesn't happen (any exception is raised).

What if instead you assert how the trace is ordered?

thunder/executors/torch_autograd.py
```diff
@@ -325,6 +325,13 @@ def make_trace(func):
     if getattr(compile_data.fn, "use_ddp", False):
         bw_extrace = sort_waits(bw_extrace)
 
+    # Importing here to avoid cyclical dependencies in future.
+    from thunder.executors.transformer_engineex import transformer_engine_ex, _rearrange_transformer_engine_linear
+
+    if transformer_engine_ex in compile_data.executors_list:
+        # NOTE: `_rearrange_transformer_engine_linear` mutates `fw_extrace`.
+        _rearrange_transformer_engine_linear(fw_extrace, bw_extrace)
```
carmocca, line 333: Maybe in the future TransformerEngine should become a transform that automatically enables the operator to avoid having this hardcoded here.

cc @t-vi since we've talked about this and it's similar to fsdp

thunder/tests/distributed/test_ddp.py
kshitij12345: This flag affects the behaviour at import time so we have to have it before the import

Also, this flag should only affect the TransformerEngine tests in this particular file as this flag would be unset at the end of test script.

https://stackoverflow.com/questions/68040747/why-does-environment-variable-get-deleted-after-the-end-of-a-python-script

thunder/tests/distributed/test_ddp.py
kshitij12345: Have updated the test, thanks!

tfogal (approved):
The minor nit about environment variables aside, I am concerned about this approach long-term. Wouldn't we also have to do something special for the FSDP transform, then? And couldn't any other arbitrary pass reorder things in such a way that it breaks this implicit requirement?
Perhaps I am misunderstanding something; apologies if so!

If TE is doing communication implicitly and that communication can cause issues, I think we need to have real dialogue with the TE team on this. I wouldn't be surprised if the answer is that we need to take over TE's communication here. Two systems scheduling comms without cognizance of the other is going to lead us towards a deadlock-prone system.

thunder/tests/distributed/test_ddp.py
kshitij12345: Yes, `NVTE_TORCH_COMPILE` is a TransformerEngine env variable. The issue is mainly that `torch.compile` doesn't work well with multiprocessing pool (see https://github.com/pytorch/pytorch/issues/97992). I will try to use `os.unsetenv` for now.

kshitij12345: My bad, unittest.mock did do the trick. Thanks @carmocca @tfogal

carmocca (approved):
LGTM and thank you!

thunder/tests/distributed/test_ddp.py
```diff
@@ -1237,6 +1250,206 @@ def finalize_pg(pg):
...
+            if bsym.sym.name.startswith("te_linear"):
+                first_ctx_name = bsym.output[1].name
+                break
+
+        for bsym in reversed(bwd_exec_trace.bound_symbols):
+            if bsym.sym.name.startswith("te_functional"):
+                assert first_ctx_name == bsym.args[-1].name, (first_ctx_name, bsym.args[-1].name)
+                break
+    except Exception as e:
+        sanity_exceptions.append(e)
```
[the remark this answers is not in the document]
kshitij12345: This is intentional. We return the caught exception from rank 0: https://github.com/Lightning-AI/lightning-thunder/blob/bf6d2759ebc82e69596b1a33089cf520e9d0ff22/thunder/tests/distributed/test_ddp.py#L1448-L1449

and then it is raised from the test process
https://github.com/Lightning-AI/lightning-thunder/blob/3bde9e695385ba37fb365c5ed1cd1cc71995aba2/thunder/tests/distributed/test_ddp.py#L1054-L1058

I think without catching the exception, the test process may crash and hang.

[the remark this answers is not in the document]
carmocca: Thank you for explaining. I wasn't familiar with this

carmocca:
Once we get past the flakes automerge will do its thing :raised_hands:

Issue #81, opened 2024-03-26
caching in make_aug_forward_and_backward breaks TE executor.


kshitij12345:
As discussed offline, Caching in `make_aug_forward_and_backward` leads to reusing the symbols created by `transformer_engine_ex` which are stateful and lead to incorrect program.
Ref:
https://github.com/Lightning-AI/lightning-thunder/blob/bdf5c3fe360a7cbaf6c1cc68eba91631e55ed190/thunder/core/vjp_utils.py#L47-L50

Sample Program
```python
import torch
import thunder
from thunder.executors.transformer_engineex import transformer_engine_ex
from transformer_engine.pytorch import fp8_autocast
dim = 256

class ThunderModel(torch.nn.Module):
    def __init__(self) -> None:
        super().__init__()
        self.fc1 = torch.nn.Linear(dim, dim, bias=False)
        self.fc2 = torch.nn.Linear(dim, dim, bias=False)

    def forward(self, x):
        return self.fc2(torch.nn.functional.relu(self.fc1(x)))

x = torch.arange(dim * dim, dtype=torch.float).view(dim, dim).cuda()

thunder_model = ThunderModel().cuda()

jit_model = thunder.jit(thunder_model, executors=(transformer_engine_ex,),)

with fp8_autocast():
    o = jit_model(x).sum()

print(thunder.last_traces(jit_model)[-1])
```

Generated Trace (`te_linear_0` is called twice):
```python
# Constructed by Delete Last Used (took 0 milliseconds)
import torch
from thunder.executors.torchex import no_autocast

@torch.no_grad()
@no_autocast()
def augmented_forward_fn(*args):
  # args: "Collection"
  t0, t1, t2, = args
  del args
  (t6, ctx_te_1) = te_linear_0(t0, t1, None)
  t7 = torch.gt(t6, 0.0)  # t7: "cuda:0 b8[256, 256]"
    # t7 = ltorch.gt(t6, 0.0)  # t7: "cuda:0 b8[256, 256]"
      # t7 = prims.gt(t6, 0.0)  # t7: "cuda:0 b8[256, 256]"
  t8 = torch.where(t7, t6, 0.0)  # t8: "cuda:0 f32[256, 256]"
    # t8 = ltorch.where(t7, t6, 0.0)  # t8: "cuda:0 f32[256, 256]"
      # t8 = prims.where(t7, t6, 0.0)  # t8: "cuda:0 f32[256, 256]"
  del t6
[...]

PR #82, opened 2024-03-26
Fix TransformerEngine grad transform

IvanYashchuk:
Today we split `grad_transform` into forward and backward with https://github.com/Lightning-AI/lightning-thunder/blob/bdf5c3fe360a7cbaf6c1cc68eba91631e55ed190/thunder/core/vjp_utils.py#L18-L36

It has built-in caching which is meant to speed up the initial trace construction, but this caching assumes the result of the split is only dependent on input metadata. This is not true in the case of TransformerEngine which creates its own BoundSymbol on the fly for each call to the grad transform, this new bound symbol must be unique and not cached because it holds a state related to fp8 scaling.
This PR introduces a skip path to disable caching when needed.

**Test plan:**
Modified test case fails on main and passes on this PR. 

Fixes https://github.com/Lightning-AI/lightning-thunder/issues/81

cc @kshitij12345

### DIFF
--- a/thunder/core/vjp_utils.py
+++ b/thunder/core/vjp_utils.py
@@ -1,11 +1,12 @@
 import inspect
+from collections.abc import Callable
+from functools import wraps
 from inspect import Parameter, Signature
 from itertools import chain
-from collections.abc import Callable
 
 from thunder.core import prims, utils
 from thunder.core.prims import PrimIDs
-from thunder.core.proxies import variableify, Proxy
+from thunder.core.proxies import Proxy, variableify
 from thunder.core.pytree import tree_flatten, tree_map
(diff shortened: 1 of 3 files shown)


kshitij12345 (approved):
LGTM! Thanks @IvanYashchuk 

I think we should not update the cache below if the `joint_forward_backward` has disabled caching. 
https://github.com/Lightning-AI/lightning-thunder/blob/bdf5c3fe360a7cbaf6c1cc68eba91631e55ed190/thunder/core/vjp_utils.py#L168

t-vi (approved):
Seems good to me. Thank you @IvanYashchuk @kshitij12345

PR #83, opened 2024-03-26
Readme improvements

rasbt:
I made some improvements to the readme. It's probably best explained by looking at the file diff.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,17 +33,17 @@ ______________________________________________________________________
 
 Thunder is a source-to-source compiler for PyTorch. It makes PyTorch programs faster by combining and using different hardware executors at once (for instance, [nvFuser](https://github.com/NVIDIA/Fuser), [torch.compile](https://pytorch.org/docs/stable/torch.compiler.html), [cuDNN](https://developer.nvidia.com/cudnn), and [TransformerEngine FP8](https://github.com/NVIDIA/TransformerEngine)).
 
-It supports single accelerators (such as GPUs and TPUs) and also works in multi-GPU settings. 
+It supports single accelerators (such as GPUs and TPUs) and also works in multi-GPU settings.
 
 Thunder aims to be usable, understandable, and extensible.
 
+&#160;
 
-&nbsp;
-
-> [!Note]
+> \[!Note\]
 > Lightning Thunder is in alpha. Feel free to get involved, but expect a few bumps along the way.
 
-&nbsp;
+&#160;
+
 ## Single-accelerator performance
 
 Thunder can achieve significant speedups over standard non-compiled PyTorch code ("PyTorch eager"), through the compounding effects of optimizations and the use of best-in-class executors. The figure below shows the pretraining throughput for Llama 2 7B as implemented in [LitGPT](https://github.com/Lightning-AI/litgpt).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -30,36 +31,50 @@ ______________________________________________________________________
 
 **Thunder makes PyTorch models Lightning fast.**
 
-Thunder is a source-to-source compiler for PyTorch. It makes PyTorch programs faster by combining and using different hardware executors at once (ie: nvFuser, torch.compile, cuDNN, and TransformerEngine FP8).
+Thunder is a source-to-source compiler for PyTorch. It makes PyTorch programs faster by combining and using different hardware executors at once (for instance, [nvFuser](https://github.com/NVIDIA/Fuser), [torch.compile](https://pytorch.org/docs/stable/torch.compiler.html), [cuDNN](https://developer.nvidia.com/cudnn), and [TransformerEngine FP8](https://github.com/NVIDIA/TransformerEngine)).
+
+It supports single accelerators (such as GPUs and TPUs) and also works in multi-GPU settings.
```
carmocca, line 36: Adding TPUs makes it sound like we have an XLA executor which we don't

README.md
```diff
@@ -30,36 +31,50 @@ ______________________________________________________________________
...
 
-Try Thunder without installing by using our [Zero to Thunder Tutorial Studio](https://lightning.ai/lightning-ai/studios/zero-to-thunder-tutorial).
+The easiest way to get started with Thunder, requiring no extra installations or setups, is by using our [Zero to Thunder Tutorial Studio](https://lightning.ai/lightning-ai/studios/zero-to-thunder-tutorial).
+
+&#160;
 
 ## Install Thunder
 
-Install [nvFuser](https://github.com/NVIDIA/Fuser) nightly, and Thunder together
+To use Thunder on your local machine, install [nvFuser](https://github.com/NVIDIA/Fuser) nightly and Thunder together as follows:
```
carmocca, line 77:
```suggestion
To use Thunder on your local machine, first install [nvFuser](https://github.com/NVIDIA/Fuser) nightly and PyTorch nightly together as follows:
```

And then we can separate the `pip install lightning-thunder` step

README.md
```diff
@@ -73,26 +88,54 @@ pip install lightning-thunder
   <summary>Advanced install options</summary>
     <!-- following section will be skipped from PyPI description -->
 
+&#160;
+
 ### Install from main
 
+Alternatively, you can install the latest version of Thunder directly from this GitHub repository as follows:
```
carmocca, line 95: Note that the nvfuser and pytorch nightly steps are required regardless of pip vs github vs local clone. We should clarify that

carmocca (approved):
LGTM!

README.md
rasbt: In that cade, why don't we just simplify "accelerators" to GPUs? More people would be familiar with that.

README.md
rasbt: Thanks. I updated this and added this as explicit step.

README.md
rasbt: Made some updates, let me know what you think.

carmocca: Looks good to me. Let me know if you'd like me to merge it or if you want Luca to check it out first

rasbt: If there's nothing wrong, I'd say let's merge. We can always make an additional pass later.

PR #84, opened 2024-03-26
Add FAQ to docs

apaz-cli:
On the hunt for more frequently asked questions. If you think of anything I haven't covered, let me know.


docs/source/basic/faq.rst
```diff
@@ -0,0 +1,58 @@
...
+
+As such, the two are not necessarily comparable. Or, they are, but you would be comparing against a default configuration, which we expect you to change.
+
+
+
+============================================
+2. How can I use torch.compile with Thunder?
+============================================
+
+The correct way to use ``torch.compile`` with Thunder is as an executor. This gives you finer grained control over which parts of your model are handled by which
```
tfogal, line 20: "by which executor."

tfogal (approved):
Thanks!

carmocca:
Another FAQ would be support for dynamic shapes

docs/source/basic/faq.rst
apaz-cli: Thanks, forgot that

docs/source/basic/faq.rst
```diff
@@ -0,0 +1,58 @@
+Thunder FAQ
+################
+
+=================================================
+1. How does Thunder compare to ``torch.compile``?
+=================================================
+
+Both `Thunder <https://github.com/Lightning-AI/lightning-thunder>`_ and `torch.compile <https://pytorch.org/docs/stable/torch.compiler.html#torch-compiler-overview>`_ are both deep learning compilers that take a pytorch module or callable, and return a callable. As such, it seems reasonable to compare them that way. With that said, the focus of the projects are completely different.
```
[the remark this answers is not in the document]
apaz-cli: Elaborated

apaz-cli:
@carmocca Added questions about dynamic shapes and inplace operations.

docs/source/basic/faq.rst
```diff
@@ -56,3 +58,24 @@ Thunder is in alpha. There will be bugs, and many torch operations are not suppo
...
+
+=======================================
+5. Does Thunder support dynamic shapes?
+=======================================
+
+No, not at the moment. Meta functions operate on the exact shapes of the tensor proxies that pass through them. This is a limitation of the current implementation, and we plan to incorporate dynamic shapes in the future. We're looking for input on how to best incorporate dynamic shapes, so if you have any ideas, please let us know by creating an issue or reaching out.
+
+
+================================================================
+6. Does Thunder support (or plan to support) inplace operations?
```
carmocca, line 71:
```suggestion
6. Does Thunder support inplace operations?
```

Since the other titles don't have this...

docs/source/basic/faq.rst
```diff
@@ -56,3 +58,24 @@ Thunder is in alpha. There will be bugs, and many torch operations are not suppo
...
+
+The common solution is to represent programs in `SSA form <https://en.wikipedia.org/wiki/Static_single-assignment_form>`_, or do some form of SSA-inspired variable renaming, but this is a much less understandable representation than a list of symbols in a trace. At the same time, it would complicate optimization passes and require rewriting many of them to handle these aliasing rules.
+
+There also exists the problem that some backends, like Triton and nvfuser, or a potential Jax backend, are not actually good at handling inplace operations either. They, like Thunder, expect functional programs, and if we simply decide to forward these operations to the executor, not every executor would be able to implement every operator. For a simple torch operator like ``x.add_(y)``, this is unsatisfactory to us. We would ideally like to be able to convert instances of ``add_()`` to ``add()`` for executors. But every trace needs to be executable.
```
carmocca, line 78: I didn't understand how "But every trace needs to be executable" ties into the previous sentence. If we did the dummy thing of functionalizing `add_` into `add`, the trace would still be executable

carmocca (approved):
Love the writing style

docs/source/basic/faq.rst
apaz-cli: My thought was that the first trace, with the `add_` still in it, would not be. We could introduce some heuristic solution for this, for example marking traces as executable or intermediate, but we're still thinking things through.

Unrelated note, I think that an SSA-style tensor/variable renaming pass, without changing the program representation, plus marking traces as executable or not, is probably the solution that I favor. But we should all talk it over at some point, and validate with the nvfuser team that it's a good direction.

carmocca: Makes sense. Maybe this is clearer then

```suggestion
There also exists the problem that some backends, like Triton and nvfuser, or a potential Jax backend, are not actually good at handling inplace operations either. They, like Thunder, expect functional programs, and if we simply decide to forward these operations to the executor, not every executor would be able to implement every operator. For a simple torch operator like ``x.add_(y)``, this is unsatisfactory to us. We would ideally like to be able to convert instances of ``add_()`` to ``add()`` for executors. But currently every intermediate trace needs to be executable and `add_()` would not be a supported bound symbol before it is replaced later.
```

docs/source/basic/faq.rst
```diff
@@ -0,0 +1,80 @@
...
+====================================================================================
+
+Why, yes! You can register it as an operator for a custom executor. See :doc:`extending thunder <../intermediate/additional_executors>` for more information.
+
+
+========================================================================
+3. Do you support custom hardware, or accelerators that aren't Nvidia's?
+========================================================================
+
+Yes, executors are device-agnostic. The python executor, for example, runs the operation with cpython on the cpu. We've been focusing on Nvidia, but we welcome contributions for executors for other accelerator backends.
```
nikitaved, line 49: Is there a way to extend device support without becoming a contributor?

t-vi: You can just do your own executors out of tree (explicit design goal), the API might not be completely stable at this point, though.

docs/source/basic/faq.rst
```diff
@@ -0,0 +1,80 @@
...
+
+
+=================================================================
+4. I ran ``thunder.jit(model)(*args)``, and my model didn't work!
+=================================================================
+
+Thunder is in alpha. There will be bugs, and many torch operations are not supported. Try to run ``thunder.examine(model)(*args)``. This will list the operations which are not supported, and if they are all supported, test the model for consistency against torch eager.
```
carmocca, line 56:
```suggestion
Thunder is in alpha. There will be bugs, and many torch operations are not supported. Try to run ``from thunder.examine import examine; examine(model, *args)``. This will list the operations which are not supported, and if they are all supported, test the model for consistency against torch eager.
```

docs/source/basic/faq.rst
```diff
@@ -0,0 +1,80 @@
...
+
+There also exists the problem that some backends, like Triton and nvfuser, or a potential Jax backend, are not actually good at handling inplace operations either. They, like Thunder, expect functional programs, and if we simply decide to forward these operations to the executor, not every executor would be able to implement every operator. For a simple torch operator like ``x.add_(y)``, this is unsatisfactory to us. We would ideally like to be able to convert instances of ``add_()`` to ``add()`` for executors. But currently, every intermediate trace needs to be executable and `add_()` would not be a supported bound symbol before it gets replaced by `add()` later.
+
+We want to support inplace operations eventually, but we are attached to traces as our program representation of choice for optimization passes. Much like with dynamic shapes, if you have any ideas on how to best incorporate inplace operations, please let us know.
```
t-vi, line 80: I don't think we lack ideas, just someone to do it. (Both 5+6).

t-vi (changes requested):

t-vi:
Supergood start on the FAQ!

docs/source/basic/faq.rst
apaz-cli: Updated it literally just now. Have another look?

docs/source/basic/faq.rst
```diff
@@ -0,0 +1,82 @@
...
+
+The common solution is to represent programs in `SSA form <https://en.wikipedia.org/wiki/Static_single-assignment_form>`_, or do some form of SSA-inspired variable renaming, but SSA is a much less understandable representation than a list of symbols in a trace. Switching to SSA would also complicate optimization passes, and require rewriting many of them to handle these aliasing rules.
+
+There also exists the problem that some backends, like Triton and nvfuser, or a potential Jax backend, are not actually good at handling inplace operations either. They, like Thunder, expect functional programs, and if we simply decide to forward these operations to the executor, not every executor would be able to implement every operation. For a simple torch operator like ``x.add_(y)``, this is unsatisfactory to us. We would ideally like to be able to convert instances of ``add_()`` to ``add()`` for executors. But currently, every intermediate trace needs to be executable and `add_()` would not be a supported bound symbol before it gets replaced by `add()` later.
```
tfogal, line 80: Again, I want to thank you for doing this. It's a very important and often thankless task, plus it's bound to generate a bunch of word-nitting requests... like the one I'm about to make ;-)

> There also exists the problem that some backends, like Triton and nvfuser, or a potential Jax backend, are not actually good at handling inplace operations either.

There's potential for those who don't know us to wonder if we're trying to be snarky and taking stabs at other projects. Plus if one of these backends added SSA overnight someday, they might be disheartened to hear us still claiming issues in their SW.

If you agree, one way to would be to just simplify out the calling out of backends, e.g.:

 > Some backend executors do not support in-place operations.

It omits a lot of details, but those details are not needed to make the point needed for the rest of the paragraph.

tfogal (approved):

docs/source/basic/faq.rst
apaz-cli: Funny enough, given the number of times I've been thanked, I don't think it's a thankless task anymore :)

You're right though, my bad. Any technical elaboration in the Thunder docs should probably only discuss Thunder, and obviously we're big fans of anything that makes models faster. That's sort of the whole point of the project, after all. Fixed.

t-vi (approved):
Let's have those now and improve later.
Big thanks @apaz-cli and also  @Borda @carmocca @nikitaved @tfogal

PR #87, opened 2024-03-27
make sure multiple tensors are passed to `pack_for_fsdp`

crcrpar:
## What does this PR do?

addressing https://github.com/Lightning-AI/lightning-thunder/pull/79#pullrequestreview-1961140778

### DIFF
--- a/thunder/tests/distributed/test_ddp.py
+++ b/thunder/tests/distributed/test_ddp.py
@@ -693,6 +693,24 @@ class CompileDDPTest(DataParallelTestCase):
                 self.assertEqual(loss, orig_loss)
                 self.assertEqual(tuple(p.grad for p in cm.parameters() if p.grad is not None), gradients)
 
+                # Make sure that at least one of "pack" takes multiple tensors.
+                from thunder.executors.torchex import pack_for_fsdp_prim_impl
+                from thunder.distributed.prims import PrimIDs as DistPrimIDs
+
+                for ex_trace in (thunder.last_traces(cm)[-1], thunder.last_backward_traces(cm)[-1]):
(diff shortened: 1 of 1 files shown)


thunder/tests/distributed/test_ddp.py
```diff
@@ -693,6 +693,24 @@ def test_fsdp_grad_parity_with_without_bucketing(
...
                 self.assertEqual(tuple(p.grad for p in cm.parameters() if p.grad is not None), gradients)
 
+                # Make sure that at least one of "pack" takes multiple tensors.
+                from thunder.executors.torchex import pack_for_fsdp_prim_impl
+                from thunder.distributed.prims import PrimIDs as DistPrimIDs
+
+                for ex_trace in (thunder.last_traces(cm)[-1], thunder.last_backward_traces(cm)[-1]):
+                    pack_bsyms = list(
+                        filter(
+                            lambda bsym: bsym.sym.id in {DistPrimIDs.PACK_FOR_FSDP, pack_for_fsdp_prim_impl.id},
```
carmocca, line 703: Question for my understanding: When would each of these 2 symbols appear?

Is it that with the torch executor, you would have `pack_for_fsdp_prim_impl.id` but without it you would have `DistPrimIDs.PACK_FOR_FSDP`?

carmocca (approved):
Thank you!

thunder/tests/distributed/test_ddp.py
crcrpar: I just lazily put both of them. In this case, we see the trace to translate and execute so `pack_for_fsdp_prim_impl.id` might be sufficient. This prim only has torchex impl,fyi

PR #88, opened 2024-03-27
topk: add as a primitive

nikitaved:
As per title. The underlying PyTorch implementation is not trivial, so it is best to model this op as a prim.

There are still some to be done follow-ups to make the op functional and useful. These are:
- backward support
- named tuple wrap so that `indices` and `values` are accessible in the jit. Currently, these are not supported and these could be accessed only with indices.

### DIFF
--- a/thunder/tests/test_jit_general.py
+++ b/thunder/tests/test_jit_general.py
@@ -613,7 +613,7 @@ def test_nanogpt():
         "falcon-7b-like",
         "falcon-40b-like",
         "codellama2-like",
-        pytest.param("mixtral-like", marks=pytest.mark.xfail(raises=NotImplementedError, reason="topk", strict=True)),
+        pytest.param("mixtral-like", marks=pytest.mark.xfail(raises=TypeError, reason="topk", strict=True)),
     ),
 )
 @pytest.mark.parametrize(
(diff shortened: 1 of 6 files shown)


nikitaved:
OK, so the issues popped after the rebase...

thunder/tests/test_jit_general.py
```diff
@@ -613,7 +613,7 @@ def test_nanogpt():
         "falcon-7b-like",
         "falcon-40b-like",
         "codellama2-like",
-        pytest.param("mixtral-like", marks=pytest.mark.xfail(raises=NotImplementedError, reason="topk", strict=True)),
+        pytest.param("mixtral-like", marks=pytest.mark.xfail(raises=TypeError, reason="topk", strict=True)),
```
carmocca, line 616: What is the error now?

carmocca (approved):
Thank you!

thunder/tests/test_jit_general.py
nikitaved: It fails with missing args for where. A yet another folloup

carmocca: @nikitaved Can you fill this in a new issue?

PR #102, opened 2024-03-29
tests: temp xfail for some DDP cases

Borda:
<details>
  <summary><b>Before submitting</b></summary>

- [x] Was this discussed/approved via a Github issue? (no need for typos and docs improvements)
- [x] Did you read the [contributor guideline](https://github.com/Lightning-AI/pytorch-lightning/blob/main/.github/CONTRIBUTING.md), Pull Request section?
- [x] Did you make sure to update the docs?
- [x] Did you write any new necessary tests?

</details>

## What does this PR do?

unblock develop as admit that these cases suddenly do not pass
followup for #99

## PR review

Anyone in the community is free to review the PR once the tests have passed.
If we didn't discuss your PR in Github issues there's a high chance it will not be merged.

## Did you have fun?

Make sure you had fun coding 🙃

PR #103, opened 2024-03-30
[test_ddp][fsdp] check strategy before check pack arg's size

crcrpar:
following #102

### DIFF
--- a/thunder/distributed/__init__.py
+++ b/thunder/distributed/__init__.py
@@ -292,6 +292,9 @@ class FSDPBucketingStrategy(Enum):
 def get_extract_bucket_name_from_tensor_proxy(granularity: FSDPBucketingStrategy):
     from thunder.core.proxies import TensorProxy
 
+    # TODO(crcrpar): Consider making `BLOCK`'s behavior more meticulous for models with simple structure
+    # as in https://github.com/Lightning-AI/lightning-thunder/blob/b24e5b23/thunder/tests/distributed/test_ddp.py#L53-L60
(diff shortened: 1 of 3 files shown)


thunder/distributed/__init__.py
```diff
@@ -278,6 +278,9 @@ class FSDPBucketingStrategy(Enum):
 def get_extract_bucket_name_from_tensor_proxy(granularity: FSDPBucketingStrategy):
     from thunder.core.proxies import TensorProxy
 
+    # TODO(crcrpar): Consider making `BLOCK`'s behavior more meticulous for models with simple structure
```
Borda, line 281: unrelated to the PR but good example, @t-vi @lantiga shall we create issues for all these todos in code, so it won't be lost, I bet (at least me) would not come back to one todos in code :)

thunder/tests/distributed/test_ddp.py
```diff
@@ -715,7 +714,10 @@ def test_fsdp_grad_parity_with_without_bucketing(
                         first_arg = bsym.args[0]
                         self.assertIsInstance(first_arg, list)
                         has_pack_multiple_tensors |= len(first_arg) > 1
-                    self.assertTrue(has_pack_multiple_tensors, msg=f"{[bsym.args[0] for bsym in pack_bsyms]=}")
+                    # note(crcrpar): The way creating a bucket name from an FQN could be better for models with simple structure
```
Borda, line 717: again shall this be todo ir like this us fine?

carmocca (approved):
LGTM. I agree that the current BLOCK implementation is not great. I wonder if we could have this use the original module type if provenance knew about which class types generated which tensors

PR #118, opened 2024-04-02
topk: backward implementation

nikitaved:
As per title. Required to unblock testing of some variants of litgpt models.

### DIFF
--- a/thunder/tests/test_grad.py
+++ b/thunder/tests/test_grad.py
@@ -1394,8 +1394,8 @@ def test_populate_grads_nanogpt(executor, device, dtype):
 
     from thunder.benchmarks import NanoGPTBenchmark, NanoGPTConfig
 
-    # NOTE Currently setting dropout to zero for reproducibility, other settings taken from gpt2 config
-    config = NanoGPTConfig(dropout=0, n_layer=12, n_head=12, n_embd=768)
+    # NOTE Currently setting dropout to zero for reproducibility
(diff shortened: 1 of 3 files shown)


t-vi (approved):
Thank you @nikitaved !

nikitaved:
@t-vi , the failures seem to be unrelated. Let me know if you beg to differ, otherwise it is good to go.

nikitaved:
I cannot reproduce OOMs locally

thunder/tests/test_grad.py
```diff
@@ -1395,7 +1395,7 @@ def test_populate_grads_nanogpt(executor, device, dtype):
     from thunder.benchmarks import NanoGPTBenchmark, NanoGPTConfig
 
     # NOTE Currently setting dropout to zero for reproducibility, other settings taken from gpt2 config
-    config = NanoGPTConfig(dropout=0, n_layer=12, n_head=12, n_embd=768)
+    config = NanoGPTConfig(dropout=0, n_layer=6, n_head=6, n_embd=768)
```
nikitaved, line 1399: @t-vi , so, it turns out topk is used in nanogpt sometimes. The OOM happens because of missing 3mb of GPU memory. I guess these could be autograd buffers tied to topk. I decided to make the model a tad smaller, which should not defeat the purpose of this test.

carmocca: I agree that 12 layers was excessive for a unit test

PR #130, opened 2024-04-04
Minor distributed test improvements

carmocca:
## What does this PR do?

Clearer output of test results.
Avoid profiler spam

### DIFF
--- a/scripts/run_standalone_tests.sh
+++ b/scripts/run_standalone_tests.sh
@@ -46,9 +46,9 @@ for test in $tests; do
   else
     status=$pytest_status
     echo "$test returned status $pytest_status"
-    echo "================TEST OUTPUT BEGIN================"
-    echo "$test-output.txt"
-    echo "================TEST OUTPUT END=================="
+    echo "============FAILED TEST OUTPUT BEGIN================"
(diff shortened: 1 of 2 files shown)


t-vi (approved):
Thank you @carmocca , really appreciate improving the test running!

scripts/run_standalone_tests.sh
```diff
@@ -40,11 +41,14 @@ status=0
 for test in $tests; do
   python -um pytest -sv "$test" --pythonwarnings ignore --junitxml="$test-results.xml" 2>&1 > "$test-output.txt"
   pytest_status=$?
-  printf "$test status >>> $pytest_status\n"
-
-  if [ $pytest_status -ne 0 ]; then
```
carmocca, line 45: Note that this can be very misleading because a skipped test also returns 0, so it can make it seem like a test passed when it didn't run

t-vi: Seems like Python 3.12 fixes that in unittest ( https://github.com/python/cpython/issues/106584 ), so there probably is a canonical "what to do" but I am not sure if there is an easy way to do this for pytest, too.

PR #142, opened 2024-04-08
TransformerEngine : Add test with FSDP (and updates to `ddp_wrapper` in test_ddp.py)

kshitij12345:
This PR adds test for using TE executor in FSDP and verifies it against Eager + TE. Also we update the `ddp_wrapper` to allow wrapping with different pytest-fixture besides `bucket_size_in_mb` (which errored when I tried to add a different pytest-fixture).

PR https://github.com/Lightning-AI/lightning-thunder/pull/80 description details of how TE automatically takes care of syncing FP8 meta-data in distributed setting. 

Also, I have verified it on a larger model using the available benchmarking script
cmd for benchmark: 
```
torchrun --nproc-per-node=2 thunder/benchmarks/benchmark_litgpt.py --compile thunder+nvfuser+transformerengine+cudnn --n_layers=10 --distributed_mode=fsdp
```

Numbers are on RTX 6000

Without TE
```
iter 41: loss 4.6562, iter time: 3180.77ms, t: 4096
iter 42: loss 4.6250, iter time: 3202.35ms, t: 4096
iter 43: loss 4.6562, iter time: 3172.88ms, t: 4096
iter 44: loss 4.6562, iter time: 3181.55ms, t: 4096
Model name: Llama-2-7b-hf
Seq Length: 4096
Micro BS: 1
Global BS: 2
Number of Layers: 10
Number of parameters: 1.14B
Distributed Mode: fsdp
Sharding Mode: zero2
Bucketing: none
Compiler: thunder+nvfuser+cudnn
Average iter time: 3187.63 ms
Memory used: 30.56 GB
Tokens/s: 2570.17
Tokens/s/GPU: 1285.09
TFLOP/s: 38.40
```

With TE
```
iter 42: loss 4.6562, iter time: 3025.66ms, t: 4096
iter 43: loss 4.6562, iter time: 3030.40ms, t: 4096
iter 44: loss 4.6562, iter time: 3018.83ms, t: 4096
Model name: Llama-2-7b-hf
Seq Length: 4096
Micro BS: 1
Global BS: 2
Number of Layers: 10
Number of parameters: 1.14B
Distributed Mode: fsdp
Sharding Mode: zero2
Bucketing: none
Compiler: thunder+nvfuser+transformerenginevfuser+cudnn
Average iter time: 3024.72 ms
Memory used: 37.25 GB
Tokens/s: 2708.12
Tokens/s/GPU: 1354.06
TFLOP/s: 40.47
```


thunder/tests/distributed/test_ddp.py
```diff
@@ -1039,12 +1041,12 @@ def __call__(self, test_stub):
         init_method = f"{FILE_SCHEMA}{file_name}"
 
         @wraps(test_stub)
-        def test_fn(executor, devices, dtype, bucket_size_in_mb=0):
```
kshitij12345, line 1042: This makes it so that we can only use this argument name in the wrapped function with pytest-fixture.

thunder/tests/distributed/test_ddp.py
```diff
@@ -1454,6 +1458,139 @@ def _test_ddp_transformer_engine_llama_sanity(input_data):
...
+                    is_same_across_ranks(thunder_fp8_meta["scaling_fwd"].scale)
+                    is_same_across_ranks(thunder_fp8_meta["scaling_fwd"].scale_inv)
+                    is_same_across_ranks(thunder_fp8_meta["scaling_fwd"].amax_history)
+                    is_same_across_ranks(thunder_fp8_meta["scaling_bwd"].scale)
+                    is_same_across_ranks(thunder_fp8_meta["scaling_bwd"].scale_inv)
+                    is_same_across_ranks(thunder_fp8_meta["scaling_bwd"].amax_history)
+                except Exception as e:
+                    # Return exceptions only for rank==0
+                    if rank == 0:
+                        comparison_exceptions.append(e)
```
crcrpar, line 1578: why would this tentatively ignore errors?

kshitij12345: IIUC, `ddp_wrapper` only re-raises Exceptions thrown from the first process (with rank 0)

https://github.com/Lightning-AI/lightning-thunder/blob/cdd43a7fc1110eec10f1854250299b84d1c3b2a8/thunder/tests/distributed/test_ddp.py#L1063-L1076

Also other tests do the same,
https://github.com/Lightning-AI/lightning-thunder/blob/cdd43a7fc1110eec10f1854250299b84d1c3b2a8/thunder/tests/distributed/test_ddp.py#L1160-L1168

crcrpar: if it really raises the first, then how about returning `comparison_exceptions` here instead of postponing it after the try-except below?

crcrpar (approved):

thunder/tests/distributed/test_ddp.py
```diff
@@ -1528,5 +1677,30 @@ def test_ddp_transformer_engine_llama_sanity(executor, devices, dtype):
...
+        pytest.mark.parametrize(
+            "thunder_fsdp_strategy",
+            (
+                FSDPType.ZERO2,
+                FSDPType.ZERO3,
+            ),
+        ),
+        pytest.mark.skipif(not TE_AVAILABLE, reason="TransformerEngine is not installed."),
+        pytest.mark.skipif(not is_fp8_supported, reason=fp8_support_reason),
+        # NOTE: Setting `NVTE_TORCH_COMPILE`
```
IvanYashchuk, line 1696: I think this note is not informative enough.
As I understand TE might use torch.compile internally and it doesn't work well with the multiprocessing setup in data parallel testing?

thunder/tests/distributed/test_ddp.py
```diff
@@ -1362,7 +1366,18 @@ def is_same_across_ranks(t):
                 # This has to be on all ranks so that the computation is not blocked
                 is_same_across_ranks(thunder_fp8_meta["scaling_fwd"].scale)
                 is_same_across_ranks(thunder_fp8_meta["scaling_fwd"].scale_inv)
-                is_same_across_ranks(thunder_fp8_meta["scaling_fwd"].amax_history)
+                # NOTE: TE forward tensor meta-data sync
```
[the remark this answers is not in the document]
IvanYashchuk: All clear, thank you!

IvanYashchuk (approved):

thunder/tests/distributed/test_ddp.py
```diff
@@ -1454,6 +1469,140 @@ def _test_ddp_transformer_engine_llama_sanity(input_data):
...
+    # Test Description: We run a dummy training loop for a simple `Linear(Relu(Linear(x)))`
+    # model with thunder (using TE executor) and with PyTorch eager + TE
+    # and verify that the weights have converged to same value and
+    # fp8 meta state is same after `n_iter`.
+    init_method, world_size, rank, executor, device, _unused_dtype, kwargs = input_data
+    thunder_fsdp_strategy = kwargs["thunder_fsdp_strategy"]
+    devicetype = devices.device_from_string(device).devicetype
+
+    # Setting LOCAL_RANK is necessary for thunder.distributed.fsdp
+    with unittest.mock.patch.dict(os.environ, {"LOCAL_RANK": str(rank)}, clear=True):
```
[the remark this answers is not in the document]
kshitij12345: This test doesn't use `DataParallelTestCase` but instead uses `ddp_wrapper` and hence we need to set it.

thunder/tests/distributed/test_ddp.py
kshitij12345: Yes that is correct. This note should actually point to the note mentioned. Will update it to say `See NOTE: Setting `NVTE_TORCH_COMPILE`

https://github.com/Lightning-AI/lightning-thunder/blob/34d21e1455bf13e0c397207b018f2835acc3870a/thunder/tests/distributed/test_ddp.py#L1497-L1505

kshitij12345:
@carmocca please have a look and stamp the PR, thanks!

thunder/tests/distributed/test_ddp.py
```diff
@@ -1453,6 +1468,139 @@ def _test_ddp_transformer_engine_llama_sanity(input_data):
...
+        class TEModel(torch.nn.Module):
+            def __init__(self) -> None:
+                super().__init__()
+                self.fc1 = TELinear(dim, dim, bias=False)
+                self.fc2 = TELinear(dim, dim, bias=False)
+
+            def forward(self, x):
+                return self.fc2(torch.nn.functional.relu(self.fc1(x)))
+
+        te_model = TEModel().cuda()
```
carmocca, line 1536:
```suggestion
        with torch.device("cuda"):
            te_model = TEModel()
```

thunder/tests/distributed/test_ddp.py
```diff
@@ -1453,6 +1468,139 @@ def _test_ddp_transformer_engine_llama_sanity(input_data):
     return None
 
 
+def _test_fsdp_transformer_engine(input_data):
+    # Test Description: We run a dummy training loop for a simple `Linear(Relu(Linear(x)))`
+    # model with thunder (using TE executor) and with PyTorch eager + TE
+    # and verify that the weights have converged to same value and
```
carmocca, line 1474: Does TransformerEngine generally guarantee this? Could this become a source of test flakiness in the future if they diverge beyond certain tolerance?

carmocca (approved):

thunder/tests/distributed/test_ddp.py
kshitij12345: Since, we are using the exact same TE layers under the hood, this should be true within floating point errors, otherwise this will signals that there is a bug in the implementation.

thunder/tests/distributed/test_ddp.py
```diff
@@ -1453,6 +1468,139 @@ def _test_ddp_transformer_engine_llama_sanity(input_data):
...
+    # fp8 meta state is same after `n_iter`.
+    init_method, world_size, rank, executor, device, _unused_dtype, kwargs = input_data
+    thunder_fsdp_strategy = kwargs["thunder_fsdp_strategy"]
+    devicetype = devices.device_from_string(device).devicetype
+
+    # Setting LOCAL_RANK is necessary for thunder.distributed.fsdp
+    with unittest.mock.patch.dict(os.environ, {"LOCAL_RANK": str(rank)}, clear=True):
```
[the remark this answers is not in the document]
kshitij12345: Nice catch, I don't think `clear=True` is a good idea, thanks!

PR #149, opened 2024-04-09
add docs about adding ops to thunder

nikitaved:
As per title.

cc @borda @apaz-cli

### DIFF
--- a/docs/source/advanced/contrib_thunder.rst
+++ b/docs/source/advanced/contrib_thunder.rst
@@ -278,7 +278,7 @@ added to `thunder/tests/opinfos.py <https://github.com/Lightning-AI/lightning-th
 
 .. code-block:: python
    :lineno-start: 2997
-   
+
    def unfold_sample_generator(op, device, dtype, requires_grad, **kwargs):
        make = partial(make_tensor, device=device, dtype=dtype, requires_grad=requires_grad)
 
@@ -451,5 +451,3 @@ We also recommend checking out
 :doc:`Defining new Thunder operations <../notebooks/adding_custom_operator>` and
 :doc:`Defining custom forward and backward for existing operators <../notebooks/adding_custom_operator_backward>`
 that cover very similar topics related to extending *Thunder* out of the tree.
-
-
--- a/docs/source/advanced/contrib_thunder.rst
+++ b/docs/source/advanced/contrib_thunder.rst
@@ -1,0 +1,455 @@
+Contributing to Thunder
+#######################
+
+We appreciate your feedback and contributions.
+If you have feature requests, questions, or want to contribute code or config files,
+please don't hesitate to use the `GitHub Issue tracker <https://github.com/Lightning-AI/lightning-thunder/issues>`_.
+
+We welcome all individual contributors, regardless of their level of experience or hardware.
+Your contributions are valuable, and we are excited to see what you can accomplish in this collaborative and supportive environment.
+
+For a simple general overview of *Thunder*, we recommend reading
+:doc:`inside Thunder <../advanced/inside_thunder>` first.
+
+
+================
+Adding operators
+================
+Adding operators might be one of the easiest and fun ways to get involved in contributing to *Thunder*.
+The `operator GitHub Issue tracker <https://github.com/Lightning-AI/lightning-thunder/issues?q=is%3Aissue+is%3Aopen+label%3Aoperators>`_
+provides a great starting point in deciding which operation to work on first.
+
+The subsections below are structured as follows
+
+* `Primitives`_, `The Core Language`_, `The Torch Language`_, `Language Context`_
+  describe the hierarchy of operations and abstractions around them in *Thunder* in general terms.
+* `Adding operations to the Torch executor`_ moves from *theory* to *practice* where we inspect real contributors' pull requests.
+
+We recommend reading the document **sequentially**!
+
+----------
+Primitives
+----------
+The *lowest* level is the primitive operations, defined in `thunder/core/prims.py <https://github.com/Lightning-AI/lightning-thunder/blob/main/thunder/core/prims.py>`_.
+Primitive operations, as seen in the `Representing operations <../advanced/inside_thunder.rst#representing-operations>`__ section,
+describe all the computation performed, and they are intended to be as simple as possible so that
+executors like `nvFuser <https://github.com/NVIDIA/Fuser>`_ find it easy to manipulate them.
+*Thunder's* primitives are similar to `PyTorch’s <https://pytorch.org>`_
+`primTorch primitives <https://github.com/pytorch/pytorch/tree/main/torch/_prims>`_,
+and are based on `JAX’s <https://github.com/google/jax>`_
+`jax.lax operations <https://jax.readthedocs.io/en/latest/jax.lax.html>`_.
+
+Primitives have several parts, as defined by the ``make_prim`` function.
+Most importantly they have an ``id``, a ``name``, and a ``meta`` function.
+The meta function performs error checking and maps the metadata (like ``dtype``, ``shape``, ``device``) of inputs to the primitive metadata of outputs.
+For operations that are part of a class, like the elementwise unary or reduction operations, they often share a common meta function.
+More unique operations, like ``slice``, define their own
+`meta functions <https://github.com/Lightning-AI/lightning-thunder/blob/888b46324462fba70f93d5017bc0d99025f05091/thunder/core/prims.py#L2812>`_.
+
+The actual execution of primitive operations is handled by executors like
+`nvFuser <https://github.com/NVIDIA/Fuser>`_ or
+`PyTorch <https://pytorch.org>`_ – more on that in a moment.
+
+Before adding a primitive, check with the team on its design.
+It might be appropriate to add primitives when necessary to describe the semantics of an operation or to improve the numerical accuracy or speed of operations.
+
+There is a tradeoff with the design of primitive operations one has to keep in mind.
+On one hand, fewer primitive operations can make program transformation, and execution easier.
+Fewer primitives means fewer transformation rules – since transformation rules are defined on primitives – and a smaller interface with executors.
+On the other hand, too few primitive operations may make it hard, or impossible, to express all the operations that users are interested in.
+Too few primitive operations may also make it difficult to execute programs quickly and numerically accurately.
+
+For example, the ``expm1`` operation can mathematically be defined in terms of the ``exp`` and subtraction operations,
+and so it does not need to be a primitive to enable any functionality.
+Many libraries, including `C++’s standard library <https://en.cppreference.com/w/c/numeric/math/expm1>`_,
+still define an ``expm1`` operation for numerical accuracy, and so ``expm1`` is a
+`primitive <https://github.com/Lightning-AI/lightning-thunder/blob/888b46324462fba70f93d5017bc0d99025f05091/thunder/core/prims.py#L1791>`_ in *Thunder*.
+
+
+-----------------
+The Core Language
+-----------------
+Above the primitives is the ``core`` language, or `clang <https://github.com/Lightning-AI/lightning-thunder/blob/main/thunder/clang/__init__.py>`_.
+Clang operations are mostly written like any other Python operation.
+They ultimately call the primitive operations, although they may call other operations before doing so
+(for example, ``clang.foo`` might call ``clang.bar`` which calls ``prims.bar``).
+
+Core language operations are intended to be common functionality that’s useful when
+defining user-facing languages like ``torch`` or ``numpy``.
+Many of these operations are just wrappers around primitive operations.
+For example, the elementwise binary primitives are as simple as possible, so they don’t perform broadcasting or type promotion.
+The core language elementwise binary operations, however, do perform broadcasting and type promotion.
+For example, take a look at the following implementation of ``add`` from `clang <https://github.com/Lightning-AI/lightning-thunder/blob/main/thunder/clang/__init__.py>`_
+
+.. code-block:: python
+  :lineno-start: 1602
+
+  def _elementwise_binary_wrapper(a, b, *, prim, type_promotion_kind=utils.ELEMENTWISE_TYPE_PROMOTION_KIND.DEFAULT):
+      computation_dtype, result_dtype = utils.elementwise_type_promotion(a, b, type_promotion_kind=type_promotion_kind)
+
+      a, b = maybe_broadcast(a, b)
+      a, b = maybe_convert_to_dtype(a, computation_dtype), maybe_convert_to_dtype(b, computation_dtype)
+
+      result = prim(a, b)
+      result = maybe_convert_to_dtype(result, result_dtype)
+
+      return result
+
+
+  @clangop(method_name="add")
+  def add(a, b):
+      return _elementwise_binary_wrapper(a, b, prim=prims.add)
+
+
+Before adding a core language operation consider if the functionality expressed is universal enough.
+
+As a style note, operations in *Thunder* should defer as much error checking as possible.
+For example, if a primitive’s meta function will perform an error check for ``X``,
+then the core language operation that calls it should generally not also check for ``X``.
+
+
+------------------
+The Torch Language
+------------------
+To translate ``torch`` operations into something that *Thunder* understands we define a
+`torch language <https://github.com/Lightning-AI/lightning-thunder/blob/main/thunder/torch/__init__.py>`_.
+Operations in the ``torch`` should reflect the behavior of their corresponding torch operations (small deviations are sometimes OK).
+
+When a program is interpreted, torch operations are remapped into these operations, which ultimately call primitive operations.
+
+
+----------------
+Language Context
+----------------
+In the ``core`` and ``torch`` languages functions are decorated to set a *language context* and – for ``torch`` operations – to describe
+how to map operations like ``torch.foo`` into ``thunder.torch.foo``.
+
+The language context determines what properties and methods tensor objects have.
+For example, when ``a + b`` is written and the first argument is an array or tensor
+(so, `TensorProxy.__add__ is invoked <https://github.com/Lightning-AI/lightning-thunder/blob/71fc8cda4e42b818372b6e9f1c99c9cc3a5c2e38/thunder/core/proxies.py#L1310>`_),
+the language context decides what that addition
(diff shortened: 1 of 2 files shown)


docs/source/advanced/contrib_thunder.rst
```diff
@@ -0,0 +1,468 @@
...
+
+For example, the ``expm1`` operation can mathematically be defined in terms of the ``exp`` and subtraction operations,
+and so it does not need to be a primitive to enable any functionality.
+Many libraries, including `C++’s standard library <https://en.cppreference.com/w/c/numeric/math/expm1>`_,
+still define an ``expm1`` operation for numerical accuracy, and so ``expm1`` is a
+`primitive <https://github.com/Lightning-AI/lightning-thunder/blob/888b46324462fba70f93d5017bc0d99025f05091/thunder/core/prims.py#L1791>`_ in *Thunder*.
+
+
+-----------------
+The Core Language
```
carmocca, line 62: This section describes core langauge very well, but it is missing describing the motivation for having this extra layer instead of merging it into prims. (I don't know what the motivation is, otherwise I would make a more precise suggestion)

docs/source/advanced/contrib_thunder.rst
```diff
@@ -0,0 +1,468 @@
...
+For example, if a primitive’s meta function will perform an error check for ``X``,
+then the core language operation that calls it should generally not also check for ``X``.
+
+
+------------------
+The Torch Language
+------------------
+To translate ``torch`` operations into something that *Thunder* understands we define a
+`torch language <https://github.com/Lightning-AI/lightning-thunder/blob/main/thunder/torch/__init__.py>`_.
+Operations in the ``torch`` language should be written as if a practitioner were writing them,
```
carmocca, line 87: What does

> as if a practitioner were writing them

mean more precisely?

docs/source/advanced/contrib_thunder.rst
```diff
@@ -0,0 +1,468 @@
...
+and they should reflect the behavior of their corresponding torch operations (small deviations are sometimes OK).
+
+When a program is interpreted, torch operations are remapped into these operations, which ultimately call primitive operations.
```
carmocca, line 90: I would expand this section a bit, also adding or linking to an easy example that shows how a torchlang op matches a pytorch operator and how it calls different prims etc. Especially since this is the most popular language for our users.

docs/source/advanced/contrib_thunder.rst
```diff
@@ -0,0 +1,468 @@
...
+
+
+----------------
+Language Context
```
carmocca, line 94: Links to code would be helpful in this section.

docs/source/advanced/contrib_thunder.rst
```diff
@@ -0,0 +1,468 @@
...
+In the ``core`` and ``torch`` languages functions are decorated to set a *language context* and – for ``torch`` operations – to describe
+how to map operations like ``torch.foo`` into ``thunder.torch.foo``.
+
+The language context determines what properties and methods tensor objects have.
+For example, when ``a + b`` is written, the language context decides what that addition means.
+Or when ``a.size`` is used, the language context determines what that means (and it’s different in PyTorch and NumPy).
+
+
+---------------------------------------
+Adding operations to the Torch executor
```
carmocca, line 105: This header could be given more weights since it marks the "Theory is over, let's look at example" boundary

docs/source/advanced/contrib_thunder.rst
```diff
@@ -0,0 +1,468 @@
...
+   SLICE = auto()
+   SQUEEZE = auto()
+   TRANSPOSE = auto()
+   UNFOLD = auto()
+   VIEW = auto()
+   # Memory layout prims (Experimental)
+   STRIDE_ORDER = auto()
+
+.. code-block:: python
+   :emphasize-lines: 1-18
```
carmocca, line 138: This is highlightning the entire snippet

docs/source/advanced/contrib_thunder.rst
```diff
@@ -0,0 +1,468 @@
...
+       shape = list(a.shape)
+       shape.append(size)
+       shape[dim] = (shape[dim] - size) // step + 1
+
+       return TensorProxy(like=a, shape=shape)
+
+
+    unfold = make_prim(PrimIDs.UNFOLD, "unfold", meta=unfold_meta, tags=(OpTags.SHAPE_OP,))
+
+register a primitive symbol ``unfold`` using ``make_prim`` with ``id=PrimIDs.UNFOLD``,
```
carmocca, line 160:
```suggestion
The above registers a primitive symbol ``unfold`` using ``make_prim`` with ``id=PrimIDs.UNFOLD``,
```

docs/source/advanced/contrib_thunder.rst
```diff
@@ -0,0 +1,468 @@
...
+of the underlying ``torch.Tensor.unfold`` operation
+(so that the primitive is directly modeled after the PyTorch operation)
+with the only exception of expecting a ``TensorProxy``
+and not the ``torch.Tensor`` as its input. The rest of the function checks the inputs and returns
+a ``TensorProxy`` of the appropriate shape. ``like=a`` means that the output will inherit the meta-data
+like ``device`` and ``dtype`` from ``a``. The primitive is also tagged with ``tags=(OpTags.SHAPE_OP,)``,
+and, therefore, is associated with shape-based operations.
+We use tags to additionally group operations for group-specific operation optimizations inside *Thunder*.
+
+Additionally, the `clang <https://github.com/Lightning-AI/lightning-thunder/blob/main/thunder/clang/__init__.py>`_ gets updated
```
carmocca, line 171: Why? What's the advantage?

docs/source/advanced/contrib_thunder.rst
carmocca: Or hint at the example below (I hadn't reached it when I posted the above) for those who want to see it already

docs/source/advanced/contrib_thunder.rst
```diff
@@ -0,0 +1,468 @@
...
+
+.. code-block:: python
+   :emphasize-lines: 1-3
```
carmocca, line 174: Again, is there a benefit to highlighting everything?

docs/source/advanced/contrib_thunder.rst
```diff
@@ -0,0 +1,468 @@
...
+   :lineno-start: 1119
+
+   @clangop()
+   def unfold(a: TensorProxy, /, dim: int, size: int, step: int) -> TensorProxy:
+       return prims.unfold(a, dim, size, step)
+
+And while this symbol directly calls into the primitive and is very simple, a more involved
+example of ``clang.slice_in_dim``
```
carmocca, line 182: This is interesting, but it distracts away from the unfold example. What if it's moved to the clang theory section?

docs/source/advanced/contrib_thunder.rst
```diff
@@ -0,0 +1,468 @@
...
+   :lineno-start: 536
+
+   _register_implementation(prims.transpose, checker=_always_executable, execution_transform=_transpose_prim_transform)
+   _register_implementation(prims.unfold, unfold, checker=_always_executable)
+   _register_implementation(prims.view, view, checker=_always_executable)
+
+the first one registers a new symbol that is directly tied to the ``torch.Tensor.unfold``, and the second
+associates this symbol with ``prims.unfold`` upon execution unless the ``checker`` fails
+(``checker`` is useful, for example, for checking PyTorch versions).
+Having ``checker=_always_executable`` always greenlights this association, and, hence,
```
carmocca, line 240: Somewhat of an offtopic question to your PR:

Why do we run checks on the meta functions for the operators instead of in the checkers?

For instance, imagine we had 2 executors who define the same operator where the first one is stricter about what it allows. If the meta function for the first one does a `utils.check(something)` then it would not give an opportunity to the second executor. As far as I understand, if this is done with a checker then it is supported.

So when do we recommend customizing the meta function vs the checker?

docs/source/advanced/contrib_thunder.rst
```diff
@@ -0,0 +1,468 @@
...
+whenever the ``torch`` executor tries to execute ``prims.unfold``, it executes ``torch.Tensor.unfold``.
+
+The mapping of the ``prims.unfold`` symbol to ``torch.Tensor.unfold`` is very simple since the inputs
+to ``prims.unfold`` can directly be passed to ``torch.Tensor.unfold`` without any additional pre-preprocessing
+(association between ``TensorProxy`` and ``torch.Tensor`` is handled automatically by the ``torch`` executor).
+This is not the case with any operation, however, and sometimes the symbol's interface has to undergo
```
carmocca, line 246: Did you mean?

```suggestion
This is not the case with all operations, however, and sometimes the symbol's interface has to undergo
```

carmocca, line 246: Would this be clearer?

```suggestion
This is not the case with any operation, however, and sometimes the primitive implementation has to undergo
```

docs/source/advanced/contrib_thunder.rst
```diff
@@ -0,0 +1,468 @@
...
+In the previous section we saw an example of adding a primitive operation.
+However, it is not guaranteed that the operation performs as expected.
+We need to test it!
+
+Operators are typically tested by adding an OpInfo for them.
+See `here <https://github.com/pytorch/pytorch/blob/7a192cc51c7172860efd35413acf9a8b9aafd2c9/torch/testing/_internal/opinfo/core.py#L341>`_
+to better understand how OpInfos work.
+OpInfo contains metadata describing an operator, a sample input generator,
+a sample generator for erroneous inputs that is used for testing handling exceptions/meta function correctness, and test directives.
+It’s used to automatically generate a variety of tests, including tests that verify the operator’s behavior is consistent with its reference implementations.
```
carmocca, line 282:
```suggestion
It’s used to automatically generate a variety of tests, most importantly tests that verify the operator’s behavior is consistent with its reference implementations.
```

docs/source/advanced/contrib_thunder.rst
```diff
@@ -0,0 +1,468 @@
...
+It is important to determine whether you need to add ``test_directives`` in order to skip tests or expect failures of tests.
+
+* Skip (``pytest.mark.skip``): Skips are needed when something is not implemented by an executor or for a device.
+* Expected Failures (``pytest.mark.xfail``): Expected failures indicate that an executor has implemented some aspect of an operation but its behavior is incorrect.
+
+An example of OpInfo for ``prims.unfold`` from the `PR #136 <https://github.com/Lightning-AI/lightning-thunder/pull/136>`_
+added to `thunder/tests/opinfos.py <https://github.com/Lightning-AI/lightning-thunder/blob/main/thunder/tests/opinfos.py>`_
+
+.. code-block:: python
+   :emphasize-lines: 1-38
```
carmocca, line 293: Again, this looks odd

docs/source/advanced/contrib_thunder.rst
```diff
@@ -0,0 +1,468 @@
...
+
+We strive for *Thunder* to be of the highest quality possible,
+so it is always a good idea to be very thorough when it comes to testing.
+
+~~~~~~~~~~~~~~~~~~~
+Adding grad support
+~~~~~~~~~~~~~~~~~~~
+Operations are not differentiable by default, unless they are implemented as compositions
+of differentiable operations (related to updating the ``torch`` language. More on that later).
+When an operation is a composition of other operations, we say that this operation is being *decomposable*.
```
carmocca, line 378:
```suggestion
When an operation is a composition of other operations, we say that this operation is *decomposable*.
```

docs/source/advanced/contrib_thunder.rst
```diff
@@ -0,0 +1,468 @@
...
+`thunder/torch/__init__.py <https://github.com/Lightning-AI/lightning-thunder/blob/main/thunder/torch/__init__.py>`_
+is where its implementation is to be placed.
+
+For example, consider the `PR #100, <https://github.com/Lightning-AI/lightning-thunder/pull/100>`_ that adds
+support for the Hardswish activation function.
+The function is implemented in `thunder/torch/__init__.py <https://github.com/Lightning-AI/lightning-thunder/blob/main/thunder/torch/__init__.py>`_
+
+.. code-block:: python
+    :lineno-start: 1211
+    :emphasize-lines: 1-9
```
carmocca, line 441: Another one

docs/source/advanced/contrib_thunder.rst
```diff
@@ -0,0 +1,468 @@
...
+        utils.check(not inplace, lambda: f"hardswish only supports inplace=False", exception_type=NotImplementedError)
+        utils.check(
+            dtypes.is_float_dtype(a.dtype),
+            lambda: f"hardswish only supports floating point dtypes, got {a.dtype}",
+            exception_type=ValueError,
+        )
+        return a * relu6(a + 3) / 6
+
+Note the checks (*Thunder* does not support in-place operations yet) and that ``hardswish`` is a composition
+of the defined in the ``torch`` language ``relu6`` operation and the language context-specific binary operations
```
carmocca, line 454:
```suggestion
Note the checks (*Thunder* does not support in-place operations yet) and that ``hardswish`` is a composition
of the ``relu6`` operation (defined in the ``torch`` language) and the language context-specific binary operations
```

carmocca, line 454: I understand that you mean binary here in the sense that you have two arguments. But for the unaware, it would be better to change the wording here for those who think of the numeral system

carmocca (approved):
Super good! My comments are minor. I don't think there's anything large missing.

For those who are also reviewing, you can see the docs in https://lightning-thunder--149.org.readthedocs.build/en/149/advanced/contrib_thunder.html

docs/source/advanced/contrib_thunder.rst
```diff
@@ -0,0 +1,468 @@
...
+For operations that are part of a class, like the elementwise unary or reduction operations, they often share a common meta function.
+More unique operations, like ``slice``, define their own
+`meta functions <https://github.com/Lightning-AI/lightning-thunder/blob/888b46324462fba70f93d5017bc0d99025f05091/thunder/core/prims.py#L2812>`_.
+
+The actual execution of primitive operations is handled by executors like
+`nvFuser <https://github.com/NVIDIA/Fuser>`_ or
+`PyTorch <https://pytorch.org>`_ – more on that in a moment.
+
+Before adding a primitive, check with the team on its design.
+It is appropriate to add primitives when necessary to describe the semantics of an operation or to improve the numerical accuracy or speed of operations.
```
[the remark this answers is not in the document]
nikitaved: Well, it is, because it is necessary :)

docs/source/advanced/contrib_thunder.rst
nikitaved: But the motivation is there, nor? Prims are simple, do not do promotions/broadcasting and such. Their goal, is, ideally, just compute. But we are not very consistent with that, to be honest. We do have some quite powerful prims.

docs/source/advanced/contrib_thunder.rst
nikitaved: So, I used it to indicate the added by the contributor lines. It is possible to get that highlighted similar to how git shows that. But I would need to manipulate Sphinx a bit more, not an expert there. What do you think, @carmocca . Or do you think that dropping any highlight is better?

docs/source/advanced/contrib_thunder.rst
nikitaved: Sorry, are these not equivalent?

nikitaved: No, I do not think so. The primitive implementation would refer to the executor's interface, right? And we map our Thunder's interface to that of the executor.

docs/source/advanced/contrib_thunder.rst
nikitaved: It is an advanced section after all :) Do you mean the usage of explicit algebraic op names?

docs/source/advanced/contrib_thunder.rst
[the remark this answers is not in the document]
carmocca: I mean would you always add a prim if it makes it faster? What if it's barely faster? This is why I ask if it's that clear.

docs/source/advanced/contrib_thunder.rst
carmocca: Having the core language as a more powerful set of operators makes sense if the other higher-level languages (torch, ...) decompose into clang ops, not just prims.
Is that the case? I was under the impression that this isn't the case.

This is why I never understood the existence of the clang. Is it in case people want to write code with these operators directly instead of torch?

docs/source/advanced/contrib_thunder.rst
carmocca: Dropping it is probably the best option in the cases where it's the whole snippet. If you want to highlight only a subset then I would keep it.

docs/source/advanced/contrib_thunder.rst
carmocca: I understand this but my question is, why do we even let it claim symbols whose inputs are not going to be accepted by the meta function? AFAIK the checker also has access to the inputs.

Maybe a question for @mruberry too

docs/source/advanced/contrib_thunder.rst
carmocca: My non-english-native brain prefers "all", but feel free to keep yours

docs/source/advanced/contrib_thunder.rst
nikitaved: Well, we can have high-level functions which makes our lives easier and not bound to the torch interface.

carmocca: Let's mention that here then, that is valid motivation

nikitaved: I did re-read the sections. I believe they explain things well enough already. So, we have simple prims, and we have the torch lang that is used directly to be mapped into when seeing a torch op. Clang and prims are used to make that map happen. This is it there is to it. All other degrees of freedom are following from the "definitions".

docs/source/advanced/contrib_thunder.rst
nikitaved: Killed off.

docs/source/advanced/contrib_thunder.rst
nikitaved: I will update outline the structure and how it is supposed to be read.

nikitaved:
Thank you so much, @carmocca , for your review! Let me know if there is something else you would like to point out!

t-vi (approved):
Love it! Let's merge it and then we can immprove later  Thank you @nikitaved  @carmocca

PR #150, opened 2024-04-09
Add `single_op_executor()`, fix a lot of random things.

apaz-cli:
Still have to figure out how to properly unproxify and reproxify. But, consider the following.


thunder/extend/__init__.py
```diff
@@ -264,8 +281,44 @@ def register_implementation(
             symbol=op, checker=checker, execution_transform=execution_transform, grad_transform=grad_transform
         )
 
-        id = sym_or_id.id if isinstance(sym_or_id, Symbol) else sym_or_id
-        self.implmap[id] = impl
+        _id = sym_or_id.id if isinstance(sym_or_id, Symbol) else sym_or_id
+        self.implmap[_id] = impl
+
+
+class SingleOpExecutor(OperatorExecutor):
```
t-vi, line 288: TBH, I'd prefer this to be a factory function rather than a subclass. It is one thing to have a convenience function (which will only affect the caller) and another to have a new datatype (which might not have grave repercussions, but has a much wider impact radius).

apaz-cli: Yeah, that makes sense to me. Come to think of it I largely feel the same. Object hierarchies are a path to pain.

thunder/__init__.py
```diff
@@ -347,13 +347,23 @@ def jit(
...
         additional_transforms = []
 
+    # Resolve names of executors
+    exc_list = executors if executors is not None else ()
+    executors = []
+    for e in exc_list:
+        if isinstance(e, str):
+            e = extend.get_executor(e)
+        if not isinstance(e, Executor):
+            raise TypeError(f"Expected an Executor or the name of one, instead got: {e}")
```
carmocca, line 357: This change is very welcome. I wrote my own version in https://github.com/Lightning-AI/litgpt/blob/main/extensions/thunder/strategies/utils.py#L7

It would be useful that you validate all the executors at once, as I do in that link. Having to rerun multiple times is inconvenient

Please also steal my test: https://github.com/Lightning-AI/litgpt/blob/main/tests/test_thunder_fsdp.py#L35-L44

thunder/extend/__init__.py
```diff
@@ -137,7 +143,7 @@ def __init__(self, name: Hashable, *, version: None | Any = None):
     # Whalley (https://dl.acm.org/doi/pdf/10.1145/186025.186103) to isolate
     # compiler bugs.
     #
-    # A FusionExecutor keeps track of its own optimization fuel as the number
+    # A FusionExecutor keeps track of itsImplInfo own optimization fuel as the number
```
carmocca, line 146: This doesnt look right

thunder/extend/__init__.py
```diff
@@ -181,12 +187,19 @@ def _meta(*args):
...
+            bsym._call_ctx = {name: fn}
 
         sym = Symbol(name=name, meta=_meta, is_fusion=True, _bind_postprocess=_bind_postprocess, executor=self)
         return sym.bind(*inputs, output=outputs)
 
 
+@run_once
+def warn_default_meta(opname: str, exname: Hashable):
+    warnings.warn(
+        f"Meta function has not been specified for operator {opname} in exexutor {exname}. Defaulting to running the operator with example tensors to determine its output metadata."
```
mruberry, line 199: "exexutor"

thunder/extend/__init__.py
apaz-cli: You're right, it doesn't. Fixed.

thunder/extend/__init__.py
apaz-cli: Removed the whole function, ripped out default meta, we can talk about it later.

thunder/__init__.py
apaz-cli: I added the test, and another to make sure it worked with custom executors. I also made `thunder.jit()` check if there were any duplicate entries in the executor list. Turns out we had a few bugs lurking.

carmocca (approved):
LGTM!

Issue #138, opened 2024-04-05
Support `is_cuda`


kshitij12345:
Something like the following should work
```python
import thunder
import torch

def foo(x):
    if not x.is_cuda:
        x = x.to('cuda')
    return x * x

x = torch.randn(3, device='cpu')
jit_foo = thunder.jit(foo)
o = jit_foo(x)

print(thunder.last_traces(jit_foo)[-1])
```

Above fails with
```python
  File "/home/kkalambarkar/lightning-thunder/thunder/core/proxies.py", line 1234, in __getattr__
    method: None | Callable = resolve_method(attr, self)
  File "/home/kkalambarkar/lightning-thunder/thunder/core/langctxs.py", line 68, in resolve_method
    method: Callable = ctx.get_method(id, *args, **kwargs)
  File "/home/kkalambarkar/lightning-thunder/thunder/torch/langctx.py", line 40, in get_method
    raise AttributeError(f"The {self.name} language context has no method {id}")
AttributeError: The torch language context has no method is_cuda
```

shaharelys:
Hey @kshitij12345! I'm a new contributor here on Thunder. Would it be okay if I look into this and try to come up with a solution?

Thx,
Shahar

kshitij12345:
@shaharelys Ofcourse, that'd be really great. Do ping here if you require any pointers.

cc: @nikitaved

k223kim:
@kshitij12345 Hi! I am also a new contributor on Thunder :) I had a quick conversation with @shaharelys regarding this issue and he was ok with me taking this issue. Just wanted to let you know that I'll be working on it! (Thank you Shahar!)
Best,
Kaeun

k223kim:
Hello team! I just wanted to check if I am on the right track regarding this issue. I have read the comments of [this issue](https://github.com/Lightning-AI/lightning-thunder/issues/64) and tried to do something similar. Just a heads up that I may make trivial mistakes and would love to learn as I go through this. It'll be awesome if anyone can take a look and provide me some guidance! :)

1.  I have checked torch.is_cuda implementation and noticed a couple of things:
- I noticed that this can be as simple as checking the tensor's device something like the following (from [pytorch](https://github.com/pytorch/pytorch/blob/1ea2f1eaa19176b8b5b778bb317203dc7f4fd3dc/torch/storage.py#L81))
```
    @property
    def is_cuda(self):
        return self.device.type == 'cuda'
```
[...]

PR #162, opened 2024-04-11
Added `is_cuda` property support

k223kim:
<details>

</details>

## What does this PR do?

Fixes [#138](https://github.com/Lightning-AI/lightning-thunder/issues/138). It implements `is_cuda` as a property.

## PR review

The implementation was discussed thoroughly the issue thread, however, I am not sure if I have written the test code in the desired way. This PR currently fails in some tests below, yet, I am not sure how I can resolve `torch.cuda.OutOfMemoryError: CUDA out of memory` error in Azure Pipeline. Also, I was not sure which part of the doc had to be updated. Any feedback and comments regarding the implementation, test code, or docs would be much appreciated.

## Did you have fun?

Absolutely! 🚀


thunder/tests/opinfos.py
```diff
@@ -542,6 +542,23 @@ def _fn(x):
     return _fn
 
 
+@wraps(torch.Tensor.is_cuda)
+def _is_cuda_torch(x: torch.Tensor):
+    # Should Number be accepted?
```
mruberry, line 547: Just accept `torch.Tensor` seems fine. 

Nit: the return value of this function could be annotated as `bool`

mruberry: This is a nice helper for the OpInfo

thunder/tests/opinfos.py
```diff
@@ -542,6 +542,23 @@ def _fn(x):
...
+    return x.is_cuda
+
+
+is_cuda_opinfo = OpInfo(
+    ltorch.is_cuda,
+    sample_input_generator=partial(elementwise_unary_generator, supports_numbers=False),
+    torch_reference=_is_cuda_torch,
+    dtypes=(datatypes.all_dtypes),
+    # Does Pytorch is_cuda support this for all dtypes?
```
mruberry, line 556: It should

thunder/torch/__init__.py
```diff
@@ -171,6 +179,11 @@ def fn_(idx: int | None = None):
     return fn_
 
 
+@torchsymbol(torch.Tensor.is_cuda, is_property=True, id="torch.is_cuda")
+def is_cuda(a: TensorLike, /) -> bool:
+    return a._device.devicetype is devices.DeviceType.CUDA
```
mruberry, line 184: It's OK to access the device property directly — `a.device.devicetype`

thunder/torch/langctx.py
```diff
@@ -36,9 +37,13 @@ def get_method(self, id: str, *args, **kwargs) -> Callable:
...
-            if method is None:
-                raise AttributeError(f"The {self.name} language context has no method {id}")
-            return method
+            property: None | Callable = _property_name_to_fn_map.get(id, None)
+            if method is None and property is None:
+                raise AttributeError(f"The {self.name} language context has no method or attribute {id}")
+            if method:
+                return method
+            else:
+                return property(inps[0])
```
mruberry, line 46: This looks good but don't use the name `property` because it conflicts with Python's `property` builtin:

https://docs.python.org/3/library/functions.html#property

That's why the name `property` is colored differently here

Another name might be `prop`

thunder/tests/opinfos.py
```diff
@@ -542,6 +542,23 @@ def _fn(x):
...
 
 
+@wraps(torch.Tensor.is_cuda)
+def _is_cuda_torch(x: torch.Tensor):
+    # Should Number be accepted?
+    return x.is_cuda
+
+
+is_cuda_opinfo = OpInfo(
+    ltorch.is_cuda,
```
mruberry, line 552: This makes a lot of sense, but what if a helper that access the property was used here and for the `torch_reference`? That way the OpInfo would test that the property was registered properly. Something like

```
def _is_cuda_property(t) -> bool:
  return t.is_cuda
```

and then

```
is_cuda_opinfo = OpInfo(
  _is_cuda_property,
  ...,
  torch_reference=_is_cuda_property,
  ...
)
```

thunder/torch/langctx.py
k223kim: Oh, I was not aware of that. Thanks! Will update it :)

thunder/tests/opinfos.py
```diff
@@ -542,6 +542,23 @@ def _fn(x):
...
+    sample_input_generator=partial(elementwise_unary_generator, supports_numbers=False),
+    torch_reference=_is_cuda_torch,
+    dtypes=(datatypes.all_dtypes),
+    # Does Pytorch is_cuda support this for all dtypes?
+)
+
+elementwise_unary_ops.append(is_cuda_opinfo)
```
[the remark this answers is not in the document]
k223kim: This is a good point. I was hesitant to add this to `elementwise_unary_ops` earlier but was not sure how this should be handled in `opinfos.py` Thanks for the detailed feedback and explanation!

thunder/tests/opinfos.py
k223kim: I am trying to wrap my head around your comment but I am not sure if I agree with your point. Is the main purpose of `OpInfo` to test the registration of the property? When referring to other `OpInfo`s in `tests/opinfos.py`, I am seeing that `op` usually refers to the `ltorch` implementation. Is there a reason why you would want this particular case to have the helper function to be assigned for both `op` and `torch_reference`?
Maybe I have misunderstood the point of `OpInfo` in the first place..
Let me know if I misunderstood anything and I would love to learn more about your opinion!

k223kim:
Thanks so much for taking your time to provide such detailed comments @mruberry ! I have updated based on your comments and resolved most of them. Would update the rest once I have a better understanding of your comment :)

thunder/tests/opinfos.py
mruberry: This is certainly an unusual case. I think in this case we are testing that the property works as expected. This is unlike existing OpInfos, which are comparing one operation with another. 

Because the property is accessed the same way in thunder and in PyTorch, we can use the same testing helper to acquire it.

k223kim: That makes sense! Thanks for the detailed explanation. Will update this as well :)

k223kim: Hmm, simply replacing `ltorch.is_cuda` to the helper function is throwing an error. Let me look into this more carefully!

thunder/tests/test_ops.py
```diff
@@ -65,13 +65,21 @@ def test_core_vs_torch_consistency(op, device: str, dtype: dtypes.dtype, executo
...
 
         tfn: Callable
-        tfn = thunder.functional.jit(
-            op.op,
-            executors=executor.executors_list(),
-            interpretation="python interpreter",
-            cache="no caching",
-            disable_torch_autograd=True,
-        )
+        # NOTE: this is needed when testing properties
```
mruberry, line 68: This is actually just skipping testing with thunder. The bug is somewhere else, and the error is still occurring for other tests

thunder/tests/opinfos.py
```diff
@@ -542,6 +542,29 @@ def _fn(x):
     return _fn
 
 
+#
+# Tensor Property OpInfos
+#
+tensor_properties: list[OpInfo] = []
+
+
+@wraps(torch.Tensor.is_cuda)
```
mruberry, line 551: Don't wrap the property

mruberry (commented):
Hey @k223kim — would you try reverting the changes to test_ops.py and no longer wrapping the `_is_cuda_torch` helper?

k223kim:
Ah, thanks so much @mruberry ! 
I realized that it was throwing an error when performing `thunder.functional.jit`. As default, it is using a python interpreter which goes through `_eager_unpacking_interpreter`, `get_siginfo`, and therefore, `inspect.signature` which is where the error occurs. Therefore, as you have mentioned, unwrapping the helper function allows it to successfully perform `inspect.signature` and resolves the error.

mruberry (approved):
Cool! Thanks @k223kim!

t-vi:
Thank you @k223kim @mruberry

Issue #156, opened 2024-04-10
Benchmark targets on test_nanogpt_cross_entropy_grad has some import issue


xwang233:
## 🐛 Bug

Benchmark targets on test_nanogpt_cross_entropy_grad has some import issue

### To Reproduce

```python
root@8d345ed01185:/opt/pytorch/lightning-thunder# pytest -vvvs thunder/benchmarks/targets.py::test_nanogpt_cross_entropy_grad[thunder+apex-grad]
============================================================================================== test session starts ==============================================================================================
platform linux -- Python 3.10.12, pytest-8.1.1, pluggy-1.4.0 -- /usr/bin/python3
cachedir: .pytest_cache
hypothesis profile 'default' -> database=DirectoryBasedExampleDatabase(PosixPath('/opt/pytorch/lightning-thunder/.hypothesis/examples'))
Test order randomisation NOT enabled. Enable with --random-order or --random-order-bucket=<bucket_type>
benchmark: 4.0.0 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
rootdir: /opt/pytorch/lightning-thunder
configfile: pyproject.toml
plugins: cov-4.1.0, hypothesis-6.100.0, random-order-1.1.1, timestamper-0.0.10, timeout-2.2.0, xdist-3.5.0, shard-0.1.2, benchmark-4.0.0
timeout: 900.0s
timeout method: signal
timeout func_only: False
collected 1 item
Running 1 items in this shard: thunder/benchmarks/targets.py::test_nanogpt_cross_entropy_grad[thunder+apex-grad]

[2024-04-10 21:48:06] thunder/benchmarks/targets.py::test_nanogpt_cross_entropy_grad[thunder+apex-grad] FAILED

=================================================================================================== FAILURES ====================================================================================================
______________________________________________________________________________ test_nanogpt_cross_entropy_grad[thunder+apex-grad] _______________________________________________________________________________

benchmark = <pytest_benchmark.fixture.BenchmarkFixture object at 0x7fe26c8f0cd0>
executor = functools.partial(<function thunder_grad_transform at 0x7fe26ca5ecb0>, compile_fn=<function thunder_apex_executor at 0x7fe26cad8b80>)

    @pytest.mark.parametrize(
        "executor,",
        (grad_executors + apex_grad_executors),
[...]

PR #168, opened 2024-04-12
Fix name 'thunder' is not defined in test_nanogpt_cross_entropy_grad[thunder+apex-grad]

IvanYashchuk:
Fixes https://github.com/Lightning-AI/lightning-thunder/issues/156


carmocca (approved):
Thanks

PR #161, opened 2024-04-11
Enable fullgraph with torch.compile

carmocca:
## What does this PR do?

Aims to enable `fullgraph=True` by default. This is useful for diagnosing unnecessary recompilation.

Since `torch.compile` has more operator coverage than we do right now, it should be safe to do this. We can revisit this decision or make it configurable in the future if that changes.

Most notably, our `no_autocast` context manager needs to be replaced because dynamo only understands the specific `torch.autocast` implementation. This also uncovered excessive recompilation due to us adding these context managers to all traces.

Closes #169

cc @borda @apaz-cli

PR #140, opened 2024-04-05
Complete `torch.compile` executor

carmocca:
## What does this PR do?

Adds an instance of `TorchCompileExecutor` named `torch_compile_Ex` that registers all of `pytorch_executor`'s and `sdpa_ex`'s operators.

**The rename of the existing executor `torch_compile_executor` to `torch_compile_cat_ex` breaks backwards compatibility.**

Playground script with litgpt:

```python
from litgpt import GPT
import thunder
import torch

from thunder import pytorch_executor
from thunder.executors.torch_compile import torch_compile_ex

with torch.device("cuda"):
    model = GPT.from_name("Llama-2-7b-hf", n_layer=1)

model = thunder.jit(model, executors=[torch_compile_ex])
x = torch.randint(model.max_seq_length, (2, 5), device="cuda")
y = model(x)

forward_trace = thunder.last_traces(model)[-1].python()
print(forward_trace)
assert "TorchCompile" in str(forward_trace)
```

Output:

```python
@torch.no_grad()
@no_autocast
def augmented_forward_fn(idx, t_lm_head_weight, t_transformer_h_0_attn_attn_weight, t_transformer_h_0_attn_proj_weight, t_transformer_h_0_mlp_fc_1_weight, t_transformer_h_0_mlp_fc_2_weight, t_transformer_h_0_mlp_proj_weight, t_transformer_h_0_norm_1_weight, t_transformer_h_0_norm_2_weight, t_transformer_ln_f_weight, t_transformer_wte_weight, tos1, t_sin):
  # idx: "cuda:0 i64[2, 5]"
  # t_lm_head_weight: "cuda:0 f32[32000, 4096]"
  # t_transformer_h_0_attn_attn_weight: "cuda:0 f32[12288, 4096]"
  # t_transformer_h_0_attn_proj_weight: "cuda:0 f32[4096, 4096]"
  # t_transformer_h_0_mlp_fc_1_weight: "cuda:0 f32[11008, 4096]"
  # t_transformer_h_0_mlp_fc_2_weight: "cuda:0 f32[11008, 4096]"
  # t_transformer_h_0_mlp_proj_weight: "cuda:0 f32[4096, 11008]"
  # t_transformer_h_0_norm_1_weight: "cuda:0 f32[4096]"
  # t_transformer_h_0_norm_2_weight: "cuda:0 f32[4096]"
  # t_transformer_ln_f_weight: "cuda:0 f32[4096]"
[...]

PR #169, opened 2024-04-12
Enable fullgraph with torch.compile (v2)

carmocca:
## What does this PR do?

Alternative implementation to #161

### DIFF
--- a/thunder/executors/torch_compile.py
+++ b/thunder/executors/torch_compile.py
@@ -1,8 +1,10 @@
+import operator
 import time
 from collections.abc import Callable, Hashable
 from typing import Any
 
 import torch
+from lightning_utilities import compare_version
 
 from thunder.core import prims, utils
 from thunder.core.proxies import Proxy, unvariableify, Variable
@@ -11,6 +13,8 @@ from thunder.core.trace import from_trace, TraceCtx, TraceProvenance
 
 from thunder.executors.utils import Region
 from thunder.extend import FusionExecutor, register_executor
+
+_TORCH_GREATER_EQUAL_2_3 = compare_version("torch", operator.ge, "2.3.0", use_base_version=True)
 
 
 def to_torch_translator(bsym: BoundSymbol) -> Callable:
@@ -75,23 +79,29 @@ def make_compiled(
     # TODO: issue "Try using _transform_for_operator_executor_execution for
     # torch.compile executor"
     torch_trace = trace(inline_trace=False)(torch_interpreted_func, *sorted_unique_inputs)
-    compiled_func = torch.compile(torch_trace.python_callable())
+    # decorators are not added because dynamo graph breaks on our custom `no_autocast` decorator.
+    # so far it only supports their native `autocast` implementation
+    trace_callable = torch_trace.python_callable(include_decorators=False)
+    trace_callable = torch.autocast(device_type="cpu", enabled=False, cache_enabled=False)(trace_callable)
+    trace_callable = torch.autocast(device_type="cuda", enabled=False, cache_enabled=False)(trace_callable)
+    trace_callable = torch.no_grad()(trace_callable)
+    compiled_func = torch.compile(trace_callable, fullgraph=True)
 
+    # The default is 8. For each of `@torch.no_grad(), and `torch.autocast(device_type="cpu"|"cuda")` torch.compile
+    # create caches with a guard for the wrapped function. Since the torch.compile caches are per code object, not
+    # frame, all the dynamic copies of these context managers share the same code cache.
+    # Since Thunder generates many traces, all of them annotated with these context managers, we need to increase
+    # the limit here. Alternatively we could pull out the context managers outside of the `torch.compile` region
+    @torch._dynamo.config.patch("cache_size_limit", 64)
     def compiled_func_wrapper(*args):
-        # PyTorch 2.1 doesn't have this attribute
-        if getattr(torch._dynamo.eval_frame, "guarded_backend_cache", None) is None:
+        if _TORCH_GREATER_EQUAL_2_3:
             return compiled_func(*args)
 
         orig = getattr(torch._dynamo.eval_frame.guarded_backend_cache, "skip_backend_check_for_run_only_mode", None)
         try:
-            # TODO: Remove this hack
-            # Dynamo doesn't recreate a guard for the compiled function called
-            # from the backward thread. This is a problem because the guard is
-            # created with the forward thread ID, and the guard is not valid
-            # for the backward thread.
-            # Issue filed: https://github.com/pytorch/pytorch/issues/114674
-            # We should be able to remove this hack once we're sure that the
-            # above fix has propagated to all supported PyTorch releases.
+            # Dynamo doesn't recreate a guard for the compiled function called from the backward thread. This is a
+            # problem because the guard is created with the forward thread ID, and the guard is not valid
+            # for the backward thread. Issue filed: https://github.com/pytorch/pytorch/issues/114674
             torch._dynamo.eval_frame.guarded_backend_cache.skip_backend_check_for_run_only_mode = True
             return compiled_func(*args)
         finally:
--- a/thunder/core/trace.py
+++ b/thunder/core/trace.py
@@ -306,7 +306,7 @@ class TraceCtx:
     # TODO issue "Add type annotations to Python function produced by traces"
     #   Consider extending the signature with type information, in particular the
     #   the type information of the return value might be interesting
-    def python(self, *, print_depth: int = 1) -> str:
+    def python(self, *, print_depth: int = 1, include_decorators: bool = True) -> str:
         token = set_tracectx(self)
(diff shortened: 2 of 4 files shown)


thunder/executors/torch_compile.py
```diff
@@ -75,23 +79,29 @@ def torch_interpreted_func(*args):
     # TODO: issue "Try using _transform_for_operator_executor_execution for
     # torch.compile executor"
     torch_trace = trace(inline_trace=False)(torch_interpreted_func, *sorted_unique_inputs)
-    compiled_func = torch.compile(torch_trace.python_callable())
-
+    # decorators are not added because dynamo graph breaks on our custom `no_autocast` decorator.
+    # so far it only supports their native `autocast` implementation
+    trace_callable = torch_trace.python_callable(include_decorators=False)
+    trace_callable = torch.autocast(device_type="cpu", enabled=False, cache_enabled=False)(trace_callable)
+    trace_callable = torch.autocast(device_type="cuda", enabled=False, cache_enabled=False)(trace_callable)
```
t-vi, line 86: I wonder if we could replace no_autocast by just these two applications of torch.autocast in a function 
https://github.com/Lightning-AI/lightning-thunder/blob/71fc8cda4e42b818372b6e9f1c99c9cc3a5c2e38/thunder/executors/torchex.py#L139

so
```
def no_autocast(fn):
    fn = torch.autocast(device_type="cpu", enabled=False, cache_enabled=False)(fn)
    fn = torch.autocast(device_type="cuda", enabled=False, cache_enabled=False)(fn)
    return fn
```

and then not having to special case anythinghere.

carmocca: Ah that's a good suggestion. Trying in https://github.com/Lightning-AI/lightning-thunder/pull/169/commits/d5614568fe833fd2ddcee5cffa39867668e4cd58.

carmocca: Actually, I need to partially revert this because even though I increased the cache size limit, this still causes extreme recompilation:

```python
V0422 14:59:42.353000 139978690109440 torch/_dynamo/guards.py:2122] [__recompiles] Recompiling function inner in /home/carlos/nightly-env/lib/python3.10/site-packages/torch/_dynamo/external_utils.py:34
V0422 14:59:42.353000 139978690109440 torch/_dynamo/guards.py:2122] [__recompiles]     triggered by the following guard failure(s):
V0422 14:59:42.353000 139978690109440 torch/_dynamo/guards.py:2122] [__recompiles]     - ___check_obj_id(L['fn'].__closure__[1].cell_contents.__closure__[0].cell_contents, 139974325237696)
V0422 14:59:42.353000 139978690109440 torch/_dynamo/guards.py:2122] [__recompiles]     - ___check_obj_id(L['fn'].__closure__[1].cell_contents.__closure__[0].cell_contents, 139974168693152)
V0422 14:59:42.353000 139978690109440 torch/_dynamo/guards.py:2122] [__recompiles]     - ___check_obj_id(L['fn'].__closure__[1].cell_contents.__closure__[0].cell_contents, 139974167264512)
V0422 14:59:42.353000 139978690109440 torch/_dynamo/guards.py:2122] [__recompiles]     - ___check_obj_id(L['fn'].__closure__[1].cell_contents.__closure__[0].cell_contents, 139974167080400)
V0422 14:59:42.353000 139978690109440 torch/_dynamo/guards.py:2122] [__recompiles]     - ___check_obj_id(L['fn'].__closure__[1].cell_contents.__closure__[0].cell_contents, 139974706455760)
V0422 14:59:42.353000 139978690109440 torch/_dynamo/guards.py:2122] [__recompiles]     - ___check_obj_id(L['fn'].__closure__[1].cell_contents.__closure__[0].cell_contents, 139974169152528)
V0422 14:59:42.353000 139978690109440 torch/_dynamo/guards.py:2122] [__recompiles]     - ___check_obj_id(L['fn'].__closure__[1].cell_contents.__closure__[0].cell_contents, 139974325238272)
```

So these context managers should definitely be left outside of the `torch.compile` region

t-vi (approved):
Looks great! Thank you @carmocca

thunder/executors/torchex.py
```diff
@@ -145,16 +145,10 @@ def _device_put_transform(a: TensorProxy, device: devices.Device) -> TensorProxy
...
-        torch.set_autocast_enabled(False)
-        torch.set_autocast_cpu_enabled(False)
-
-    def __exit__(self, exc_type, exc_value, exc_traceback):
-        torch.set_autocast_enabled(self.was_cuda_enabled)
-        torch.set_autocast_cpu_enabled(self.was_cpu_enabled)
+def no_autocast(fn):
+    fn = torch.autocast(device_type="cpu", enabled=False, cache_enabled=False)(fn)
+    fn = torch.autocast(device_type="cuda", enabled=False, cache_enabled=False)(fn)
+    return fn
```
IvanYashchuk, line 151: Could you add a test to prevent anyone from changing to a previous version?

IvanYashchuk, line 151: Would the following work equivalently but with the effect that there would be zero change for the notebooks?
```py
def no_autocast():
    def decorator(fn):
        fn = torch.autocast(device_type="cpu", enabled=False, cache_enabled=False)(fn)
        fn = torch.autocast(device_type="cuda", enabled=False, cache_enabled=False)(fn)
        return fn
    return decorator
```

thunder/executors/torch_compile.py
```diff
@@ -75,23 +80,25 @@ def torch_interpreted_func(*args):
...
-    compiled_func = torch.compile(torch_trace.python_callable())
+    trace_callable = torch_trace.python_callable(include_decorators=False)
+    compiled_func = torch.compile(trace_callable, fullgraph=True)
+    # For each of `@torch.no_grad(), and `torch.autocast(device_type="cpu"|"cuda")` torch.compile
+    # create caches with a guard for the wrapped function. Since the torch.compile caches are per code object, not
+    # frame, all the dynamic copies of these context managers share the same code cache.
+    # Since Thunder generates many traces, all of them annotated with these context managers, we must put these context
+    # managers outside the `torch.compile` region
+    compiled_func = no_autocast(compiled_func)
+    compiled_func = torch.no_grad()(compiled_func)
```
IvanYashchuk, line 91: This is not needed because these compiled functions would be executed as part of the larger trace with these contexts set up.

thunder/executors/torch_compile.py
```diff
@@ -75,23 +80,25 @@ def torch_interpreted_func(*args):
     # TODO: issue "Try using _transform_for_operator_executor_execution for
     # torch.compile executor"
     torch_trace = trace(inline_trace=False)(torch_interpreted_func, *sorted_unique_inputs)
-    compiled_func = torch.compile(torch_trace.python_callable())
+    trace_callable = torch_trace.python_callable(include_decorators=False)
+    compiled_func = torch.compile(trace_callable, fullgraph=True)
```
IvanYashchuk, line 84: What's the benefit of `fullgraph=True`? Should we use some other flags like reduce-overhead?

thunder/executors/torchex.py
carmocca: Let me try. Compiling a full LLaMA (which is what required this changes) is excessive for CI but I could try compiling a small sized LLaMA and try to assert that there was no `torch.compile()` recompilation

carmocca: I can try this but from my experience, `torch.compile` has poor support for generic decorators

thunder/executors/torch_compile.py
carmocca: You are correct. I kept them to not change the functionality of the trace. I only wanted to exclude them from the `torch.compile` region

thunder/executors/torch_compile.py
carmocca: My end goal is getting #140 to work. I was facing issues with that + LitGPT and my first reaction was to try to get fullgraph compilation to work.

Since torch.compile has good operator coverage, it should be safe to set this today. That should also enable more opportunities for optimization on their side and surface decisions in our implementation that disable it.

In the future I would suggest that we expose fullgraph and the other backend flags (like reduce-overhead) to the thunder user

carmocca: Opened #281

thunder/executors/torchex.py
carmocca: It does work. Do you want me to revert this change?

IvanYashchuk:
Does this PR interfere with the decorator added in https://github.com/Lightning-AI/lightning-thunder/pull/48?

carmocca:
What do you mean by "interfere" precisely? I don't foresee people using both the torch.compile and TE engine executors together soon. But as long as we are not torch.compiling the TE autocast context manager it should be okay.

thunder/executors/torchex.py
IvanYashchuk: No, I was checking my understanding.

PR #180, opened 2024-04-12
Fix Tensor indexing with `None`

jjsjann123:
## What does this PR do?

Fix cases where we have
```
# tensor = torch.tensor(...)
tensor[None]
```

### DIFF
--- a/thunder/clang/__init__.py
+++ b/thunder/clang/__init__.py
@@ -566,7 +566,7 @@ def _basic_indexing(a: TensorLike, /, key) -> TensorLike:
     specified_slices = 0
     ellipsis_idx = None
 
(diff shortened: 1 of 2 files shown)


jjsjann123:
For more context here is where it's coming from: https://github.com/huggingface/diffusers/blob/2523390c269a37209fbb96359d494369aff89361/src/diffusers/models/unets/unet_2d_condition.py#L927

tfogal (approved):
The semantics of tensor indexing are certainly unique!
Thanks for the fix.

jjsjann123:
Yeah, numpy also has this, I'm assuming that's where it came from... :exploding_head:

carmocca (approved):
Neat

mruberry:
fyi @nikitaved

Issue #175, opened 2024-04-12
`TensorBase.cuda`


tfogal:
## 🚀 Feature

Implement [Tensor.cuda](https://pytorch.org/docs/stable/generated/torch.Tensor.cuda.html#torch.Tensor.cuda) that returns a cuda-backed copy of the given tensor.

### Motivation

NeMo [text-to-image model](https://github.com/NVIDIA/NeMo/blob/main/examples/multimodal/text_to_image/imagen/imagen_training.py). It's plausible that the source tensor used in the model is a GPU tensor already, so we might be able to get by with just returning a tensor copy without worrying about movement between devices.

PR #181, opened 2024-04-15
add cuda method

kshitij12345:
Fixes https://github.com/Lightning-AI/lightning-thunder/issues/175

### DIFF
--- a/thunder/tests/opinfos.py
+++ b/thunder/tests/opinfos.py
@@ -2697,6 +2697,12 @@ cuda_opinfo = OpInfo(
     sample_input_generator=cuda_sample_generator,
     error_input_generator=cuda_error_generator,
     torch_reference=torch.Tensor.cuda,
+    test_directives=(
+        DecorateInfo(
+            pytest.mark.skip,
+            active_if=not torch.cuda.is_available(),
+        ),
+    ),
 )
 data_movement_ops.append(cuda_opinfo)
 
--- a/thunder/tests/opinfos.py
+++ b/thunder/tests/opinfos.py
@@ -2668,6 +2668,39 @@ to_opinfo = OpInfo(
 data_movement_ops.append(to_opinfo)
 
 
+def cuda_sample_generator(op, device, dtype, requires_grad, **kwargs):
+    make = partial(make_tensor, device=device, dtype=dtype, requires_grad=requires_grad)
+
+    # None
+    # cpu -> cuda
+    # default cuda -> default cuda (no-op)
+    yield SampleInput(make(4, 4))
+
+    # Explictly pass device
+    # default cuda -> default cuda
+    if torch.device(device).type == "cuda":
+        yield SampleInput(make(4, 4), device)
+
+
+def cuda_error_generator(op, device, **kwargs):
+    make = partial(make_tensor, device="cpu", dtype=torch.float, requires_grad=False)
+
+    err_msg = "`non_blocking==True` is currently not supported."
+    yield SampleInput(make(3, 3), non_blocking=True), RuntimeError, err_msg
+
+    err_msg = "Invalid device cpu, must be cuda device"
+    yield SampleInput(make(3, 3), device="cpu"), RuntimeError, err_msg
+
+
+cuda_opinfo = OpInfo(
+    ltorch.cuda,
+    sample_input_generator=cuda_sample_generator,
+    error_input_generator=cuda_error_generator,
+    torch_reference=torch.Tensor.cuda,
+)
+data_movement_ops.append(cuda_opinfo)
+
+
 def type_as_sample_generator(op, device, dtype, requires_grad, **kwargs):
     make = partial(make_tensor, low=-1, high=1, device=device, dtype=dtype, requires_grad=requires_grad)
 
--- a/thunder/tests/test_grad.py
+++ b/thunder/tests/test_grad.py
@@ -36,6 +36,8 @@ op_skip = {
     "embedding",
     "index_put",
     "batch_norm",
+    # Requires CUDA runtime to be available (fails on CPU only runtimes).
+    "cuda",
 }
 
 # Don't rely on the generated list of supported ops.
--- a/thunder/tests/test_grad.py
+++ b/thunder/tests/test_grad.py
@@ -36,9 +36,11 @@ op_skip = {
     "embedding",
     "index_put",
     "batch_norm",
-    # Requires CUDA runtime to be available (fails on CPU only runtimes).
-    "cuda",
 }
+
+if not torch.cuda.is_available():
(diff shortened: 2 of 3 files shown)


thunder/tests/opinfos.py
```diff
@@ -2668,6 +2668,45 @@ def to_sample_generator(op, device, dtype, requires_grad, **kwargs):
...
+    # default cuda -> default cuda (no-op)
+    yield SampleInput(make(4, 4))
+
+    # Explictly pass device
+    # default cuda -> default cuda
+    if torch.device(device).type == "cuda":
+        yield SampleInput(make(4, 4), device)
+
+
+def cuda_error_generator(op, device, **kwargs):
```
mruberry, line 2685: Nice error cases

thunder/torch/__init__.py
```diff
@@ -287,6 +287,30 @@ def to(
     return a
 
 
+@torchsymbol(torch.Tensor.cuda, is_method=True)
+def cuda(a: TensorProxy, /, device=None, non_blocking: bool = False, memory_format: None | torch.memory_format = None):
+    # Modeled similar to PyTorch:
+    # https://github.com/pytorch/pytorch/blob/e3ac61587aa368c613ef01df1f328a396b64cd5d/tools/autograd/templates/python_variable_methods.cpp#L496-L501
+    # If `device` is None, this function defaults `device` to current CUDA device
```
mruberry, line 294: Nice comment

thunder/torch/__init__.py
```diff
@@ -287,6 +287,30 @@ def to(
...
+    # and delegates actual data-movement and layout ordering to `Tensor.to`.
+
+    # NOTE: `Tensor.to` doesn't model `non_blocking` currently.
```
mruberry, line 297: What about...

```
def cuda(a: TensorProxy, /, device: None | DeviceLike=None, **kwargs):
  if device is None:
    cuda_idx: int = torch.cuda.current_device()
    device: devices.Device = devices.Device(devices.DeviceType.CUDA, cuda_idx)

  return to(a, device=device, **kwargs)
```

and let `to` handle the `non_blocking` error message (or not)?

kshitij12345: `to` doesn't handle `non_blocking` currently.

https://github.com/Lightning-AI/lightning-thunder/blob/139cc2246b87f3e58d80b13897286a61d55826d8/thunder/torch/__init__.py#L239-L252

And leads to error below which seems to be confusing.
```python
  File "/home/kkalambarkar/lightning-thunder/thunder/core/langctxs.py", line 124, in _fn
    result = fn(*args, **kwargs)
TypeError: to() got an unexpected keyword argument 'non_blocking'
```

kshitij12345:
@mruberry 

In the snippet below, `samples = iter(opinfo.sample_inputs("cpu", dtypes.float64, requires_grad=True))` assumes that the ops can always run on CPU even if CUDA is not enabled. However, for method `cuda` this is not true. This line executes when the file is loaded and leads to failure. Should we just add the op in `op_skip` but this would mean we won't be able to run the grad tests when CUDA is available. We can't use OpInfo.test_directives this code is not tied to a particular test.

https://github.com/Lightning-AI/lightning-thunder/blob/139cc2246b87f3e58d80b13897286a61d55826d8/thunder/tests/test_grad.py#L76-L93

thunder/torch/__init__.py
mruberry: Good point! Although I suppose in the future we should add the `non_blocking` argument to `thunder.torch.to`, since PyTorch's `Tensor.to` supports it.

mruberry:
> @mruberry
> 
> In the snippet below, `samples = iter(opinfo.sample_inputs("cpu", dtypes.float64, requires_grad=True))` assumes that the ops can always run on CPU even if CUDA is not enabled. However, for method `cuda` this is not true. This line executes when the file is loaded and leads to failure. Should we just add the op in `op_skip` but this would mean we won't be able to run the grad tests when CUDA is available. We can't use OpInfo.test_directives as this code is not tied to a particular test.
> 
> https://github.com/Lightning-AI/lightning-thunder/blob/139cc2246b87f3e58d80b13897286a61d55826d8/thunder/tests/test_grad.py#L76-L93

fyi @IvanYashchuk 

I think the real fix for this is to not query operator properties when a test file is imported, but instead have the OpInfo specify properties like whether the operator supports the grad transform (or not), or to use a list of operators that are supported (when testing an extension that can't modify the OpInfos directly). 

@IvanYashchuk may have a better suggestion for what to do in this case. Skipping the operation for now (and filing a follow-up issue) sounds reasonable to me.

IvanYashchuk:
Do we have a way to specify supported devices for OpInfo? Can we query the next supported device in the test generation if the CPU is not supported?

mruberry:
> Do we have a way to specify supported devices for OpInfo? Can we query the next supported device in the test generation if the CPU is not supported?

Yes, we have a devicetypes attribute on the OpInfo.

https://github.com/Lightning-AI/lightning-thunder/blob/139cc2246b87f3e58d80b13897286a61d55826d8/thunder/tests/opinfos.py#L323

But in this case the `.cuda()` method supports inputs on all device types, and just requires CUDA be available to execute.

thunder/torch/__init__.py
kshitij12345: Have filed an issue to support `non_blocking` in `Tensor.to`. See https://github.com/Lightning-AI/lightning-thunder/issues/197

kshitij12345:
> Skipping the operation for now (and filing a follow-up issue) sounds reasonable to me.

For now, will skip this op and file an issue for the same.

thunder/tests/test_grad.py
```diff
@@ -36,6 +36,8 @@
     "embedding",
     "index_put",
     "batch_norm",
+    # Requires CUDA runtime to be available (fails on CPU only runtimes).
+    "cuda",
```
mruberry, line 40: I suppose one fix would be to test if cuda is available and only if that's true add the string to this set? @IvanYashchuk

thunder/torch/__init__.py
```diff
@@ -287,6 +287,30 @@ def to(
     return a
 
 
+@torchsymbol(torch.Tensor.cuda, is_method=True)
+def cuda(a: TensorProxy, /, device=None, non_blocking: bool = False, memory_format: None | torch.memory_format = None):
```
mruberry, line 291: Nit: annotate the return type `-> TensorLike`

mruberry: Nit: annotate the `device` parameter, `device: None | DeviceLike=None`

mruberry (approved):
LGTM!

Issue #191, opened 2024-04-15
Disable bookend optimization in nvfuser


jjsjann123:
## 🚀 Feature

Bookend optimization was implemented when nvfuser makes unnecessary copy of tensor going through meta operations (view, reshape, transpose) that could be resolved through view.

Since nvfuser now does support alias operations, we would want to turn bookend optimization off by default.

### Motivation

It is beneficial to have meta operations being inside fusion. `reshape` is a good example here.
```
  # t_2: "cuda:0 f32[2, 16, 8, 16]"
  t1 = torch.permute(t_2, [0, 2, 1, 3])  # t1: "cuda:0 f32[2, 8, 16, 16]"
    # t1 = ltorch.permute(t_2, [0, 2, 1, 3])  # t1: "cuda:0 f32[2, 8, 16, 16]"
      # t1 = prims.transpose(t_2, (0, 2, 1, 3))  # t1: "cuda:0 f32[2, 8, 16, 16]"
  del t_2
  t3 = torch.reshape(t1, (2, 8, 256))  # t3: "cuda:0 f32[2, 8, 256]"
    # t3 = ltorch.reshape(t1, (2, 8, 256))  # t3: "cuda:0 f32[2, 8, 256]"
      # t3 = prims.reshape(t1, (2, 8, 256))  # t3: "cuda:0 f32[2, 8, 256]"
```

e.g. Reshape could ended up being a real data copy, we would want to fuse that with other operations. If `t3` is input to a fusion, we can fuse the permute/reshape with other operations; similarly, if `t_2` is an output from fusion, codegen could make a better decision on whether to fuse the reshape with previous operation or to produce `t_2` in a format that the following permute+reshape does't trigger memory copy.

### Pitch

The promise we (@wujingyue) made earlier:

> we'll bring more clarity to CPU latency and disable bookending with blast radius controlled.

Actionable item:

- [ ] check functional issue with bookend disabled;
- [ ] track einsum performance after disabling bookend optimization #193 ;
- [ ] nvfuser backend host latency.

cc @tfogal @apaz-cli

wujingyue:
For the first action item, https://github.com/Lightning-AI/lightning-thunder/pull/206 triggers the following CI errors. 

```
FAILED thunder/tests/test_examine_memory.py::test_nanogpt_block_nvfuser_cuda_float32 - AssertionError: assert 235985920 == 242277376
[...]

PR #206, opened 2024-04-16
Disable bookend by default.

wujingyue:


lantiga:
@wujingyue should we keep this open?

wujingyue:
> @wujingyue should we keep this open?

It's a draft PR to fix #191, so it feels natural to me to keep it in draft state. Do you suggest we organize it differently?

lantiga:
Tagging as `later` to keep track

PR #214, opened 2024-04-17
Refactor integration with PyTorch's Autograd

IvanYashchuk:
One of the complications with the previous PyTorch Autograd integration is that we dealt with two traces: forward and backward. They were stitched together and connected by `ThunderFunction.apply` somewhere in the `thunder.jit` code. We want to support the composable application of trace transformations in any order as `transform3(transform2(transform1(computation_trace)))`, each transform might potentially change the meaning of backward so it must also be applied to the backward pass and it becomes complicated to handle two distinct objects and keep them in sync.

This is a first step in the refactoring process. I think the simplest way to transparently have different transforms to interact with PyTorch Autograd registration is to have a special symbol that represents registration into Autograd, one of the inputs to this symbol is a Trace that should be modified when a transform is applied. The details of what exactly the interface of this symbol should be are not clear. Maybe the subsymbols of this symbol could represent the backward trace.

todo:
- [ ] Fill in more details in the description.


thunder/tests/test_examine_memory.py
IvanYashchuk, line 1: @kiya00, could you please review the changes to these tests? The forward trace now includes a special symbol that registers the backward function into PyTorch's Autograd for given forward tensors.

thunder/executors/torch_autograd.py
```diff
@@ -259,21 +314,36 @@ def split_forward_backward(computation_trc: TraceCtx, compile_data, compile_stat
...
     bw_traces.append(bw_extrace)
 
-    bw_trace = rename_bwd_trace_outputs(bw_extrace, fw_extrace)
+    for transform in compile_data.post_optimization_transforms:
+        utils.check_type(transform, PostOptimizationTransform)
+        bw_extrace = transform.transform_trace(bw_extrace, executors_list=compile_data.executors_list)
+        bw_traces.append(bw_extrace)
+
+    # Update the forward trace with the correct backward function
+    if any(bsym.sym.name == "connect_to_autograd_impl" for bsym in reversed(fw_extrace.bound_symbols)):
```
t-vi, line 329: Why could this not be `bsym.sym == connect_to_autograd`?
(I guess it's two questions, one why it is impl and why we compare the name rather than the sym.)

t-vi:
Very interesting and exciting, thank you for working on this!

> The details of what exactly the interface of this symbol should be are not clear. Maybe the subsymbols of this symbol could represent the backward trace.
I think it would be quite a departure from what subsymbols currently are, unless we manage to have a "build trace from" or so.

thunder/executors/torch_autograd.py
IvanYashchuk: > Why could this not be bsym.sym == connect_to_autograd?

At this stage "transform for execution" phase has replaced Thunder's general symbol `connect_to_autograd` which works on TensorProxy with PyTorch executor-specific symbol `connect_to_autograd_impl` that works with torch.Tensors.

As for string comparison, I probably didn't want to import anything from `torch.executors.torchex` due to possible cyclic dependencies. I'll revisit this.

IvanYashchuk:
Current failures to fix:
```py
FAILED thunder/tests/test_networks.py::test_nanogpt_complete_cuda_graphs_autograd_nvfuser_cuda_float32 - RuntimeError: one of the variables needed for gradient computation has been modified by an inplace operation: [torch.cuda.LongTensor [4, 64]] is at version 1; expected version 0 instead. Hint: enable anomaly detection to find the operation that failed to compute its gradient, with torch.autograd.set_detect_anomaly(True).
FAILED thunder/tests/test_networks.py::test_nanogpt_complete_cuda_graphs_autograd_torch_cuda_float32 - RuntimeError: one of the variables needed for gradient computation has been modified by an inplace operation: [torch.cuda.LongTensor [4, 64]] is at version 1; expected version 0 instead. Hint: enable anomaly detection to find the operation that failed to compute its gradient, with torch.autograd.set_detect_anomaly(True).
============= 2 failed, 15 passed, 3 warnings in 79.89s (0:01:19) ==============
```

thunder/executors/torch_autograd.py
t-vi: If needed, importing in the function should work and be preferable to string comparison.

IvanYashchuk:
Current failure to fix:
```
FAILED thunder/tests/test_core.py::test_dataclass_output[True] - AssertionError: argument layout mismatch: [<TensorProxy(name="x", dtype=thunder.dtypes.float32, shape=(3, 3))>, <TensorProxy(name="t2", dtype=thunder.dtypes.float32, shape=(3, 3))>] (<TensorProxy(name="t1", dtype=thunder.dtypes.float32, shape=(3, 3))>, <TensorProxy(name="t4", dtype=thunder.dtypes.float32, shape=(3, 3))>, <TensorProxy(name="t5", dtype=thunder.dtypes.float32, shape=(3, 3))>, <TensorProxy(name="t6", dtype=thunder.dtypes.float32, shape=(3, 3))>, <TensorProxy(name="t7", dtype=thunder.dtypes.float32, shape=(3, 3))>, <TensorProxy(name="t8", dtype=thunder.dtypes.float32, shape=(3, 3))>)
``

File: thunder/executors/cudagraphex.py [hunk 2]
```diff
@@ -138,8 +138,21 @@ def fuse(self, region: Region, fusion_counter: int, num_static_inputs: None | in
...
+        # is deleted before the return statement.
+        bsyms = [
+            bsym
+            for bsym in region.bound_symbols
+            if bsym.sym.id != prims.PrimIDs.DEL
+            or not (
+                bsym.sym.id == prims.PrimIDs.DEL
+                and any(a in region.outputs for a in bsym.flat_variableified_proxy_args)
+            )
+        ]
```
nikitaved, line 152: @IvanYashchuk , I am a bit confused. Could you please elaborate on the comment?

thunder/executors/cudagraphex.py
```diff
@@ -138,8 +138,21 @@ def fuse(self, region: Region, fusion_counter: int, num_static_inputs: None | in
         inputs = [unvariableify(inp) for inp in sorted(region.inputs, key=lambda var: var.proxy.name)]
         outputs = [unvariableify(out) for out in sorted(region.outputs, key=lambda var: var.proxy.name)]
 
+        # Remove del operations from the fusion region if they are part of the
+        # output of the region. This is to avoid the case where the output tensor
+        # is deleted before the return statement.
+        bsyms = [
```
nikitaved, line 144: @IvanYashchuk , is this even possible? If so, it looks like we have a bug somewhere?

IvanYashchuk: The generated Python callable that was used for CUDA Graphs execution had statements like
```py
...
del t0, t1
return ..., t0, t1
```
This was happening because the `del` statements were consumed here https://github.com/Lightning-AI/lightning-thunder/blob/e9cfc1a5b6b211e03fb0ec906344f85bf4f9ae43/thunder/executors/cudagraphex.py#L168-L169
This bit of code tries to remove the `del` operations that try to remove variables that are returned from this CUDA Graph region.

IvanYashchuk: I can make a minimal example and submit an issue for this bug.

nikitaved: Do you have a specific example this was taking place that I could reproduce?

nikitaved: Yeah, well, that could happen given that backward pass is definitely less straightforward compared to the forward pass.

IvanYashchuk: Here's the promised example https://github.com/Lightning-AI/lightning-thunder/issues/728.

IvanYashchuk: Another approach is to remove all `del` operations from the trace first and then reintroduce them again after CUDA Graphs regions are formed.

t-vi:
Seems to be superseded. Please reopen if you disagree.

PR #219, opened 2024-04-18
add large_mem test marker for memory-intense tests

t-vi:
Tests requiring a lot of (GPU) mem can be marked with @pytest.mark.large_mem .
Those will be run with 4 workers rather than the default of 9 in the GPU "test regular" step.

### DIFF
--- a/.azure/gpu-tests.yml
+++ b/.azure/gpu-tests.yml
@@ -95,7 +95,7 @@ jobs:
               --durations=250 \
               --timeout=240 \
               --numprocesses=9 \
-	      -m "not large_mem"\
+              -m "not large_mem"\
               --ignore=thunder/tests/distributed --ignore=thunder/tests/test_networks.py
           coverage run --source thunder -m \
             pytest thunder/tests/ \
@@ -106,7 +106,7 @@ jobs:
               --durations=250 \
               --timeout=240 \
               --numprocesses=4 \
-	      -m "large_mem"\
+              -m "large_mem"\
               --ignore=thunder/tests/distributed --ignore=thunder/tests/test_networks.py
           # compile coverage results
           python -m coverage report
--- a/.azure/gpu-tests.yml
+++ b/.azure/gpu-tests.yml
@@ -95,6 +95,18 @@ jobs:
               --durations=250 \
               --timeout=240 \
               --numprocesses=9 \
+	      -m "not large_mem"\
+              --ignore=thunder/tests/distributed --ignore=thunder/tests/test_networks.py
+          coverage run --source thunder -m \
+            pytest thunder/tests/ \
+              -m "not standalone" \
+              -v --datefmt="%Y%m%d-%H:%M:%S.%f" \
+              --timeout=240 \
+              --random-order-seed=42 \
+              --durations=250 \
+              --timeout=240 \
+              --numprocesses=4 \
+	      -m "large_mem"\
               --ignore=thunder/tests/distributed --ignore=thunder/tests/test_networks.py
           # compile coverage results
           python -m coverage report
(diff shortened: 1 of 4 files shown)


pyproject.toml
crcrpar, line 1: [question]
pytorch has [`largeTensorTest`](https://github.com/pytorch/pytorch/blob/d59f1da62f6a74d815dede5ac5513b37c49733a3/torch/testing/_internal/common_device_type.py#L1106) which lets us specify the required memory size.
Would we get more specific in the required memory size in the future?

.azure/gpu-tests.yml
```diff
@@ -95,6 +95,18 @@ jobs:
...
+              --ignore=thunder/tests/distributed --ignore=thunder/tests/test_networks.py
+          coverage run --source thunder -m \
+            pytest thunder/tests/ \
+              -m "not standalone" \
+              -v --datefmt="%Y%m%d-%H:%M:%S.%f" \
+              --timeout=240 \
+              --random-order-seed=42 \
+              --durations=250 \
+              --timeout=240 \
+              --numprocesses=4 \
```
crcrpar, line 108: [question] what's the reasoning/heuristic to set this to 4?

t-vi: half of 9

pyproject.toml
t-vi: Oh, cool. So I'd be totally happy if someone adapted that for Thunder, it's just that what I do myself is limited to "get things reliably back to green for now".

Borda: A think that these can be just makes as stand-alone so will be executed in sequence

.azure/gpu-tests.yml
Borda: and the rest 0.5 is just tax :)

pyproject.toml
Borda: https://github.com/Lightning-AI/lightning-thunder/blob/649c3d71de9af13a3247694a4e23289221d0ff72/pyproject.toml#L35

t-vi: I wanted to strike a balance between having longer test runs and avoiding OOM.

pyproject.toml
```diff
@@ -33,6 +33,7 @@ addopts = [
 ]
 markers = [
     "standalone: mark a test as standalone",
+    'large_mem: needs a lot of gpu memory (select/deselect with -m large_mem / -m "not large_mem")',
```
Borda, line 36: Does this mean that only one model/test is running at a time?
as large memory could have different parameters... also does this apply to GPU or also CPU?

t-vi: Currently just GPU and currently we halve the number of pythread threads running these tests.

thunder/tests/test_jit_general.py
```diff
@@ -631,6 +631,7 @@ def test_nanogpt():
     "device",
     ("cpu", "cuda"),
 )
+@pytest.mark.large_mem
```
carmocca, line 634: There are some nicer alternative solutions that we could pursue:

- Make the "like" configs smaller.
- Run these tests on 16bit precision.

For reference, this is how much the params take today:
```
gpt-neox-like 6532864 26.131456 MB
llama1-like 602432 2.409728 MB
long-context-like 4301120 17.20448 MB
llama2-like 602432 2.409728 MB
falcon-7b-like 4359040 17.43616 MB
falcon-40b-like 1589760 6.35904 MB
codellama2-like 819520 3.27808 MB
mixtral-like 3308800 13.2352 MB
```

(in litgpt_model.py)
```python
for name, config in name_to_config.items():
    if not name.endswith("-like"):
        continue
    config = Config(**config)
    with torch.device("meta"):
        model = GPT(config)
    n_params = sum(p.numel() for p in model.parameters())
    print(name, n_params, (n_params * 4) / 1e6, "MB")
```

IvanYashchuk: These reported numbers look small. Maybe there should be a better clean-up after each test?

t-vi:
Seems like this did not fly.

PR #222, opened 2024-04-18
Move Grad AllReduce Bucketing to inside of `thunder.executors.passes.transform_for_execution` from `torch_autograd.split_forward_backward`

### LINKED ISSUE #184: Refactor bucketing of AllReduce to be applied to a generic trace
## 🚀 Feature

Today bucketing of AllReduce operation is done when the following conditions are met:
* `disable_torch_autograd` is set to False
https://github.com/Lightning-AI/lightning-thunder/blob/139cc2246b87f3e58d80b13897286a61d55826d8/thunder/__init__.py#L328
* `compile_data.fn.use_ddp` is set to True
https://github.com/Lightning-AI/lightning-thunder/blob/139cc2246b87f3e58d80b13897286a61d55826d8/thunder/executors/torch_autograd.py#L280-L283

We need to move the bucketing transformation out of the "torch autograd" code path to be applied in the generic thunder.jit pipeline with no information about "forward" or "backward".

This transformation could be placed to happen before "transform_for_execution" as part of the "additional_transforms":
https://github.com/Lightning-AI/lightning-thunder/blob/139cc2246b87f3e58d80b13897286a61d55826d8/thunder/__init__.py#L608-L610

or as the first step inside "transform_for_execution".

Currently, in the torch_autograd.py file, the allreduce bucketing transform is placed right before the "transform_for_execution":
https://github.com/Lightning-AI/lightning-thunder/blob/139cc2246b87f3e58d80b13897286a61d55826d8/thunder/executors/torch_autograd.py#L283-L292

crcrpar:
## What does this PR do?

Fixes #184

As per title. The changes are addition of a logic to tell whether or not the input `TraceCtx` represents DDP backward.

### DIFF
--- a/thunder/distributed/transforms/ddp.py
+++ b/thunder/distributed/transforms/ddp.py
@@ -32,6 +32,7 @@ if TYPE_CHECKING:
 
 
 __all__ = [
+    "apply_bucketing_to_grad_allreduce",
     "optimize_allreduce_in_ddp_backward",
 ]
 
@@ -238,3 +239,40 @@ def optimize_allreduce_in_ddp_backward(
     )
     utils.check(visit_callable.has_replaced_return, lambda: "")
     return updated_bwd_trace
+
+
+def apply_bucketing_to_grad_allreduce(
+    trace: TraceCtx,
+    compile_data: CompileData | None = None,
+) -> tuple[TraceCtx, bool]:
+    if get_skip_data_parallel_grad_sync():
+        return trace, False
+
+    if compile_data is None:
+        import thunder
+
+        compile_data = thunder.compile_data(trace)
+
+    producers, consumers = utils.producers_and_consumers(trace)
+    flat_trace_output, flat_trace_output_spec = tree_flatten(trace.output)
+    output_tensor_to_index_and_prod_bsym = utils.ProxyDict()  # dict[Proxy, tuple[int, BoundSymbol]]
+    for index, output in enumerate(flat_trace_output):
+        if isinstance(output, TensorProxy):
+            output_tensor_to_index_and_prod_bsym[output] = (index, producers[output])
+
+    grad_before_after_allreduce = utils.ProxyDict()
+    bsym: BoundSymbol
+    for key in output_tensor_to_index_and_prod_bsym._dict:
+        _, bsym = output_tensor_to_index_and_prod_bsym[key]
+        if bsym.sym.id == dist_prims.PrimIDs.WAIT:
+            bsym_of_allreduce: BoundSymbol = producers[bsym.flat_proxy_args[0]]
+            utils.check(
+                bsym_of_allreduce.sym.id,
+                dist_prims.PrimIDs.ALL_REDUCE,
+                lambda: f"{bsym.sym.id=}, {bsym_of_allreduce.sym.id=}",
+            )
+            grad_before_after_allreduce[bsym.flat_proxy_outs[0]] = bsym_of_allreduce.flat_proxy_args[0]
+    if len(grad_before_after_allreduce._dict) == 0:
+        return trace, False
+
+    return optimize_allreduce_in_ddp_backward(trace, compile_data=compile_data), True
--- a/thunder/distributed/transforms/ddp.py
+++ b/thunder/distributed/transforms/ddp.py
@@ -267,6 +267,9 @@ def apply_bucketing_to_grad_allreduce(trace: TraceCtx) -> TraceCtx:
         import thunder
 
         compile_data = thunder.compile_data(trace)
+    # There's no ways to move forward if `compile_data` is None, so early exit.
+    if compile_data is None:
+        return trace
     utils.check_type(compile_data, CompileData)
 
     if not compile_data.is_module:
--- a/thunder/distributed/transforms/ddp.py
+++ b/thunder/distributed/transforms/ddp.py
@@ -3,6 +3,7 @@ from collections import defaultdict
 from dataclasses import dataclass
 from typing import TYPE_CHECKING
 
+from thunder.common import CompileData
 from thunder.core import dtypes
 from thunder.core import devices
 from thunder.core import prims
@@ -28,12 +29,10 @@ if TYPE_CHECKING:
     from thunder.core.symbol import BoundSymbol
     from thunder.core.trace import TraceCtx
     from thunder.core.trace import VariableInterface
-    from thunder.common import CompileData
 
 
 __all__ = [
     "apply_bucketing_to_grad_allreduce",
-    "optimize_allreduce_in_ddp_backward",
 ]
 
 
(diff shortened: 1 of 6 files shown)


thunder/distributed/transforms/ddp.py
```diff
@@ -202,3 +202,66 @@ def optimize_allreduce_in_ddp_backward(
...
+        if bsym.sym.id == dist_prims.PrimIDs.WAIT:
+            bsym_of_allreduce: BoundSymbol = producers[bsym.flat_proxy_args[0]]
+            utils.check(
+                bsym_of_allreduce.sym.id,
+                dist_prims.PrimIDs.ALL_REDUCE,
+                lambda: f"{bsym.sym.id=}, {bsym_of_allreduce.sym.id=}",
+            )
+            grad_before_after_allreduce[bsym.flat_proxy_outs[0]] = bsym_of_allreduce.flat_proxy_args[0]
+    if len(grad_before_after_allreduce._dict) == 0:
+        return trace
```
crcrpar, line 265: Should we expect `len(grad_before_after_allreduce._dict` to be a natural number?

thunder/distributed/transforms/ddp.py
```diff
@@ -202,3 +202,66 @@ def optimize_allreduce_in_ddp_backward(
...
+
+    Returns:
+        :class:`thunder.core.trace.TraceCtx`
+    """
+
+    from torch.distributed import ProcessGroup
+    from thunder.core.compile_data import get_compile_data
+
+    if get_skip_data_parallel_grad_sync():
+        return trace
```
IvanYashchuk, line 227: It should be possible to create and jit a program that includes allreduce in a non-data parallel setting. This pass should work if there are allreduces in the trace.
What do you think of scanning the trace probably from the end of the bound symbols list to optimize for the backward case to check whether there's an allreduce in the trace or not. Return early if there's less than one allreduce.

thunder/distributed/transforms/ddp.py
IvanYashchuk: Is 0 a natural number? :smile:  https://math.stackexchange.com/questions/283/is-0-a-natural-number

The length of an empty dict is 0.

thunder/executors/passes.py
```diff
@@ -129,8 +129,16 @@ def visit_(bsym: BoundSymbol) -> transforms.VISIT_TYPE:
 
 
 def transform_for_execution(trace: TraceCtx, executors_list: Sequence[Executor]) -> TraceCtx:
+    import torch
+
```
IvanYashchuk, line 133: It doesn't seem that this import is needed here.

thunder/distributed/transforms/ddp.py
```diff
@@ -202,3 +202,66 @@ def optimize_allreduce_in_ddp_backward(
     )
     utils.check(visit_callable.has_replaced_return, lambda: "")
     return updated_bwd_trace
+
+
+def apply_bucketing_to_grad_allreduce(trace: TraceCtx) -> TraceCtx:
```
IvanYashchuk, line 207: It looks like this trace can work not only with gradients but any allreduce inputs?
```suggestion
def apply_bucketing_to_allreduce(trace: TraceCtx) -> TraceCtx:
```

crcrpar: I still think this is only for gradient all-reduce as it currently (a) checks `use_ddp` attr of `compile_data.fn` and (b) calls the method to batching gradient all-reduce's or to convert the existing ones into in-place ones.

thunder/distributed/transforms/ddp.py
crcrpar: > a program that includes allreduce in a non-data parallel setting.

iirc the bucketing got puzzled by such trace so that I added a check in https://github.com/Lightning-AI/lightning-thunder/commit/896b40131484de8637586bd2de9be2a55fd9496f.

> What do you think of scanning the trace probably from the end of the bound symbols list to optimize for the backward case to check whether there's an allreduce in the trace or not. Return early if there's less than one allreduce

I don't think I understand this as I thought the following part does this.

thunder/executors/passes.py
crcrpar: really?

IvanYashchuk: My bad I see it's used for `torch.distributed.is_available`.

crcrpar:
The rebase to https://github.com/Lightning-AI/lightning-thunder/pull/222/commits/b855247e171527d4bc523cd4e2ca44b8461460c4 seems to give me a bug which at glance is unintelligible where comms are there even under `no_sync`. The change of this pr doesn't look that related per se...

t-vi:
This is the oldest non-draft PR with all green CI. Do we want this as is or do we want to change things yet, @crcrpar @IvanYashchuk ?

crcrpar:
I don't have any out of my head

IvanYashchuk:
I'm sorry for the delay. I'll review the changes tomorrow.

lantiga:
Should we try to get this in? @crcrpar @IvanYashchuk

lantiga:
@t-vi does this interact with the fsdp work?

t-vi:
I think it is orthogonal.

crcrpar:
This is for ddp. We could reuse this for fsdp backward as well but the semantic would be different from what we have today

thunder/distributed/transforms/ddp.py
IvanYashchuk: I'm sorry what I meant is that we should make (in the future, not now) the bucketing transform applicable to any trace. As you pointed out in the other comment (https://github.com/Lightning-AI/lightning-thunder/pull/222/files#r1570536852) this pass works only if it's marked for ddp today. We should allow people to write Thunder code involving allreduce in the user script to be optimized by Thunder.

thunder/executors/torch_autograd.py
IvanYashchuk, line 1: This is the main change of this PR to me. Thank you!
This file needs to be lean.

thunder/executors/passes.py
```diff
@@ -129,8 +129,16 @@ def visit_(bsym: BoundSymbol) -> transforms.VISIT_TYPE:
...
     start_time_ns = time.time_ns()
 
+    if torch.distributed.is_available():
+        # Apply AllReduce bucketing if possible & needed
+        from thunder.distributed.transforms.ddp import apply_bucketing_to_grad_allreduce
```
IvanYashchuk, line 138: Is `apply_bucketing_to_grad_allreduce` importable with PyTorch build without distributed available? If yes could this if available statement be moved inside the function to keep the indentation in `transform_for_execution` cleaner?

IvanYashchuk (approved):

IvanYashchuk:
> Should we try to get this in? @crcrpar @IvanYashchuk

@lantiga, yes, let's get this in

t-vi (commented):

t-vi (approved):

Issue #183, opened 2024-04-15
Automated benchmarking


riccardofelluga:
## 🚀 Feature

I would like to have automated benchmarks for selected models to allow for performance tracking.

### Motivation

This enables us to check for performance issues and track development over time.

IvanYashchuk:
Automation is always interesting. Could you please expand on what you mean by "automated" specifically? What are the manual steps that you'd like to see automated?

riccardofelluga:
Sure! I've updated the description with more info

xwang233:
How is this different from running 
```bash
pytest thunder/benchmarks/targets.py
python thunder/benchmarks/distributed.py
```

cc @crcrpar @tfogal

IvanYashchuk:
@riccardofelluga, can you answer the question above from Xiao? Providing detailed information on what's on your mind and what you would like to achieve would be very helpful here.

PR #225, opened 2024-04-18
Add benchmark reporting tool and mistral example

### LINKED ISSUE #224: Benchmarking suite that runs scripts
As part of the automated benchmarking I would like the benchmarks to automatically report metrics and save them for comparison.

cc @crcrpar

riccardofelluga:
## What does this PR do?

As part of #183 this fixes #224 by introducing an abseil based template that can be used to setup benchmarks and report results in from of JSONs. Also this PR includes an example benchmark of Mistral.

## How does this work?
This contribution consist of two main parts:
- It defines a template for creating standalone benchmarking scripts and provides the functionality record info and metrics. At the end of the run the recorded information is stored in the form of a JSON file.
-  The `Benchmarking` class provides a place where practitioners can register previously written scripts and parametrize them. Doing so allows to automate running benchmarks.

### Benchmark scripts template
The idea here is that developers can write scripts with the following structure:
``` python
from absl.testing import absltest
from benchmark_scripts import Target

class MyBenchmarkClass(Target):
    def setUp(self):
        super().setUp()
        # setup benchmark here

    def tearDown(self) -> None:
        super().tearDown()
        # teardown for the benchmark just ran

    def test_my_benchmark(self):
        # code to be benchmarked
        pass

if __name__ == "__main__":
    absltest.main()
```
The reason behind having a separate script for each benchmark is that it helps with isolation. Running multiple benchmarks on the same process can interfere on some metrics like peak memory.
Within this class practitioners can report metrics and parameters with the methods `report_metrics` and `report_flags` respectively. This reported metrics together with commit/version information will then be saved as in the form of a JSON file like the following:
``` json
{
    "test_class": "MistralBenchmark",
    "timestamp": "2024-04-19_10-30-23",
    "head": "0534668a",
    "succeeded": true,
    "iter_time": 93333096.0,
    "max_memory_allocated": 5.098717696,
    "input_len": 4096,
    "batch_size": 1,
    "max_iters": 100,
    "warmup_iters": 20,
    "compile": "eager"
}
```
(example from `mistral.py` ran with `python mistral.py --compile=eager --input_len=4096 --batch_size=1`)

### Benchmark automation
The `Benchmarking` class in `benchmark_scripts.py` file provides with a way to automate and parametrize the previously described scripts. In particular, practitioners can register an parametrize a script by adding a method to this class that starts with `test_`:
``` python 
    @parameterized.product(
        compile=(
            "eager",
            "inductor",
            "thunder",
        ),
        input_len=(4_096,),
        batch_size=(1,),
    )
    def test_my_script(self, **kwargs):
        self.run_standalone_script("my_benchmark.py", kwargs)
```
It will pass the parameters as arguments when calling `my_benchmark.py` file. 

## Design decisions
In both cases I've used abseil built in argument parser for convenience.

### DIFF
--- a/thunder/benchmarks/benchmark_scripts.py
+++ b/thunder/benchmarks/benchmark_scripts.py
@@ -85,8 +85,8 @@ class Benchmarks(parameterized.TestCase):
     def setUpClass(cls):
         super().setUpClass()
 
-        if flags.FLAGS.benchmark_output_dir and not os.path.exists(flags.FLAGS.benchmark_output_dir):
-            os.makedirs(flags.FLAGS.benchmark_output_dir)
+        if flags.FLAGS.benchmark_output_dir:
+            os.makedirs(flags.FLAGS.benchmark_output_dir, exist_ok=True)
 
     @classmethod
     def tearDownClass(cls) -> None:
--- a/thunder/benchmarks/benchmark_scripts.py
+++ b/thunder/benchmarks/benchmark_scripts.py
@@ -80,6 +80,7 @@ class Benchmarks(parameterized.TestCase):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self._summary_metrics = {}
+        self._benchmakr_dir = os.path.dirname(__file__)
 
     @classmethod
     def setUpClass(cls):
@@ -93,7 +94,7 @@ class Benchmarks(parameterized.TestCase):
         super().tearDownClass()
 
     def run_standalone_script(self, filename: str, kwargs: dict[str, Any]) -> None:
-        cmd_string = ["python", f"{filename}"]
+        cmd_string = ["python", "".join([self._benchmakr_dir, "/", filename])]
         for key, val in kwargs.items():
             cmd_string.append("--" + str(key) + "=" + str(val))
 
--- a/thunder/benchmarks/benchmark_scripts.py
+++ b/thunder/benchmarks/benchmark_scripts.py
@@ -1,0 +1,117 @@
+import json
+import os
+import subprocess
+
+from typing import Any
+from datetime import datetime
+from functools import partial
+
+from absl import flags
+from absl import logging
+from absl.testing import absltest, parameterized
+
+flags.DEFINE_string("benchmark_output_dir", default="./benchmark_results", help="Output directory for benchmark JSONs")
+
+
+class Target(absltest.TestCase):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._failed_tests = []
+
+        for fn_name in dir(self):
+            if fn_name.startswith("test"):
+                fn = getattr(self, fn_name)
+                wrapped_fn = partial(self._test_wrapper, fn=fn)
+                setattr(self, fn_name, wrapped_fn)
+
+    # Add timestamp and defer exceptions so we can save the test status.
+    def _test_wrapper(self, *args, fn=None, **kwargs):
+        self._start_time = datetime.now().strftime("%Y-%m-%d_%H-%M-%S")
+        try:
+            return fn(*args, **kwargs)
+        except Exception as ex:
+            self._failed_tests.append(ex)
+
+    def _get_git_revision(self) -> str:
+        cmd = ["git", "rev-parse", "--short", "HEAD"]
+        return subprocess.check_output(cmd).decode("ascii").strip()
+
+    def _report_benchmark_results(self):
+        out_dir = flags.FLAGS.benchmark_output_dir
+        filename = os.path.join(out_dir, f"{self._name}_{self._start_time}.json")
+        git_revision = self._get_git_revision()
+
+        with open(filename, mode="w") as json_report:
+            json.dump(
+                {
+                    "test_class": self._name,
+                    "timestamp": self._start_time,
+                    "head": git_revision,
+                    "succeeded": not self.has_failed_tests(),
+                    **self._reported_metrics,
+                    **self._flags,
+                },
+                json_report,
+            )
+
+    def has_failed_tests(self):
+        return len(self._failed_tests) > 0
+
+    def setUp(self) -> None:
+        super().setUp()
+        self._name = type(self).__name__
+        self._reported_metrics: dict = {}
+        self._flags: dict = {}
+
+    def tearDown(self) -> None:
+        super().tearDown()
+        self._report_benchmark_results()
+        for exception in self._failed_tests:
+            raise exception
+
+    def report_flag(self, name: str, value: Any) -> None:
+        self._flags[name] = value
+
+    def report_metrics(self, metrics: dict[str, Any]) -> None:
+        self._reported_metrics.update(metrics)
+
+
+class Benchmarks(parameterized.TestCase):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._summary_metrics = {}
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+
+        if flags.FLAGS.benchmark_output_dir and not os.path.exists(flags.FLAGS.benchmark_output_dir):
+            os.makedirs(flags.FLAGS.benchmark_output_dir)
+
+    @classmethod
+    def tearDownClass(cls) -> None:
+        super().tearDownClass()
+
+    def run_standalone_script(self, filename: str, kwargs: dict[str, Any]) -> None:
+        cmd_string = ["python", f"{filename}"]
+        for key, val in kwargs.items():
+            cmd_string.append("--" + str(key) + "=" + str(val))
+
+        logging.info(f'Running: {" ".join(cmd_string)!r}')
+        self.assertCommandSucceeds(cmd_string, env=os.environ)
+
+    @parameterized.product(
+        compile=(
+            "eager",
+            "inductor",
+            "thunder",
+        ),
+        input_len=(4_096,),
+        batch_size=(1,),
+    )
+    def test_mistral(self, **kwargs):
+        self.run_standalone_script("mistral.py", kwargs)
+
+
+if __name__ == "__main__":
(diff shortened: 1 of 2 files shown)


thunder/benchmarks/benchmark_scripts.py
crcrpar, line 1: would you have plans to generalize this script to other files of `thunder/benchmarks` directory?

thunder/benchmarks/benchmark_scripts.py
```diff
@@ -0,0 +1,117 @@
...
+    def _test_wrapper(self, *args, fn=None, **kwargs):
+        self._start_time = datetime.now().strftime("%Y-%m-%d_%H-%M-%S")
+        try:
+            return fn(*args, **kwargs)
+        except Exception as ex:
+            self._failed_tests.append(ex)
+
+    def _get_git_revision(self) -> str:
+        cmd = ["git", "rev-parse", "--short", "HEAD"]
+        return subprocess.check_output(cmd).decode("ascii").strip()
```
crcrpar, line 37: I thought it'd make sense to return either thunder's version or git commit hash

thunder/benchmarks/benchmark_scripts.py
```diff
@@ -0,0 +1,117 @@
...
+        compile=(
+            "eager",
+            "inductor",
+            "thunder",
+        ),
+        input_len=(4_096,),
+        batch_size=(1,),
+    )
+    def test_mistral(self, **kwargs):
+        self.run_standalone_script("mistral.py", kwargs)
```
crcrpar, line 113: it seems that this requires us to run this script inside of `thunder/benchmarks` directory?

thunder/benchmarks/mistral.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+def shift_targets_to_left(targets: torch.Tensor) -> torch.Tensor:
+    targets = torch.roll(targets, dims=-1, shifts=-1)
+    targets[:, -1] = 0
+    return targets
+
+
+def get_model(vocab_size: int) -> GPT:
+    config = Config.from_name("Mistral-7B-v0.1")
```
crcrpar, line 81: would there be any other mistral models that we might want to run?

thunder/benchmarks/mistral.py
```diff
@@ -0,0 +1,176 @@
...
+from absl.testing import absltest
+
+import torch
+from torch.optim import AdamW
+from torch.utils.data import DataLoader, IterableDataset
+
+from thunder.tests.litgpt_model import Config, GPT
+from transformers import AutoTokenizer
+
+flags.DEFINE_string("compile", default="eager", help="Specify compile option: thunder|inductor|eager")
```
crcrpar, line 17: would there be an equivalent of argparse's [`choices`](https://docs.python.org/3/library/argparse.html#choices)?

thunder/benchmarks/mistral.py
crcrpar, line 1: At glance I'm not sure why this needs to be a separate file from https://github.com/Lightning-AI/lightning-thunder/blob/649c3d71de9af13a3247694a4e23289221d0ff72/thunder/benchmarks/benchmark_litgpt.py given the use of lit_gpt.

Could you kindly educate me?

thunder/benchmarks/benchmark_scripts.py
```diff
@@ -0,0 +1,117 @@
...
+    def _get_git_revision(self) -> str:
+        cmd = ["git", "rev-parse", "--short", "HEAD"]
+        return subprocess.check_output(cmd).decode("ascii").strip()
+
+    def _report_benchmark_results(self):
+        out_dir = flags.FLAGS.benchmark_output_dir
+        filename = os.path.join(out_dir, f"{self._name}_{self._start_time}.json")
+        git_revision = self._get_git_revision()
+
+        with open(filename, mode="w") as json_report:
```
mruberry, line 44: What does an example output look like?

mruberry:
A notebook explaining how to add a benchmark, and another notebook showing how to run and understand the results of a benchmark, would be very interesting.  They don't need to be in this PR. That said, this PR could have more comments explaining how to use its functionality

thunder/benchmarks/benchmark_scripts.py
```diff
@@ -0,0 +1,117 @@
...
+
+from typing import Any
+from datetime import datetime
+from functools import partial
+
+from absl import flags
+from absl import logging
+from absl.testing import absltest, parameterized
+
+flags.DEFINE_string("benchmark_output_dir", default="./benchmark_results", help="Output directory for benchmark JSONs")
```
Borda, line 13:
```suggestion
flags.DEFINE_string("output_dir", default="./benchmark_results", help="Output directory for benchmark JSONs")
```
As it is passed to the benchmark script...

thunder/benchmarks/benchmark_scripts.py
```diff
@@ -0,0 +1,117 @@
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._summary_metrics = {}
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+
+        if flags.FLAGS.benchmark_output_dir and not os.path.exists(flags.FLAGS.benchmark_output_dir):
+            os.makedirs(flags.FLAGS.benchmark_output_dir)
```
Borda, line 89:
```suggestion
        if flags.FLAGS.benchmark_output_dir:
            os.makedirs(flags.FLAGS.benchmark_output_dir. exist_ok=True)
```

thunder/benchmarks/benchmark_scripts.py
riccardofelluga: Yes if it can be useful for the team I think that would be a great path to continue on. For now this is more to see how the pieces fit together and then improve over time.

thunder/benchmarks/mistral.py
riccardofelluga: Yes it can be done with `lags.DEFINE_multi_string("multi_choice", ["a", "b", "c"] ...`.

thunder/benchmarks/mistral.py
riccardofelluga: You're right they are similar, I think the main difference here is that that script is not meant to be run nightly. At the moment the Mistral script is just an example of benchmarking and reporting metrics for a model. The main idea being that later on we can integrate this tools with more models and more metrics. 
If later on when the benchmarking tools are more solid, we still see that this files are very similar and serve the same purpose I agree on the fact that we could try to unify them

thunder/benchmarks/benchmark_scripts.py
riccardofelluga: Yes the main idea for the future is to use this template for multiple models and with different metrics

thunder/benchmarks/benchmark_scripts.py
riccardofelluga: Below an example output, I'll add it in the description of the PR too:

``` json
{
    "test_class": "MistralBenchmark",
    "timestamp": "2024-04-19_10-30-23",
    "head": "0534668a",
    "succeeded": true,
    "iter_time": 93333096.0,
    "max_memory_allocated": 5.098717696,
    "input_len": 4096,
    "batch_size": 1,
    "max_iters": 100,
    "warmup_iters": 20,
    "compile": "eager"
}
```

thunder/benchmarks/benchmark_scripts.py
riccardofelluga: It's true! Let me fix that so that we can run from other dirs

riccardofelluga:
I've updated the description of the PR, let me know if there is something else that you would like to know that I might have missed :)

thunder/benchmarks/mistral.py
```diff
@@ -0,0 +1,176 @@
...
+        tokenizer = AutoTokenizer.from_pretrained("mistralai/Mistral-7B-v0.1")
+        tokenizer.add_special_tokens({"pad_token": "[PAD]"})
+        vocab_size = len(tokenizer.get_vocab())
+
+        model = get_model(vocab_size)
+        model.to(dtype=torch.bfloat16)
+        model.to(device=self.device)
+        self.model = setup_compile(model)
+
+        self.mixed_ctx = torch.amp.autocast(device_type="cuda", dtype=torch.bfloat16)
```
crcrpar, line 121: why would autocast be enabled, though the model is already bfloat16?

thunder/benchmarks/mistral.py
```diff
@@ -0,0 +1,176 @@
...
+            self.optimizer.zero_grad()
+
+            if step > flags.warmup_iters:
+                self.metric.add_batch()
+
+        logging.info(f"Iter {step} with {self.metric.compute()*1e-6:.4f} ms per iteration.")
+
+
+if __name__ == "__main__":
+    absltest.main()
```
crcrpar, line 176: would this support distributed or only single device run?

thunder/benchmarks/benchmark_scripts.py
```diff
@@ -0,0 +1,118 @@
...
+        super().__init__(*args, **kwargs)
+        self._summary_metrics = {}
+        self._benchmakr_dir = os.path.dirname(__file__)
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+
+        if flags.FLAGS.output_dir:
+            os.makedirs(flags.FLAGS.output_dir, exist_ok=True)
```
crcrpar, line 90: it feels like this should be `setUp` not `setUpClass` as the former is called per each test method while the latter, per class.

thunder/benchmarks/benchmark_scripts.py
```diff
@@ -0,0 +1,118 @@
...
+
+    @classmethod
+    def tearDownClass(cls) -> None:
+        super().tearDownClass()
```
crcrpar, line 94: could you tell me what's expected to be defined in `tearDownClass` or `tearDown`?

riccardofelluga:
Closing for re-evaluation

PR #228, opened 2024-04-18
Change prologue details to prepare for fsdp as a transform

t-vi:
- support tracing a model on meta device (to enable tracing before materializing / sharding)
- use model.get_parameter / model.get_buffer to acquire parameters and buffers to more easily divert these.

### DIFF
--- a/thunder/core/devices.py
+++ b/thunder/core/devices.py
@@ -14,6 +14,7 @@ import thunder.core.baseutils as baseutils
 class DeviceType(Enum):
     CPU = auto()
     CUDA = auto()
+    META = auto()
 
 
 all_devicetypes = (DeviceType.CPU, DeviceType.CUDA)
@@ -21,6 +22,7 @@ all_devicetypes = (DeviceType.CPU, DeviceType.CUDA)
 _devicetype_prettyprint_map = {
     DeviceType.CPU: "cpu",
     DeviceType.CUDA: "cuda",
+    DeviceType.META: "meta",
 }
 _inverse_devicetype_prettyprint_map = {v: k for k, v in _devicetype_prettyprint_map.items()}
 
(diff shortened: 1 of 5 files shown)


thunder/core/jit_ext.py
IvanYashchuk, line 1: I must confess I don't understand what's going on in this file and I'm glad I don't need to.

thunder/tests/test_jit_general.py
```diff
@@ -629,7 +629,7 @@ def test_nanogpt():
 )
 @pytest.mark.parametrize(
     "device",
-    ("cpu", "cuda"),
+    ("cpu", "cuda", "meta"),
```
IvanYashchuk, line 632: Do we call PyTorch's meta functions anywhere in the pipeline? What's the state of meta tensors in PyTorch?

IvanYashchuk (approved):
How do prologue traces change with this PR?

I don't see yet how this will be connected with the "fsdp as a transform" work and I'm very curious to see where this leads to!

thunder/core/devices.py
```diff
@@ -14,13 +14,15 @@
...
     CUDA = auto()
+    META = auto()
 
 
 all_devicetypes = (DeviceType.CPU, DeviceType.CUDA)
 
 _devicetype_prettyprint_map = {
     DeviceType.CPU: "cpu",
     DeviceType.CUDA: "cuda",
+    DeviceType.META: "meta",
```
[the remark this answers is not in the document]
t-vi: I'm not attached to it any particular way, this is inspired by the thunder.distributed.fsdp 's use of meta.

t-vi:
Thank you @IvanYashchuk @crcrpar . Let's think about the meta bits separately. I'm happy to switch to whatever works best for fsdp.

PR #232, opened 2024-04-18
Allow duplicate executors

apaz-cli:
Disallowing them found a lot of bugs, but was ultimately a bad idea.


t-vi (approved):
Seems good. Thank you @apaz-cli

PR #239, opened 2024-04-20
Add `record_history` flag to `thunder.jit`. Off by default. Fixes memory issues.

apaz-cli:
Bikeshedding on the name is appreciated.

Also, recently we started calling it the "interpreter log." Going to end up renaming it either way.


t-vi:
I like `interpreter_log` (and this is user facing), so using that is likely a good idea.

t-vi (commented):
Let's have it. Thank you @apaz-cli

t-vi (approved):

PR #242, opened 2024-04-22
[benchmark] call `distributed_c10d.destroy_process_group` in benchmark_litgpt

crcrpar:
as per title.

cc @crcrpar @carmocca @awaelchli

### DIFF
--- a/thunder/benchmarks/benchmark_litgpt.py
+++ b/thunder/benchmarks/benchmark_litgpt.py
@@ -510,3 +510,7 @@ if __name__ == "__main__":
     from jsonargparse import CLI
 
     CLI(benchmark_main)
+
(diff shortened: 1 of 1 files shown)


thunder/benchmarks/benchmark_litgpt.py
```diff
@@ -496,3 +496,7 @@ def benchmark_main(return_metrics_as_json=False, json_path="", **kwargs) -> None
     from jsonargparse import CLI
 
     CLI(benchmark_main)
+
+    # ref: https://github.com/pytorch/pytorch/blame/3af12447f85dfede191a113c052e58fa7b21a8b3/torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp#L1110-L1116
```
carmocca, line 500: nit: shorter link that is not a git blame

```suggestion
    # ref: https://github.com/pytorch/pytorch/blob/3af12447/torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp#L1110-L1116
```

carmocca (approved):
Thanks!

PR #244, opened 2024-04-22
Use runtime value of NumberProxy if it is produced by an operator (#231)

### LINKED ISSUE #231: If saved_for_backward returns NumberProxy, the value is taken from compile time, not runtime
## 🐛 Bug

When I add an operator that returns numbers, the values in the saved_for_backward are the compile time value defined in `Symbol.meta`, not the real value computed at runtime.

An example trace is like:
```py
@torch.no_grad()
@no_autocast()
def augmented_forward_fn(a):
  # a: "cuda:0 f32[2, 2]"
  t2 = get_rng_state_prim_impl(None, devices.Device("cuda:0"))  # t2: "cpu ui8[16]"
  (i3, i4) = unpack_rng_state_prim_impl(t2)
  del t2
  [t1] = nvFusion0(a, i3, i4)
    # t0 = prims.uniform_philox((2, 2), 0.0, 1.0, device=devices.Device("cuda:0"), dtype=dtypes.float32, seed=i3, offset=i4)  # t0: "cuda:0 f32[2, 2]"
    # t1 = prims.mul(t0, a)  # t1: "cuda:0 f32[2, 2]"
  i6 = operator.add(i4, 4)  # i6: "int 11"
    # i6 = prims.add(i4, 4)  # i6: "int 11"
  del i4
  t7 = pack_rng_state_prim_impl(i3, i6)  # t7: "cpu ui8[16]"
  del i3, i6
  set_rng_state_prim_impl(t7, devices.Device("cuda:0"))
  del t7
  ######### i3, i4 is not passed to backward, but the constant value 7 in the meta function
  return {'output': t1, 'flat_args': [a], 'flat_output': (t1,)}, ((), (7, 7))  

@torch.no_grad()
@no_autocast()
  # cotangents: "Collection"
  _, C1, = saved_for_backward
  clear_collection(saved_for_backward)
  del saved_for_backward
  t2, = cotangents
  clear_collection(cotangents)
  del cotangents
  i3, i4, = C1
  clear_collection(C1)
  del C1
  [t12] = nvFusion0(i3, i4, t2)
    # t0 = prims.uniform_philox((2, 2), 0.0, 1.0, device=devices.Device("cuda:0"), dtype=dtypes.float32, seed=i3, offset=i4)  # t0: "cuda:0 f32[2, 2]"
    # t12 = prims.mul(t0, t2)  # t12: "cuda:0 f32[2, 2]"
  del i3, i4, t2
  return (t12,)
```

### To Reproduce
Reproduction is based on branch `uniform_rng`

```py
import torch
import thunder

def func(a):
    b = thunder.torch.uniform_like(a, device=a.device, dtype=a.dtype)
    return b*a

a = torch.randn(2, 2, device="cuda", requires_grad=True)

jfunc = thunder.jit(func)
out = jfunc(a)

print(thunder.last_traces(jfunc)[-1])
print(thunder.last_backward_traces(jfunc)[-1])
```

kiya00:
## What does this PR do?

Try to fix #231.

The changes in this PR go through all the NumberProxy in `saved_for_backward`, replace it with the `value` in NumberProxy if it has no producer in the forward trace, otherwise keep the NumberProxy


IvanYashchuk:
Could you please explain in the issue why is it a problem and what do you want to achieve?

thunder/core/transforms.py
```diff
@@ -3702,7 +3702,10 @@ def _update_forward_with_new_saved_for_backward(forward_trace: Trace, saved_for_
         saved_for_backward (Sequence[Variable]): Saved_for_backward to use to
             update the forward trace.
     """
-    saved_for_backward = tree_map(lambda x: x.value if isinstance(x, NumberProxy) else x, saved_for_backward)
+    forward_trace_producers = utils.producers(forward_trace)
+    saved_for_backward = tree_map(
```
jjsjann123, line 3706: nitpick: a comment explaining why we'd skip baking in compile time constant `x.value` would be great.

A TODO would be even better: Why we'd ever want to bake in number proxies and that's a future target that we should resolve.

IvanYashchuk:
@jjsjann123, with this PR your example script from https://github.com/Lightning-AI/lightning-thunder/issues/541#issuecomment-2159734931 works!

thunder/core/transforms.py
IvanYashchuk: I agree, the comment on why is needed here. It's missing in a lot of places in Thunder in general...

This line of code is responsible of sending incorrect `saved_for_backward` to this function:
https://github.com/Lightning-AI/lightning-thunder/blob/067f15aae47ad71229732ca6c35a5d190135e48c/thunder/core/rematerialization.py#L643

There's another buggy usage https://github.com/Lightning-AI/lightning-thunder/blob/067f15aae47ad71229732ca6c35a5d190135e48c/thunder/core/rematerialization.py#L501

The fix can live inside `_update_forward_with_new_saved_for_backward` or `saved_for_backward` argument should be fixed at the callsites of `_update_forward_with_new_saved_for_backward`. Currently "saved_for_backward" is taken from the backward trace which has all inputs proxified and it's a mistake to put this information into the forward trace output because not all numbers are actually produced in forward.

IvanYashchuk (changes requested):
@kiya00, I have reopened this pull request because I think this is the correct fix for the bugs we have. I'm sorry that I missed it somehow.

Could you please add this example https://github.com/Lightning-AI/lightning-thunder/issues/541#issuecomment-2159734931 as a test case.

thunder/core/transforms.py
jjsjann123: Let's move forward with this PR. I stand corrected, my [earlier comment](https://github.com/Lightning-AI/lightning-thunder/issues/541#issuecomment-2161521027) was mistaken :stuck_out_tongue:

jjsjann123 (approved):

kiya00:
Hi @IvanYashchuk , I have added the test, please take another look~

IvanYashchuk (approved):

IvanYashchuk:

t-vi (approved):
Thank you @kiya00 @jjsjann123 @IvanYashchuk

PR #245, opened 2024-04-22
[docs] add some classes/methods, fix ref, fix `no_sync` equations

crcrpar:
cc @borda @apaz-cli

### DIFF
--- a/thunder/common.py
+++ b/thunder/common.py
@@ -56,6 +56,12 @@ import numpy as np
 class CompileStats:
     """A class holding statistics and caches for a compiled function.
 
+    .. note::
+        It is highly recommended that some attributes such as :attr:`CompileStats.last_traces` and
+        :attr:`CompileStats.last_backward_traces` via :func:`thunder.last_traces` and
+        :func:`thunder.last_backward_traces`, respectively.
+        See :mod:`thunder` for more of such utility functions.
+
     Attributes:
         last_executed:
         last_traces (Sequence[TraceCtx]):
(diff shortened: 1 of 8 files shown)


thunder/common.py
```diff
@@ -137,6 +172,17 @@ def last_computation_execution_time(self, /) -> int:
 #   like additional_param_names
 # TODO RC1 Rename this to CompileOptions
 class CompileData:
+    """A class holding data about the compiled object.
+
+    Data include statistics about how it's been called such as ``additional_param_names``.
+
+    Attributes:
+        additional_param_names:
```
t-vi, line 180: Let's not have these, they're on their way out.

thunder/common.py
```diff
@@ -53,6 +54,40 @@
 # TODO RC1 Update last_executed to last_computation
 # TODO RC1 Review how autograd traces are presented
 class CompileStats:
+    """A class holding statistics and caches for a compiled function.
+
+    Attributes:
+        last_executed:
```
t-vi, line 60: Thank you for looking at these. Could you add a comment that it is preferable to use the top-level thunder.last_traces etc. functions where applicable, please?

thunder/common.py
crcrpar: um, like when? really shortly?

t-vi: Sorry, I missed this. They're only used with thunder.compile.

t-vi (approved):
Thank you @crcrpar

PR #249, opened 2024-04-22
support `nan_to_num`

### LINKED ISSUE #176: Implement `nan_to_num`
## 🚀 Feature

Implement [Tensor.nan_to_num](https://pytorch.org/docs/stable/generated/torch.nan_to_num.html#torch-nan-to-num)

### Motivation

NeMo [text-to-image model](https://github.com/NVIDIA/NeMo/blob/main/examples/multimodal/text_to_image/imagen/imagen_training.py).

### Additional context

Note that the model actually uses the in-place `nan_to_num_` variant, but it's possible that we can work with the NeMo team to translate this to the out-of-place version for the near term.

k223kim:
<details>

</details>

## What does this PR do?

Fixes #176.

## PR review

## Did you have fun?

One of the best part of my day!


thunder/tests/opinfos.py
```diff
@@ -2493,6 +2493,54 @@ def where_sample_generator(op, device, dtype, requires_grad, **kwargs):
 conditional_and_mask_ops.append(where_opinfo)
 
 
+def nan_to_num_sample_generator(op, device, dtype, requires_grad, **kwargs):
+    make = partial(make_tensor, device=device, dtype=dtype, requires_grad=requires_grad)
+
+    cases = (
```
mruberry, line 2499: Nit: a comment describing the format of these tuples would be cool

thunder/tests/opinfos.py
```diff
@@ -2493,6 +2493,54 @@ def where_sample_generator(op, device, dtype, requires_grad, **kwargs):
...
+        ((4, 4), None, 1.0, None),
+        ((4, 4), None, None, 1.0),
+        ((5,), None, 1.0, 0.0),
+        ((8, 1, 6), 1, None, None),
+        ((8, 7, 5, 1), 1, 1.0, None),
+        ((8, 7, 5, 1), 1, None, 0.0),
+        ((8, 7, 5, 1), 1, 1.0, 0.0),
+    )
+
+    for a_shape, nan, posinf, neginf in cases:
```
mruberry, line 2510: By default, `make_tensor` will only generate values in the interval [-9, 9) for floating point types. See the PyTorch documentation here:

https://pytorch.org/docs/stable/testing.html#torch.testing.make_tensor


When testing `nan_to_num`, this PR should generate tensors by hand (AKA "fixtures") that include NaN, inf, and -inf values for the inexact dtypes. For example:

```
# if the dtype is a floating point type
t = torch.tensor((nan, inf, -inf, inf, ...))
```

```
# if the dtype is a complex type
t = torch.tensor((complex(nan, nan), complex(inf, 0), ...))
```

That way we'll be the sure the function is actually doing something. It might be interesting to look at what `nan_to_num` does with complex values, btw.

thunder/tests/opinfos.py
```diff
@@ -2493,6 +2493,54 @@ def where_sample_generator(op, device, dtype, requires_grad, **kwargs):
...
+        yield SampleInput(make(a_shape, dtype=dtype, requires_grad=requires_grad), nan, posinf, neginf)
+
+
+def nan_to_num_error_generator(op, device, dtype=torch.float32, **kwargs):
+    make = partial(make_tensor, device=device, dtype=dtype)
+    err_msg = "out is not None which is currently unsupported"
```
mruberry, line 2516: Nice error generator

thunder/torch/__init__.py
```diff
@@ -1525,6 +1525,37 @@ def where(pred: TensorLike, a: Number | TensorLike, b: Number | TensorLike, /) -
...
+    posinf: None | Number = None,
+    neginf: None | Number = None,
+    /,
+    out: None | TensorLike = None,
+):
+
+    utils.check(out is None, lambda: "out is not None which is currently unsupported", NotImplementedError)
+
+    if dtypes.is_boolean_dtype(a.dtype) or dtypes.is_integer_dtype(a.dtype):
+        # NOTE PyTorch returns a.clone()
```
mruberry, line 1541: Nice note

thunder/tests/opinfos.py
```diff
@@ -2493,6 +2493,54 @@ def where_sample_generator(op, device, dtype, requires_grad, **kwargs):
...
+            ),
+        ),
+        NotImplementedError,
+        err_msg,
+    )
+
+
+nan_to_num_opinfo = OpInfo(
+    ltorch.nan_to_num,
+    supports_grad=True,
```
mruberry, line 2536: The test failures are due to missing functionality in the "phantom grad" transform, and those tests are controlled by this setting (which has a very confusing name!). fyi @IvanYashchuk, who's looking at unifying grad transforms, to maybe remove/replace/rename this setting so it's not so confusing.

If this line is removed and the test inputs are updated to include extremal values, then I think this PR will work as expected. Let's make those updates and mark it ready for review!

mruberry (commented):
Great start, @k223kim! I made some comments for you review

thunder/tests/opinfos.py
k223kim: Wow, can't believe I missed this. Thanks so much! 🚀

thunder/torch/__init__.py
```diff
@@ -1589,6 +1589,37 @@ def where(
...
+        return a
```
nikitaved, line 1606: To make it a clone, what about returning `a - 0` for integer types and `a | a` for boolean types?

thunder/torch/__init__.py
```diff
@@ -1589,6 +1589,37 @@ def where(
...
+
+    if nan is None:
+        nan = 0.0
+
+    if posinf is None:
+        posinf = torch.finfo(to_torch_dtype(a.dtype)).max
```
nikitaved, line 1612: Do you know what happens in PyTorch if the underlying type is, say, `float`, but a max for `double` is provided?

k223kim: Ah interesting! In such case, `nan_to_num` doesn't replace the values. 
```python
a = torch.randn(3)
a[0] = float("inf")
b = a.double()
b_max = torch.finfo(b.dtype).max
a_out = torch.nan_to_num(a, nan=None, posinf=b_max)
print(a == a_out) # tensor([True, True, True])
```

thunder/torch/__init__.py
k223kim: I can definitely do that!

thunder/torch/__init__.py
```diff
@@ -1589,6 +1589,56 @@ def where(
...
+
+    utils.check(out is None, lambda: "out is not None which is currently unsupported", NotImplementedError)
+
+    if dtypes.is_boolean_dtype(a.dtype):
+        # NOTE PyTorch returns a.clone()
+        return a - 0
+
+    if dtypes.is_integer_dtype(a.dtype):
+        # NOTE PyTorch returns a.clone()
+        return a | a
```
nikitaved, line 1610: Shouldn't it be the other way around? Also, it seems like `a | a` works even with integer types in PyTorch, not sure about Thunder though...

k223kim: Oh wow, my bad. I was blind for a second. Let me fix that in like 2 hours.

thunder/tests/opinfos.py
```diff
@@ -2531,6 +2531,67 @@ def where_error_generator(op, device, dtype=torch.float32, **kwargs):
...
+def nan_to_num_sample_generator(op, device, dtype, requires_grad, **kwargs):
+    make = partial(make_tensor, device=device, dtype=dtype, requires_grad=requires_grad)
+
+    a = make((4, 4), dtype=dtype, requires_grad=requires_grad)
+    double_max = torch.finfo(torch.float64).max
+    double_min = torch.finfo(torch.float64).max
+    if dtype == torch.FloatType:
+        a = torch.tensor((0, float("nan"), float("inf"), -float("inf")))
+    elif dtype == torch.ComplexType:
+        a = torch.tensor((complex(0, 0), complex(float("nan"), float("nan")), complex(float("inf"), -float("inf"))))
```
mruberry, line 2543: nit: this code would probably be more readable and easier to write by assigning to the names `nan` `inf` and `neginf` like `nan = float('nan')`

Let's also add the following numbers to the complex tensor:

- `complex(nan, 0)`
- `complex(nan, inf)`
- `complex(inf, 0)`
-  `complex(0, inf)`
- `complex(-inf, 0)` 
- `complex(nan, 5)`
- `complex(inf, 3)`

thunder/tests/opinfos.py
```diff
@@ -2531,6 +2531,67 @@ def where_error_generator(op, device, dtype=torch.float32, **kwargs):
 conditional_and_mask_ops.append(where_opinfo)
 
 
+def nan_to_num_sample_generator(op, device, dtype, requires_grad, **kwargs):
+    make = partial(make_tensor, device=device, dtype=dtype, requires_grad=requires_grad)
+
+    a = make((4, 4), dtype=dtype, requires_grad=requires_grad)
+    double_max = torch.finfo(torch.float64).max
+    double_min = torch.finfo(torch.float64).max
+    if dtype == torch.FloatType:
```
mruberry, line 2540: These checks aren't quite write. This wants to check if the dtype is a float dtype or a complex dtype. The value of `dtype` will be something like `torch.float32`, which will not compare equal to `torch.FloatType`. Instead, this can query the `is_floating_point` and `is_complex` attributes of PyTorch dtypes, like this:

```
if dtype.is_floating_point:
  ...
elif dtype.is_complex:
  ...
```

thunder/torch/__init__.py
```diff
@@ -1589,6 +1589,56 @@ def where(
...
+    a_dtype_min = torch.finfo(to_torch_dtype(a.dtype)).min
+
+    if nan is None:
+        nan = 0.0
+    elif nan > a_dtype_max:
+        nan = float("inf")
+    elif nan < a_dtype_min:
+        nan = -float("inf")
+
+    if posinf is None:
```
mruberry, line 1622: Add a comment for what this line of code does, and how it's different from PyTorch (maybe with a simple example)

thunder/tests/opinfos.py
k223kim: I completely missed that, thanks so much!

thunder/torch/__init__.py
k223kim: I found it confusing to add a single comment to explain that part of the code. Rather, I have added a docstring for `nan_to_num` to explain its functionality. Also, as far as I am aware, this is identical to PyTorch's `nan_to_num`.

thunder/tests/opinfos.py
```diff
@@ -2531,6 +2531,82 @@ def where_error_generator(op, device, dtype=torch.float32, **kwargs):
 conditional_and_mask_ops.append(where_opinfo)
 
 
+def nan_to_num_sample_generator(op, device, dtype, requires_grad, **kwargs):
+    make = partial(make_tensor, device=device, dtype=dtype, requires_grad=requires_grad)
+
+    a = make((4, 4), dtype=dtype, requires_grad=requires_grad)
+    double_max = torch.finfo(torch.float64).max
+    double_min = torch.finfo(torch.float64).max
```
mruberry, line 2539: This line incorrectly sets `double_min` to the maximum representable double value

k223kim: Thanks for catching that. Fixed!

thunder/torch/__init__.py
```diff
@@ -1589,6 +1589,79 @@ def where(
...
+
+    if dtypes.is_integer_dtype(a.dtype):
+        # NOTE PyTorch returns a.clone()
+        return a - 0
+
+    a_dtype_max = torch.finfo(to_torch_dtype(a.dtype)).max
+    a_dtype_min = torch.finfo(to_torch_dtype(a.dtype)).min
+    inf = float("inf")
+
+    if nan is None:
```
mruberry, line 1638: Nit: setting nan, posinf, and neginf like this might be more readable with a helper function defined inline

mruberry (commented):
Looks good, @k223kim! I added one last comment about a small error

thunder/torch/__init__.py
k223kim: @mruberry Do you mean something like:
```python
nan = 0.0 if nan is None else inf if nan > a_dtype_max else -inf if a < a_dtype_min else nan
```

mruberry: More like 

```
def convert(x, if_none):
  if x is None:
    return if_none

  if x > a_dtype_max:
    return inf

  if x < a_dtype_min:
    return -inf

  return x

nan = convert(nan, 0)
posinf = convert(posinf, a_dtype_max)
neginf = convert(neginf, a_dtype_min)
```

What do you think of something like that?

k223kim: Oh, I see what you mean. I do agree that your code is more readable :) will update!

mruberry (approved):

mruberry:
Cool! Thanks @k223kim!

PR #253, opened 2024-04-23
arg of `v`, not `value`

crcrpar:


crcrpar:
embarrassingly I'm quite not sure if this primitive is used or not

t-vi (approved):
Just like the unpack prims, I think the only use of them is through printing (in epilogues), so this is unused.
But clearly, it's better to have working unused then non-working.

PR #259, opened 2024-04-24
[fsdp] Add option of AllGather rate limiting (mainly for ZERO2)

crcrpar:
Adding the new argument of `apply_rate_limting` to `thunder.distributed.fsdp` so that we can try rate limiting of AllGather, for especially when ZERO2 is used.

The major changes this pr brings are options of (a) trying rate limiting with zero2 and (b) turning off it for zero3.

Previously I tried this with llama-2-7b-hf-ish models with zero2 but the perf gain wasn't seen.
I however think it might make sense to have an option to switch on/off it.


lantiga:
Hey @crcrpar, should we take this out of draft?

PR #273, opened 2024-04-25
Implement `Tensor.type`

### LINKED ISSUE #177: Implement `Tensor.type`
## 🚀 Feature

Implement [Tensor.type](https://pytorch.org/docs/stable/generated/torch.Tensor.type.html#torch.Tensor.type)

### Motivation

NeMo [text-to-image model](https://github.com/NVIDIA/NeMo/blob/main/examples/multimodal/text_to_image/imagen/imagen_training.py).

cc @tfogal

k223kim:
<details>

</details>

## What does this PR do?

Fixes #177. At the moment, I did not implement `non_blocking=True`. Any suggestions would be appreciated! Also, the current CI test that fails seems irrelevant to this PR.

## PR review

## Did you have fun?

As always 🚀

### DIFF
--- a/thunder/tests/opinfos.py
+++ b/thunder/tests/opinfos.py
@@ -2752,27 +2752,46 @@ data_movement_ops = []
 def type_sample_generator_tensor(op, device, dtype, requires_grad, **kwargs):
     # dtype is not None
     # expected to return tensor
-    make = partial(make_tensor, device=device, dtype=dtype, requires_grad=requires_grad)
-    DTYPE_STR = {
-        torch.float32: "torch.FloatTensor",
-        torch.float64: "torch.DoubleTensor",
-        torch.float16: "torch.HalfTensor",
-        torch.bfloat16: "torch.BFloat16Tensor",
-        torch.uint8: "torch.ByteTensor",
-        torch.int8: "torch.CharTensor",
-        torch.int16: "torch.ShortTensor",
-        torch.int32: "torch.IntTensor",
-        torch.long: "torch.LongTensor",
-        torch.bool: "torch.BoolTensor",
+
+    _torch_dtype_to_old_torch_typestring_map = {
+        torch.float32: "FloatTensor",
+        torch.float64: "DoubleTensor",
+        torch.float16: "HalfTensor",
+        torch.bfloat16: "BFloat16Tensor",
+        torch.uint8: "ByteTensor",
+        torch.int8: "CharTensor",
+        torch.int16: "ShortTensor",
+        torch.int32: "IntTensor",
+        torch.long: "LongTensor",
+        torch.bool: "BoolTensor",
     }
 
-    yield SampleInput(make(4, 4), dtype)
-    yield SampleInput(make(4, 4), DTYPE_STR[dtype])
+    make = partial(make_tensor, device=device, dtype=dtype, requires_grad=requires_grad)
+
+    to_dtype = torch.complex128 if dtype.is_complex else torch.float64
+
+    yield SampleInput(make(4, 4, device=device), dtype)
+    yield SampleInput(make(4, 4, device=device), dtype=to_dtype)
+    # below can be deleted if we don't support strings
+    yield SampleInput(make(4, 4, device=device), f"torch.{_torch_dtype_to_old_torch_typestring_map[dtype]}")
+
+    # Explictly pass device
+    if torch.device(device).type == "cuda":
+        yield SampleInput(make(4, 4, device=device), f"torch.cuda.{_torch_dtype_to_old_torch_typestring_map[dtype]}")
+
+
+# kind of redundant?
+def type_error_generator_tensor(op, device, dtype=torch.float32, **kwargs):
+    make = partial(make_tensor, device=device, dtype=dtype, requires_grad=False)
+
+    err_msg = r"type\(\): `non_blocking==True` is currently not supported."
+    yield SampleInput(make(3, 3), dtype, non_blocking=True), RuntimeError, err_msg
 
 
 type_opinfo_tensor = OpInfo(
     ltorch.type,
     sample_input_generator=type_sample_generator_tensor,
+    error_input_generator=type_error_generator_tensor,
     torch_reference=torch.Tensor.type,
 )
 
@@ -2787,6 +2806,14 @@ def type_sample_generator_str(op, device, dtype, requires_grad, **kwargs):
     yield SampleInput(make(4, 4))
 
 
+# kind of redundant?
+def type_error_generator_str(op, device, dtype=torch.float32, **kwargs):
+    make = partial(make_tensor, device=device, dtype=dtype, requires_grad=False)
+
+    err_msg = r"type\(\): `non_blocking==True` is currently not supported."
+    yield SampleInput(make(3, 3), non_blocking=True), RuntimeError, err_msg
+
+
 # comparing strings (NOTE: assert_close does not support string comparison)
 def string_compare(actual, expected, **kwargs):
     assert actual == expected
@@ -2795,6 +2822,7 @@ def string_compare(actual, expected, **kwargs):
 type_opinfo_str = OpInfo(
     ltorch.type,
     sample_input_generator=type_sample_generator_str,
+    error_input_generator=type_error_generator_str,
     torch_reference=torch.Tensor.type,
     test_directives=(
         DecorateInfo(
--- a/thunder/tests/opinfos.py
+++ b/thunder/tests/opinfos.py
@@ -2749,6 +2749,64 @@ data_movement_ops = []
 # data_movement_ops.append(convert_element_type_opinfo)
 
 
+def type_sample_generator_tensor(op, device, dtype, requires_grad, **kwargs):
+    # dtype is not None
+    # expected to return tensor
+    make = partial(make_tensor, device=device, dtype=dtype, requires_grad=requires_grad)
+    DTYPE_STR = {
+        torch.float32: "torch.FloatTensor",
+        torch.float64: "torch.DoubleTensor",
+        torch.float16: "torch.HalfTensor",
+        torch.bfloat16: "torch.BFloat16Tensor",
+        torch.uint8: "torch.ByteTensor",
+        torch.int8: "torch.CharTensor",
+        torch.int16: "torch.ShortTensor",
+        torch.int32: "torch.IntTensor",
+        torch.long: "torch.LongTensor",
+        torch.bool: "torch.BoolTensor",
+    }
+
+    yield SampleInput(make(4, 4), dtype)
+    yield SampleInput(make(4, 4), DTYPE_STR[dtype])
+
+
+type_opinfo_tensor = OpInfo(
+    ltorch.type,
+    sample_input_generator=type_sample_generator_tensor,
+    torch_reference=torch.Tensor.type,
+)
+
+data_movement_ops.append(type_opinfo_tensor)
+
+
+def type_sample_generator_str(op, device, dtype, requires_grad, **kwargs):
+    # dtype is None
+    # expected to return string
+    make = partial(make_tensor, device=device, dtype=dtype, requires_grad=requires_grad)
+
+    yield SampleInput(make(4, 4))
+
+
+# comparing strings (NOTE: assert_close does not support string comparison)
+def string_compare(actual, expected, **kwargs):
+    assert actual == expected
+
+
+type_opinfo_str = OpInfo(
+    ltorch.type,
+    sample_input_generator=type_sample_generator_str,
+    torch_reference=torch.Tensor.type,
+    test_directives=(
+        DecorateInfo(
+            custom_comparator(string_compare),
+            "test_core_vs_torch_consistency",
+        ),
+    ),
+)
+
+data_movement_ops.append(type_opinfo_str)
+
+
 def to_sample_generator(op, device, dtype, requires_grad, **kwargs):
     make = partial(make_tensor, device=device, dtype=dtype, requires_grad=requires_grad)
 
--- a/thunder/tests/opinfos.py
+++ b/thunder/tests/opinfos.py
@@ -2773,11 +2773,11 @@ def type_sample_generator_tensor(op, device, dtype, requires_grad, **kwargs):
     yield SampleInput(make(4, 4, device=device), dtype)
     yield SampleInput(make(4, 4, device=device), dtype=to_dtype)
     # below can be deleted if we don't support strings
-    yield SampleInput(make(4, 4, device=device), f"torch.{_torch_dtype_to_old_torch_typestring_map[dtype]}")
+    yield SampleInput(make(4, 4, device=device), f"torch.{_torch_dtype_to_old_torch_typestring_map[to_dtype]}")
 
     # Explictly pass device
     if torch.device(device).type == "cuda":
-        yield SampleInput(make(4, 4, device=device), f"torch.cuda.{_torch_dtype_to_old_torch_typestring_map[dtype]}")
+        yield SampleInput(make(4, 4, device=device), f"torch.cuda.{_torch_dtype_to_old_torch_typestring_map[to_dtype]}")
 
 
 # kind of redundant?
--- a/thunder/torch/__init__.py
+++ b/thunder/torch/__init__.py
@@ -187,6 +187,19 @@ def is_cuda(a: TensorLike, /) -> bool:
 register_method("size", size)
 
 
+@torchsymbol(torch.Tensor.type, is_method=True)
+def type(a: TensorLike, dtype: None | dtypeLike, non_blocking: bool = False, /) -> TensorLike:
+    utils.check(
+        not non_blocking,
+        lambda: f"type(): `non_blocking==True` is currently not supported.",
+        exception_type=NotImplementedError,
+    )
+    if dtype is None:
+        return a.dtype
+    return clang.maybe_convert_to_dtype(a, to_dtype(dtype))
+
+
+register_method("type", type)
 #
 # Data movement and transformation operations
 #
--- a/thunder/torch/__init__.py
+++ b/thunder/torch/__init__.py
@@ -188,15 +188,63 @@ register_method("size", size)
 
 
 @torchsymbol(torch.Tensor.type, is_method=True)
-def type(a: TensorLike, dtype: None | dtypeLike = None, non_blocking: bool = False, /) -> TensorLike:
+def type(a: TensorLike, dtype: None | str | dtypeLike = None, non_blocking: bool = False, /) -> str | TensorLike:
     utils.check(
         not non_blocking,
         lambda: f"type(): `non_blocking==True` is currently not supported.",
         exception_type=NotImplementedError,
     )
+
+    DTYPE_STR = {
+        torch.float32: "torch.FloatTensor",
+        torch.float64: "torch.DoubleTensor",
+        torch.float16: "torch.HalfTensor",
+        torch.bfloat16: "torch.BFloat16Tensor",
+        torch.uint8: "torch.ByteTensor",
+        torch.int8: "torch.CharTensor",
+        torch.int16: "torch.ShortTensor",
+        torch.int32: "torch.IntTensor",
+        torch.long: "torch.LongTensor",
+        torch.bool: "torch.BoolTensor",
+    }
     if dtype is None:
-        return to_torch_dtype(a.dtype)
-    return clang.maybe_convert_to_dtype(a, to_dtype(dtype))
+        # returns the type of the input tensor in string
+        torch_dtype = to_torch_dtype(a.dtype)
+        torch_dtype = DTYPE_STR.get(torch_dtype)
+        if a.device.devicetype is devices.DeviceType.CUDA:
+            t, _dtype = torch_dtype.split(".")
+            torch_dtype = f"{t}.cuda.{_dtype}"
+        return torch_dtype
+
+    TORCH_DTYPES = {
+        "torch.FloatTensor": torch.float32,
+        "torch.DoubleTensor": torch.float64,
+        "torch.HalfTensor": torch.float16,
+        "torch.BFloat16Tensor": torch.bfloat16,
+        "torch.ByteTensor": torch.uint8,
+        "torch.CharTensor": torch.int8,
+        "torch.ShortTensor": torch.int16,
+        "torch.IntTensor": torch.int32,
+        "torch.LongTensor": torch.long,
+        "torch.BoolTensor": torch.bool,
+    }
+    if isinstance(dtype, str):
+        parse_dtype = dtype.split(".")
+        if len(parse_dtype) == 2:
+            utils.check(dtype in TORCH_DTYPES, lambda: f"type(): invalid type: {dtype}.", exception_type=ValueError)
+            dtype = TORCH_DTYPES.get(dtype)
+        else:
+            t, device, torch_dtype = dtype.split(".")
+            utils.check(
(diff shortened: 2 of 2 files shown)


k223kim:
Hey @mruberry, would you be able to check this draft out? Would appreciate it :)

thunder/torch/__init__.py
```diff
@@ -187,6 +187,68 @@ def is_cuda(a: TensorLike, /) -> bool:
...
 
+@torchsymbol(torch.Tensor.type, is_method=True)
+def type(a: TensorLike, dtype: None | str | dtypeLike = None, non_blocking: bool = False, /) -> str | TensorLike:
+    utils.check(
+        not non_blocking,
+        lambda: f"type(): `non_blocking==True` is currently not supported.",
+        exception_type=NotImplementedError,
+    )
+
+    DTYPE_STR = {
```
mruberry, line 198: This is to interesting. This PR is absolutely correct that PyTorch does support putting in a string for `dtype`, and that behavior is used by some programs, but not many. 

I wonder if we should just throw a `NotImplementedError` if a string is passed as the `dtype` argument for now, and see if we get a request to support it

thunder/torch/__init__.py
```diff
@@ -187,6 +187,68 @@ def is_cuda(a: TensorLike, /) -> bool:
 register_method("size", size)
 
 
+@torchsymbol(torch.Tensor.type, is_method=True)
+def type(a: TensorLike, dtype: None | str | dtypeLike = None, non_blocking: bool = False, /) -> str | TensorLike:
+    utils.check(
```
mruberry, line 192: Nice `NotImplementedError`

thunder/torch/__init__.py
```diff
@@ -187,6 +187,68 @@ def is_cuda(a: TensorLike, /) -> bool:
...
+        torch.bfloat16: "torch.BFloat16Tensor",
+        torch.uint8: "torch.ByteTensor",
+        torch.int8: "torch.CharTensor",
+        torch.int16: "torch.ShortTensor",
+        torch.int32: "torch.IntTensor",
+        torch.long: "torch.LongTensor",
+        torch.bool: "torch.BoolTensor",
+    }
+
+    if dtype is None:
```
mruberry, line 211: I suggest separating this function into a few functions for readability.

If `dtype is None` then how about calling a helper function, `_dtype_to_old_torch_typestring(a.dtype)`? Then the logic for that can go there

thunder/torch/__init__.py
k223kim: Yeah, that is also possible and reasonable. However as per the issue tag, it seems that it is trying to implement Megatron? When I do a quick search about the usage of `type()` it is something like the [following](https://arc.net/l/quote/fmtstkye):
```python
                    if param.type() in ['torch.cuda.HalfTensor', 'torch.cuda.BFloat16Tensor']:
                        float16_params_this_group.append(param)
```
What do you think? When observing the use case, it seems like it is better to support string.

thunder/torch/__init__.py
```diff
@@ -187,6 +187,68 @@ def is_cuda(a: TensorLike, /) -> bool:
...
+        # returns the type of the input tensor in string
+        torch_dtype = to_torch_dtype(a.dtype)
+        torch_dtype = DTYPE_STR.get(torch_dtype)
+        if a.device.devicetype is devices.DeviceType.CUDA:
```
mruberry, line 215: Instead of splitting the string, map like this:

```
torch.bool: "BoolTensor"
```

and construct the final string like this: 

```
f"torch.{devicetype_str}{dtype_str}"
```

where `devicetype_str` can be acquired like this:

```
devicetype_str: str = ""
if a.device.devicetype is not devices.DeviceType.CPU:
  devicetype_str = f"{devices.devicetype_string(a.device.devicetype)}."
```

thunder/torch/__init__.py
```diff
@@ -187,6 +187,68 @@ def is_cuda(a: TensorLike, /) -> bool:
...
+            t, _dtype = torch_dtype.split(".")
+            torch_dtype = f"{t}.cuda.{_dtype}"
+        return torch_dtype
+
+    TORCH_DTYPES = {
```
mruberry, line 220: Let's put this logic in a helper like `_old_torch_typestring_to_device_and_dtype(str)`

thunder/torch/__init__.py
k223kim: Wow! You are amazing.

thunder/torch/__init__.py
t-vi: Yeah, you have a point. I think it's terrible code compared to `param.is_cuda and param.dtype in (torch.float16, torch.bfloat16)`  but I guess our mission is not to run only good code. :)

thunder/torch/__init__.py
```diff
@@ -187,6 +187,68 @@ def is_cuda(a: TensorLike, /) -> bool:
...
+        "torch.HalfTensor": torch.float16,
+        "torch.BFloat16Tensor": torch.bfloat16,
+        "torch.ByteTensor": torch.uint8,
+        "torch.CharTensor": torch.int8,
+        "torch.ShortTensor": torch.int16,
+        "torch.IntTensor": torch.int32,
+        "torch.LongTensor": torch.long,
+        "torch.BoolTensor": torch.bool,
+    }
+    if isinstance(dtype, str):
```
mruberry, line 232: Let's refine this logic to look like this:

```
if isinstance(dtype, str):
  dev, dtyp = _old_torch_typestring_to_device_and_dtype(dtype)
else:
  # Question here -- if a device is not specified and the tensor is a CUDA tensor, will this create a tensor
  # on a CPU device, or leave the tensor on a CUDA device? This would change how dev is set here
  dev = a.device
  dtyp = dtype

return to(a, dev, dtyp)
```

thunder/torch/__init__.py
mruberry: This function should have a comment showing some examples of these old type strings, too, since many readers may not be familiar with them.

The logic might look like this:

```
_torch_dtype_to_old_torch_typestring_map = {
  torch.float16: "HalfTensor",
  ...
}

_old_torch_typestring_to_torch_dtype_map = {v: k for k, v in _torch_dtype_to_old_torch_typestring_map.items()}

def _old_torch_typestring_to_devicetype_and_dtype(typestring: str) -> tuple[DeviceLike, dtypeLike]:

  # Two cases:
  #    - torch.DtypeTensor
  #    - torch.device.DtypeTensor 

  _, *dev_and_dtype = typestring.split('.')

  if len(dev_and_dtype) == 1:
    dtype_str, = dev_and_dtype
    return <cpu device>,  _old_torch_typestring_to_torch_dtype_map[dtype_str]

  if len(dev_and_dtype) == 2:
    dtype_str, devicetype_str = dev_and_dtype
    return <convert devicetype_str to devicetype>, _old_torch_typestring_to_torch_dtype_map[dtype_str]

  # Assertion error -- expected the string to split into one or two elements
  baseutils.check(False, ...)
```

thunder/torch/__init__.py
```diff
@@ -187,6 +187,68 @@ def is_cuda(a: TensorLike, /) -> bool:
 register_method("size", size)
 
 
+@torchsymbol(torch.Tensor.type, is_method=True)
+def type(a: TensorLike, dtype: None | str | dtypeLike = None, non_blocking: bool = False, /) -> str | TensorLike:
```
mruberry, line 191: A first version of this PR might be:

```
@torchsymbol(torch.Tensor.type, is_method=True)
def type(a: TensorLike, dtype: None | str | dtypeLike = None, non_blocking: bool = False, /) -> str | TensorLike:
  <NotImplementedError if dtype is a str>
  <NotImplementedError if dtype is None>
  <NotImplementedError if non_blocking is True>

  return to(a, dtype)
```

But there's already a lot of the needed work to parse the old torch typestrings, so if that's interesting then we can include it. I made some comments for how this PR might refactor that logic to be a little easier to work with

thunder/tests/opinfos.py
```diff
@@ -2749,6 +2749,64 @@ def tril_sample_generator(op, device, dtype, requires_grad, **kwargs):
 # data_movement_ops.append(convert_element_type_opinfo)
 
 
+def type_sample_generator_tensor(op, device, dtype, requires_grad, **kwargs):
```
mruberry, line 2752: Take a look at the sample generator for `to`

https://github.com/k223kim/lightning-thunder/blob/2430d477ed6bf70e3cfd58aa52dffdb21d89fb95/thunder/tests/opinfos.py#L2810

This will currently just check that a tensor can be cast to its current dtype, but it'd be interesting to verify it can be upcast (to double or complex double), too

thunder/tests/opinfos.py
```diff
@@ -2749,6 +2749,64 @@ def tril_sample_generator(op, device, dtype, requires_grad, **kwargs):
...
+    assert actual == expected
+
+
+type_opinfo_str = OpInfo(
+    ltorch.type,
+    sample_input_generator=type_sample_generator_str,
+    torch_reference=torch.Tensor.type,
+    test_directives=(
+        DecorateInfo(
+            custom_comparator(string_compare),
```
mruberry, line 2801: This is a really, really nice use of `custom_comparator`. It's a pain that torch's `assert_close` does not handle strings.

mruberry (commented):
Cool stuff, @k223kim! I left some comments for you review. These old torch typestrings are a pain, right? I'm glad torch moved away from them.

I look forward to hearing your thoughts!

thunder/torch/__init__.py
mruberry: Awesome to look at the actual use case -- thanks @k223kim

k223kim:
Hi @mruberry thank you for carefully going through my code despite the fact that it may not be needed at all. I should have discussed this earlier before implementing it, but again, thanks for the detailed comments.
I do agree that handling strings are a pain and is not good coding (as pr @t-vi ). Let me remove the current implementation and throw an `NotImplementedError` for cases where `dtype` is string. I guess now,  I am really questioning if this implementation is needed at all. Cause, if we don't accept strings, how is it different from `to`? 🤔

mruberry:
> Hi @mruberry thank you for carefully going through my code despite the fact that it may not be needed at all. I should have discussed this earlier before implementing it, but again, thanks for the detailed comments. I do agree that handling strings is a pain and is not good coding (as pr @t-vi ). Let me remove the current implementation and throw an `NotImplementedError` for cases where `dtype` is string. I guess now, I am really questioning if this implementation is needed at all. Cause, if we don't accept strings, how is it different from `to`? 🤔

I think you pointed out that the particular use case we're interested in DOES use strings, and this PR already does a lot of the work, so maybe let's keep them? Just because something is rare doesn't mean it never happens

k223kim:
Hey @mruberry! I have updated this PR based on your comments :) Although I am questioning if the string support is needed or not, I have included them. Feel free to drop it if that is not necessary. Would appreciate if you can check this out!

thunder/tests/opinfos.py
```diff
@@ -2749,6 +2749,92 @@ def tril_sample_generator(op, device, dtype, requires_grad, **kwargs):
...
+    }
+
+    make = partial(make_tensor, device=device, dtype=dtype, requires_grad=requires_grad)
+
+    to_dtype = torch.complex128 if dtype.is_complex else torch.float64
+
+    yield SampleInput(make(4, 4, device=device), dtype)
+    yield SampleInput(make(4, 4, device=device), dtype=to_dtype)
+    # below can be deleted if we don't support strings
+    yield SampleInput(make(4, 4, device=device), f"torch.{_torch_dtype_to_old_torch_typestring_map[dtype]}")
```
mruberry, line 2776: Use `to_dtype` for the key to the map here, too

thunder/tests/opinfos.py
```diff
@@ -2749,6 +2749,92 @@ def tril_sample_generator(op, device, dtype, requires_grad, **kwargs):
...
+
+    # Explictly pass device
+    if torch.device(device).type == "cuda":
+        yield SampleInput(make(4, 4, device=device), f"torch.cuda.{_torch_dtype_to_old_torch_typestring_map[dtype]}")
```
mruberry, line 2780: Use `to_dtype` for the key to the map here, too

thunder/torch/__init__.py
```diff
@@ -186,6 +186,92 @@ def is_cuda(a: TensorLike, /) -> bool:
...
+    torch.int16: "ShortTensor",
+    torch.int32: "IntTensor",
+    torch.long: "LongTensor",
+    torch.bool: "BoolTensor",
+}
+
+_old_torch_typestring_to_torch_dtype_map = {v: k for k, v in _torch_dtype_to_old_torch_typestring_map.items()}
+
+
+def _device_and_dtype_to_old_torch_typestring(device: DeviceLike, dtype: dtypeLike):
```
mruberry, line 205: nit: add the return annotation `-> str:`

thunder/torch/__init__.py
```diff
@@ -186,6 +186,92 @@ def is_cuda(a: TensorLike, /) -> bool:
...
+    #    - torch.device.DtypeTensor
+
+    _, *dev_and_dtype = typestring.split(".")
+
+    if len(dev_and_dtype) == 1:
+        (dtype_str,) = dev_and_dtype
+        return "cpu", _old_torch_typestring_to_torch_dtype_map[dtype_str]
+
+    if len(dev_and_dtype) == 2:
+        dtype_str, devicetype_str = dev_and_dtype
```
mruberry, line 227: Aren't these two swapped? Won't `dtype_str` be 'cuda'?

mruberry: If these two are swapped, then how are the OpInfo tests passing?

thunder/torch/__init__.py
```diff
@@ -186,6 +186,92 @@ def is_cuda(a: TensorLike, /) -> bool:
...
+        return dtype_str, _old_torch_typestring_to_torch_dtype_map[devicetype_str]
+
+    # Assertion error -- expected the string to split into one or two elements
+    utils.check(
+        False,
+        lambda: f"type(): dtype format does not match torch.Tensor dtype nor old torch typestring format",
```
mruberry, line 233: `f"type(): unrecognized torch typestring {typestring}`

thunder/torch/__init__.py
```diff
@@ -186,6 +186,92 @@ def is_cuda(a: TensorLike, /) -> bool:
...
+        exception_type=ValueError,
+    )
+
+
+@torchsymbol(torch.Tensor.type, is_method=True)
+def type(
+    a: TensorLike, /, dtype: None | str | dtypeLike = None, non_blocking: bool = False, **kwargs
```
mruberry, line 240: I don't think this wants to accept `**kwargs`?

thunder/torch/__init__.py
```diff
@@ -186,6 +186,92 @@ def is_cuda(a: TensorLike, /) -> bool:
...
+        # on CUDA device 0 be created?
+
+        if devtype == a.device.type:
+            dev = a.device
+        elif devtype == "cpu":
+            dev = devices.DeviceType.CPU
+        else:
+            dev = devices.DeviceType.CUDA
+    else:
+        # Question here -- if a device is not specified and the tensor is a CUDA tensor, will this create a tensor
```
mruberry, line 266: Let's answer this question and update the PR

thunder/torch/__init__.py
```diff
@@ -186,6 +186,92 @@ def is_cuda(a: TensorLike, /) -> bool:
...
+        # Follow-up question: what if the old string says "CUDA" but the tensor is on CUDA device 1, will a tensor
+        # on CUDA device 0 be created?
+
+        if devtype == a.device.type:
+            dev = a.device
+        elif devtype == "cpu":
+            dev = devices.DeviceType.CPU
+        else:
+            dev = devices.DeviceType.CUDA
+    else:
```
mruberry, line 265: In this `else` branch, what is `dtype` assumed to be? A comment explaning this would be helpful

thunder/torch/__init__.py
```diff
@@ -186,6 +186,92 @@ def is_cuda(a: TensorLike, /) -> bool:
...
+    )
+
+    if dtype is None:
+        # returns the type of the input tensor in string
+        return _device_and_dtype_to_old_torch_typestring(a.device, a.dtype)
+
+    if isinstance(dtype, str):
+        devtype, dtype = _old_torch_typestring_to_devicetype_and_dtype(dtype)
+
+        # below if-statement handles the following case that you have mentioned:
```
mruberry, line 255: Let's answer this question and update the PR

thunder/torch/__init__.py
```diff
@@ -186,6 +186,92 @@ def is_cuda(a: TensorLike, /) -> bool:
...
+        # Follow-up question: what if the old string says "CUDA" but the tensor is on CUDA device 1, will a tensor
+        # on CUDA device 0 be created?
+
+        if devtype == a.device.type:
+            dev = a.device
+        elif devtype == "cpu":
+            dev = devices.DeviceType.CPU
+        else:
+            dev = devices.DeviceType.CUDA
```
mruberry, line 264: Let's not assume that the fallthrough is `CUDA`. I think this can work like this:

```
if devtype == a.device.type:
  # This handles the case that a tensor is already on a CUDA device, and the device type string is CUDA. In this
  # case the tensor remains on its current device.
  dev = a.device
else:
  dev = device_from_string(devtype)
```

thunder/torch/__init__.py
mruberry: I checked this -- CUDA tensor inputs will create CPU tensors if the devicetype is omitted from the typestring

mruberry: Which I think is consistent with the logic below, so this comment can probably be deleted?

mruberry (commented):
Hey @k223kim!

Cool stuff! I made some additional comments for refinements. I'm curious how a couple things work currently. Maybe there's an issue with testing? Check it out and let me know!

thunder/torch/__init__.py
k223kim: Your comments are really aesthetic.

thunder/torch/__init__.py
k223kim: As per [PyTorch](https://pytorch.org/docs/stable/generated/torch.Tensor.type.html#torch.Tensor.type), some may use the keyword `async` and PyTorch implementation accepts that keyword although it is deprecated. Thought Thunder might want to do this as well.
Should I remove it as `async` is deprecated?

thunder/torch/__init__.py
k223kim: Oh, I answered this question right below this comment. I should have made that clearer. But, thanks for answering it :)

thunder/torch/__init__.py
k223kim: I don't think these two are swapped. For instance, when the user inputs `torch.IntTensor` to a CUDA tensor, this will result in a tensor on a CPU device. So, if `len(dev_and_dtype) == 1`, I think it is safe to assume it is going to copy the tensor on a CPU device.

thunder/torch/__init__.py
k223kim: The issue with this is that, `device_from_string` requires a device number in CUDA. However, the weird PyTorch typestring does not include that.

thunder/torch/__init__.py
k223kim: Cool, that is why I think the below implementation makes sense as well (in `_old_torch_typestring_to_devicetype_and_dtype`):
```
    if len(dev_and_dtype) == 1:
        (dtype_str,) = dev_and_dtype
        return "cpu", _old_torch_typestring_to_torch_dtype_map[dtype_str]
```

thunder/torch/__init__.py
k223kim: Let me do something like this to avoid such assumption.
```python
        if devtype == a.device.type:
            dev = a.device
        elif devtype == "cpu":
            dev = devices.DeviceType.CPU
        elif devtype == "cuda":
            dev = devices.DeviceType.CUDA
        else:
            raise ValueError(f"type(): unrecognized torch typestring {dtype}")
```

thunder/torch/__init__.py
k223kim: Great idea!

k223kim:
Hi @mruberry! Thanks so much for the review! I have answered some comments and updated the code accordingly. Please let me know if there is a better way to do this :) Appreciate your help as always!

thunder/torch/__init__.py
mruberry: I understand what you're saying, but I think what I said was unclear, sorry. Let me rephrase my earlier comment. The original code is:

```
# Two cases:
    #    - torch.DtypeTensor
    #    - torch.device.DtypeTensor

    _, *dev_and_dtype = typestring.split(".")
```

at this point, won't `dev_and_dtype` be `(device, DtypeTensor)` or `(DtypeTensor,)`?

```
    if len(dev_and_dtype) == 1:
        (dtype_str,) = dev_and_dtype
        return "cpu", _old_torch_typestring_to_torch_dtype_map[dtype_str]
```

this block handles the case that `dev_and_dtype` is `(DtypeTensor,)`, and then the next block

```
    if len(dev_and_dtype) == 2:
        dtype_str, devicetype_str = dev_and_dtype
```

wouldn't `dev_and_dtype` be `(device, DtypeTensor)` here, but it's unpacked as `(dtype_str, devicetype_str)`, and that's what I'm confused by. I would expect the unpacking to be `devicetype_str, dtype_str`?

thunder/torch/__init__.py
mruberry: I appreciate your detailed looking at this function, including how it will show errors. I agree with you the current error message for that case is not very good.

However, if possible, I think we should improve the `_old_torch_typestring_to_devicetype_and_dtype` function to throw a better error message when given something ridiculous. That way any future callers of the function will also throw better error messages, so the code will be more maintainable. Otherwise each caller would have to include logic like the above.

So may before returning from `_old_torch_typestring_to_devicetype_and_dtype`, it should check that the devicetype string is inferred to be `cpu` or `cuda`?

thunder/torch/__init__.py
mruberry: Interesting! Yes — let's try not supporting it for now and see if anyone asks for `async`

thunder/torch/__init__.py
mruberry: yes it does!

thunder/torch/__init__.py
k223kim: Sounds good!

thunder/torch/__init__.py
k223kim: Oh my bad, I completely misunderstood your point. Yes, you are right! let me update this. They were swapped indeed 😅

k223kim: And to answer your above question, as I was doing something like this:
```
_old_torch_typestring_to_torch_dtype_map[devicetype_str]
```
it was passing the test! My apologies for the confusion.

thunder/torch/__init__.py
k223kim: I see what you mean. Thanks for the detailed explanation. Let me add that check in `_old_torch_typestring_to_devicetype_and_dtype`!

k223kim:
Hey @mruberry, thanks for the great explanations and my apologies for the confusion. Hope my code is now easier to understand and maintain :) Would appreciate if you can take a look! 🚀

mruberry (approved):
Nice work with a tricky operator, @k223kim!

@Borda or @t-vi may need to merge this

mruberry:
ping @t-vi

t-vi:
Thank you @k223kim @mruberry

PR #275, opened 2024-04-25
Add support for `scatter_add` gradient

rdspring1:
This PR adds support for `scatter_add` gradient. 

**Note:** The value gradient is only defined when `value.shape == index.shape`.
See https://pytorch.org/docs/stable/generated/torch.Tensor.scatter_add_.html#torch.Tensor.scatter_add_

### DIFF
--- a/thunder/core/transforms.py
+++ b/thunder/core/transforms.py
@@ -833,6 +833,21 @@ register_grad(pids.GATHER, _gather_prim_grad)
 
 
 @torchctx
+def _scatter_add_prim_grad(a: TensorProxy, /, index: TensorProxy, value: TensorProxy, dim: int) -> TensorProxy:
+    fwd = prims.scatter_add(a, index, value, dim)
+
+    g = get_grad(fwd)
+    # NOTE The value gradient is only valid when src.shape == index.shape.
+    value_grad = prims.gather(g, index, dim)
+    put_grads((a, value), (g, value_grad))
+
+    return fwd
+
+
+register_grad(pids.SCATTER_ADD, _scatter_add_prim_grad)
+
+
+@torchctx
 def _take_along_axis_prim_grad(a: TensorProxy, index: TensorProxy, dim: int) -> TensorProxy:
     fwd = prims.take_along_axis(a, index, dim)
 
--- a/thunder/tests/opinfos.py
+++ b/thunder/tests/opinfos.py
@@ -4496,10 +4496,32 @@ def scatter_add_sample_generator(op, device, dtype, requires_grad, **kwargs):
         yield SampleInput(a, index=b, src=c, dim=dim)
 
 
+def scatter_add_error_generator(op, device, dtype=torch.float32, requires_grad=True, **kwargs):
+    make = partial(make_tensor, device=device, dtype=dtype, requires_grad=requires_grad)
+    # torch.scatter_add expects index to be long but not int
+    # Index is not differentiable! Marking requires_grad as False
+    make_index = partial(make_tensor, device=device, dtype=torch.long, requires_grad=False)
+    make_source = partial(make_tensor, device=device, dtype=dtype, requires_grad=requires_grad)
+
+    # Gradient definition for src tensor is valid only if src.shape == index.shape
+    shape_a = (4, 5, 3)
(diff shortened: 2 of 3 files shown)


thunder/tests/opinfos.py
```diff
@@ -4679,13 +4679,17 @@ def scatter_add_sample_generator(op, device, dtype, requires_grad, **kwargs):
     # torch.scatter_add expects index to be long but not int
     # Index is not differentiable! Marking requires_grad as False
     make_index = partial(make_tensor, device=device, dtype=torch.long, requires_grad=False)
-    # Not sure if we need to consider higher order gradient, marking requires_grad as False
-    make_source = partial(make_tensor, device=device, dtype=dtype, requires_grad=False)
+    make_source = partial(make_tensor, device=device, dtype=dtype, requires_grad=requires_grad)
 
+    # NOTE The value gradient is only valid when src.shape == index.shape.
```
mruberry, line 4684: This is a really nice comment for future readers

thunder/tests/opinfos.py
```diff
@@ -4705,13 +4709,38 @@ def scatter_add_sample_generator(op, device, dtype, requires_grad, **kwargs):
...
+    shape_b = (3, 2, 3)
+    shape_source = (4, 3, 9)
+
+    a = make(shape_a)
+    b = make_index(shape_b, low=0, high=shape_a[dim])
+    c = make_source(shape_source)
+    yield (
+        SampleInput(a, index=b, src=c, dim=dim),
+        RuntimeError,
+        "The gradient for the src Tensor is implemented only when src.shape == index.shape. src shape is (.*?) while index shape is (.*?).",
```
mruberry, line 4735: Is the gradient well-defined when the source and index have different shapes? If so, why do we only support this case?

thunder/core/transforms.py
```diff
@@ -853,6 +853,21 @@ def _gather_prim_grad(a: TensorProxy, index: TensorProxy, dim: int) -> TensorPro
 register_grad(pids.GATHER, _gather_prim_grad)
 
 
+@torchctx
+def _scatter_add_prim_grad(a: TensorProxy, /, index: TensorProxy, value: TensorProxy, dim: int) -> TensorProxy:
+    fwd = prims.scatter_add(a, index, value, dim)
+
+    g = get_grad(fwd)
+    # NOTE The value gradient is only valid when src.shape == index.shape.
```
mruberry, line 861: (which was checked in forward)

mruberry (approved):
Cool!

@IvanYashchuk and @nikitaved , would you like to take a look, too?

thunder/core/transforms.py
nikitaved: No, forward does not have this lim. This is the "hidden" requirement to make sure the grad is correct. It is an issue with a long history :)

thunder/tests/opinfos.py
nikitaved: It is defined, but not with this implementation. See https://github.com/pytorch/pytorch/issues/27614#issuecomment-564648819, for example.

thunder/tests/opinfos.py
nikitaved: super minor nit: I would probably change "valid" to "correct"

thunder/torch/__init__.py
```diff
@@ -2100,6 +2100,11 @@ def gather(a: TensorLike, /, dim: int, index: TensorLike) -> TensorLike:
 # NOTE PyTorch's scatter_add has a parameter named 'src', not 'source'
 @torchsymbol(torch.scatter_add, is_method=True)
 def scatter_add(a: TensorLike, /, dim: int, index: TensorLike, src: TensorLike) -> TensorLike:
+    utils.check(
```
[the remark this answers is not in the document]
nikitaved: It is a limitation of the backward implementation - so I would suggest putting this check there.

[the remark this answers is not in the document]
rdspring1: I moved the check to the grad transform.

mruberry:
@t-vi @Borda Looks like this will require overriding the CI to merge

PR #276, opened 2024-04-25
Add SDPA benchmark for Llama 2 13B, Llama 2 70B, Llama 3 8B, Llama 3 70B

IvanYashchuk:
New parametrized microbenchmark to test the performance of several configurations of the scaled dot product attention call:
```
pytest thunder/benchmarks/targets.py -k "test_litgpt_sdpa_grad" --collect-only
<Module thunder/benchmarks/targets.py>
  <Function test_litgpt_sdpa_grad[Llama-2-7b-hf-torch]>
  <Function test_litgpt_sdpa_grad[Llama-2-7b-hf-torch.compile]>
  <Function test_litgpt_sdpa_grad[Llama-2-7b-hf-thunder]>
  <Function test_litgpt_sdpa_grad[Llama-2-7b-hf-thunder+cudnn]>
  <Function test_litgpt_sdpa_grad[Llama-2-13b-hf-torch]>
  <Function test_litgpt_sdpa_grad[Llama-2-13b-hf-torch.compile]>
  <Function test_litgpt_sdpa_grad[Llama-2-13b-hf-thunder]>
  <Function test_litgpt_sdpa_grad[Llama-2-13b-hf-thunder+cudnn]>
  <Function test_litgpt_sdpa_grad[Llama-2-70b-hf-torch]>
  <Function test_litgpt_sdpa_grad[Llama-2-70b-hf-torch.compile]>
  <Function test_litgpt_sdpa_grad[Llama-2-70b-hf-thunder]>
  <Function test_litgpt_sdpa_grad[Llama-2-70b-hf-thunder+cudnn]>
  <Function test_litgpt_sdpa_grad[Llama-3-8B-torch]>
  <Function test_litgpt_sdpa_grad[Llama-3-8B-torch.compile]>
  <Function test_litgpt_sdpa_grad[Llama-3-8B-thunder]>
  <Function test_litgpt_sdpa_grad[Llama-3-8B-thunder+cudnn]>
  <Function test_litgpt_sdpa_grad[Llama-3-70B-torch]>
  <Function test_litgpt_sdpa_grad[Llama-3-70B-torch.compile]>
  <Function test_litgpt_sdpa_grad[Llama-3-70B-thunder]>
  <Function test_litgpt_sdpa_grad[Llama-3-70B-thunder+cudnn]>
```

H100 results:
```py
TODO
```

I have updated the litgpt pin to include the Llama 3 config and added it to the general jit tests.

### DIFF
--- a/thunder/tests/test_jit_general.py
+++ b/thunder/tests/test_jit_general.py
@@ -621,6 +621,7 @@ def test_nanogpt():
         "falcon-7b-like",
         "falcon-40b-like",
         "codellama2-like",
+        "Llama-3-8B",
         pytest.param(
             "mixtral-like",
             marks=pytest.mark.xfail(raises=(NotImplementedError, TypeError), reason="topk and where", strict=True),
@@ -639,6 +640,7 @@ def test_litgpt_variants(name, device):
 
     x = torch.randint(0, 200, (5, 5), device=device)
     config = litgpt_model.Config.from_name(name)
+    config.n_layer = 1
 
     with device:
         reference = litgpt_model.GPT(config)
--- a/requirements/test.txt
+++ b/requirements/test.txt
@@ -11,7 +11,7 @@ expecttest ==0.2.1  # for test_ddp.py
 hypothesis ==6.100.0  # for test_ddp.py
 numpy  # for test_ops.py
 einops  # for test_einops.py
-litgpt @ git+https://github.com/Lightning-AI/lit-gpt@940ffc96f7214bca24aa77479bc7c33900aaef28
+litgpt @ git+https://github.com/Lightning-AI/lit-gpt@54628ecbdf72b2ceb42bb0ef5f74f1a89ec13d2f
 absl-py # thunder/benchmarks/test_benchmark_litgpt.py
 pandas # thunder/benchmarks/test_benchmark_litgpt.py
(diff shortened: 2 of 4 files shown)


thunder/tests/test_jit_general.py
```diff
@@ -639,6 +640,7 @@ def test_litgpt_variants(name, device):
 
     x = torch.randint(0, 200, (5, 5), device=device)
     config = litgpt_model.Config.from_name(name)
+    config.n_layer = 1
```
nikitaved, line 643: I am just curious. Was this value set to `1` before?

IvanYashchuk: No, I think it's better to set it to 1 to make the test run faster, there's no need to run the full for-loop of the transformer. I don't want to create another "-like" variant as done for other configs https://github.com/Lightning-AI/lightning-thunder/blob/4d9fa6057dbff420491861e82985f0cfc3dcc31e/thunder/tests/litgpt_model.py#L8-L116

t-vi: Given that we might use them in various places, creating a like might be worthwhile.
We're usually using 2 layers because it includes layer->layer data flow.

IvanYashchuk: I'll revert this change but will keep the requirements bump.

carmocca (approved):
LGTM! Thanks

PR #279, opened 2024-04-25
route model access through the thunder module

t-vi:
...then we can override parameters etc. on the thunder module.

### DIFF
--- a/thunder/__init__.py
+++ b/thunder/__init__.py
@@ -203,7 +203,10 @@ class ThunderModule(pytorch.nn.Module):
         super().__init__()
         self._model = model
 
+        # We delete self.training in order for training to be used from
+        # the model itself through `__getattr__`.
         del self.training
+
         self._forward_fn = compiled_model_call
 
     def get_buffer(self, name):
--- a/thunder/__init__.py
+++ b/thunder/__init__.py
@@ -43,7 +43,7 @@ from thunder.common import (
 )
 import thunder.extend as extend
 from thunder.extend import Executor, add_default_executor
-from thunder.core.compile_data import compile_data_and_stats
+from thunder.core.compile_data import compile_data_and_stats, get_compile_data
 from thunder.core.langctxs import LanguageContext
 import thunder.core.langctxs as langctxs
 from thunder.core.baseutils import run_once, check
@@ -205,6 +205,15 @@ class ThunderModule(pytorch.nn.Module):
 
         self._forward_fn = compiled_model_call
 
+    def get_buffer(self, name):
+        return self._model.get_buffer(name)
+
+    def get_parameter(self, name):
+        return self._model.get_parameter(name)
+
+    def get_submodule(self, name):
+        return self._model.get_submodule(name)
+
     def forward(self, *args, **kwargs):
         res = self._forward_fn(*args, **kwargs)
         return res
@@ -284,6 +293,19 @@ def _with_cache_info_ctx(fn):
 
 def _get_cache_info():
     return _cache_info_ctx.get()
+
+
+def _get_thunder_module(model):
+    cd = get_compile_data()
+    tm = cd._thunder_module_map.get(id(model))
+    if tm and tm._model is not model:
+        tm = None
+    if tm is None:
+        # TODO: we would like to raise an error here, but this would require
+        #       us wrapping models that are passed in closures etc.
+        # raise RuntimeError("could not find ThunderModule")
+        return model
+    return tm
 
 
 def add_executor_lists(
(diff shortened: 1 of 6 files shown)


nikitaved:
I was just about to file an issue about that :)

t-vi:
I'll need to fix that some things (such as `__globals__`) still need to go directly onto the original modules.

thunder/__init__.py
```diff
@@ -203,8 +203,18 @@ def __init__(self, model, compiled_model_call):
         super().__init__()
         self._model = model
 
+        del self.training
```
carmocca, line 206: Why is this here? A comment would be helpful

thunder/__init__.py
```diff
@@ -203,8 +203,18 @@ def __init__(self, model, compiled_model_call):
...
         self._forward_fn = compiled_model_call
 
+    def get_buffer(self, name):
+        return self._model.get_buffer(name)
+
+    def get_parameter(self, name):
+        return self._model.get_parameter(name)
+
+    def get_submodule(self, name):
+        return self._model.get_submodule(name)
```
carmocca, line 216: We might want to add logic to `__getattr__` so that if the method comes from `nn.Module` and is not overridden, we automatically route to `self._model`

Avoiding the need for these manual additions

thunder/__init__.py
crcrpar, line 206: at glance this reads a tad aggressive. why and how would this be needed?

thunder/__init__.py
```diff
@@ -448,43 +471,44 @@ def get_computation_and_inputs(*args, **kwargs):
...
+                    cs.last_interpreted_instructions = None
+                    cs.last_interpreter_log = None
+                    cs.last_prologue_traces = pro_traces
+                    cs.last_prologue = pro
+                    cs.last_prologue_transformation_start = 0
+                    cs.last_prologue_transformation_stop = 0
+                    cs.last_computation_transformation_start = 0
+                    cs.last_computation_transformation_stop = 0
+
+                    return cache_entry, inps, pro_to_epi
```
crcrpar, line 511: the diff of these lines is just one indent due to the addition of `compile_data_and_stats` contextmanager?

crcrpar (commented):
not a feedback, just a few questions

thunder/core/prims.py
```diff
@@ -689,6 +691,43 @@ def unpack_function_obj_printer(
...
+    )
+    utils.check(
+        len(kwarg_printables) == 0,
+        lambda: f"Expected no kwargs for unpack_thunder_module but got {kwarg_printables}",
+        exception_type=AssertionError,
+    )
+
+    result_str = "_" if bsym.output is None else f"{codeutils.prettyprint(out_printables, with_type=True)}"
+    arg_str = codeutils.prettyprint(arg_printables[0])
+    s = f"{result_str} = thunder._get_thunder_module({arg_str})"
```
carmocca, line 718: If this is meant to appear in traces, it's not nice that it's a protected function

thunder/__init__.py
```diff
@@ -692,6 +716,7 @@ def fn_(*args, **kwargs) -> Any:
 
     if isinstance(fn, pytorch.nn.Module):
         fn_ = ThunderModule(fn, fn_)
+        cd._thunder_module_map[id(fn)] = fn_
```
carmocca, line 719: What about the case that you jit a function that takes the module? We won't have this reference so easily

thunder/__init__.py
t-vi: I'll add a comment. This is to automatically take the `self._model.training` instead of `self.training`.

thunder/__init__.py
t-vi: Yes.

thunder/__init__.py
t-vi: Indeed!

thunder/__init__.py
t-vi: The idea here is that the next PR will change those to include overrides.

thunder/__init__.py
t-vi: Yeah, but it needs extension to be of full use, but it is not a regression as is. I added this as a map because it enables the case where we do have modules. We still need to add the mechanism to populate this map.

thunder/core/prims.py
t-vi: I'm not keen on having it as a "public api" directly under thunder. Maybe we should just move ThudnerModule to core/module.py and do it there?

t-vi: I moved things to thunder.core.module now.

t-vi:
Thank you @carmocca @crcrpar

PR #282, opened 2024-04-26
Add `_stacklevel` to `thunder.torch.softmax`'s kwarg

### LINKED ISSUE #258: `thunder.jit` fails with `nn.Softmax` raising `got an unexpected keyword argument '_stacklevel'`
## 🐛 Bug

`thunder.jit` fails with `nn.Softmax` raising:
```python
TypeError: softmax() got an unexpected keyword argument '_stacklevel'
```
pointing to: https://github.com/pytorch/pytorch/blob/a21327e0b03cc18850a0608be2d9c5bd38fd4646/torch/nn/functional.py#L1883

### To Reproduce
```python
import torch

examine(torch.nn.Softmax(1).cuda(), torch.randn(1, 10, device="cuda"))
#TypeError: softmax() got an unexpected keyword argument '_stacklevel'
```

### Environment

- PyTorch Version (e.g., 1.0): `2.4.0.dev20240423+cu121`
- `thunder` build from source using at https://github.com/Lightning-AI/lightning-thunder/commit/e0ab64867a5be914d0548c195a3f850a76c8c397

CC @Fuzzkatt as you've encountered this issue first.

crcrpar:
## What does this PR do?

Fixes #258.

`torch.nn.Softmax`'s forward method calls `torch.nn.functional.softmax` with `_stacklevel` kwarg in https://github.com/pytorch/pytorch/blob/b2f521f3769a9545b7c9df57569b1cba6116745b/torch/nn/modules/activation.py#L1545
and it doesn't seem that we have a test case of this way of calling [`torch.nn.functional.softmax`](https://github.com/pytorch/pytorch/blob/b2f521f3769a9545b7c9df57569b1cba6116745b/torch/nn/functional.py#L1857)

cc @ptrblck @Fuzzkatt

### DIFF
--- a/thunder/executors/torchex.py
+++ b/thunder/executors/torchex.py
@@ -1329,7 +1329,6 @@ max_pool3d_with_indices_backward = ex.register_operator(
 nll_loss = _register_torch_operation("nll_loss", module=torch.nn.functional)
 pad = _register_torch_operation("pad", module=torch.nn.functional)
 scaled_dot_product_attention = _register_torch_operation("scaled_dot_product_attention", module=torch.nn.functional)
-_softmax = _register_torch_operation("_softmax")
 softmax = _register_torch_operation("softmax", like=ltorch._softmax)
 
 
--- a/thunder/executors/torchex.py
+++ b/thunder/executors/torchex.py
@@ -1329,7 +1329,8 @@ max_pool3d_with_indices_backward = ex.register_operator(
 nll_loss = _register_torch_operation("nll_loss", module=torch.nn.functional)
 pad = _register_torch_operation("pad", module=torch.nn.functional)
 scaled_dot_product_attention = _register_torch_operation("scaled_dot_product_attention", module=torch.nn.functional)
-softmax = _register_torch_operation("softmax")
+_softmax = _register_torch_operation("_softmax")
+softmax = _register_torch_operation("softmax", like=ltorch._softmax)
 
 
 # NOTE This transform translates number proxies to boolean values
@@ -1632,7 +1633,7 @@ _register_implementation(ltorch.nll_loss_backward, nll_loss_backward, checker=_a
 _register_implementation(ltorch.pad, pad, checker=_always_executable)
 pad_prim_impl = ex.register_operator("torch_pad_prim_impl", meta=prims.pad.meta, fn=_pad_prim_impl)
 _register_implementation(prims.pad, pad_prim_impl, checker=_always_executable)
-_register_implementation(ltorch.softmax, checker=_always_executable, execution_transform=_softmax_transform)
+_register_implementation(ltorch._softmax, checker=_always_executable, execution_transform=_softmax_transform)
 _register_implementation(ltorch.scaled_dot_product_attention, scaled_dot_product_attention, checker=_always_executable)
 
 
--- a/thunder/torch/__init__.py
+++ b/thunder/torch/__init__.py
@@ -4118,6 +4118,9 @@ def _softmax(
     return converted
 
 
+register_method("softmax", _softmax)
+
+
 # A wrapper to support `torch.nn.Softmax` whose `forward` passes the kwarg of `_stacklevel=5` to `torch.nn.functional.softmax`.
 # ref: https://github.com/pytorch/pytorch/blob/8d12ba9acfa20ed7df438a8892c9bf8e6bef5775/torch/nn/modules/activation.py#L1545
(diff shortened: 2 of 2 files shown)


mruberry:
Accepting this argument and ignoring it seems pragmatic. I suppose we could test if it's ever set to something that isn't its default value, but that's very, very niche. The test failures are related. It looks like there are some additional calls that want to pass the stacklevel argument.

I'll make a comment suggesting how to deal with this

thunder/torch/__init__.py
```diff
@@ -4070,7 +4070,14 @@ def sigmoid(a: TensorLike, /) -> TensorLike:
 
 # CompositeImplicitAutograd - don't register decomp
 @torchsymbol(torch.softmax, torch.nn.functional.softmax, is_method=True)
-def softmax(a: TensorLike, /, dim: int, *, dtype: None | dtypeLike = None) -> TensorLike:
+def softmax(
```
mruberry, line 4073: Instead of modifying the symbol, what about creating a non-symbol function that calls the symbol, like this:

```
@torchsymbol(torch.softmax, torch.nn.functional.softmax, is_method=True)
def _softmax(...):
  # The symbol doesn't take _stacklevel, so things don't need to reason about it
  ...

def softmax(..., _stacklevel:int = 3):
  # Doesn't pass _stacklevel, just ignores it
  return _softmax(...)
```

Would something like that work?

crcrpar: [`torch.softmax`](https://pytorch.org/docs/stable/generated/torch.softmax.html#torch-softmax) doesn't have `_stacklevel` as one of its arguments but [`torch.nn.functional.softmax`](https://pytorch.org/docs/stable/generated/torch.nn.functional.softmax.html#torch.nn.functional.softmax) does while the former is said to be an alias for the latter. I'll take your suggestion.

thunder/torch/__init__.py
```diff
@@ -4070,7 +4070,13 @@ def sigmoid(a: TensorLike, /) -> TensorLike:
 
 # CompositeImplicitAutograd - don't register decomp
 @torchsymbol(torch.softmax, torch.nn.functional.softmax, is_method=True)
-def softmax(a: TensorLike, /, dim: int, *, dtype: None | dtypeLike = None) -> TensorLike:
+def _softmax(
```
mruberry, line 4073: With apologies, @crcrpar, I think my suggestion was slightly wrong. I think the `torchsymbol` decorator now has to set the `id` explicitly, otherwise the name `_softmax` is used

crcrpar:
as to https://dev.azure.com/Lightning-AI/lightning/_build/results?buildId=200805&view=logs&j=3f274fac-2e11-54ca-487e-194c91f3ae9f&t=7d03ab59-8b52-555c-a216-77e14d28deec&l=10564, the forward extrace with this PR does not seem to have any bound symbols of softmax unlike one generated with main branch

mruberry:
That is really interesting — I wonder what's happening?

thunder/torch/__init__.py
mruberry: I think the registration is a little more complicated. There are actually places we use `fn.__name__` directly in:

https://github.com/Lightning-AI/lightning-thunder/blob/cfc9ea747aac5e7e64279b7a5d07cf543aa1ab1a/thunder/torch/__init__.py#L73

I think we need to add an override for the `fn.__name__` lookup so we see "softmax" instead of "_softmax"

crcrpar:
> That is really interesting — I wonder what's happening?

the `id` passed to `torchsymbol` has to come with `torch` prefix as torchsymbol does not give any namespace accordingly to the given `module`.
see: https://github.com/Lightning-AI/lightning-thunder/pull/282/commits/5d24a5dbd122403d1994dd94955931dcc7bdf467

thunder/executors/torchex.py
```diff
@@ -1329,7 +1329,8 @@ def max_pool_with_indices_backward_meta(
 nll_loss = _register_torch_operation("nll_loss", module=torch.nn.functional)
 pad = _register_torch_operation("pad", module=torch.nn.functional)
 scaled_dot_product_attention = _register_torch_operation("scaled_dot_product_attention", module=torch.nn.functional)
-softmax = _register_torch_operation("softmax")
+_softmax = _register_torch_operation("_softmax")
```
mruberry, line 1332: What does this line do?

mruberry (approved):
Thanks @crcrpar! @t-vi, this will probably require a manual merge because of flaky ci

Issue #265, opened 2024-04-24
Add sanity check for primitive inplace copy operator


kiya00:
## 🚀 Feature

Add a sanity check to ensure the `copy_to` argument of `prims.copy_` is not used as input for any of its subsequent operators

### Motivation

Nvfuser's `fd.add_output(output, input)` API has the rule "do not use the input after it has been updated". We need the same sanity check for `prims.copy_(copy_from, copy_to)`.

cc: @IvanYashchuk @jjsjann123

jjsjann123:
wondering if this is part of #145 , or just a simple patch to catch the obvious issues?

kiya00:
> wondering if this is part of #145 , or just a simple patch to catch the obvious issues?

I think that's a simple patch and could be removed when the full inplace analysis is ready

PR #285, opened 2024-04-26
Add sanity check for inplace copy

kiya00:
<details>

</details>

## What does this PR do?

Fixes  (#265).
Because it's not safe to use nvfuser's inplace update API. A check is added to ensure that the copy_to argument (and output) of `prims.copy_` is not used as input for any of its subsequent operators in a nvFusion.

cc @apaz-cli


thunder/tests/test_inplace_copy.py
```diff
@@ -121,3 +121,42 @@ def forward(self, x):
...
+        thunder.core.prims.copy_(x, y)
+        return y
+
+    def func4(x, y):
+        z = x * y
+        o = thunder.core.prims.copy_(z, x)
+        thunder.core.prims.copy_(o, y)
+        return y
+
+    import pytest
```
crcrpar, line 151: I guess it's fine to move this to top of the file

thunder/core/transform_common.py
```diff
@@ -33,6 +33,40 @@ def _remove_noop_subsymbols(bsym: BoundSymbol) -> None:
...
+                        f"{bsym} trying to use {input} (the 'copy_to' argument of 'prims.copy_') as input, which is not supported"
+                    )
+            if bsym.sym.id in inplace_copy_symbol_id:
+                copy_to_arg = bsym.flat_proxy_args[1]
+                vcopy_to_arg = variableify(copy_to_arg)
+                out = bsym.flat_proxy_outs
+                if out:
+                    vcopy_to_arg_ = variableify(out[0])
+                    inplace_copy_to_arg.add(vcopy_to_arg_)
+                inplace_copy_to_arg.add(vcopy_to_arg)
```
crcrpar, line 64: [question] as an alternative, could it be implemented by checking `consumers[bsym_of_copy_to.flat_proxy_outs[0]]` where consumers is [`thunder.core.utils.consumers(extrace)`](https://github.com/Lightning-AI/lightning-thunder/blob/3a2b27e9b38e639133913c42c03acdf0e136ec4f/thunder/core/utils.py#L982)?

kiya00: yes, it works for the output. For the `copy_to` input arg, it needs the one-by-one check to distinguish the consumer before and after the inplace copy operator

IvanYashchuk: We have a linear order of operations in symbols if we call first `consumers(trace)`, assign indices to each symbol, and filter out the ones appearing before the inplace copy operator, would it work? Would the implementation be simpler?

thunder/core/transform_common.py
```diff
@@ -33,6 +33,40 @@ def _remove_noop_subsymbols(bsym: BoundSymbol) -> None:
     bsym.subsymbols = nsbsyms
 
 
+def _inplace_copy_sanity_check(extrace: Trace):
+    """Make sure that the copy_to argument of prims.copy_ is not used as input for any of its subsequent operators, except for the Return and Del operators."""
```
IvanYashchuk, line 37: Please expand this docstring to include more details about the necessity of this check, why it's important to have this check, the types of neural network patterns that are permissible, the patterns this check aims to block, and the issues associated with those patterns.

thunder/tests/test_inplace_copy.py
```diff
@@ -121,3 +122,40 @@ def forward(self, x):
     assert_close(net.state_dict()["dense1_bn.running_mean"], torch_net.state_dict()["dense1_bn.running_mean"])
     assert_close(net.state_dict()["dense1_bn.running_var"], torch_net.state_dict()["dense1_bn.running_var"])
     assert_close(x.grad, x1.grad)
+
+
+@instantiate(dtypes=(thunder.float32,))
+def test_inplace_copy_sanity_check(executor, device, dtype):
```
IvanYashchuk, line 128: @jjsjann123, could you please take a look at these test cases and tell us if these are sufficient restrictions to generate correctly functioning nvFuser code?

thunder/core/transform_common.py
```diff
@@ -33,6 +33,58 @@ def _remove_noop_subsymbols(bsym: BoundSymbol) -> None:
...
+    from thunder.core.trace import VariableInterface
+
+    inplace_copy_symbol_id = ("copy_", prims.PrimIDs.COPY_)
+    symbol_id_skip_list = (prims.PrimIDs.RETURN, prims.PrimIDs.DEL)
+    inplace_copy_to_arg: set[VariableInterface] = set()
+
+    def check_symbol(bsym):
+        if bsym.sym.id in symbol_id_skip_list:
+            return
+        elif bsym.sym.is_fusion:
```
jjsjann123, line 65: torch.compile also has `is_fusion == True`. do they need this check as well?

thunder/core/transform_common.py
```diff
@@ -33,6 +33,58 @@ def _remove_noop_subsymbols(bsym: BoundSymbol) -> None:
...
+
+        c = prims.copy_(a, b)
+        d = torch.add(a, a)
+        return c
+    """
+    from thunder.core.trace import VariableInterface
+
+    inplace_copy_symbol_id = ("copy_", prims.PrimIDs.COPY_)
+    symbol_id_skip_list = (prims.PrimIDs.RETURN, prims.PrimIDs.DEL)
+    inplace_copy_to_arg: set[VariableInterface] = set()
```
jjsjann123, line 60: Do we want to keep a global `inplace_copy_to_arg` set?

I understand as the race only happens within a kernel, so the set should be reset when we check for a fusion symbol and clears afterwards?

thunder/tests/test_inplace_copy.py
jjsjann123: tests looks fine~ thanks for the ping.

jjsjann123 (commented):
Thanks for putting the effort into writing the check, it's great to have those for a codegen block!

I'm a bit unclear about our expectation from this check function though.
Feels almost like we are hoping that running this check globally would *alway* ensure that we have a correct function. But I don't think it's sufficient.

A user script could be using an argument after it has been inplace updated (see example below). That's a legit program, our check will reject it. A false negative check is not ideal but I think it's acceptable.

```
def fn(a, b):
  c = a + 1
  d = copy_(c, a)
  e = b + a
  return e
```

The worse scenario is that, our optimization/transformation could re-order things. For the program above, `e = b+a` doesn't depend on anything and we could move it before the inplace copy. Which alters the semantics of the function, but it would appear as legit program in our check.
False positive check and silent wrong result is bad.

thunder/__init__.py
```diff
@@ -554,6 +554,7 @@ def get_computation_and_inputs(*args, **kwargs):
                 computation_trc = extraces[-1]
                 cs.last_computation_transformation_stop = time.time_ns()
 
+            thunder.core.transform_common._inplace_copy_sanity_check(computation_trc)
```
jjsjann123, line 557: should we add a compile option to switch off the check here?

thunder/core/transform_common.py
jjsjann123: since we do not have nested fusion, I'm thinking that we can clear the set of inplace_copy_to_arg here?

thunder/core/transform_common.py
```diff
@@ -33,6 +33,58 @@ def _remove_noop_subsymbols(bsym: BoundSymbol) -> None:
...
+            return
+        elif bsym.sym.is_fusion:
+            for subbsym in bsym.subsymbols:
+                check_symbol(subbsym)
+        else:
+            for input in bsym.flat_proxy_args:
+                vinput = variableify(input)
+                if vinput in inplace_copy_to_arg:
+                    raise NotImplementedError(
+                        f"{bsym} trying to use {input} (the 'copy_to' argument of 'prims.copy_') as input, which is not supported"
```
jjsjann123, line 73: we can have the error message here mentioning about the compile option and some note about the risks of disabling the check.

jjsjann123 (commented):
BTW, I'm not blocking this PR. It's an improvement to what we have.

thunder/core/transform_common.py
kiya00: oh yes, thanks for pointing out this mistake. I misunderstood. The limitation of using `copy_to` is only inside one nvFusion. I'll fix it.

kiya00:
> False positive check and silent wrong result is bad.

yes, I met this problem too
```
def func1(x, y):
        z = x * y
        o = thunder.core.prims.copy_(z, x)
        return x + y
```
generate the trace:
```
@torch.no_grad()
@no_autocast
def computation(x, y):
  # x: "cuda:0 f32[4, 4]"
  # y: "cuda:0 f32[4, 4]"
  [t2] = nvFusion0(x, y)
    # t2 = prims.add(x, y)  # t2: "cuda:0 f32[4, 4]"
    # result = prims.mul(x, y)  # result: "cuda:0 f32[4, 4]"
    # prims.copy_(result, x)
  del x, y
  return t2
```
which is wrong result.
I think this sanity check is for checking the final execution trace, not the original function. The reorder problem should be solved by the general inplace and aliasing plan(maybe?) cc: @IvanYashchuk

kiya00:
> BTW, I'm not blocking this PR. It's an improvement to what we have.

Not at all, some of the logic is not right, thanks for pointing out @jjsjann123 , I convert it to draft and I'll fix it on Monday :)

jjsjann123:
> > False positive check and silent wrong result is bad.
> 
> yes, I met this problem too
> 
> ```
> def func1(x, y):
>         z = x * y
>         o = thunder.core.prims.copy_(z, x)
>         return x + y
> ```
> 
> generate the trace:
> 
> ```
> @torch.no_grad()
> @no_autocast
> def computation(x, y):
>   # x: "cuda:0 f32[4, 4]"
>   # y: "cuda:0 f32[4, 4]"
>   [t2] = nvFusion0(x, y)
>     # t2 = prims.add(x, y)  # t2: "cuda:0 f32[4, 4]"
>     # result = prims.mul(x, y)  # result: "cuda:0 f32[4, 4]"
>     # prims.copy_(result, x)
>   del x, y
>   return t2
> ```
> 
> which is wrong result. I think this sanity check is for checking the final execution trace, not the original function. The reorder problem should be solved by the general inplace and aliasing plan(maybe?) cc: @IvanYashchuk

Got'ya. Sounds like our understanding is aligned and that's all that I wanted to confirm. Feel free to ping me when it's ready for another round of review.

kiya00:
followed the comments, the modifications are:

- check the sharp edge for each `nvFusion`

- use the `utils.consumers` and check the index for the relative position

- added a `disable_inplace_copy_check` in the `thunder.jit`

@jjsjann123 @IvanYashchuk , could you take a look again

thunder/core/transform_common.py
```diff
@@ -33,6 +33,52 @@ def _remove_noop_subsymbols(bsym: BoundSymbol) -> None:
...
+    for bsym in nvfuser_symbols:
+        consumer_dict = consumers(list(bsym.subsymbols), _map_to_numbers=True)
+        inplace_copy_idx = [(idx, sym) for idx, sym in enumerate(bsym.subsymbols) if sym.sym.id == prims.PrimIDs.COPY_]
+        for idx, subbsym in inplace_copy_idx:
+            copy_to_arg = subbsym.flat_args[1]
+            copy_to_out = subbsym.output
+
+            def check(inp, log_str):
+                if inp is not None and inp in consumer_dict:
+                    last_used_idx = consumer_dict[inp][-1]
```
jjsjann123, line 71: nitpick: this assumes that consumers is in ascending order, which is true for `thunder.core.utils.consumers`. But that is implementation dependent. Can we simply do a `max` since that's easy.

jjsjann123 (approved):
LGTM. Thanks for updating this!

thunder/__init__.py
```diff
@@ -263,6 +263,7 @@ def jit(
     early_transforms: list | None = None,
     additional_transforms: list | None = None,
     record_history: bool = False,
+    disable_inplace_copy_check: bool = False,
```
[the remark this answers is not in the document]
kiya00: added it to implicit `compile_options`

kiya00:
Hi @t-vi @mruberry , this PR is a simple sanity check on the final trace for nvfuser `copy_`, could you take a look?

t-vi (approved):
Thank you @kiya00 @IvanYashchuk @jjsjann123

PR #250, opened 2024-04-22
Number proxies

jjsjann123:
<details>
  <summary><b>Before submitting</b></summary>

- [ ] Was this discussed/approved via a Github issue? (no need for typos and docs improvements)
- [ ] Did you read the [contributor guideline](https://github.com/Lightning-AI/pytorch-lightning/blob/main/.github/CONTRIBUTING.md), Pull Request section?
- [ ] Did you make sure to update the docs?
- [ ] Did you write any new necessary tests?

</details>

## What does this PR do?

Fixes # (issue).

## PR review

Anyone in the community is free to review the PR once the tests have passed.
If we didn't discuss your PR in Github issues there's a high chance it will not be merged.

## Did you have fun?

Make sure you had fun coding 🙃

PR #286, opened 2024-04-26
Number proxies is not a number

### LINKED ISSUE #272: NumberProxy is no Number
## 🚀 Feature

Derivatives of `NumberProxy` should not inherit from `Number` directly.

### Motivation

It's error prone to have derivatives of `NumberProxy` identified as `Number` at the same time. We would want to identify `NumberProxy` as dynamic value, so interpreter would know not to bake in numbers or accidentally constant fold things away.

### Pitch

We should remove the inheritance of `float` from `FloatProxy` here: https://github.com/Lightning-AI/lightning-thunder/blob/34d21e1455bf13e0c397207b018f2835acc3870a/thunder/core/proxies.py#L973

Same thing would apply to `IntegerProxy` / `ComplexProxy`. After this change, we will be able to distinguish `NumberProxy` from constant number in the program.

### Alternatives

Alternative: we'll have to update existing checks where `isinstance(x, Number)` to `not isinstance(x, NumberProxy) and isinstance(x, Number)` to properly identify runtime values. The worse part is, if we miss anything in the code base, it'll ended up as an unintended behavior (like bake in static number) and possibly a silent wrong result.

comparing to the proposed approach here, if we change the inheritance hierarchy, existing legacy check `isinstance(x, Number)` would return false and we would NOT treating a `NumberProxy` as a regular compile time constant. This likely would lead us to some different code path where we'll try to use it as a `TensorProxy` instead, that'll give us an exception, which is much easier to patch.

### Additional context

This is an incremental work supporting #262

jjsjann123:
</details>

## What does this PR do?

Fixes #272.

Changes in this PR:
1. Removing the inheritance of Number kinds in NumberProxy derivatives. The motivation is to avoid mistakenly identify a NumberProxy as Numbers. Specifically, we are removing `complex` / `int` / `float` from the inheritance of `ComplexProxy` / `IntegerProxy` / `FloatProxy`.
2. changing existing checks from `isinstance(x, Number)` to `isinstance(x, (Number, NumberProxy))`.
3. Handling proper type promotion in NumberProxy's binary/unary operations.


jjsjann123:
1. Indexing is fun. but we can probably use `__index__` defined in IntegerProxy to work around that.
2. We are going to have trace that tracks NumberProxy with binary operations across them and we need to have proper type promotion, which we don't have it handled properly yet.

jjsjann123:
Looks like grad is wrong... `getitem`. let me try to repro that.
But other than that, I think the draft is serving a reasonable evaluation on the workload of making the switch. `+243 - 133` feels like not that big of an endeavor. I'll summarize it in the issue for our discussion.

jjsjann123:
grad failure is only on windows nightly. Can't seem to repro it locally, that's a bit strange.

12.1 failures is coming from OOM. <- wondering if it's just flaky or there's a real impact. I'm a bit nervous since I have to update the memory examine, but I wasn't expecting any significant impact, since the added trace are only for numbers...

jjsjann123:
Windows failure can be seen here: https://github.com/Lightning-AI/lightning-thunder/pull/291

Maybe I missed something obvious?

jjsjann123:
> Windows failure can be seen here: #291
> 
> Maybe I missed something obvious?

failing tests is reported on #296 and is not related to code change here.

jjsjann123:
Want to put a note here that the current snapshot [06cbd60](https://github.com/Lightning-AI/lightning-thunder/pull/286/commits/06cbd604864f2d832f9e135bfe6ca41097b1c410) seems to be passing existing CI tests.

thunder/tests/test_examine_memory.py
```diff
@@ -150,7 +150,8 @@ def test_nanogpt_block(executor, device, dtype):
         expected_return_calculated_mem = get_return_memory(fw_extrace.bound_symbols[-1]) - 4 * 2 * 1024 * 768
         assert expected_return_calculated_mem == sum(fw_alloc_mem[1].values())
 
-        assert bw_alloc_mem[0] == 361783296
+        # NOTE: Questionable memory bump and haven't investigated yet
+        assert bw_alloc_mem[0] == 361881600
```
jjsjann123, line 154: This change seems to be strange. Note that I need to further investigate this.

thunder/clang/__init__.py
```diff
@@ -189,11 +189,13 @@ def arange(*, start: Number, step: Number, stop: Number, device: DeviceLike, dty
...
-    utils.check(
-        (start == stop) or (step < 0 and stop < start) or (step > 0 and stop > start),
-        lambda: f"step={step} must make progress from start={start} to stop={stop}",
-    )
+    # TODO support NumberProxy check
+    if all(isinstance(arg, Number) for arg in [start, stop, step]):
+        # Checks that step makes progress
+        utils.check(
+            (start == stop) or (step < 0 and stop < start) or (step > 0 and stop > start),
+            lambda: f"step={step} must make progress from start={start} to stop={stop}",
```
jjsjann123: I think I just missed a small spot in __eq__ & __ne__ method where it's doing a `isinstance(other, Number)` check.

jjsjann123: Looks like that's indeed the problem. :crossed_fingers:

thunder/core/transforms.py
```diff
@@ -1042,9 +1042,24 @@ def _div_prim_grad(a: Number | TensorProxy, b: Number | TensorProxy, /) -> Numbe
 
 register_grad(pids.DIV, _div_prim_grad)
 
+
+# NOTE not differentiable, but it would trigger a flatten failure without a grad function
+# NOTE that's probably a bad error message that we should fix.
+@torchctx
+def _py_floordiv_prim_grad(a: Number | TensorProxy, b: Number | TensorProxy, /) -> Number | TensorProxy:
```
jjsjann123, line 1049: want to point out a somewhat orthogonal changes here.

Is it expected that changing things to NumberProxy would push more requirement on having a grad rule?

thunder/tests/test_examine_memory.py
jjsjann123: I noticed that in the bwd trace, we are seeing number proxies showing up in the trace. I'm imagining that's the reason of the bump in analysis.

My naive understanding is that, currently since we are not registering numbers as proxies. number proxies should not show up in trace through `thunder.jit` and operations should still back in. I'm not totally sure why backward is seeing these.

jjsjann123: But the problem with this bump is actually in an change in `nvfuserex_impl.py`. where we are rejecting `torch.sum` with a reduction axes as NumberProxies.

Somehow after grad transform, the trace is taking a ProxyNumber as the reduction axis and the added logic is rejecting that. So the change in behavior is a right thing to do. But do we understand why grad transform is doing that?

I get this bwd trace in this test from running our main branch
```
  t411 = torch.sum(t409, i51, True, dtype=None)  # t411: "cuda:0 f32[2, 12, 1024, 1]"
    # t411 = ltorch.sum(t409, i51, True, dtype=None)  # t411: "cuda:0 f32[2, 12, 1024, 1]"
      # t552 = prims.sum(t409, (3,))  # t552: "cuda:0 f32[2, 12, 1024]"
      # t411 = prims.broadcast_in_dim(t552, [2, 12, 1024, 1], [0, 1, 2])  # t411: "cuda:0 f32[2, 12, 1024, 1]"
  del t409, i51
```

```
commit d6237c88d26aa4a5541f3ad2e955140264afb068 (HEAD -> main, origin/main, origin/HEAD)
Author: Ivan Yashchuk <IvanYashchuk@users.noreply.github.com>
Date:   Fri May 10 18:26:46 2024 +0300

    Fix flaky test_prim_inplace_copy_fwd (#399)
```

jjsjann123: I'll open an issue to track this.

jjsjann123: Following up with #403

jjsjann123:
I think this PR is ready for review cc'ing @t-vi

thunder/core/proxies.py
```diff
@@ -626,7 +626,22 @@ def _elementwise_unary_helper(a, name, fn):
         except Exception as e:
             return fn(vala)
 
-        return method(a)
+        if method is None:
```
mruberry, line 629: We should delete these implicit NumberProxy to number translations in the future, right?

thunder/core/proxies.py
```diff
@@ -684,7 +699,28 @@ def _elementwise_binary_helper(a, b, name, fn):
...
+        try:
+            method = resolve_method(name, a, b)
+        except Exception as e:
+            return fn(vala, valb)
+
+        if method is None:
+            return fn(vala, valb)
+
+        # circular import
+        from thunder.core.utils import elementwise_type_promotion, dtype_to_numbertype, are_same_dtypes
```
mruberry, line 711: Do we expect the type promotion logic here to be deleted in the future? What will that require?

thunder/core/transforms.py
mruberry: I think @IvanYashchuk has had some thoughts about producing gradients for numbers, and not just tensors.

fyi @IvanYashchuk to review this particular change

@jjsjann123, is the correct interpretation of this grad rule that it just stops grad from going from the output of the floor division to its inputs?

Are we calling floor division on tensors now, btw? Is the comment here:

https://github.com/Lightning-AI/lightning-thunder/blob/eb6565b15ed8f199c4cc8e1b2fa91d256aa8d975/thunder/core/prims.py#L2311

out of date?

mruberry:
Looks pretty great, @jjsjann123! And test failures are unrelated. I asked a couple questions.

I'm curious what the next step here is. Do we need to be more careful about producing NumberProxy objects when operations are performed on them?

thunder/core/proxies.py
jjsjann123: We do need the type promotion logic. Maybe not explicitly in the trace if that's what you meant?

My naive understanding is that since we are using `prims` for NumberProxy, it's natural to follow the same protocol that type promotion is handling out side of `prim`. Otherwise, we'll need some special code path inside `prim`'s meta function for NumberProxy. That felts a bit messy to me.

jjsjann123:
> I'm curious what the next step here is. Do we need to be more careful about producing NumberProxy objects when operations are performed on them?

A couple items for next step:

1. Yes. I'm surprised to see NumberProxy popping up all the time in CI. I'll go figure out what's going on with that as a side thing.

2.1 Meanwhile, I tried to merge this branch with the other draft where I'm enabling dynamic numbers as a real thing and I'm surprised to see that PR (#250) passing all CI :laughing: . I'll start cleaning up that PR as well.
Initially we can faithfully treat all symbolic value as-is and reject it in nvfuser when we cannot support those.

2.2 Next step is to have a smarter caching logic even for symbolic numbers. i.e. we can probably bake in some NumberProxy and properly insert check in prologue_trace to allow more aggressive fusion.

3. dynamic shape should be something I need to start thinking about how they would show up in our trace as well as how it's going to be handled. I'll try to get some concrete examples to demonstrate the ideas. This needs to be cleared up before we start working on caching for symbolic numbers in step 2.2. I think there'll be quite some overlapped topics here.

thunder/core/proxies.py
mruberry: Maybe we should be calling `clang` operators that do the type promotion?

thunder/core/transforms.py
jjsjann123: > @jjsjann123, is the correct interpretation of this grad rule that it just stops grad from going from the output of the floor division to its inputs?

yes. Since we are not supposed to propagate it back and are not expecting tensor inputs as well.

> Are we calling floor division on tensors now, btw? 

No we are *NOT* calling floordiv with tensors. I can update the annotation if you think we should restrict it. And I should probably also put an assert there that a and b are not TensorProxy.

The reason we needed this is for
`test_phantom_grad_vs_torch_consistency_reshape_`

sequences are not baked in, so our reshape handling of `-1` shows up in trace.

e.g. something like this vvv.
```
@torch.no_grad()
@no_autocast
def reshape(a, shape):
  # a: "cpu f16[4, 2]"
  # shape: "Collection"
  i0, _, i2, = shape
  i5 = prims.mul(1, i0)  # i5: "int 2"
  i9 = prims.mul(i5, i2)  # i9: "int 4"
  i12 = prims.py_floordiv(8, i9)  # i12: "int 2"
  t20 = prims.reshape(a, (i0, i12, i2))  # t20: "cpu f16[2, 2, 2]"
  return t20
```

jjsjann123: looks like the tests are still using `thunder.compile`, which handles sequence differently than `thunder.jit` that bakes in scalar values in a sequence. I think this explains why I have to navigate through the CI for that.

thunder/core/proxies.py
jjsjann123: I think you are saying that `resolve_method(...)` should return `clang.xxx` instead of `prim.xxx` so we don't have to duplicate the type promotion logic here?

Let me dig a bit to see how easy that is :bow:

jjsjann123:
err. strangely I'm not able to repro the failure on my local machine.
```
FAILED thunder/tests/test_nvfuser_remat.py::test_find_cut_dropout_nvfuser_cuda_None - AssertionError: assert ('t0', 't11', 't7') == ('t0', 't5', 't9')
```

thunder/core/proxies.py
mruberry: I think so, right? If we want to make these operations on numbers more consistent with operations on tensors.

One thing I'm curious about is what happens in the prim meta operations. Don't those operations have a check that we don't call any symbols from them, but we do a lot of manipulations on numbers? Do we need to disable recording symbols in those operations maybe, and/or check that we don't perform tensor operations?

mruberry: Looks like the check is that the operation is just not on a tensor proxy, so I think we'll be OK with NumberProxies:

https://github.com/Lightning-AI/lightning-thunder/blob/73ca5dc79179773c542d71b3f81d02ae0a220329/thunder/core/prims.py#L41

mruberry: But follow-up question: do operations on NumberProxies called from primitive meta functions get recorded somewhere?

mruberry: Like in the meta operation for reshape:

https://github.com/Lightning-AI/lightning-thunder/blob/73ca5dc79179773c542d71b3f81d02ae0a220329/thunder/core/prims.py#L2976

if the elements of `shape` are NumberProxies, will the multiplications used to compute `numel` be recorded? Is that good or bad?

mruberry: Currently all operations that occur in primitive meta functions are ignored:

https://github.com/Lightning-AI/lightning-thunder/blob/73ca5dc79179773c542d71b3f81d02ae0a220329/thunder/core/symbol.py#L252

jjsjann123: > if the elements of shape are NumberProxies, will the multiplications used to compute numel be recorded? Is that good or bad?
> Currently all operations that occur in primitive meta functions are ignored:

That does sound like the right behavior in my naive mind. `prims` are the primitive in our trace and we shouldn't further trace anything inside.
In your example, `prims.reshape` does shape manipulation just as a sanity check and that shouldn't be exposed. if we do want things to show up in trace, that should be put inside `clang/__init__.py` I think?
But `prim`'s behavior on NumberProxy needs to be updated... i.e. NumberProxies are going to be dynamic and checks on Meta functions are no longer meaningful. I think it just means certain prim that does a hard check on NumberProxies shouldn't be accepting number proxies any more and we should move those checks to `clang/__init__.py` instead?


Interestingly `clang.reshape` inserts semantic check for cases with `-1` in new shape but not with concrete shapes. I guess it's just assuming `prim` is handling the shape check.
We need to start making conscious decision on what "check" to be baked into the trace.

jjsjann123:
test failures are false positive.

Here's the trace from CI:
```
@torch.no_grad()
@no_autocast
def _value_and_grad(*args):
  # args: "Collection"
  [t1] = nvFusion0()
    # t1 = prims.full((2, 2), 1, device=devices.Device("cuda:0"), dtype=dtypes.float32)  # t1: "cuda:0 f32[2, 2]"
  t0, = args
  del args
  [t11, t7] = nvFusion1(t0)
    # t2 = prims.exp(t0)  # t2: "cuda:0 f32[2, 2]"
    # t3 = prims.exp(t2)  # t3: "cuda:0 f32[2, 2]"
    # t6 = prims.uniform((2, 2), 0.0, 1.0, device=devices.Device("cuda:0"), dtype=dtypes.float32)  # t6: "cuda:0 f32[2, 2]"
    # t7 = prims.lt(t6, 0.9)  # t7: "cuda:0 b8[2, 2]"
    # t8 = prims.convert_element_type(t7, dtypes.float32)  # t8: "cuda:0 f32[2, 2]"
    # t9 = prims.mul(t3, t8)  # t9: "cuda:0 f32[2, 2]"
    # t10 = prims.mul(t9, 1.1111111111111112)  # t10: "cuda:0 f32[2, 2]"
    # t11 = prims.exp(t10)  # t11: "cuda:0 f32[2, 2]"
  t13 = torch.permute(t11, (1, 0))  # t13: "cuda:0 f32[2, 2]"
    # t13 = ltorch.permute(t11, (1, 0))  # t13: "cuda:0 f32[2, 2]"
      # t13 = prims.transpose(t11, (1, 0))  # t13: "cuda:0 f32[2, 2]"
  t12 = torch.matmul(t11, t11)  # t12: "cuda:0 f32[2, 2]"
    # t12 = ltorch.matmul(t11, t11)  # t12: "cuda:0 f32[2, 2]"
      # t12 = prims.matmul(t11, t11)  # t12: "cuda:0 f32[2, 2]"
  t14 = torch.matmul(t1, t13)  # t14: "cuda:0 f32[2, 2]"
    # t14 = ltorch.matmul(t1, t13)  # t14: "cuda:0 f32[2, 2]"
      # t14 = prims.matmul(t1, t13)  # t14: "cuda:0 f32[2, 2]"
  t16 = torch.matmul(t13, t1)  # t16: "cuda:0 f32[2, 2]"
    # t16 = ltorch.matmul(t15, t1)  # t16: "cuda:0 f32[2, 2]"
      # t16 = prims.matmul(t15, t1)  # t16: "cuda:0 f32[2, 2]"
  del t13, t1
  [t25] = nvFusion2(t0, t11, t14, t16, t7)
    # t2 = prims.exp(t0)  # t2: "cuda:0 f32[2, 2]"
    # t3 = prims.exp(t2)  # t3: "cuda:0 f32[2, 2]"
    # t8 = prims.convert_element_type(t7, dtypes.float32)  # t8: "cuda:0 f32[2, 2]"
    # t17 = prims.add(t14, t16)  # t17: "cuda:0 f32[2, 2]"
    # t18 = prims.mul(t17, t11)  # t18: "cuda:0 f32[2, 2]"
    # t20 = prims.mul(1.1111111111111112, t18)  # t20: "cuda:0 f32[2, 2]"
    # t21 = prims.mul(t8, t20)  # t21: "cuda:0 f32[2, 2]"
    # t24 = prims.mul(t21, t3)  # t24: "cuda:0 f32[2, 2]"
    # t25 = prims.mul(t24, t2)  # t25: "cuda:0 f32[2, 2]"
  del t0, t11, t14, t16, t7
  return (t12, (t25,))
```

Here's my local test
```
@torch.no_grad()
@no_autocast
def _value_and_grad(*args):
  # args: "Collection"
  [t1] = nvFusion0()
    # t1 = prims.full((2, 2), 1, device=devices.Device("cuda:0"), dtype=dtypes.float32)  # t1: "cuda:0 f32[2, 2]"
  t0, = args
  del args
  [t5, t9] = nvFusion1(t0)
    # t2 = prims.exp(t0)  # t2: "cuda:0 f32[2, 2]"
    # t3 = prims.exp(t2)  # t3: "cuda:0 f32[2, 2]"
    # t4 = prims.uniform((2, 2), 0.0, 1.0, device=devices.Device("cuda:0"), dtype=dtypes.float32)  # t4: "cuda:0 f32[2, 2]"
    # t5 = prims.lt(t4, 0.9)  # t5: "cuda:0 b8[2, 2]"
    # t6 = prims.convert_element_type(t5, dtypes.float32)  # t6: "cuda:0 f32[2, 2]"
    # t7 = prims.mul(t3, t6)  # t7: "cuda:0 f32[2, 2]"
    # t8 = prims.mul(t7, 1.1111111111111112)  # t8: "cuda:0 f32[2, 2]"
    # t9 = prims.exp(t8)  # t9: "cuda:0 f32[2, 2]"
  t11 = torch.permute(t9, (1, 0))  # t11: "cuda:0 f32[2, 2]"
    # t11 = ltorch.permute(t9, (1, 0))  # t11: "cuda:0 f32[2, 2]"
      # t11 = prims.transpose(t9, (1, 0))  # t11: "cuda:0 f32[2, 2]"
  t10 = torch.matmul(t9, t9)  # t10: "cuda:0 f32[2, 2]"
    # t10 = ltorch.matmul(t9, t9)  # t10: "cuda:0 f32[2, 2]"
      # t10 = prims.matmul(t9, t9)  # t10: "cuda:0 f32[2, 2]"
  t12 = torch.matmul(t1, t11)  # t12: "cuda:0 f32[2, 2]"
    # t12 = ltorch.matmul(t1, t11)  # t12: "cuda:0 f32[2, 2]"
      # t12 = prims.matmul(t1, t11)  # t12: "cuda:0 f32[2, 2]"
  t14 = torch.matmul(t11, t1)  # t14: "cuda:0 f32[2, 2]"
    # t14 = ltorch.matmul(t13, t1)  # t14: "cuda:0 f32[2, 2]"
      # t14 = prims.matmul(t13, t1)  # t14: "cuda:0 f32[2, 2]"
  del t11, t1
  [t23] = nvFusion2(t0, t12, t14, t5, t9)
    # t2 = prims.exp(t0)  # t2: "cuda:0 f32[2, 2]"
    # t3 = prims.exp(t2)  # t3: "cuda:0 f32[2, 2]"
    # t6 = prims.convert_element_type(t5, dtypes.float32)  # t6: "cuda:0 f32[2, 2]"
    # t15 = prims.add(t12, t14)  # t15: "cuda:0 f32[2, 2]"
    # t16 = prims.mul(t15, t9)  # t16: "cuda:0 f32[2, 2]"
    # t18 = prims.mul(1.1111111111111112, t16)  # t18: "cuda:0 f32[2, 2]"
    # t19 = prims.mul(t6, t18)  # t19: "cuda:0 f32[2, 2]"
    # t22 = prims.mul(t19, t3)  # t22: "cuda:0 f32[2, 2]"
    # t23 = prims.mul(t22, t2)  # t23: "cuda:0 f32[2, 2]"
  del t0, t12, t14, t5, t9
  return (t10, (t23,))
```

For some reason the tensor names are changed between runs, which is strange. Is that expected behavior? I'll open an issue for that.

Meanwhile, the test asserting on variable name is bad practice as well. I'll patch that.

thunder/clang/langctx.py
```diff
@@ -30,19 +30,19 @@ def get_method(self, id: str, *args, **kwargs) -> Callable:
...
+        has_proxy_input: bool = False
         for x in inps:
-            if isinstance(x, TensorProxy):
-                has_tensor_input = True
+            if isinstance(x, TensorProxy) or isinstance(x, NumberProxy):
+                has_proxy_input = True
                 break
 
-        if has_tensor_input:
+        if has_proxy_input:
```
jjsjann123, line 39: This is where I'm using `clang` to explicitly handle dtype promotion for number proxies.

thunder/tests/test_nvfuser_remat.py
```diff
@@ -309,10 +309,12 @@ def test_find_cut_dropout(executor, device, _):
...
-    # Note t5 is the boolean mask for dropout. It should be chosen over the t6
-    # that is the float32 mask. See this issue: "The Recomputation Algorithm on
-    # Dropout choses a float32 mask to save"
-    assert cut == ("t0", "t5", "t9")
+    assert cut[0] == producer.args[0].name
+    # Note cut[1]/producer.output[0] is the boolean mask for dropout. It should
+    # be chosen over the float32 mask. See this issue: "The Recomputation
+    # Algorithm on Dropout choses a float32 mask to save"
+    assert cut[1] == producer.output[0].name
+    assert cut[2] == producer.output[1].name
```
jjsjann123, line 317: I don't know why CI's trace context is changing the name of the tensors.... I cannot repro this on my local machine, which is very strange.

thunder/torch/langctx.py
```diff
@@ -46,9 +46,9 @@ def get_method(self, id: str, *args, **kwargs) -> Callable:
                 return prop(inps[0])
 
         # has_tensor_input is False
-        # Defers to the primitive language context when there are no tensor inputs=
-        #   (the primitive language context handles operations on numbers)
-        primsctx: LanguageContext = resolve_language(Languages.PRIMS)
+        # Defers to the CLANG language context when there are no tensor inputs=
+        #   (the clang language context handles operations on NumberProxies and Numbers)
+        primsctx: LanguageContext = resolve_language(Languages.CLANG)
```
jjsjann123, line 51: similarly, instead of PRIMs. dispatch through CLANG so we have a chance to handle type promotion for number proxies.

thunder/clang/__init__.py
```diff
@@ -19,7 +19,7 @@
 import thunder.core.dtypes as dtypes
 from thunder.core import utils
 import thunder.core.prims as prims
-from thunder.core.proxies import TensorProxy, pyval, pytype, proxy, AnyProxy, Proxy
+from thunder.core.proxies import IntegerProxy, NumberProxy, TensorProxy, pyval, pytype, proxy, AnyProxy, Proxy
```
jjsjann123, line 22: Note to myself. NumberProxy needs to be added on signature here. I'll do a clean up.

jjsjann123:
@mruberry This one is good for another round of review~

I removed the typo promotion logic in proxies as suggested and now it's relying on clang/__init__.py 's logic. Made a few comment above hopefully that'll be helpful with the review. :bow:

thunder/core/transforms.py
IvanYashchuk: A year ago Thunder supported differentiating function with operations on numbers and then it got broken. But it's only for floats. Non-differentiable inputs should be taken care as constants, so it's a quick of current transform code.
Could you please try this again to check if this "rule" is still needed after https://github.com/Lightning-AI/lightning-thunder/pull/364 is merged?

IvanYashchuk: I removed this piece of code and the tests still passed.
```
pytest thunder/tests/test_grad.py -k "test_phantom_grad_vs_torch_consistency_reshape_"
```

mruberry (approved):
Looks like a good step forward. Let's address @IvanYashchuk's last concern and then get this merged ASAP because it's a large PR that touches a lot of the system. We can follow-up with more fixes/improvements/details in future PRs

thunder/core/transforms.py
jjsjann123: Ha thanks a lot. I'll clean it up.

jjsjann123:
ci looks green now and I have addressed @IvanYashchuk 's request regarding the grad rule above: https://github.com/Lightning-AI/lightning-thunder/pull/286#discussion_r1603248237

This one should be good to go @t-vi

thunder/core/proxies.py
jjsjann123: Marking as resolved and opened #428 to track

jjsjann123:
I realized that the merge button is green now. I'm gonna press that myself.

PR #291, opened 2024-04-29
moving grad rule `NE` , `GT`, `LE` to register_grad from augmented fwd/bwd

jjsjann123:
missing grad rule in `should_flatten_for_grad` triggers unintentional flatten.

### DIFF
--- a/thunder/core/transforms.py
+++ b/thunder/core/transforms.py
@@ -1046,6 +1046,9 @@ register_grad(pids.DIV, _div_prim_grad)
 register_grad(pids.EQ, prims.eq)
 register_grad(pids.GE, prims.ge)
 register_grad(pids.LT, prims.lt)
+register_grad(pids.NE, prims.ne)
+register_grad(pids.GT, prims.gt)
+register_grad(pids.LE, prims.le)
 
 
 @torchctx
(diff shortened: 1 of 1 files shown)


jjsjann123:
failing tests are from #296

jjsjann123:
IIUC, we should move everything to the new `register_grad`?!

IvanYashchuk (approved):
Thank you!

t-vi:
> IIUC, we should move everything to the new register_grad?!

Probably, I'm a big fan of the "There should be one-- and preferably only one --obvious way to do it." adage.

t-vi (approved):
Thank you @jjsjann123 @IvanYashchuk

jjsjann123:
> > IIUC, we should move everything to the new `register_grad`?!
> 
> I can't say we should, but we can. We support either way of registering backward formulas. We have a test for avoiding code duplication:
> 
> https://github.com/Lightning-AI/lightning-thunder/blob/843b2c265ec335aa1406f3d5650324567b88bf78/thunder/tests/test_grad.py#L902-L906

I don't have a preference on which form to use. The only reason that I moved it over is that because in `grad` we are using only the ones registered in `register_grad` but not from the old augmented fwd/bwd.
https://github.com/Lightning-AI/lightning-thunder/blob/b2b75c8818e210a327d02f1a920053fb9d5a2228/thunder/core/transforms.py#L1353-L1359

Maybe the right fix here is to actually modify the `should_flatten_for_grad` function to query augemented list instead of moving the grad function over, if we are going to keep both and use them in a combined fashion.

IvanYashchuk:
> The only reason that I moved it over is that because in grad we are using only the ones registered in register_grad but not from the old augmented fwd/bwd.

Right, the `grad` function was a prototype to showcase a possible way of going forward with grad rules and registering transforms, it's not used in the main code path and I need to remove it.

> Maybe the right fix here is to actually modify the should_flatten_for_grad function to query augemented list instead of moving the grad function over, if we are going to keep both and use them in a combined fashion.

We could go with this route but I think a simpler option is to rename "grad_v1" -> "grad" and all existing "phantom_grad" tests should still pass then remove the original "grad" function code.

Issue #119, opened 2024-04-02
test phantom grad fails with ops that return tensors that do not require grad


nikitaved:
## 🐛 Bug

As per title. To reproduce, in https://github.com/Lightning-AI/lightning-thunder/pull/118 one can just replace the calls `topk(...)[0]` with `topk(...)` to get:

```
RuntimeError: element 1 of tensors does not require grad and does not have a grad_fn
```
which is generated in `torch.autograd.grad`.

PR #293, opened 2024-04-29
test_phantom_grad: support ops which return non-differentiable outputs

kshitij12345:
Fixes: https://github.com/Lightning-AI/lightning-thunder/issues/119 (and also add xfail for getitem grad tests on Windows)


kshitij12345:
Windows failure looks related to : https://github.com/Lightning-AI/lightning-thunder/issues/296

thunder/tests/test_grad.py
```diff
@@ -1183,9 +1183,26 @@ def snippet_phantom_grad_vs_torch_consistency(op, torch_op, sample, comp, singul
 
     args, kwargs = sample.args, sample.kwargs
 
+    def is_output_differentiable(x):
+        # These are the differentiable dtypes in PyTorch.
+        return x.dtype.is_floating_point or x.dtype.is_complex
```
nikitaved, line 1188: Is that really sufficient? Shall we test binary ops with only one input requiring grad?

thunder/tests/test_grad.py
```diff
@@ -1218,6 +1237,7 @@ def upcast_tensors(x: Any) -> Any:
         f for f in reference_flats if isinstance(f, torch.Tensor) and f.requires_grad
     )
     reference_result = torch_op(*reference_args, **reference_kwargs)
+    reference_result = filter_differentiable_outputs(reference_result)
     reference_grad_result = torch.autograd.grad(reference_result, reference_tensors_requiring_grad, grads)
```
nikitaved, line 1241: Shouldn't we check that `len(grads) == len(reference_result)`? Or are we fine with `torch.autograd.grad` complaining?

thunder/tests/opinfos.py
```diff
@@ -3407,6 +3407,8 @@ def make_idx(dim_length: int, indices: int):
             executors=("nvfuser",),
             active_if=nvfuser_version < LooseVersion("0.1.4"),
         ),
+        DecorateInfo(pytest.mark.xfail, "test_vjp_correctness", active_if=IS_WINDOWS),
+        DecorateInfo(pytest.mark.xfail, "test_phantom_grad_vs_torch_consistency", active_if=IS_WINDOWS),
```
nikitaved, line 3411: What's the deal with Windows I wonder?

nikitaved: Ah, I see :)

thunder/tests/test_grad.py
kshitij12345: I don't think that will be a problem. Even in that case, we can get the grad for one of the inputs. Did you have something else in mind?

```python
>>> import torch
>>> x = torch.rand(3, 3, requires_grad=True)
>>> y = torch.rand(3, 3, requires_grad=False)
>>> o = x + y
>>> o
tensor([[1.8362, 0.3462, 0.8109],
        [0.1791, 1.6089, 1.4538],
        [1.2674, 1.1876, 1.6030]], grad_fn=<AddBackward0>)
```

kshitij12345: Also, we filter the inputs which require_grad before passing it to `torch.autograd.grad` 

Ref: https://github.com/Lightning-AI/lightning-thunder/blob/843b2c265ec335aa1406f3d5650324567b88bf78/thunder/tests/test_grad.py#L1203-L1204

nikitaved: The filtering takes place in the tests. What happens if I define an op with two floating point inputs and `put_grad` for only one of them?

kshitij12345: I think we can also check `grad_fn is not None` on the output. Will update to this.

```python
>>> import torch
>>> x = torch.ones(3, requires_grad=True)
>>> x + x
tensor([2., 2., 2.], grad_fn=<AddBackward0>)
>>> (x + x).grad_fn
<AddBackward0 object at 0x7f0502edcf40>
>>> y = torch.ones(3, requires_grad=False)
>>> (y + y).grad_fn
>>> (y + x).grad_fn
<AddBackward0 object at 0x7f0502e21060>
>>> x = torch.ones(3, requires_grad=True)
>>> x < 1
tensor([False, False, False])
>>> (x < 1).grad_fn
>>> torch.topk(x, k=2)
torch.return_types.topk(
values=tensor([1., 1.], grad_fn=<TopkBackward0>),
indices=tensor([0, 1]))
>>> torch.topk(torch.ones(3, requires_grad=False), k=2)
torch.return_types.topk(
values=tensor([1., 1.]),
indices=tensor([0, 1]))
```

kshitij12345: > The filtering takes place in the tests. What happens if I define an op with two floating point inputs and put_grad for only one of them?

In that case, we compute `grad` for only the outputs which are differentiable. In case of `topk`, we will pass the gradients of input w.r.t. `values` and not the indices.

kshitij12345: Ping @nikitaved

kshitij12345: Gentle ping @nikitaved

t-vi (approved):
Thank you  @kshitij12345 @nikitaved

PR #301, opened 2024-04-30
Print the complete dataframes

carmocca:
## What does this PR do?

Avoids this trimmed print

```text
AVERAGE ITERATION TIME (ms)
compiler       model_name  Num GPUS  Seq Len  Micro BS  Global BS  ...       eager   inductor     thunder
0         tiny-llama-1.1b         1     2048         1          1  ...  XXXXXXXXXX  XXXXXXXXX  XXXXXXXXXX

[1 rows x 13 columns]
```


t-vi (approved):
Sounds good! Thank you @carmocca

PR #302, opened 2024-04-30
Add `torch.unflatten`

### LINKED ISSUE #288: torch.unflatten not supported by thunder.jit
torch.unflatten is not supported by thunder.jit. However, we see that torch.reshape is supported by thunder.jit and we can rewrite torch.unflatten with torch.reshape to accomplish the same final result. 

Minimal repro:

```
import torch
import thunder

a = torch.randn(100, 128*3, device='cuda')

final_shape = list(a.shape)
final_shape[-1] = 3
final_shape.append(128)

# works
examine(torch.reshape, *(a, final_shape))

# fails
examine(torch.unflatten, *(a, -1, (3, 128)))
```

cc @apaz-cli

k223kim:
<details>

</details>

## What does this PR do?

Fixes #288. FYI I used `view` rather than `reshape`.

## PR review

## Did you have fun?

🚀🚀🚀

### DIFF
--- a/thunder/tests/opinfos.py
+++ b/thunder/tests/opinfos.py
@@ -3609,6 +3609,47 @@ reshape_opinfo = OpInfo(
 shape_ops.append(reshape_opinfo)
 
 
+def unflatten_sample_generator(op, device, dtype, requires_grad, **kwargs):
+    make = partial(make_tensor, device=device, dtype=dtype, requires_grad=requires_grad)
+
+    cases = (
+        ((4,), 0, (4,)),  # no-op
+        ((2, 2), -1, (2,)),  # no-op
+        ((1, 2, 1), 1, (2,)),  # no-op
+        ((4, 2), 0, (2, 2)),
+        ((125, 3), 0, (25, 5)),
+        ((25, 25), -1, (1, 5, 5)),
+        ((16, 32), -1, (2, 4, 1, 4)),
+        ((4, 5, 6), -1, (6, 1, 1, 1)),
+    )
+
+    for tensor_shape, dim, shape in cases:
+        yield SampleInput(make(tensor_shape), dim, shape)
+
+
+def unflatten_error_generator(op, device, dtype=torch.float32, **kwargs):
+    make = partial(make_tensor, dtype=dtype, device=device)
+    input_tensor = make(4, 4)
+    dim = 3
+    yield (SampleInput(input_tensor, 0, ()), RuntimeError, r"unflatten\(\) sizes must be non-empty")
+    # should check for dim?
+    yield (
+        SampleInput(input_tensor, dim, (2, 2)),
+        IndexError,
+        rf"Dimension out of range \(expected to be in range of \[{-len(input_tensor.shape)}, {len(input_tensor.shape)-1}\], but got {dim}\)",
+    )
+
+
+unflatten_opinfo = OpInfo(
+    ltorch.unflatten,
+    sample_input_generator=unflatten_sample_generator,
+    error_input_generator=unflatten_error_generator,
+    torch_reference=torch.unflatten,
+)
+
+shape_ops.append(unflatten_opinfo)
+
+
 def repeat_sample_generator(op, device, dtype, requires_grad, **kwargs):
     make = partial(make_tensor, device=device, dtype=dtype, requires_grad=requires_grad)
 
--- a/thunder/tests/opinfos.py
+++ b/thunder/tests/opinfos.py
@@ -3612,6 +3612,7 @@ shape_ops.append(reshape_opinfo)
 def unflatten_sample_generator(op, device, dtype, requires_grad, **kwargs):
     make = partial(make_tensor, device=device, dtype=dtype, requires_grad=requires_grad)
 
+    # shape, dim, unflatten_shape
     cases = (
         ((4,), 0, (4,)),  # no-op
         ((2, 2), -1, (2,)),  # no-op
@@ -3621,6 +3622,11 @@ def unflatten_sample_generator(op, device, dtype, requires_grad, **kwargs):
         ((25, 25), -1, (1, 5, 5)),
         ((16, 32), -1, (2, 4, 1, 4)),
         ((4, 5, 6), -1, (6, 1, 1, 1)),
+        ((5, 125, 5), 1, (5, 5, 5)),
+        ((4, 12, 6), 1, (2, 2, 3)),
+        ((12, 2), 0, (-1, 6)),
+        ((4, 12, 6), 1, (12, -1)),
+        ((4, 12, 6), 1, (6, -1)),
     )
 
     for tensor_shape, dim, shape in cases:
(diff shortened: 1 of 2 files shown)


thunder/tests/opinfos.py
```diff
@@ -3609,6 +3609,47 @@ def reshape_sample_generator(op, device, dtype, requires_grad, **kwargs):
...
+
+    cases = (
+        ((4,), 0, (4,)),  # no-op
+        ((2, 2), -1, (2,)),  # no-op
+        ((1, 2, 1), 1, (2,)),  # no-op
+        ((4, 2), 0, (2, 2)),
+        ((125, 3), 0, (25, 5)),
+        ((25, 25), -1, (1, 5, 5)),
+        ((16, 32), -1, (2, 4, 1, 4)),
+        ((4, 5, 6), -1, (6, 1, 1, 1)),
```
nikitaved, line 3623: PyTorch allows `-1` sizes. Let's test these as well?

thunder/torch/__init__.py
```diff
@@ -782,6 +782,17 @@ def reshape(a: TensorLike, /, *shape: int) -> TensorLike:
     return clang.reshape(a, shape)
 
 
+@torchsymbol(torch.unflatten, is_method=True)
+def unflatten(a: TensorLike, /, dim: int, sizes=tuple[int, ...]) -> TensorLike:
```
mruberry, line 786: This signature looks pretty nice, but `sizes-tuple[int, ...]` is assigning the type annotation, and the type annotation is overly strict because any `Sequence` is OK, so it should be:

```
sizes: Sequence[int]
```

thunder/torch/__init__.py
```diff
@@ -782,6 +782,17 @@ def reshape(a: TensorLike, /, *shape: int) -> TensorLike:
...
+    utils.check(
```
mruberry, line 787: Is this error check necessary, or will the error be caught by `view`?

thunder/torch/__init__.py
```diff
@@ -782,6 +782,17 @@ def reshape(a: TensorLike, /, *shape: int) -> TensorLike:
...
+        len(sizes) > 0,
+        lambda: f"unflatten() sizes must be non-empty",
+        RuntimeError,
+    )
+    dim = utils.canonicalize_dim(a.ndim, dim)
```
mruberry, line 792: Nice canonicalization

thunder/tests/opinfos.py
```diff
@@ -3609,6 +3609,47 @@ def reshape_sample_generator(op, device, dtype, requires_grad, **kwargs):
...
+
+    for tensor_shape, dim, shape in cases:
+        yield SampleInput(make(tensor_shape), dim, shape)
+
+
+def unflatten_error_generator(op, device, dtype=torch.float32, **kwargs):
+    make = partial(make_tensor, dtype=dtype, device=device)
+    input_tensor = make(4, 4)
+    dim = 3
+    yield (SampleInput(input_tensor, 0, ()), RuntimeError, r"unflatten\(\) sizes must be non-empty")
```
mruberry, line 3634: Two more error cases that are interesting:

- the unflattened shape is fully specified and invalid. For example, a dimension of length 12 and a request to unflatten it to `(4, 4)`
- the unflattened shape is partially specified and invalid. For example, a dimension of length 12 and a request to unflatten it to `(-1, 5)`.

I imagine that the call to `view()` will handle these errors properly for these cases, but it's still nice to check them from `unflatten()`

thunder/tests/opinfos.py
mruberry: Adding tests for partial shape specification is great. Let's also add some tests of unflattening dimensions which are not the outermost or innermost, like `(4, 12, 6), 1, ...`

thunder/tests/opinfos.py
```diff
@@ -3609,6 +3609,47 @@ def reshape_sample_generator(op, device, dtype, requires_grad, **kwargs):
 shape_ops.append(reshape_opinfo)
 
 
+def unflatten_sample_generator(op, device, dtype, requires_grad, **kwargs):
+    make = partial(make_tensor, device=device, dtype=dtype, requires_grad=requires_grad)
+
+    cases = (
```
mruberry, line 3615: A short comment like 

```
# shape, dim, unflatten_shape
```

before `cases` would improve readability

thunder/tests/opinfos.py
```diff
@@ -3609,6 +3609,47 @@ def reshape_sample_generator(op, device, dtype, requires_grad, **kwargs):
...
+        ((4,), 0, (4,)),  # no-op
+        ((2, 2), -1, (2,)),  # no-op
+        ((1, 2, 1), 1, (2,)),  # no-op
```
mruberry, line 3618: Nice no-op testing

mruberry (commented):
Looking good so far, @k223kim! I made some comments for your review

thunder/tests/opinfos.py
k223kim: Awesome idea! I thought error cases should only include the ones caught by this specific function (unflatten) but it makes sense to check these.

thunder/tests/opinfos.py
k223kim: Sure thing! I wasn't aware that `-1` sizes were allowed. Thanks so much!

thunder/torch/__init__.py
k223kim: You are right! Didn't know PyTorch supported lists and other sequence as well. Will update :) Thanks!

thunder/torch/__init__.py
k223kim: This error will be caught by `view`, specifically, in `reshape`! However, it throws an error that seems specific for `reshape`. For instance, in the case of `SampleInput(input_tensor, 0, ())`, the error message will be something like the following:
```
Attempting to reshape a.shape=(4, 4) to shape=(4,), but a.numel=16 is different from the number of elements in shape, 4
```
I thought this error message is clearer? What do you think? If not, I can definitely remove this.

thunder/torch/__init__.py
```diff
@@ -782,6 +782,17 @@ def reshape(a: TensorLike, /, *shape: int) -> TensorLike:
...
+    return a.view(tuple(a.shape[:dim]) + tuple(sizes) + tuple(a.shape[dim + 1 :]))
```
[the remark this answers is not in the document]
k223kim: Hmm, if I am understanding this correctly, when doing `a.shape`, it returns `torch.Size([...])` which is a `torch.Size` class. Won't we want an explicit conversion to `tuple`?

k223kim:
Hey @mruberry ! Thanks for the detailed feedback as always. Updated some lines and I have added more test cases as per the comments. I do have some questions regarding some of your comments and would appreciate it if you can take a look!

thunder/torch/__init__.py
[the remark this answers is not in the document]
k223kim: Oh I see. You are right. I completely missed that. Will update this right now :) Thank you!

mruberry (approved):
Awesome! Nice reuse of `view`

fyi @Borda and @t-vi to merge

Issue #284, opened 2024-04-26
implement zip lookaside in Python interpreter (enables e.g. thunder.jit with zip from LitGPT LLaMAMoE)


IvanYashchuk:
## 🐛 Bug

Here's a simplified version of LitGPT's LLaMAMoE without data-dependent shapes and it fails somewhere in the general jit:
```py
NotImplementedError: unpacking from OPAQUE <slot wrapper '__next__' of 'zip' objects> ProvenanceRecord(
```
To reproduce:
```py
import torch
import thunder
from torch import nn

class Test(nn.Module):
    def __init__(self) -> None:
        super().__init__()
        self.n_expert = 8
        self.n_expert_per_token = 2
        self.C = 2
        self.gate = nn.Linear(self.C, self.n_expert, bias=False)
        self.experts = nn.ModuleList(nn.Linear(2, 2, bias=False) for _ in range(self.n_expert))

    def forward(self, x: torch.Tensor) -> torch.Tensor:
        B, T, C = x.size()  # batch size, sequence length, embedding dimensionality (n_embd)
        x = x.view(-1, C)  # (B*T, C)
        router = self.gate(x)  # (B*T, n_expert)
        probs, indices = torch.topk(router, self.n_expert_per_token)  # (B*T, n_expert_per_token)
        probs = probs.softmax(dim=1, dtype=torch.float).to(dtype=x.dtype)
        masks = indices.unsqueeze(-1) == torch.arange(self.n_expert, device=x.device)
        masks = masks.permute(2, 0, 1)  # (n_expert, B*T, n_expert_per_token)
        y = torch.zeros_like(x)  # (B*T, C)
        for (mask, expert) in zip(masks, self.experts):
            token_idx, expert_idx = torch.arange(B*T, device=x.device), torch.arange(B*T, device=x.device)
            pprobs = probs[token_idx, expert_idx]
            pprobs = pprobs.unsqueeze(-1)
            eexpert = expert(x[token_idx])
            y = torch.index_add(y, 0, token_idx, pprobs * eexpert)
        return y.view(B, T, C)

model = Test()
model = thunder.jit(model)

x = torch.randn(2, 3, 2)
y = model(x)
```
raises:
```py
---------------------------------------------------------------------------
NotImplementedError                       Traceback (most recent call last)
File ~/dev/lightning-thunder/thunder/core/jit_ext.py:1272, in unpack_inputs.<locals>.unpack(v)
   1271 try:
[...]

PR #303, opened 2024-04-30
Add MoE layer example

IvanYashchuk:
**DRAFT MODE TO PREVENT MERGES** The approach and code is ready for experimentation and review.

The main result of this PR is that Thunder can run a variant of the MoE layer from LitGPT. There are three modifications
1. zip is replaced with a for-loop with explicit indexing into lists (blocked by https://github.com/Lightning-AI/lightning-thunder/issues/284).
2. Thunder doesn't support advanced indexing with `None` (need to create an issue). The workaround is to use unsqueeze instead of `None` when indexing.
3. Inplace addition (`+=`) is replaced with `index_add`.

The main missing operator is `nonzero(x, as_tuple=True)`. The problem with this operator is that the output shape is unknown at compile time and it's dynamic at runtime. I tried using NumberProxy with None, NumberProxy with a custom int subclass as value, using a custom int subclass directly. But simple `-1` in the shape worked best.

The forward pass worked just with 14ce0978aaf32803884fe66b891b6a4ccd2fca7d. The backward pass required more of `-1`-special handling.

Currently, `index_add`, `index_select`, `topk` are not fused with any of Thunder's fusing executors.


mruberry:
Super exciting! Really looking forward to discuss this in more detail at a design review!

mruberry:
@t-vi and @carmocca, I think you'll be interested in this

apaz-cli:
Do we have any broader ideas for how this fits into the strategy for handling dynamic and data dependent shapes? I was under the impression that this was just something we were completely incapable of doing with the way that we're modeling traces.

thunder/core/utils.py
```diff
@@ -1048,6 +1053,7 @@ def find_producer_symbols(trace: TraceCtx, proxies: Sequence[Proxy], stop_proxie
         (__b = ltorch.sub(x, y)
         # __b = prims.sub(x, y),)
     """
+    stop_proxies = filter(lambda x: isinstance(x, Proxy), stop_proxies)
```
IvanYashchuk, line 1056:
```suggestion
    stop_proxies = tuple(filter(lambda x: isinstance(x, Proxy), stop_proxies))
```

jjsjann123:
> Thunder doesn't support advanced indexing with None (need to create an issue). The workaround is to use unsqueeze instead of None when indexing.

Vaguely remember that I have run into `None` in indexing, but I think I was just seeing that with basic indexing....

PR #208, opened 2024-04-17
[transformer_engine executor] Update args to `_Linear.forward` and `Context`

crcrpar:
Switch to a dict from a sequence of arguments for a bit better solidity, also some optimistic patch hoping TransformerEngine executor a bit more nicely supports nvidia/transformerengine repo.

Issue #296, opened 2024-04-29
getitem grad is not calculated properly on Windows 


mruberry:
See https://github.com/Lightning-AI/lightning-thunder/actions/runs/8873124592/job/24358511214?pr=273 and https://github.com/Lightning-AI/lightning-thunder/actions/runs/8871047445/job/24353547585?pr=278

PR #325, opened 2024-05-01
TE: simplify context

kshitij12345:
> we should simplify Context as following, so that we don't need to care about exact information being saved for backward. I have verified this on v1.5 (stable), v1.6 (next release) and v1.7 (main) and it passes the tests on all three.

Ref: https://github.com/Lightning-AI/lightning-thunder/pull/208#issuecomment-2063770993

It works on TE v1.5 (stable), v1.6 (next release) and v1.7 (main)


kshitij12345:
Failures are related to: https://github.com/Lightning-AI/lightning-thunder/issues/296

carmocca (approved):

Issue #349, opened 2024-05-02
using size() twice in a row on same tensor causes interpreter out of sync error with thunder jit


Fuzzkatt:
Using size() twice in a row on the same tensor causes the interpreter to hit an out of sync error with thunder jit.

Minimal repro:
```
import torch
import thunder

def func(a):
    b = a.size()
    c = a.size() # expect error here
    return a

a = torch.randn(100, 100, device='cuda')
jfunc = thunder.jit(func)
jfunc(a)
```

Full error log:
```
Traceback (most recent call last):
  File "/opt/pytorch/lightning-thunder/thunder/core/interpreter.py", line 6683, in fn_
    interpretation_result: Any = _interpret_call(wrapped_fn_2, args, kwargs)
  File "/opt/pytorch/lightning-thunder/thunder/core/interpreter.py", line 6009, in _interpret_call
    rval = _call_dispatch(compilectx, runtimectx, fn, *args, **kwargs)  # type: ignore
  File "/opt/pytorch/lightning-thunder/thunder/core/interpreter.py", line 6239, in _call_dispatch
    return _setup_frame_and_run_python_function(compilectx, runtimectx, wrapped_fn, *args, **kwargs)
  File "/opt/pytorch/lightning-thunder/thunder/core/interpreter.py", line 6368, in _setup_frame_and_run_python_function
    raise e.with_traceback(tb)
  File "/opt/pytorch/lightning-thunder/thunder/core/interpreter.py", line 6669, in fn_2
    return fn(*args, **kwargs)
  File "/opt/pytorch/lightning-thunder/thunder/core/interpreter.py", line 6363, in _setup_frame_and_run_python_function
    res, status = _run_frame(frame, compilectx, runtimectx)
  File "/opt/pytorch/lightning-thunder/thunder/core/interpreter.py", line 6409, in _run_frame
    interpretation_result: None | int | INTERPRETER_SIGNALS = compilectx.interpret(
  File "/opt/pytorch/lightning-thunder/thunder/core/interpreter.py", line 412, in interpret
    return self._opcode_interpreter(inst, **interpreter_state)
  File "/opt/pytorch/lightning-thunder/thunder/core/interpreter.py", line 1231, in default_opcode_interpreter
    return handler(inst, **interpreter_state)
  File "/opt/pytorch/lightning-thunder/thunder/core/interpreter.py", line 3631, in _call_function_ex_handler
[...]

Issue #289, opened 2024-04-27
torch.nn.TransformerEncoder hits error with thunder.jit


Fuzzkatt:
Tracking my attempt to make torch.nn.Transformer work with thunder.jit. Currently after applying a series of workarounds, we can get torch.nn.TransformerEncoderLayer to work with thunder.jit. However, we hit a new error with torch.nn.TransformerEncoder:

`transformer: thunder.core.interpreter.InterpreterError: Encountered exception AssertionError: attribute size of TensorProxy object out of sync: <function prop_lookaside_wrap.<locals>.fn.<locals>.fn_ at 0x7fe1c457a440> vs. <function prop_lookaside_wrap.<locals>.fn.<locals>.fn_ at 0x7fe1c457b0a0> while tracing Transformer(`

which mostly just stacks n copies of TransformerEncoderLayers into one module. 

Workarounds: 
* comment out is_nested check in https://github.com/pytorch/pytorch/blob/main/torch/nn/modules/activation.py#L1236-L1238
* rewrite unflatten as reshape in https://github.com/pytorch/pytorch/blob/main/torch/nn/functional.py#L4909 (Ex. https://github.com/Lightning-AI/lightning-thunder/issues/288) 
* comment out is_nested check in https://github.com/pytorch/pytorch/blob/main/torch/nn/modules/transformer.py#L42

Minimal repro:

```
import torch
import thunder
from thunder.examine import examine

src = torch.randn(10, 32, 512, device='cuda')
encoder_layer = torch.nn.TransformerEncoderLayer(d_model=512, nhead=8).to('cuda')
encoder = torch.nn.TransformerEncoder(encoder_layer, num_layers=6).to('cuda')

# works after applying workarounds
jfunc = thunder.jit(encoder_layer)
jfunc(src)
print('encoder layer working')

# fails even after applying workarounds
jfunc = thunder.jit(encoder)
jfunc(src)

```

Fuzzkatt:
https://github.com/pytorch/pytorch/blob/main/torch/nn/modules/transformer.py#L921 using copy.deep_copy which is called here https://github.com/pytorch/pytorch/blob/main/torch/nn/modules/transformer.py#L276 looks suspicious; how would this work with thunder.jit?

mruberry:
triage review — @t-vi can add more details about this issue

Fuzzkatt:
[...]

PR #352, opened 2024-05-03
add size fix

Fuzzkatt:
Fixes https://github.com/Lightning-AI/lightning-thunder/issues/349 and by extension https://github.com/Lightning-AI/lightning-thunder/issues/289 which is blocked by it. Rewrites size in a way that is consistent with the other ops and removes the special logic around it that previously existed.

### DIFF
--- a/thunder/torch/__init__.py
+++ b/thunder/torch/__init__.py
@@ -170,21 +170,16 @@ def is_floating_point(a: TensorLike, /) -> bool:
 
 
 # Handles the size method
-def size(a):
-    def fn_(idx: int | None = None):
-        if idx is None:
-            return a.shape
-        return a.shape[idx]
-
-    return fn_
+@torchsymbol(torch.Tensor.size, is_method=True, id="torch.Tensor.size")
+def size(a: TensorLike, dim: None | int = None) -> int | TensorLike:
+    if dim:
+        return a.shape[dim]
+    return a.shape
 
 
 @torchsymbol(torch.Tensor.is_cuda, is_property=True, id="torch.is_cuda")
 def is_cuda(a: TensorLike, /) -> bool:
     return a.device.devicetype is devices.DeviceType.CUDA
-
-
-register_method("size", size)
 
 
 #
--- a/thunder/core/proxies.py
+++ b/thunder/core/proxies.py
@@ -1128,11 +1128,6 @@ class FutureTensorProxy(Proxy, TensorProxyInterface):
     def type_string(self):
         return f"FUTURE {self.device} {self.dtype.shortname()}{list(self.shape)}"
 
-    @property
-    def size(self, /) -> Any:
-        fn = resolve_method("size", self)
-        return fn(self)
-
     def wait(self) -> TensorProxy:
         from thunder.distributed.prims import wait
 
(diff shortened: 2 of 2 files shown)


t-vi:
Thank you @Fuzzkatt .

These errors relate to `size`, so I think if they might well be related to the change:
```
FAILED thunder/tests/test_cudnn_executor.py::test_vjp_correctness_cudnn_sdpa[float16-may-cat-grad-qkv] - AssertionError: A symbol [Symbol name=size] was called while processing a primitive
FAILED thunder/tests/test_cudnn_executor.py::test_vjp_correctness_cudnn_sdpa[bfloat16-may-cat-grad-qkv] - AssertionError: A symbol [Symbol name=size] was called while processing a primitive
```

thunder/torch/__init__.py
```diff
@@ -170,23 +170,18 @@ def is_floating_point(a: TensorLike, /) -> bool:
...
 # Handles the size method
-def size(a):
-    def fn_(idx: int | None = None):
-        if idx is None:
-            return a.shape
-        return a.shape[idx]
-
-    return fn_
+@torchsymbol(torch.Tensor.size, is_method=True, id="torch.Tensor.size")
+def size(a: TensorLike, dim: None | int = None) -> int | TensorLike:
```
mruberry, line 174: Nit: 

```
(a: TensorLike, /, dim: None | int = None) -> int | Sequence[int]...
```

The shape isn't stored as a tensor, but a sequence of integers (it actually might always be a tuple of integers)

fyi @jjsjann123

thunder/torch/__init__.py
```diff
@@ -170,23 +170,18 @@ def is_floating_point(a: TensorLike, /) -> bool:
...
 
 # Handles the size method
-def size(a):
-    def fn_(idx: int | None = None):
-        if idx is None:
-            return a.shape
-        return a.shape[idx]
-
-    return fn_
+@torchsymbol(torch.Tensor.size, is_method=True, id="torch.Tensor.size")
```
mruberry, line 173: This PR is a good overall direction to go in. I think the issue is that we've been too aggressive at labeling some functions symbols.

Could we update this function, `is_floating_point`, and `is_cuda` to not be symbols? Updating all of them may be beyond the scope of this PR, so maybe this should just modify the changes to `size` so it's not a symbol.

I think we don't want to make this a symbol because it's just extracting the shape from a tensor, which is currently constexpr (so the symbol is constexpr, and doesn't need to be evaluated), and in the future will just return an existing proxy, anyway.

fyi @jjsjann123, @IvanYashchuk and @nikitaved on this thinking

mruberry (commented):
Good idea, @Fuzzkatt! Made a couple comments about refining this approach

thunder/torch/__init__.py
Fuzzkatt: Fixed

thunder/torch/__init__.py
Fuzzkatt: Fixed

Fuzzkatt:
> Thank you @Fuzzkatt .
> 
> These errors relate to `size`, so I think if they might well be related to the change:
> 
> ```
> FAILED thunder/tests/test_cudnn_executor.py::test_vjp_correctness_cudnn_sdpa[float16-may-cat-grad-qkv] - AssertionError: A symbol [Symbol name=size] was called while processing a primitive
> FAILED thunder/tests/test_cudnn_executor.py::test_vjp_correctness_cudnn_sdpa[bfloat16-may-cat-grad-qkv] - AssertionError: A symbol [Symbol name=size] was called while processing a primitive
> ```

@t-vi, just ran these tests manually, they should be fixed as of the latest commit as I removed the @torchsymbol decorator from size based on @mruberry's suggestion.

PR #354, opened 2024-05-03
Reenable vjp nvfuser ge ne le

IvanYashchuk:
Enable skipped and xfailed tests for test_vjp_correctness of comparison operators as suggested in https://github.com/Lightning-AI/lightning-thunder/pull/291#issuecomment-2090509548.


carmocca (approved):
Thanks!

PR #318, opened 2024-04-30
Enable `linear` for nvFuser

Priya2698:
## What does this PR do?

Enables linear in nvFuser. Part of resolving https://github.com/NVIDIA/Fuser/issues/2053

PR #357, opened 2024-05-03
Update test_linear_nvfuser

IvanYashchuk:
Thunder's nvFuser executor got support for `prims.linear` in https://github.com/Lightning-AI/lightning-thunder/pull/318. This PR updates the test
* to use the sample generator from OpInfos
* to reject linear execution if inputs are not bf16, fp16
* to skip the test if the nvFuser version is not new enough

### DIFF
--- a/thunder/executors/nvfuserex_impl.py
+++ b/thunder/executors/nvfuserex_impl.py
@@ -2209,6 +2209,12 @@ def _linear_check(a: TensorProxy, b: TensorProxy, bias: TensorProxy | None) -> b
     if bias is not None and not is_supported_tensor(bias):
         return False
 
+    # nvFuser supports only fp16 and bf16 inputs. Only checking the first tensor
+    # dtype, as all tensors should have the same dtype which is checked by
+    # linear_meta.
+    if a.dtype not in (dtypes.float16, dtypes.bfloat16):
+        return False
+
     # nvFuser only supports 2D inputs in v0.2.3.
     if not a.ndim == 2:
         return False
--- a/thunder/tests/test_nvfuser.py
+++ b/thunder/tests/test_nvfuser.py
@@ -854,7 +854,12 @@ def test_optimization_fuel(executor, device, _):
 
 
 @instantiate(
-    dtypes=(thunder.float16, thunder.bfloat16), devicetypes=(devices.DeviceType.CUDA,), executors=(nvFuserExecutor,)
+    dtypes=(thunder.float16, thunder.bfloat16),
+    devicetypes=(devices.DeviceType.CUDA,),
+    executors=(nvFuserExecutor,),
+    decorators=(
+        pytest.mark.skipif(nvfuser_version() < LooseVersion("0.2.3"), reason="Requires nvFuser version 0.2.3 or later"),
+    ),
 )
 def test_linear(executor, device: str, dtype: dtypes.dtype):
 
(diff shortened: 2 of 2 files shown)


crcrpar (approved):

thunder/tests/test_nvfuser.py
```diff
@@ -854,32 +854,35 @@ def get_num_fusions(cfn):
...
 def test_linear(executor, device: str, dtype: dtypes.dtype):
 
     def fn(a, b, bias=None):
         return torch.nn.functional.linear(a, b, bias)
 
-    m, n, k = 128, 64, 32
-    torch_dtype = ltorch.to_torch_dtype(dtype)
-    a = torch.randn((m, k), dtype=torch_dtype, device=device)
-    b = torch.randn((n, k), dtype=torch_dtype, device=device)
+    at_least_one_tested = False
```
nikitaved, line 870: What does that mean?

IvanYashchuk: There's a for loop over sample inputs, inside the for loop there's a continue statement to skip inputs that are not 2D. If someone goes ahead and modifies the sample input generator for linear to not generate a 2D sample this test won't test anything silently. With assert on "at least one sample input tested," there would be a failure.

thunder/executors/nvfuserex_impl.py
```diff
@@ -2209,6 +2209,12 @@ def _linear_check(a: TensorProxy, b: TensorProxy, bias: TensorProxy | None) -> b
     if bias is not None and not is_supported_tensor(bias):
         return False
 
+    # nvFuser supports only fp16 and bf16 inputs. Only checking the first tensor
+    # dtype, as all tensors should have the same dtype which is checked by
+    # linear_meta.
+    if a.dtype not in (dtypes.float16, dtypes.bfloat16):
+        return False
```
jjsjann123, line 2216: Wasn't aware there's such restriction.

cc'ing @Priya2698 is this an intended behavior? I thought we are just dispatching to aten now and don't see why it's rejecting full precision.

Priya2698: At the moment, we do have that restriction since we use `fusedMultiplySum` and `MmaOp` that only accept bf16/fp16.

Priya2698: This restriction will although be lifted with the new IR nodes.

Priya2698 (approved):
Thanks for creating this example. I'll update the matmul test similarly once we have complete support.

thunder/executors/nvfuserex_impl.py
jjsjann123: Got'ya.

Thanks @IvanYashchuk for catching this. @Priya2698 remember to do a version bump when that restriction is lifted, so we can have a proper behavior inserted here.

thunder/tests/test_nvfuser.py
IvanYashchuk: Yes, it's a viable option.

lantiga:
@IvanYashchuk can we merge this after resolving the conflicts?

IvanYashchuk:
This PR is superseded by https://github.com/Lightning-AI/lightning-thunder/pull/434 and https://github.com/Lightning-AI/lightning-thunder/pull/419.

Issue #347, opened 2024-05-02
`use_cudagraphs=True` is silently ignored.


nikitaved:
## 🐛 Bug

`thunder.jit` sets it to `False` and explicit setting with `thunder.jit(..., use_cudagraphs=True)` is ignored.

IvanYashchuk:
This problem is not fixed https://github.com/Lightning-AI/lightning-thunder/pull/358#issuecomment-2092851744

PR #358, opened 2024-05-03
thunder.jit - fix ignoring compile options

nikitaved:
As per title. This enables using `use_cudagraphs` and other options that were previously ignored.
Fixes https://github.com/Lightning-AI/lightning-thunder/issues/347

### DIFF
--- a/thunder/__init__.py
+++ b/thunder/__init__.py
@@ -407,11 +407,7 @@ def jit(
     )
     # TODO: rewrite the code below once options are finalized
     # NOTE: the following options might be absorbed by `compile_options`
-    absorbed_options = {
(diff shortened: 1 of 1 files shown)


t-vi (commented):
Thank you n Nik!

IvanYashchuk:
`use_cudagraphs`, `use_torch_compile`, and `use_rematerialization` are still ignored because there's no code path that actually uses these flags in `thunder.jit`. 

These flags are used only in `thuner.common._create_callable` that is only used in `thunder.compile`.

nikitaved:
@IvanYashchuk , yes, it is a follow-up!

PR #359, opened 2024-05-03
FSDP as a regular transform

t-vi:
User facing: do `fsdp(jit(m))` instead of `jit(fsdp(m))`.

Key properties:
- This introduces "early transforms", i.e. the code transformation potentially affects prologue, computation, and epilogue, 
- The ThunderModule is adapted (here with sharding) and matches the code changes,
- The original module is left alone, we are using overrides on the ThunderModule.
- Things are intended to be composable with other early transforms (and other "classic" transforms). 

The plan is to merge this and then incrementally switch all tests.
In the medium term, we will want abstractions for some of the rewriting, but it would be useful to inform this endeavor by the next transforms (quantization, lora) of this nature. 


t-vi:
@crcrpar @IvanYashchuk

t-vi:
This builds on #279 , I'll rebase once it is in.

IvanYashchuk:
The error is expected because `thunder.distributed.prims.synchronize` was not meant to be executed with real PyTorch tensors. It's only a Thunder operation to work with Thunder's TensorProxies, it should be replaced with nothing or allgather in the transformations that should happen before "transform_for_execution".

You could register it in the torch executor with "execution_transform" checking the ddp_type attribute and dispatching it to either nothing or allgather like here: https://github.com/Lightning-AI/lightning-thunder/blob/7ac5684e274c9c14336759451add371221011c8b/thunder/distributed/prims.py#L291-L308

thunder/__init__.py
```diff
@@ -211,6 +211,19 @@ def _get_cache_info():
     return _cache_info_ctx.get()
 
 
+def _get_thunder_module(model):
+    cd = get_compile_data()
+    tm = cd._thunder_module_map.get(id(model))
```
crcrpar, line 216: purely my taste but
```python
    if (model_hash := id(model) not in cd._thunder_module_map):
        return model
    else:
        tm = cd._thunder_module_map[model_hash]
        if tm._model is not model:
            return model
        return tm
```

thunder/__init__.py
```diff
@@ -260,6 +273,7 @@ def jit(
     interpretation: None | INTERPRETATION_OPTIONS | str = None,
     cache: None | CACHE_OPTIONS | str = None,
     disable_torch_autograd: bool = False,  # TODO Revisit this UX for RC1
+    early_transforms: list | None = None,
```
crcrpar, line 276: in what sense "early"?

thunder/__init__.py
```diff
@@ -471,25 +488,41 @@ def get_computation_and_inputs(*args, **kwargs):
...
-                epilogue = None
 
             cs.last_trace_tracing_stop = time.time_ns()
 
             # Makes the prologue callable
             cs.last_prologue_transformation_start = time.time_ns()
-            protraces = transform_for_execution(
+
+            for transform in early_transforms:
+                prologue_trc, computation_trc, epilogue_trc = transform(
```
crcrpar, line 505: Q: would we have a spec of `transform: Callable`?

thunder/core/module.py
```diff
@@ -19,16 +19,22 @@ def __init__(self, model, compiled_model_call):
         super().__init__()
         self._model = model
 
-        # We delete self.training in order for training to be used from
-        # the model itself through `__getattr__`.
         del self.training
-
         self._forward_fn = compiled_model_call
+        self._overrides = {}
```
crcrpar, line 24: a docstring or a comment would be great

thunder/core/module.py
```diff
@@ -19,16 +19,22 @@ def __init__(self, model, compiled_model_call):
...
 
     def get_buffer(self, name):
+        null = object()
+        p = self._overrides.get(name, null)
+        if p is not null:
+            return p
```
crcrpar, line 30: why not
```python
if name in self._overrides:
    return self._overrides[name]
return self._model.get_buffer(name)
```
?

thunder/core/module.py
```diff
@@ -19,16 +19,22 @@ def __init__(self, model, compiled_model_call):
...
         return self._model.get_buffer(name)
 
     def get_parameter(self, name):
+        null = object()
+        p = self._overrides.get(name, null)
+        if p is not null:
+            return p
```
crcrpar, line 37: ditto

thunder/distributed/__init__.py
```diff
@@ -349,6 +350,162 @@ def f(tensor: TensorProxy) -> str:
...
 
 
+def fsdp_transform_module(
+    thunder_model: "thunder.ThunderModule",
+    *,
+    device: torch.device | None = None,
+    broadcast_from: int | None = None,
+    sharding_strategy: FSDPType = FSDPType.ZERO2,
+    bucketing_strategy: FSDPBucketingStrategy = FSDPBucketingStrategy.NONE,
+) -> "thunder.ThunderModule":
```
crcrpar, line 360: i'm speculating that the following would let us remove double quotes
```python
from __future__ import annotations
...
if TYPE_CHECKING:
    from thunder import ThunderModule
```

thunder/tests/distributed/test_ddp.py
```diff
@@ -1228,7 +1228,7 @@ def _test_native_ddp_helper(input_data):
     return None
 
 
-def _test_native_fsdp_helper(input_data):
+def _test_native_fsdp_helper(input_data, new_fsdp=True):
```
crcrpar, line 1231: where is paramtrization of `new_fsdp` defined?

thunder/distributed/__init__.py
```diff
@@ -349,6 +350,162 @@ def f(tensor: TensorProxy) -> str:
...
+
+        new_computation_trace.set_provenance(thunder.core.trace.TraceProvenance("fsdp pass"))
+
+        return prologue_trace, new_computation_trace, epilogue_trace
+
+    # add prologue + compute transform
+    thunder_model = thunder.core.transforms.add_transform(thunder_model, early_transform=prologue_and_compute_transform)
+
+    # modidfy module
+    import copy
```
crcrpar, line 456: move this to the header

thunder/distributed/__init__.py
```diff
@@ -349,6 +350,162 @@ def f(tensor: TensorProxy) -> str:
...
+    *,
+    device: torch.device | None = None,
+    broadcast_from: int | None = None,
+    sharding_strategy: FSDPType = FSDPType.ZERO2,
+    bucketing_strategy: FSDPBucketingStrategy = FSDPBucketingStrategy.NONE,
+) -> "thunder.ThunderModule":
+    import thunder
+
+    cd = thunder.compile_data(thunder_model)
+    cd.use_fsdp = True
```
crcrpar, line 364: wouldn't it be high time we promoted `use_fsdp` and `use_ddp` to an public member of
 https://github.com/Lightning-AI/lightning-thunder/blob/ef3bcd07e55579e2b3d57da75aabb38fe1ae6862/thunder/common.py#L139-L158?

thunder/distributed/__init__.py
```diff
@@ -349,6 +350,162 @@ def f(tensor: TensorProxy) -> str:
...
+
+        modules_and_thunder_modules = [
+            (bsym.args[0], bsym.output)
+            for bsym in prologue_trace.bound_symbols
+            if bsym.sym is thunder.prims.unpack_thunder_module
+        ]
+
+        if (
+            len(modules_and_thunder_modules) != 1
+            or prologue_producers[modules_and_thunder_modules[0][0]].sym is not thunder.prims.unpack_function_obj
```
crcrpar, line 388: could this get more readable?
The hardcoded [0][0] feels not straightforward to me

thunder/__init__.py
t-vi: I should add a comment, but I would view this as a sort of "weak reference", so 217+218 are part of the lookup.

thunder/__init__.py
t-vi: before the prologue is run. I'm starting to wonder whether we should do this to all transforms, but this is a second step.

thunder/core/module.py
t-vi: It does one lookup more / is inherently not thread safe (not of matter here). Probably not that bad in the grand scheme of things (and compared to creating the null object).

thunder/distributed/__init__.py
t-vi: Good call, thank you. I'll move the destructuring up and do the checking afterwards.

thunder/tests/distributed/test_ddp.py
t-vi: :grin: I should just use new_fsdp unconditionally. The plan is to switch a first test to fsdp in this PR, and then switch over the other tests and finally remove the old fsdp.

thunder/__init__.py
```diff
@@ -211,6 +211,19 @@ def _get_cache_info():
     return _cache_info_ctx.get()
 
 
+def _get_thunder_module(model):
```
carmocca, line 214: Aren't you duplicating this? Since you moved it to a different module in the previous PR

thunder/core/module.py
```diff
@@ -19,16 +19,22 @@ def __init__(self, model, compiled_model_call):
         super().__init__()
         self._model = model
 
-        # We delete self.training in order for training to be used from
-        # the model itself through `__getattr__`.
```
carmocca, line 23: Accidental removal?

thunder/distributed/__init__.py
```diff
@@ -349,6 +350,162 @@ def f(tensor: TensorProxy) -> str:
...
+        submodule = thunder_model.get_submodule(module_name)
+        module_copy = copy.copy(submodule)
+
+        # Materialize meta-parameters on-device if necessary.
+        # This is done before sharding in case the materialization logic depends on the tensor shape.
+        # The tradeoff is that all of a module's direct parameters need to fit in device.
+        # Each module only initializes its own parameters and not those of its children (recurse=False)
+        if any(t.is_meta for t in chain(module_copy.parameters(recurse=False), module_copy.buffers(recurse=False))):
+            # TODO: we could also support calling a "param_init_fn" argument like PyTorch
+            # thunder.distributed._materialize(submodule, device)
```
carmocca, line 470: Why did you inline the materialize code here?

thunder/distributed/__init__.py
```diff
@@ -349,6 +350,162 @@ def f(tensor: TensorProxy) -> str:
...
+            # Cannot `submodule.to(device)` because we don't want it to recurse
+            for n, p in module_copy.named_parameters(recurse=False, prefix=module_name):
+                if p.device != device:
+                    thunder_model._overrides[n] = torch.nn.Parameter(p.to(device=device), requires_grad=p.requires_grad)
+                    device_adjustments[n] = device
+            for n, b in module_copy.named_buffers(recurse=False, prefix=module_name):
+                if p.device != device:
+                    thunder_model._overrides[n] = b.to(device=device)
+                    device_adjustments[n] = device
+
```
carmocca, line 496: I think you are missing the broadcast here

thunder/distributed/__init__.py
```diff
@@ -349,6 +350,162 @@ def f(tensor: TensorProxy) -> str:
...
+    thunder_model = thunder.core.transforms.add_transform(thunder_model, early_transform=prologue_and_compute_transform)
+
+    # modidfy module
+    import copy
+
+    sharded_params = {}
+    device_adjustments = {}
+    for module_name, _ in thunder_model._model.named_modules():
+        submodule = thunder_model.get_submodule(module_name)
+        module_copy = copy.copy(submodule)
```
carmocca, line 462: Why does the new code use a copy?

thunder/distributed/__init__.py
```diff
@@ -349,6 +350,162 @@ def f(tensor: TensorProxy) -> str:
...
+                    device_adjustments[n] = device
+
+        for pn, p in submodule.named_parameters(recurse=False, prefix=module_name):
+            if pn not in thunder_model._overrides:
+                thunder_model._overrides[pn] = copy.copy(p)
+            # we collect shapes and devices because we do not know if other transforms also change it...
+            old_shape = thunder_model._overrides[pn].shape
+            thunder.distributed._shard_param(thunder_model._overrides[pn], global_rank, world_size, pn)
+            new_shape = thunder_model._overrides[pn].shape
+            sharded_params[pn] = (old_shape, new_shape, thunder_model._overrides[pn].device)
```
carmocca, line 504: Where is this information used?

thunder/distributed/__init__.py
t-vi: We want to leave the original module untouched.

thunder/distributed/__init__.py
t-vi: Originally, I had hoped to get by without the module copy, but you are right, I'm should be doing this as a call.

thunder/distributed/__init__.py
t-vi: This is used in the `prologue_and_compute_transform` callback. The thing is that we need to transform the `ThunderModule` when the fsdp(...) is applied but the `prologue_and_compute_transform` needs to run when the ThunderModule is actually called (and we have a cache miss).

thunder/distributed/__init__.py
t-vi: added a todo

thunder/core/transforms.py
```diff
@@ -397,27 +397,48 @@ def visitor_transform(trace_from: Trace, visit: Callable, *, provenance: None |
...
 
 # Helper function to add a transform
-def add_transform(cfn: Callable, transform: Callable) -> Callable:
+def add_transform(
+    cfn: Callable,
+    *,
+    transform: Callable | None = None,
+    early_transform: Callable | None = None,
+    disable_torch_autograd_support=False,
+) -> Callable:
```
crcrpar, line 406: could you give this method a docstring?

thunder/__init__.py
crcrpar: I meant to ask whether or not we have a specification of a transform (and optionally a check)

thunder/core/module.py
```diff
@@ -25,10 +25,21 @@ def __init__(self, model, compiled_model_call):
 
         self._forward_fn = compiled_model_call
 
+        # overrides for parameters and buffers (see get_buffer/get_parameter)
+        self._overrides = {}
+
+        self._null = object()
```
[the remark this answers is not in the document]
t-vi: because you can legitimately override a buffer / parameter with `None`.

thunder/distributed/__init__.py
```diff
@@ -349,6 +355,176 @@ def f(tensor: TensorProxy) -> str:
...
+                    device_adjustments[n] = device
+
+        # Broadcast parameters if requested
+        if broadcast_from is not None:
+            for pn, _ in submodule.named_parameters(recurse=False, prefix=module_name):
+                tdist.broadcast(
+                    thunder_model.get_parameter(pn), src=broadcast_from, group=process_group, async_op=False
+                )
+            for pn, _ in submodule.named_buffers(recurse=False, prefix=module_name):
+                tdist.broadcast(thunder_model.get_buffer(pn), src=broadcast_from, group=process_group, async_op=False)
```
carmocca, line 514: Cool, you added buffers

thunder/distributed/__init__.py
carmocca: I'm not convinced by this because the current implementation doesn't modify the original module, only the parameters. And those need to be modified by reference because to avoid blowing up the memory of course.

What was your motivation for this originally? Do you expect this from transforms in general? For example, if you had a LoRA transform that replaces Linears, wouldn't you expect it to swap the classes?

carmocca (approved):
LGTM. I expect future PRs to surface bugs (if any 😉).

thunder/distributed/__init__.py
```diff
@@ -349,6 +355,176 @@ def f(tensor: TensorProxy) -> str:
...
+    world_size = tdist.get_world_size(group=process_group)
+    if device is None:
+        local_rank = int(os.environ["LOCAL_RANK"])
+        device = torch.device("cuda", local_rank)
+
+    def prologue_and_compute_transform(prologue_trace, computation_trace, epilogue_trace, **kwargs):
+        import thunder
+
+        prologue_producers, prologue_consumers = thunder.core.utils.producers_and_consumers(prologue_trace)
+        computation_producers, computation_consumers = thunder.core.utils.producers_and_consumers(computation_trace)
```
crcrpar, line 396: nits
```suggestion
        prologue_producers, prologue_consumers = utils.producers_and_consumers(prologue_trace)
        computation_producers, computation_consumers = utils.producers_and_consumers(computation_trace)
```

thunder/distributed/__init__.py
```diff
@@ -349,6 +355,176 @@ def f(tensor: TensorProxy) -> str:
...
+            bsym = prologue_producers[pro_out_p]
+            if bsym.sym == thunder.prims.unpack_parameter:
+                param_thunder_module, param_name = bsym.args
+                assert param_thunder_module is thunder_module_proxy
+                if param_name in sharded_params:
+                    old_shape, new_shape, new_torch_device = sharded_params[param_name]
+                    thunder_device = thunder.core.devices.to_device(new_torch_device)
+                    thunder_device_str = str(thunder_device)
+
+                    pro_out_p._ddp_type = thunder.core.proxies.DDPType.FULLY_SHARDED
```
crcrpar, line 425: nits
```suggestion
                    pro_out_p._ddp_type = DDPType.FULLY_SHARDED
```

thunder/distributed/__init__.py
```diff
@@ -349,6 +355,176 @@ def f(tensor: TensorProxy) -> str:
...
+                    pro_out_p._shape = tuple(new_shape)
+                    pro_out_p._device = thunder_device
+                    if comp_inp_p is not pro_out_p:
+                        comp_inp_p._ddp_type = thunder.core.proxies.DDPType.FULLY_SHARDED
```
crcrpar, line 429: nits
```suggestion
                        comp_inp_p._ddp_type = DDPType.FULLY_SHARDED
```

thunder/distributed/__init__.py
```diff
@@ -349,6 +355,176 @@ def f(tensor: TensorProxy) -> str:
...
+# When the user calls fsdp(jitted_module), this function does the following
+# - It transforms the ThunderModule jitted_module, materializing and sharding the parameters as `overrides`
+#   in the ThunderModule.
+# - While doing that, it leaves the original user module alone.
+# - It then registers an early transform (callback that runs before prologue is executed) that transforms the
+#   prologue and compute trace.
+#
+# Note that for doing so, there are a few constraints / caveats:
+# - We do not have prologues/compute traces when we transform the module.
+# - We need to record the info from the module transformations because a later transform might modify the module further.
```
crcrpar, line 367: could you update the docstring of `fsdp` below?
I'd like to some explanation e.g. when the actual sharding happens

thunder/distributed/__init__.py
```diff
@@ -349,6 +355,176 @@ def f(tensor: TensorProxy) -> str:
...
+    process_group = tdist.distributed_c10d._get_default_group()
+    utils.check(process_group is not None, lambda: "The default process group is None")
+    global_rank = tdist.get_rank(group=process_group)
+    world_size = tdist.get_world_size(group=process_group)
+    if device is None:
+        local_rank = int(os.environ["LOCAL_RANK"])
+        device = torch.device("cuda", local_rank)
+
+    def prologue_and_compute_transform(prologue_trace, computation_trace, epilogue_trace, **kwargs):
+        import thunder
```
crcrpar, line 393: Q: would this be necessary even L378 does this?

thunder/core/transforms.py
```diff
@@ -397,27 +397,48 @@ def visitor_transform(trace_from: Trace, visit: Callable, *, provenance: None |
...
             sharp_edges=cd.sharp_edges,
             # cache, interpretation?
-            additional_transforms=cfn._lc_transforms + [transform],
-            disable_torch_autograd=True,  # cd.disable_torch_autograd_support,
+            early_transforms=early_transforms,
+            additional_transforms=additional_transforms,
+            disable_torch_autograd=cd.disable_torch_autograd_support or disable_torch_autograd_support,
             **cd.compile_options,
         )
+        from thunder import ThunderModule
```
crcrpar, line 437: shouldn't this be ?
```suggestion
        from thunder.core.module import ThunderModule
```

thunder/distributed/__init__.py
```diff
@@ -349,6 +355,176 @@ def f(tensor: TensorProxy) -> str:
...
+# - It then registers an early transform (callback that runs before prologue is executed) that transforms the
+#   prologue and compute trace.
+#
+# Note that for doing so, there are a few constraints / caveats:
+# - We do not have prologues/compute traces when we transform the module.
+# - We need to record the info from the module transformations because a later transform might modify the module further.
+
+
+def fsdp_transform_module(
+    thunder_model: thunder.ThunderModule,
```
crcrpar, line 371:
```suggestion
    thunder_model: ThunderModule,
```

thunder/distributed/__init__.py
```diff
@@ -349,6 +355,176 @@ def f(tensor: TensorProxy) -> str:
...
+    *,
+    device: torch.device | None = None,
+    broadcast_from: int | None = None,
+    sharding_strategy: FSDPType = FSDPType.ZERO2,
+    bucketing_strategy: FSDPBucketingStrategy = FSDPBucketingStrategy.NONE,
+) -> thunder.ThunderModule:
```
crcrpar, line 377:
```suggestion
) -> ThunderModule:
```

thunder/distributed/__init__.py
```diff
@@ -388,12 +564,23 @@ def fsdp(
...
     """
+    import thunder
+
     utils.check(isinstance(sharding_strategy, FSDPType), lambda: f"FSDPType.ZERO2 and FSDPType.ZERO3 are supported.")
     utils.check(
         tdist.is_available(),
         lambda: "fsdp requires torch distributed to be available (but it's not)",
     )
 
+    if isinstance(model, thunder.ThunderModule):
```
crcrpar, line 575:
```suggestion
    if isinstance(model, ThunderModule):
```

thunder/distributed/__init__.py
```diff
@@ -16,6 +20,8 @@
 
 if TYPE_CHECKING:
     from torch.distributed import ProcessGroup
+    import thunder
```
crcrpar, line 23: it feels lik eL20 should be
```
from thunder.core.module import ThunderModule
```
and this should be removed since I don't think `ThunderModule` is a class residing thunder namespace as per https://github.com/Lightning-AI/lightning-thunder/blob/cb489c9532e3fdf37d684a92f13f843130a7d0a5/thunder/__init__.py#L14

thunder/distributed/__init__.py
t-vi: :)

t-vi:
@carmocca  makes the following comments on replacing parameters only in the overrides:

> I'm not convinced by this because the current implementation doesn't modify the original module, only the parameters. And those need to be modified by reference because to avoid blowing up the memory of course.

> What was your motivation for this originally? Do you expect this from transforms in general? For example, if you had a LoRA transform that replaces Linears, wouldn't you expect it to swap the classes?

Which we should look at more closely soon.

PR #364, opened 2024-05-03
Update "phantom grad" to use the same forward-backward transformation code as "vjp"

IvanYashchuk:
This PR swaps `thunder.transforms.grad` with `thunder.transforms.grad_v1`. The difference between the two is that grad_v1 uses the same code as the main code path with PyTorch Autograd integration, they differ only in the implementation but do the same thing.

After this PR, I'm planning to remove the `grad` function code that is now unused and renamed in this PR to `__grad`.

The grad transform was added in https://github.com/Lightning-AI/lightning-thunder/commit/6069e6f20b8bd42fae8a3e2ebde692adfebc6faf and it included a new way of defining forward-backward rules in one same Python function and a new way of registering a transformation to the jitted function.
The new way of defining forward-backward rules is now also supported in the main code path; we no longer need that prototype.

Test plan:
* All existing "phantom grad" tests should still pass: `pytest thunder/tests/test_grad.py -k "phantom_grad"`.

### DIFF
--- a/thunder/core/transforms.py
+++ b/thunder/core/transforms.py
@@ -1337,7 +1337,7 @@ def _grad_out_specifier_default(pytree: Any) -> list[TensorProxy]:
 # The algorithm for modifying the program has the following steps:
 #   1) Flattens the original trace for the grad transform -- ensuing that all top-level symbols have
 #       a grad function.
-def grad(
+def __grad(
     cfn, grad_specifier: Callable = _grad_specifier_default, grad_out_specifier: Callable = _grad_out_specifier_default
 ) -> Callable:
     # Creates a custom transform callable that binds the additional arguments to the grad transform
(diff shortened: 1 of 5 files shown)


thunder/core/transforms.py
```diff
@@ -1337,7 +1337,7 @@ def _grad_out_specifier_default(pytree: Any) -> list[TensorProxy]:
 # The algorithm for modifying the program has the following steps:
 #   1) Flattens the original trace for the grad transform -- ensuing that all top-level symbols have
 #       a grad function.
-def grad(
+def __grad(
```
IvanYashchuk, line 1340: "register_grad" rules are still being used but now though the same code path as the rest of the thunder.jit with disable_torch_autograd=False.

IvanYashchuk:
All checks have passed.

@mruberry, could you please help merge this PR if the changes make sense?

mruberry (approved):

Issue #366, opened 2024-05-06
transformer_engine: clear_collection doesn't clear the tensors saved by transformer_engine_ex ops for backward.


kshitij12345:
`clear_collection` only clears a list, but `transformer_engine_ex` saves a collection of dict which holds a tuple of tensor. Because of this we hold onto to the tensors longer leading to higher memory usage.

https://github.com/Lightning-AI/lightning-thunder/blob/40e9c3ec7530cfc109a86301cb58ea62309da28b/thunder/executors/pythonex.py#L228-L230

cc: @IvanYashchuk

kshitij12345:
Per offline discussion:
* We should update `clear_collection` to also clear dictionary.
* We should update the name `clear_collection` to `clear_mutable_collection` to make it clear that this won't do anything if `tuple` or other immutable collection is passed.

PR #367, opened 2024-05-06
clear_collection: also clear other mutable collections besides list

kshitij12345:
Fixes: #366 

Tested with `thunder/tests/test_transformer_engine.py` and `thunder/tests/distributed/test_ddp.py -k transformer`.

Benchmark with `test_benchmark_litgpt` for `open_llama_3b` - FSDP Zero3 - Microbatch Size 1

thunder : Memory used: 25.07 GB
(Before PR) thunder + transformerengine : Memory used: 44.41 GB
(After PR) thunder + transformerengine : Memory used: 33.97 GB

### DIFF
--- a/thunder/executors/pythonex.py
+++ b/thunder/executors/pythonex.py
@@ -226,7 +226,7 @@ def _clear_collection_meta(coll: CollectionProxy) -> None:
 
 
 def _clear_collection_prim_impl(a: Sequence) -> None:
-    if isinstance(a, list):
+    if isinstance(a, (list, dict)):
         a.clear()
 
 
--- a/thunder/executors/pythonex.py
+++ b/thunder/executors/pythonex.py
@@ -1,7 +1,7 @@
 from typing import List, Any, Dict, Tuple, Union, Type
 from collections.abc import Callable
 from collections.abc import Hashable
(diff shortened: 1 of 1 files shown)


thunder/executors/pythonex.py
```diff
@@ -226,7 +226,7 @@ def _clear_collection_meta(coll: CollectionProxy) -> None:
 
 
 def _clear_collection_prim_impl(a: Sequence) -> None:
-    if isinstance(a, list):
+    if isinstance(a, (list, dict)):
```
IvanYashchuk, line 229: There are [MutableSequence](https://docs.python.org/3/library/collections.abc.html#collections.abc.MutableSequence), [MutableMapping](https://docs.python.org/3/library/collections.abc.html#collections.abc.MutableMapping), and [MutableSet](https://docs.python.org/3/library/collections.abc.html#collections.abc.MutableSet) types that we can probably use here for the isinstance check.

IvanYashchuk (approved):

carmocca (approved):
LGTM!

IvanYashchuk:
> docs-make (html) Expected — Waiting for status to be reported

Is it stuck?

Issue #369, opened 2024-05-06
Benchmark: Running benchmark_litgpt with `transformerengine` leads to `nullcontext` not defined.


kshitij12345:
Run: 

```
torchrun --nproc_per_node=2 --nnodes=1 thunder/benchmarks/benchmark_litgpt.py --return_metrics_as_json=True --json_path=/tmp/benchmark_litgpt_data.json --distributed_mode=fsdp --shard_mode=zero3 --model_name=pythia-14m --micro_batch_size=1 --compile=thunder_inductor_transformerengine --nsys_enabled=False --dynamic=False
```
Error
```
An error occurred: UnboundLocalError – local variable 'nullcontext' referenced before assignment
```

PR #370, opened 2024-05-06
fix benchmark script with TE and also remove te_ctx as we automatically wrap the trace with te_ctx.

kshitij12345:
Fixes: #369 

For case where, `transformerengine` is in `self.compile` and `self.skip_data_sync=False`, `nullcontext` is undefined.

https://github.com/Lightning-AI/lightning-thunder/blob/40e9c3ec7530cfc109a86301cb58ea62309da28b/thunder/benchmarks/benchmark_litgpt.py#L354-L367

Also, we remove `te_ctx` as it is not required since https://github.com/Lightning-AI/lightning-thunder/pull/48


carmocca (approved):
Thanks

PR #372, opened 2024-05-06
increase timeout for macos

t-vi:


carmocca:
In https://github.com/Lightning-AI/lightning-thunder/actions/runs/8974221177/job/24646125828, the issue was this test (notice the 9 min difference)
image
which is curious because the test is not that complex: https://github.com/Lightning-AI/lightning-thunder/blob/main/thunder/tests/test_patterns.py#L59-L101

If you have other failed builds I can take a look. This might be caused by test interactions. Worst case I can SSH into the CI runner to try to repro there, but it will be useful to find recurring patterns first

t-vi:
seems to have solved itself.

t-vi:
I'll take a look though for more occurences.

PR #379, opened 2024-05-08
transformer_engine: Control backward FP8 metadata synchronization (v1.6 onwards)

kshitij12345:
Fixes: https://github.com/Lightning-AI/lightning-thunder/issues/368

Failure in above issue happens because we don't always correctly reorder the inputs to the `te_linear` in distributed setting.

This PR takes a different approach and instead inserts a sync operation in the backward trace which makes us agnostic to the previous constraint.

Details of the approach can be found at : `NOTE: Backward FP8 metadata sync` in `transformer_engineex.py`.

NOTE: Once v1.6 is released, I am planning to drop the support for older versions and clean-up the code.


thunder/executors/transformer_engineex.py
```diff
@@ -458,6 +468,36 @@ def _get_te_wrapper_string():
     return TE_CTX_STR
 
 
+def te_sync_fp8_meta_bwd_meta():
+    pass
+
+
+def te_sync_fp8_meta_bwd_impl():
+    FP8GlobalStateManager.reduce_and_update_fp8_tensors(forward=False)
```
IvanYashchuk, line 476: Is this available in TE version < 1.6?

IvanYashchuk (approved):

thunder/executors/transformer_engineex.py
kshitij12345: Previously it was https://github.com/NVIDIA/TransformerEngine/blob/6a9edc38bf9b941b7d369af5103fa8fe0b121d61/transformer_engine/pytorch/fp8.py#L343

As per the talks with TE team, this is the new function and should stay stable most likely.

thunder/tests/distributed/test_ddp.py
```diff
@@ -1501,21 +1507,34 @@ def _test_ddp_transformer_engine_llama_sanity(input_data):
...
+            # Verify that the symbol to sync backward
+            # fp8 metadata is present in backward trace.
+            found_bwd_sync_symbol = False
+            for bsym in reversed(bwd_exec_trace.bound_symbols):
+                if bsym.sym.id == te_sync_fp8_meta_bwd.id:
+                    found_bwd_sync_symbol = True
+                    break
+
+            if not found_bwd_sync_symbol:
+                raise RuntimeError("Backward sync symbol not found.")
```
carmocca, line 1520: nit: neat python magic https://book.pythontips.com/en/latest/for_-_else.html

```suggestion
            for bsym in reversed(bwd_exec_trace.bound_symbols):
                if bsym.sym.id == te_sync_fp8_meta_bwd.id:
                    break
            else:
                raise RuntimeError("Backward sync symbol not found.")
```

carmocca (approved):
LGTM

thunder/tests/distributed/test_ddp.py
kshitij12345: Nice, thank you!

Issue #205, opened 2024-04-16
benchmarking — create a notebook showing how to work with the single gpu benchmarks


mruberry:
The notebook should

- show how to understand what the benchmark's options are
- show how to programmatically run the benchmark
- show how to create the callable that the benchmark is invoking
- show how to acquire the input generator the benchmark is using to create the inputs to the callable

This way practitioners will be able to run the benchmark callable in their own contexts, which should support profiling.

fyi @kevinstephano — is this what you were thinking of when talking about extracting a callable from a benchmark?

cc @borda @apaz-cli @crcrpar

kevinstephano:
This sounds good!

This might already be in mind with these bullet points but it would be helpful to show explicitly how to:
* Show the trace
* Execute specific fusions
* Dump the debug for specific fusions

riccardofelluga:
Regarding the debug for a specific fusion I've created #387. If that goes through I will show in this notebook how to use that information to dump it for specific fusions so that then it's just a matter of reading the output and creating the repro script

PR #388, opened 2024-05-08
Add  input information to fusion definitions for trace inspection and debugging

### LINKED ISSUE #387: Add more nvFuser debug information
## 🚀 Feature

For debugging purposed, I would like to be able to quickly retrieve input information for a fusion.

### Motivation

When debugging performance issues an important aspect is to be able to narrow the search space quickly and efficiently. 

### Pitch

At the time of writing, debugging nvFuser fusions is very approachable already as it is possible to retrieve fusion definitions from the trace. It is however not trivial to retrieve the input information. Therefore, to improve the experience even further, I propose to add a mechanism to retrieve input information for a fusion definition. 

### Alternatives

- Trigger an error in nvFuser and let it print the debugging information.
- Meticulously read a trace and write down input shapes.

### Additional context

This idea came out after starting to work on #205 where an idea by @kevinstephano is to show how to dump the debug for specific fusions and execute them in a notebook. This issue works toward reaching that goal.

riccardofelluga:
<details>

- [x ] Did you read the [contributor guideline](https://github.com/Lightning-AI/pytorch-lightning/blob/main/.github/CONTRIBUTING.md), Pull Request section?

</details>

## What does this PR do?

Fixes #387.
This PR adds information about the inputs for a fusion definition such that it can be retrieved by inspecting the trace. A tutorial on how to read this information will be published as part of #205.

Also this PR is in preparation for #205

Quickly, from a trace:
``` python
trace = thunder.last_traces(fn)[-1]
trace_ctx = trace.python_ctx()
print(trace_ctx['nvFusion0'].last_inputs())
```

will print something like:
``` shell
inputs = [
    torch.randn((2048,), dtype=torch.float32, device='cuda:0').as_strided((1, 2048), (2048, 1)),
    torch.randn((4096,), dtype=torch.bfloat16, device='cuda:0').as_strided((1, 2048, 4096), (4096, 0, 1))
]
```

I'm open to change the `str` output to returning a list of tensors, however for debugging it's usually enough to have a string.

### DIFF
--- a/thunder/executors/nvfuserex_impl.py
+++ b/thunder/executors/nvfuserex_impl.py
@@ -441,7 +441,6 @@ class FusionDefinitionWrapper:
         return msg
 
 
-
 # Group bookend meta operations into separate regions
 # This function returns a List[Region] which changes the executor of meta regions to torchex
 #
--- a/thunder/executors/nvfuserex_impl.py
+++ b/thunder/executors/nvfuserex_impl.py
@@ -396,10 +396,12 @@ class FusionDefinitionWrapper:
     cache_info: None | Callable = None
     cache_clear: None | Callable = None
     last_used: None | FusionDefinition = None
+    last_inputs_meta: None | Sequence[tuple] = None
 
     def __call__(self, *args):
         fd = self.get_fd(to_descriptors(args))
         self.last_used = fd
+        self.last_inputs_meta = [(i.size(), i.stride(), i.dtype, i.device) for i in args if isinstance(i, torch.Tensor)]
 
         # Set device if set in one of the "factory" methods like full, iota, or uniform
         kwargs = (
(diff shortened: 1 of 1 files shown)


IvanYashchuk (commented):
That's a really interesting idea and it would be a very useful addition to query the last used inputs for a specific fusion region. Since this an nvFuser executor specific change I don't think Thunder is the right place to put these queries.

How would you generalize it to all other execution functions and regions?

Thunder allows getting the `nvfuser.FusionDefinition` object with `execution_trace.python_ctx()["nvFusion0"].last_used` and maybe this particular code should be a method of `nvfuser.FusionDefinition`?

thunder/executors/nvfuserex_impl.py
```diff
@@ -413,6 +415,31 @@ def __call__(self, *args):
     def __repr__(self):
         return f"FusionDefinitionWrapper({self.name})"
 
+    def last_inputs(self) -> str:
```
mruberry, line 418: + @kevinstephano to consider if nvfuser itself would like to expose something like this. I think we discussed this briefly before, but maybe we could query nvFuser's Python API for something like "last reproduction" and that would include this information? Maybe the reproduction could be structured so it could be a Python string + contain structured information about the inputs?

For `last_inputs` here, I think we should consider just saving non-tensor inputs and only saving the metadata of tensor inputs. Non-tensor inputs can only be numbers or sequences of numbers, right? It'd be nice if last_inputs would return a valid set of inputs to the fusion, even creating the requested tensors based on the metadata. We could also rename this function to something like `print_last_inputs`.

I'm OK with this logic being in Thunder for now if it's helpful, even if we decide to expose it through nvfuser. When the nvfuser exposure is available we can remove this. Let's just add a warning when it's used that it's experimental and may change in the future.

How does that sound, @riccardofelluga? @IvanYashchuk? @kevinstephano?

IvanYashchuk: Yes, I agree that it should be possible to query from nvFuser's Python API the metadata of the last executed inputs.

I like the `print_last_inputs` name more, the current name suggests like real last inputs are captured, which is not the case. Currently, only tensor metadata is printed here, but nvFuser fusions also accept numbers.

lantiga:
Hey, what's the next step?

riccardofelluga:
Wait, so there has been some other discussion in the issue linked to this. I close this PR and I will open a new one with the results

Issue #409, opened 2024-05-13
torch.Tensor.size is broken on main


kshitij12345:
```python
import thunder
import torch

def foo(x):
    return x.size(0)

x = torch.randn(3, 3)

jfoo = thunder.jit(foo)

print(jfoo(x))  # (3, 3)
print(foo(x))  # 3

print(thunder.last_traces(jfoo)[-1])
# # Constructed by Delete Last Used (took 0 milliseconds)
# import torch
# from thunder.executors.torchex import no_autocast

# @torch.no_grad()
# @no_autocast
# def computation():
#   return (3, 3)
```

It is failing since: 463487c8b8063969b812c49b8e2c1625a39f281d

kshitij12345:
This affects transformer_engine_ex which uses `size` in its checker function.

PR #411, opened 2024-05-13
fix size for dim-0

kshitij12345:
Fixes: https://github.com/Lightning-AI/lightning-thunder/issues/409

`bool(0)` is False and we end up taking incorrect branch.

TODO: 
* [x] Add test

### DIFF
--- a/thunder/tests/opinfos.py
+++ b/thunder/tests/opinfos.py
@@ -583,6 +583,38 @@ numel_opinfo = OpInfo(
 )
 tensor_properties.append(numel_opinfo)
 
+
+def size_sample_generator(op, device, dtype, requires_grad, **kwargs):
+    make_t = partial(make_tensor, device=device, dtype=dtype, requires_grad=requires_grad)
+
+    shapes = (
+        (),
+        (1,),
+        (
+            2,
+            2,
+        ),
+        (0, 2, 1),
+        (2, 0, 1),
+    )
+
+    for shape in shapes:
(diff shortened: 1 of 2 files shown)


thunder/torch/__init__.py
```diff
@@ -173,7 +173,7 @@ def is_floating_point(a: TensorLike, /) -> bool:
 
 # Handles the size method
 def size(a: TensorLike, /, dim: None | int = None) -> int | Sequence[int]:
-    if dim:
+    if dim is not None:
```
mruberry, line 176: Nice explicit test against None

thunder/tests/opinfos.py
```diff
@@ -7970,3 +7970,34 @@ def item_sample_generator(op, device, dtype, requires_grad, **kwargs):
...
+
+    for shape in shapes:
+        t = make_t(shape)
+        yield SampleInput(t)
+
+        for d in range(len(shape)):
+            yield SampleInput(t, d)
+
+
+size_opinfo = OpInfo(
```
mruberry, line 7997: Maybe we can put this here:

https://github.com/kshitij12345/lightning-thunder/blob/d30195d85ca6782a58553f0fe8677f930f1a9fa5/thunder/tests/opinfos.py#L565

as a tensor property?

kshitij12345: Right, I missed that. Thanks!

mruberry (approved):

PR #412, opened 2024-05-13
Avoid deprecated autocast APIs

carmocca:
## What does this PR do?

Avoids deprecation warnings from https://github.com/pytorch/pytorch/pull/124359


carmocca:
Actually it's too soon to do this. This deprecation is only in the nightlies. Closing till another day

PR #413, opened 2024-05-13
Fix nsys profiling in benchmark_litgpt

parthmannan:
<details>
</details>

## What does this PR do?

Fixes nsys profiling bug in benchmark_litgpt.py introduced in #45 
When `gradient_accumulation_steps==1`, the nsys profiler never starts the capture as that is called only if we are use higher GA steps.

PR #417, opened 2024-05-14
[Benchmak][LitGPT] Relocate profiler_start

Aidyn-A:
I had an experience when `cudaProfilerStart()` was never called, ending up with no profile generated. I propose to call `cudaProfilerStart()` a bit earlier.

cc @crcrpar


carmocca:
Thanks for the fix! However, there is #413 open already

Aidyn-A:
Guess I am not the only one, who faced the same issue. Closing in favor of #413.

Issue #396, opened 2024-05-10
Regression on thunder/benchmarks/targets.py


IvanYashchuk:
## 🐛 Bug

Changes added in https://github.com/Lightning-AI/lightning-thunder/commit/b1f447022b0732e83c11661c30746568280834f7 (https://github.com/Lightning-AI/lightning-thunder/pull/279) introduced a regression on the benchmarks, for example,
```
pytest thunder/benchmarks/targets.py -vs -k 'test_nanogpt_gpt2_fwd[thunder]'
```
has a median time of about 4.3ms before the commit and 6.15ms with b1f447022b0732e83c11661c30746568280834f7.

Many other benchmark cases from benchmarks/targets.py are affected by the change.

`git revert b1f447022b0732e83c11661c30746568280834f7` on the current main restores the performance.

@t-vi, do you know what could be causing the slowdown?

cc @apaz-cli @crcrpar

t-vi:
For my own reference, on my reference machine:
```
b1f447022b0732e83c11661c30746568280834f7^
test_nanogpt_gpt2_fwd[thunder]     13.0015  13.6725  13.1576  0.2891  13.0281  0.2124       1;1  76.0015       5           1

b1f447022b0732e83c11661c30746568280834f7
test_nanogpt_gpt2_fwd[thunder]     14.7374  15.5216  14.9018  0.3466  14.7441  0.2112       1;1  67.1059       5           1

4663b87e67955edec371c6f37a763a7ec358d835 (current main)
test_nanogpt_gpt2_fwd[thunder]     14.7783  15.2227  14.8889  0.1873  14.8131  0.1166       1;1  67.1642       5           1
```

t-vi:
Has been reverted, so I'm reopening.

PR #421, opened 2024-05-15
cache params/buffers/submodules in ThunderModule

t-vi:
Switching to `.get_parameter` and friends (starting from the root module) is too expensive, so we cache all parameters etc.

Fixes: #396 

@IvanYashchuk

### DIFF
--- a/thunder/core/module.py
+++ b/thunder/core/module.py
@@ -1,4 +1,5 @@
 from contextlib import contextmanager
+import itertools
 from typing import Any
 
 import torch as pytorch
@@ -26,7 +27,13 @@ class ThunderModule(pytorch.nn.Module):
         self._forward_fn = compiled_model_call
 
         # overrides for parameters and buffers (see get_buffer/get_parameter)
-        self._overrides = {}
+        # we populate these here for performance reasons (sam as module cache),
+        # a single dict lookup is cheaper than traversin the module
+        # hierarchy, see https://github.com/Lightning-AI/lightning-thunder/issues/396#issuecomment-2113231498
+        self._overrides = {
+            k: v for k, v in itertools.chain(self._model.named_parameters(), self._model.named_buffers())
+        }
+        self._module_cache = {k: v for k, v in self._model.named_modules()}
 
         self._null = object()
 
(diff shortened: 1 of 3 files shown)


t-vi:
The errors are real and I'll fix them.

thunder/core/module.py
```diff
@@ -26,7 +27,13 @@ def __init__(self, model, compiled_model_call):
         self._forward_fn = compiled_model_call
 
         # overrides for parameters and buffers (see get_buffer/get_parameter)
-        self._overrides = {}
+        # we populate these here for performance reasons (sam as module cache),
+        # a single dict lookup is cheaper than traversin the module
+        # hierarchy, see https://github.com/Lightning-AI/lightning-thunder/issues/396#issuecomment-2113231498
+        self._overrides = {
+            k: v for k, v in itertools.chain(self._model.named_parameters(), self._model.named_buffers())
+        }
```
IvanYashchuk, line 35: Have you considered using `nn.Module.state_dict()`? The only difference I see from the current implementation is that here both persistent and non-persistent buffers are included, while state_dict includes only persistent buffers.
Why is LitGPT using non-persistent buffers for `cos` and `sin`? https://github.com/Lightning-AI/litgpt/blob/5dabf5f6be507e4da4d7be69813e6f31e799aefd/litgpt/model.py#L52-L53
```suggestion
        self._overrides = self._model.state_dict() | {k: v for k, v in self._model.named_buffers()}
```

IvanYashchuk (commented):
I thought `.get_parameter` is used only for FSDP changes from https://github.com/Lightning-AI/lightning-thunder/pull/359. How does it affect single process runs?

thunder/core/module.py
t-vi: state_dict can actually contain all sorts of things (versions, ...), so I think tailoring to parameters and buffers seems the right thing.
I think we don't want to have cos and sin in the state dict (why should we...)

IvanYashchuk (approved):
I have verified that the perf is restored. Thanks!

t-vi:
Thank you @IvanYashchuk

nikitaved:
This seems to break CIs on `pytest -sv thunder/tests/distributed/test_ddp.py -k test_fsdp_grad_parity_with_without_bucketing_executor_torch_bucketing_block_zero3_fsdp_jit`

PR #424, opened 2024-05-16
[fsdp] Propagate `bucketing_strategy` and `sharding_strategy` via original module

crcrpar:
It turned out that the lack of `bucketing_strategy` and `sharding_strategy` of the original `nn.Module` kept fsdp comm bucketing away from `fsdp(jit(model))` path.

### DIFF
--- a/thunder/distributed/__init__.py
+++ b/thunder/distributed/__init__.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 import os
 
 from itertools import chain
-import collections
 from contextlib import contextmanager
 from contextvars import ContextVar, Token
 import copy
@@ -20,7 +19,7 @@ from thunder.core.proxies import DDPType
 
 if TYPE_CHECKING:
     from torch.distributed import ProcessGroup
-    import thunder
+    from thunder.core.module import ThunderModule
 
 
 __all__ = [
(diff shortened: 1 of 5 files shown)


thunder/distributed/transforms/fsdp_v2.py
crcrpar, line 1: welcome any suggestions as I don't quite like the name of `fsdp_v2`

t-vi: `fsdp_on_traces` ?

thunder/distributed/__init__.py
```diff
@@ -389,85 +387,18 @@ def fsdp_transform_module(
...
-    thunder_model = thunder.core.transforms.add_transform(thunder_model, early_transform=prologue_and_compute_transform)
+    cd = get_compile_data(thunder_model)
+    # TODO: promote use_fsdp and use_ddp to public members of CompileData
+    cd.use_fsdp = True
+    orig_module: torch.nn.Module = cd.fn
+    utils.check(
+        isinstance(orig_module, torch.nn.Module) and not isinstance(orig_module, ThunderModule),
+        lambda: f"CompileData.fn expected to be `nn.Module` but {type(orig_module)}",
+    )
+    orig_module.use_fsdp = True
```
t-vi, line 398: Why can we not keep / take this on the ThunderModule, though?

crcrpar: because bucketing was implemented before fsdp as a trace transform and it is applied to a trace passed to `torch_autograd.split_forward_backward` thus it was straightforward to query the values to the original module through CompileData.

t-vi (approved):
Thank you @crcrpar !

In the medium term, I would like to see if we can attach the information in the compile data, but let's not block on that. If you want to rename v2 or change anything else, please do, or you can just hit the merge.

thunder/distributed/__init__.py
t-vi: OK, but so it's not completely unthinkable to get it from `compile_data` instead of `compile_data.fn`?

crcrpar: right

thunder/distributed/transforms/fsdp_v2.py
crcrpar: i'm afraid trace doesn't sound much clearer than `v2` :sweat_smile:

PR #429, opened 2024-05-17
Implement clone()

### LINKED ISSUE #342: Implement TensorBase.clone
### 🚀 Feature

Implement [`TensorBase.clone`](https://pytorch.org/docs/stable/generated/torch.clone.html)

### Motivation

NeMo's [MegatronNeVaModel](https://github.com/NVIDIA/NeMo/blob/main/examples/multimodal/multimodal_llm/neva/neva_pretrain.py)

cc @tfogal

tfogal:
Fixes #342.

## What does this PR do?

Fixes #342 

## PR review

## Did you have fun?

Always ;-)

### DIFF
--- a/thunder/tests/test_core.py
+++ b/thunder/tests/test_core.py
@@ -2313,6 +2313,42 @@ def test_preserve_weight_names(executor, device: str, dtype: dtypes.dtype):
     assert "t_fc2_weight" in sig.parameters
 
 
+@requiresCUDA
+def test_clone():
+    def foo(a):
+        return a.clone()
+
+    jfoo = thunder.jit(foo)
+    for shp in ((3, 5), [7], (8, 6, 4)):
+        for dev in (torch.device("cpu"), torch.device("cuda:0")):
+            for dt in (torch.float32, torch.float16, torch.bfloat16):
+                # there are issues with layouts other than strided; see
+                # test_clone_sparse_coo.
+                lout = torch.strided
+                b = jfoo(torch.randn(shp, device=dev, layout=lout, dtype=dt))
+                assert b.dtype == dt
+                assert b.layout == lout
+                assert b.device == dev
+                assert b.shape == torch.Size(shp)
+
+
+@pytest.mark.xfail(reason="randn() call fails, before getting to thunder")
+def test_clone_sparse_coo():
+    def foo(a):
+        return a.clone()
+
+    jfoo = thunder.jit(foo)
+    shp = (3, 5)
+    dev = "cpu"
+    lout = torch.sparse_coo
+    dt = torch.float32
+    b = jfoo(torch.randn(shp, device=dev, layout=lout, dtype=dt))
+    assert b.dtype == dt
+    assert b.layout == lout
+    assert b.device == dev
+    assert b.shape == torch.Size(shp)
+
+
 @instantiate(dtypes=(thunder.float32,))
 def test_default_method(executor, device: str, dtype: dtypes.dtype):
     # This test ensures that when no language context is given, it will fallback to the default implementation.
--- a/thunder/tests/test_core.py
+++ b/thunder/tests/test_core.py
@@ -2332,19 +2332,20 @@ def test_clone():
                 assert b.shape == torch.Size(shp)
 
 
-@pytest.mark.xfail(reason="randn() call fails, before getting to thunder")
+@pytest.mark.xfail(reason="missing torch support for repeat'ing sparse tensors")
 def test_clone_sparse_coo():
     def foo(a):
         return a.clone()
 
     jfoo = thunder.jit(foo)
     shp = (3, 5)
-    dev = "cpu"
-    lout = torch.sparse_coo
+    dev = torch.device("cpu")
     dt = torch.float32
-    b = jfoo(torch.randn(shp, device=dev, layout=lout, dtype=dt))
+    # randn(layout=torch.sparse_coo, ...) will fail, so we use to_sparse()
+    # from a dense tensor to get a sparse one.
+    b = jfoo(torch.randn(shp, device=dev, dtype=dt).to_sparse())
     assert b.dtype == dt
-    assert b.layout == lout
+    assert b.layout == torch.sparse_coo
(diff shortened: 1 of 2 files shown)


thunder/torch/__init__.py
```diff
@@ -2001,6 +2001,11 @@ def amin(a, /, dim=None, keepdim: bool = False):
     )
 
 
+@torchsymbol(torch.clone, is_method=True)
+def clone(a: TensorProxy, *, memory_format=torch.preserve_format) -> TensorProxy:
+    return a
```
nikitaved, line 2006: Could that be `a + 0`?

tfogal: It could be `a+0` of course... tell me more about where you're coming from here, I'm not quite following the 'why'.

I actually wasn't at the triage meeting. @mruberry could you give more insight into #342's triage comment?

tfogal:
The failures are distributed notebook flakiness; unrelated, so marking ready for review.

I didn't understand the #342 comment about not making this a symbol. Without the `@torchsymbol` decorator, `clone` calls won't map to this implementation, of course. Maybe someone could help educate me on how else this could be done?

thunder/tests/test_core.py
```diff
@@ -2313,6 +2313,42 @@ def forward(self, x):
...
+                # test_clone_sparse_coo.
+                lout = torch.strided
+                b = jfoo(torch.randn(shp, device=dev, layout=lout, dtype=dt))
+                assert b.dtype == dt
+                assert b.layout == lout
+                assert b.device == dev
+                assert b.shape == torch.Size(shp)
+
+
+@pytest.mark.xfail(reason="randn() call fails, before getting to thunder")
```
mruberry, line 2335: I don't think that thunder supports sparse tensors, but to create one you can call `to_sparse` on a non-sparse tensor

see https://pytorch.org/docs/stable/sparse.html

thunder/torch/__init__.py
nikitaved: `a+0` will be a copy. Returning just `a` will produce a tensor with the same storage.

tfogal: Good point. I addressed this with a cheeky `repeat` usage, which avoids the issue of an appropriate type for `0` s.t. this doesn't introduce a cast.

Please 'resolve' if this works for you.

thunder/tests/test_core.py
tfogal: I fiddled around with this a bit but hit another missing torch feature; it seems `repeat` uses `expand` under the hood, and `expand` doesn't work on sparse tensors.

I get the impression that dense/sparse layouts are not in wide use, and I don't think having our support go wider than torch's is valuable at this time. I'll leave the test just to document our shortcoming but ... let's move on.

tfogal:
The test failures are for distributed things:
https://dev.azure.com/Lightning-AI/lightning/_build/results?buildId=202766&view=logs&j=b97dbf6d-98bd-5b68-7c01-878b39c3da28&t=3c72ede2-92c1-5cd2-2bac-ad2411af2aea&l=306
which seem unrelated. Let's try merging main into this...

thunder/torch/__init__.py
nikitaved: I am not sure this will preserve the memory format, though. It's not uncommon to see use cases with channels last and alike.

tfogal: AFAICT [the implementation of repeat](https://github.com/Lightning-AI/lightning-thunder/blob/54bb6146ff757905925f8d9ea2197870c4971011/thunder/torch/__init__.py#L919) does not do anything to reorder dimensions or similar. it's basically just a broadcast from one tensor to another. Given that the dimensions would be the same, I imagine this just ends up being a copy.

For other reasons I do think we want a `copy` primitive (which would be a more natural mapping, here), but I'm not sure others agree. Reasonable might argue that "broadcast to the same dimensions *is* the way to do copy", and I'm not sure I would disagree.

@t-vi thoughts on what we should do here?

nikitaved: Sure, but then it works only if memory format is selected. Is it tested? What about different memory formats?

mruberry:
triage review — just `return a` for the impl seems OK for now — we can follow-up if that confuses practitioners or we want more clarity

thunder/torch/__init__.py
tfogal: Hey @nikitaved we had a chat and I missed something earlier:

if we just `return a` here without adding the `@torchsymbol` annotation then the clone won't actually appear in the trace, it'll just give back the same tensor. in triage review we noted the issue that this could result in weirdness if something modifies the return value of `clone`, but this actually falls under the general rubric of in-place ops and functionalizing the trace, and the sentiment was that we shouldn't try to boil the ocean in a PR that just adds an op.

Your point on memory formats is a good one though; for now I'll just add a conditional and check that preserve_format is used, so we at least get some notification if this is awry.

tfogal:
The failures are just the distributed tests failing, issue #432. This is ready for re-review / merging.

tfogal:
poke @mruberry @t-vi for review/merging

thunder/torch/__init__.py
```diff
@@ -2001,6 +2001,37 @@ def amin(a, /, dim=None, keepdim: bool = False):
...
+    """
+    # Our implementation currently does not introduce a copy, and so nothing
+    # except preserve_format is feasible to support.
+    # If you're hitting this you could try commenting this check out; if your
+    # model does not actually rely on specified memory formats then it should
+    # be fine.
+    if memory_format is not torch.preserve_format:
+        raise NotImplementedError("only preserve_format is currently supported")
+    # This implementation just creates an alias instead of a copy. This may
+    # introduce problems; such problems would be fixable when we get to adding an
```
lantiga, line 2022: Can we add a test that xfails on aliasing vs actually copying? So we don't forget when we get to SSA.

tfogal: Good idea! I've added one.

lantiga (approved):
Ship it!

PR #430, opened 2024-05-17
CUDA Graphs: enable in forward

nikitaved:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #430

### DIFF
--- a/thunder/__init__.py
+++ b/thunder/__init__.py
@@ -322,6 +322,9 @@ def jit(
     assert type(record_history) is bool
 
     # TODO RC1 Refine the compile data option to remove unused options
+    # TODO: refine options
+    # NOTE(fixme): use_cudagraphs is being absorbed into compile_options
+    use_cudagraphs = compile_options.get('use_cudagraphs', False)
     cd = CompileData(
         fn=fn,
         langctx=langctx,
@@ -329,7 +332,7 @@ def jit(
         cache_option=cache,
         sharp_edges=sharp_edges,
         using_jit=True,
-        use_cudagraphs=False,
+        use_cudagraphs=use_cudagraphs,
         disable_torch_autograd_support=disable_torch_autograd,
         use_rematerialization=False,
         only_execute_prims=False,
@@ -578,6 +581,9 @@ def jit(
                 backward_fn = backward_trc.python_callable()
             else:
                 backward_fn = None
+
+            if cd.use_cudagraphs:
+                comp = CUDAGraphExecutor(comp)
 
             # TODO RC1 Update the cache
             cache_entry = CacheEntry(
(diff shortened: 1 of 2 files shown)


t-vi (commented):
Supergood! Should we have a test that checks for some example that cudagraph is used when registered?

nikitaved:
The issue with distributed seems unrelated - it fails for me either way

mruberry:
triage review —

- if we're interested in enabling CUDA Graph support, let's do a design review! We could consider implementing it as a transform and handling forward and backward and dealing with operations in PyTorch that cause CUDA graphs to fail
- this PR can enable CUDA graphs using `use_cudagraphs` but needs to emit a warning that the feature is unstable
- this PR can remove the option entirely so practitioners aren't confused and we revisit CUDA graphs later during a design review

nikitaved:
It operates over a callable. Does it really make sense to have it as a transform?

nikitaved:
I would like to keep to be able to see whether there is benefit to it after all.

mruberry:
> It operates over a callable. Does it really make sense to have it as a transform?

An interesting question to consider during a design review!

nikitaved:
@mruberry , it seems it actually does makes a sense to have it as a transform if it is implemented as an executor, and if we can inspect `thunder.cuda.make_graph_callable` to be able to implement the claiming strategy, and/or we can define "static" subsections on our own. Modulo edge cases when the whole compute is actually "graphable" and consequences for other executors.

t-vi (approved):
Seems good in terms of getting thunder.jit to parity of the old thunder.compile, so to not let the better be the enemy of the good, LGTM.

Thank you @nikitaved @mruberry

thunder/__init__.py
```diff
@@ -581,6 +584,12 @@ def get_computation_and_inputs(*args, **kwargs):
             else:
                 backward_fn = None
 
+            # TODO: using vanilla CUDAGraphExecutor is not safe unless the graph is always static!
+            # (fixme): inspect torch.cuda.make_graph_callables and/or use it instead!
+            # See https://github.com/Lightning-AI/lightning-thunder/issues/433
+            if cd.use_cudagraphs:
+                comp = CUDAGraphExecutor(comp)
```
IvanYashchuk, line 591: How are network parameters treated? Usually, they can be treated as having static memory and do not require a copy before executing the CUDA graph. Have you verified that it works as expected and no large copy is made before the recorded kernels execution?

nikitaved: I did not modify that bit and it is inherited from the original implementation. Which, unless the mask is provided (not the case), marks parameters for not copying.

IvanYashchuk: Are nn.Parameters preserved when passed through the prologue trace? I hope they are not unwrapped into tensors.

nikitaved: Yes, it is all good there. The mask looks as expected.

IvanYashchuk: Thanks a lot for checking!

PR #445, opened 2024-05-22
Add support for torch FP8 dtypes

### LINKED ISSUE #254: Add support for FP8E4M3 and FP8E5M2 dtypes
## 🚀 Feature

Currently supported dtypes are listed in this conversion dict:
https://github.com/Lightning-AI/lightning-thunder/blob/17024fa1c805c4c7bc37673be26fac4ef7a1b64f/thunder/core/dtypes.py#L504-L535

PyTorch has recently added support for FP8E4M3 and FP8E5M2 dtypes and we need to add them to Thunder for native FP8 mixed precision training.

The first step is to be able to compile a function that does only dtype conversion.

riccardofelluga:
<details>

</details>

## What does this PR do?
This PR fixes #254 and adds native thunder support for the following dtypes:
``` python
torch.float8_e5m2
torch.float8_e5m2fnuz
torch.float8_e4m3fn
torch.float8_e4m3fnuz
```

Since the float8 dtype is implemented in 4 different variants I added  the variant mechanism for Thunder dtypes such that we can differentiate between them.

This PR also adds the option to create test fp8 tensors with `make_tensor` so that we can start testing fp8 operations. After running the existing operators tests it is evident that the support for this dtype in torch is scarce since the majority of tests fail with "not implemented" runtime errors. With that I decided to skip the operator testing for all the fp8.

Furthermore, I updated the type promotion table, please get a look and don't hesitate to comment if you think some promotions are not in the right place.

## Did you have fun?
Oh yes!


thunder/core/dtypes.py
```diff
@@ -154,12 +162,22 @@ class inexact(dtype):
 class floating(inexact):
     """Base class for the floating dtypes: bfloat16, float16, float32, float64."""
```
crcrpar, line 163: what do you say to adding fp8 dtypes to this line?

thunder/core/dtypes.py
```diff
@@ -551,7 +591,7 @@ def to_torch_dtype(x: None | torch.dtype | dtype) -> None | torch.dtype:
 
 # Converts NumPy dtypes to and from thunder dtypes
 
-# NOTE NumPy does not support the bfloat16 or complexhalf (complex32) datatypes
+# NOTE NumPy does not support the bfloat16 or complexhalf (complex32) or float8 datatypes
```
crcrpar, line 594:
```suggestion
# NOTE NumPy does not support the bfloat16, complexhalf (complex32) or float8 datatypes
```

thunder/core/dtypes.py
riccardofelluga: Oh yes! I missed it!

thunder/core/dtypes.py
```diff
@@ -80,23 +81,30 @@ def is_weak(self):
...
+        return (
+            f"{self._name}{8 * self._bytes}{f'_{self._variant}' if self._variant else ''}{'_' if self._is_weak else ''}"
+        )
 
     def __str__(self):
         return self.__repr__()
 
     def __hash__(self) -> int:
-        return hash((self._name, self._bytes, self._is_weak))
+        return hash((self._name, self._bytes, self._is_weak, f"{self._variant if self._variant else ''}"))
```
kshitij12345, line 96: I think `hash(None)` should work so this can probably written as `hash((self._name, self._bytes, self._is_weak, self._variant))`

thunder/core/dtypes.py
```diff
@@ -80,23 +81,30 @@ def is_weak(self):
...
 
     def shortname(self):
-        return f"{self._shortname}{8 * self._bytes}"
+        return f"{self._shortname}{8 * self._bytes}{f'_{self._variant}' if self._variant else ''}"
 
     # TODO Fix name printing
     def __repr__(self):
-        return f"{self._name}{8 * self._bytes}{'_' if self._is_weak else ''}"
+        return (
+            f"{self._name}{8 * self._bytes}{f'_{self._variant}' if self._variant else ''}{'_' if self._is_weak else ''}"
```
kshitij12345, line 89: Given that `self._variant` is Optional[str], I think we can write this as 
```python
f"{self._name}{8 * self._bytes}{self._variant if self._variant else ''}{'_' if self._is_weak else ''}"
```

thunder/core/dtypes.py
```diff
@@ -242,6 +268,8 @@ def __init__(self, name, shortname, *, bytes, is_weak):
 
 float_dtypes = {d for d in all_dtypes if isinstance(d, floating)} | {float}
 
+float_8bit_dtypes = {d for d in all_dtypes if (isinstance(d, inexact) and d.bytes <= 1)}
```
kshitij12345, line 271: Is `isinstance(d, inexact)` required instead of `isinstance(d, floating)` (which is used in the above line for `float_dtypes`)? Also, I think we should keep `d.bytes = 1`, as in future we may have dtypes smaller than a byte.

kshitij12345 (commented):
Thanks @riccardofelluga for the PR, this will enable a lot of interesting opportunities for FP8 support in `thunder`.

Overall the PR looks really good, I just think FP8 should not be involved in type promotion for the following reasons:
1. For gradient computation of linear, it is common to use E5M2 for gradient (to avoid underflow/overflow) and weights/input in `E4M3` for higher precision, in such case, we don't want `E4M3` to be upcasted as there are GEMM kernels for mixed FP8 dtypes.
2. When we are using an operator with FP8, we want to enforce that FP8 inputs will stay in FP8 for performance reasons (we won't want the input to be upcasted to higher precision and not use special FP8 GEMMs).
3. Except for `matmul`, there aren't any math operation support for FP8. (Also, I don't think the coverage would increase soon).

So, I think we should remove the type promotion logic. (cc: @mruberry for his thoughts on the same)

Also, I think we should not enable testing with `FP8` by default, as FP8 will have a very limited op support and all tests will have to disable it for now. It is likely that we will have separate tests for selected ops to test with fp8. So we should probably tweak `instantiate` to not enable FP8 types by default for testing and only use them if they were specifically passed to `dtypes` argument.

https://github.com/Lightning-AI/lightning-thunder/blob/82185e3a55d5b3f0bea8a7366d74a275dbe34acd/thunder/tests/framework.py#L416-L438

thunder/core/dtypes.py
mruberry: Agreed; `inexact` includes the complex types, which probably doesn't matter practically, bit it's nice to be more consistent with the construction of `float_dtypes` by querying `floating`

thunder/core/dtypes.py
```diff
@@ -551,7 +591,7 @@ def to_torch_dtype(x: None | torch.dtype | dtype) -> None | torch.dtype:
 
 # Converts NumPy dtypes to and from thunder dtypes
 
-# NOTE NumPy does not support the bfloat16 or complexhalf (complex32) datatypes
+# NOTE NumPy does not support the bfloat16, complexhalf (complex32) or float8 datatypes
```
mruberry, line 594: Nice comment update

thunder/core/utils.py
```diff
@@ -327,23 +338,31 @@ def check_same_dtype(*args):
...
-    [ complex,  c64_,  c32_,  c64_, c128_,  c64,  c32,  c64, c128, complex,  c32_,  c64_, c128_,  c32,  c64, c128], # complex
-    [    c32_,  c64_,  c32_,  c64_, c128_,  c64,  c32,  c64, c128,    c32_,  c32_,  c64_, c128_,  c32,  c64, c128], # c32_
-    [    c64_,  c64_,  c64_,  c64_, c128_,  c64,  c32,  c64, c128,    c64_,  c64_,  c64_, c128_,  c32,  c64, c128], # c64_
-    [   c128_, c128_, c128_, c128_, c128_,  c64,  c32,  c64, c128,   c128_, c128_, c128_, c128_,  c32,  c64, c128], # c128_
-    [     c32,   c32,   c32,   c32,   c32,  c64,  c32,  c64, c128,     c32,   c32,   c32,   c32,  c32,  c64, c128], # c32
-    [     c64,   c64,   c64,   c64,   c64,  c64,  c64,  c64, c128,     c64,   c64,   c64,   c64,  c64,  c64, c128], # c64
-    [    c128,  c128,  c128,  c128,  c128, c128, c128, c128, c128,    c128,  c128,  c128,  c128, c128, c128, c128], # c128
+    #           f   f8_e4m3fn_  f8_e4m3fnuz_     f8_e5m2_  f8_e5m2fnuz_          bf_         f16_         f32_          f64_   f8_e4m3fn  f8_e4m3fnuz      f8_e5m2   f8_e5m2fnuz     bf   f16   f32   f64  complex   c32_   c64_  c128_   c32   c64  c128
```
mruberry, line 341: This table is getting out of hand!

mruberry: Serious question for now: what if we just didn't support type promotion for these datatypes for the moment?

Like if we ever try to type promote them we just have a special check that all the tensors are the same fp8 dtype, and if they're not we fail?

thunder/core/utils.py
```diff
@@ -327,23 +338,31 @@ def check_same_dtype(*args):
...
+    [       float,  f8_e4m3fn_, f8_e4m3fnuz_,    f8_e5m2_, f8_e5m2fnuz_,         bf_,        f16_,        f32_,        f64_,   f8_e4m3fn, f8_e4m3fnuz,     f8_e5m2, f8_e5m2fnuz,    bf,  f16,  f32,  f64, complex,  c32_,  c64_, c128_,  c32,  c64, c128], # f
+    [       float,  f8_e4m3fn_, f8_e4m3fnuz_,    f8_e5m2_, f8_e5m2fnuz_,         bf_,        f16_,        f32_,        f64_,   f8_e4m3fn, f8_e4m3fnuz,     f8_e5m2, f8_e5m2fnuz,    bf,  f16,  f32,  f64, complex,  c32_,  c64_, c128_,  c32,  c64, c128], # f8_e4m3fn_
```
mruberry, line 343: One of the reasons we may not want to define type promotion now is that we don't have to think if different fp8 variants should promote to fp16, for example

thunder/tests/make_tensor.py
```diff
@@ -117,9 +117,15 @@ def clamp(a, l, h):
     shape = cast(tuple[int, ...], tuple(shape))
 
     _integral_types = [torch.uint8, torch.int8, torch.int16, torch.int32, torch.int64]
+    _floating_8bit_types = [torch.float8_e4m3fn, torch.float8_e4m3fnuz, torch.float8_e5m2, torch.float8_e5m2fnuz]
```
mruberry, line 120: This is a really cool extension of `make_tensor`

mruberry (approved):
Hey @riccardofelluga! Cool stuff. I made a few small suggestions. The big things I'm curious about before merging is:

- should we delay implementing type promotion logic for fp8 dtypes? The logic update seems pretty reasonable, but maybe different fp8 dtypes should promote to fp16 for now?
- a convenient set of datatypes so test authors can select the floating point types except the fp8 types would be nice for now

Curious to hear your thoughts!

thunder/core/dtypes.py
riccardofelluga: This line would make from `f8_variant` to `f8variant`. I think for readability it's maybe better to keep the underscore. What do you think?

kshitij12345: Ah right, my bad. Thanks for clearing this!

riccardofelluga:
Thanks everybody for your comments! I though about it a bit and I mainly agree with your points, the support for this dtype in torch is so low that I agree with you, we should delay the type promotion.

Regarding the tests, I also had the idea to disable them from the start but I didn't realize how ugly the `pytest.skip` everywhere ended up looking. I introduce the `float_math_dtypes` set of floating datatypes so that it is possible to `@instantiate` the test with only the floating dtypes that are worth testing for now.

thunder/core/dtypes.py
kshitij12345: I think we should revert it back to `f"{self._variant if self._variant else ''}"` as hash of empty string is always `0`. 

I have verified that `hash(None)` is not same across different Python process. This is maybe the reason why the pytest sharding is failing. A PR fixed this to be constant across different processes https://github.com/python/cpython/pull/99541 but as per the comment https://github.com/python/cpython/pull/99541#issuecomment-1974098703, it is not guaranteed to be constant by the language spec.

My bad 😓!

t-vi:
Thank you @riccardofelluga @crcrpar @kshitij12345 @mruberry @lantiga

PR #447, opened 2024-05-23
patch proxify on sequence and dictionary

jjsjann123:
fixing duplicated proxy for `dict` / `Sequence` in `proxify`


t-vi (approved):
Looks reasonable. Thank you, @jjsjann123

PR #448, opened 2024-05-23
fsdp_v2: use correct variable

kshitij12345:
Following snippet fails

```python
import os
import torch
import thunder
import thunder.distributed
from thunder.tests.litgpt_model import GPT, Config

if __name__ == '__main__':

    class Model(torch.nn.Module):
        def __init__(self, *args, **kwargs) -> None:
            self.register_buffer("foo", torch.empty(4, 4))
            self.fc = torch.nn.Linear(4, 4)
        
        def forward(self, x):
            return self.fc(x)

    with device:
        model = Model()
        x = torch.randn(4, 4)

    model(x)
```

Cmd: 
```
CUDA_LAUNCH_BLOCKING=1 torchrun --nproc-per-node 2 --local-ranks-filter 0 test.py
```

Fails with
```
[rank0]:   File "lightning-thunder/thunder/distributed/__init__.py", line 434, in fsdp_transform_module
[rank0]:     if p.device != device:
[rank0]: UnboundLocalError: local variable 'p' referenced before assignment
```

### DIFF
--- a/thunder/distributed/__init__.py
+++ b/thunder/distributed/__init__.py
@@ -431,7 +431,7 @@ def fsdp_transform_module(
                     thunder_model._overrides[n] = torch.nn.Parameter(p.to(device=device), requires_grad=p.requires_grad)
                     device_adjustments[n] = device
             for n, b in module_copy.named_buffers(recurse=False, prefix=module_name):
-                if p.device != device:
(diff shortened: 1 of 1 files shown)


t-vi (approved):
Thank you @kshitij12345

PR #449, opened 2024-05-23
re-land Caching all parameters for perf

t-vi:
Relative to #421 , this fixes copying the parameters  before sharding when the cache has been preseeded with the values.

Fixes: #396 
(again)


carmocca (approved):
Thank you for your work Tom!

PR #452, opened 2024-05-24
disable calling torchrun in the zero to thunder notebook

t-vi:
...trying to fix the CI.


t-vi:
Does not seem to help. :( Apparently, the torchrun either runs well or just waits for something until the timeout.

PR #459, opened 2024-05-24
pad: support alternate `value` types.

### LINKED ISSUE #458: TypeError with torch.nn.functional.pad
### 🐛 Bug

NeMo's Stable Diffusion uses `torch.nn.functional.pad` as below:

`x = torch.nn.functional.pad(x, pad, mode="constant", value=0)`

When run with `thunder.jit`, it gives a `TypeError`:
`Expected the numbertype <class 'float'>, corresponding to the dtype float32, but found the numbertype <class 'int'>!`
although `torch.nn.functional.pad` supports `value=0` along with float `value=0.0`

CC: @tfogal

tfogal:
PyTorch actually allows the `value` argument to have a different type than the input type, as long as the type of `value` is convertible to the type of `input`. This contradicts [the documentation](https://pytorch.org/docs/stable/generated/torch.nn.functional.pad.html) which indicates that the padding value, if specified, must be a float. 🤷

Fixes #458.

cc @tfogal

### DIFF
--- a/thunder/tests/test_elementwise.py
+++ b/thunder/tests/test_elementwise.py
@@ -139,35 +139,3 @@ def test_where(executor, device, dtype):
     # int64 x complex
     # int32 x complex64
     # predicate as number (True/False)
-
-def test_pad_cast_value_itof():
-    """
-    Pad should cast the given value to the type of tensor and pad that value.
-    """
-    def fqn():
-        x = torch.tensor([2,3], dtype=torch.int32)
-        y = torch.nn.functional.pad(x, pad=(1,2), value=6.4)
-        return y
-    th_fqn = thunder.jit(fqn)
-    v = th_fqn()
-    assert v[0] == 6
-    assert v[1] == 2
-    assert v[2] == 3
-    assert v[3] == 6
-    assert v[4] == 6
-
-def test_pad_cast_value_ftoi():
-    """
-    Pad should cast the given value to the type of tensor and pad that value.
-    """
-    def fqn():
-        x = torch.tensor([2.4, 3.8])
-        y = torch.nn.functional.pad(x, pad=(1,2), value=1)
-        return y
-    th_fqn = thunder.jit(fqn)
-    v = th_fqn()
-    assert v[0] == 1.0
-    assert v[1] == 2.4
-    assert v[2] == 3.8
-    assert v[3] == 1.0
-    assert v[4] == 1.0
--- a/thunder/tests/test_elementwise.py
+++ b/thunder/tests/test_elementwise.py
@@ -139,3 +139,35 @@ def test_where(executor, device, dtype):
     # int64 x complex
     # int32 x complex64
     # predicate as number (True/False)
+
+def test_pad_cast_value_itof():
+    """
+    Pad should cast the given value to the type of tensor and pad that value.
+    """
+    def fqn():
+        x = torch.tensor([2,3], dtype=torch.int32)
+        y = torch.nn.functional.pad(x, pad=(1,2), value=6.4)
+        return y
+    th_fqn = thunder.jit(fqn)
+    v = th_fqn()
+    assert v[0] == 6
+    assert v[1] == 2
+    assert v[2] == 3
+    assert v[3] == 6
+    assert v[4] == 6
+
+def test_pad_cast_value_ftoi():
+    """
+    Pad should cast the given value to the type of tensor and pad that value.
+    """
+    def fqn():
+        x = torch.tensor([2.4, 3.8])
+        y = torch.nn.functional.pad(x, pad=(1,2), value=1)
+        return y
+    th_fqn = thunder.jit(fqn)
+    v = th_fqn()
+    assert v[0] == 1.0
+    assert v[1] == 2.4
+    assert v[2] == 3.8
+    assert v[3] == 1.0
+    assert v[4] == 1.0
--- a/thunder/torch/__init__.py
+++ b/thunder/torch/__init__.py
@@ -906,6 +906,12 @@ def pad(a: TensorProxy, /, pad: tuple[int, ...], mode: str | None = "constant", 
     if value is None:
         value = 0
     a_typ = to_dtype(a, true_dtype=True)
+    # Note that this can be unsafe. This can happen, for example, if `a` is an
+    # integer tensor and `value` is a float. It can also be more subtle, where
+    # `a` is a lower-precision float than `value`.
+    if a_typ is not to_dtype(value, true_dtype=True):
+      warnings.warn("`value` and Tensor input are of different types. This "
+                    "may create numeric issues.")
     v2 = clang.maybe_convert_to_dtype(value, a_typ)
 
     return clang.pad(a, v2, pad_config)
--- a/thunder/torch/__init__.py
+++ b/thunder/torch/__init__.py
@@ -905,8 +905,10 @@ def pad(a: TensorProxy, /, pad: tuple[int, ...], mode: str | None = "constant", 
 
     if value is None:
         value = 0
+    a_typ = to_dtype(a, true_dtype=True)
+    v2 = clang.maybe_convert_to_dtype(value, a_typ)
 
-    return clang.pad(a, value, pad_config)
+    return clang.pad(a, v2, pad_config)
 
 
 @torchsymbol(torch.permute, is_method=True)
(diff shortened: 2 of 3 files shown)


thunder/torch/__init__.py
```diff
@@ -905,8 +905,10 @@ def pad(a: TensorProxy, /, pad: tuple[int, ...], mode: str | None = "constant",
 
     if value is None:
         value = 0
+    a_typ = to_dtype(a, true_dtype=True)
+    v2 = clang.maybe_convert_to_dtype(value, a_typ)
```
mruberry, line 909: `enforce_safe_casting=True`?

thunder/tests/test_elementwise.py
```diff
@@ -139,3 +139,41 @@ def test_where(executor, device, dtype):
     # int64 x complex
     # int32 x complex64
     # predicate as number (True/False)
+
+
+def test_pad_cast_value_itof():
```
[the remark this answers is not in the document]
mruberry: So pad doesn't require the cast is safe? Interesting -- let's add a comment to the pad implementation for that

thunder/tests/test_elementwise.py
```diff
@@ -139,3 +139,41 @@ def test_where(executor, device, dtype):
...
+    """
+    Pad should cast the given value to the type of tensor and pad that value.
+    """
+
+    def fqn():
```
mruberry, line 149: Instead of a separate test, you may want to extend 

https://github.com/Lightning-AI/lightning-thunder/blob/3034ef9efad3db43f9e94e3e9013ce853c0b3680/thunder/tests/opinfos.py#L3634

with two more samples, one where value is an integer and another where value is float

thunder/torch/__init__.py
tfogal: Sadly, no; the cast is unsafe in PyTorch:
```
>>> x = torch.tensor([5,2])
>>> torch.nn.functional.pad(x, [2,1], value=3.14159)
tensor([3, 3, 5, 2, 3])
```

thunder/tests/test_elementwise.py
tfogal: Yeah... sadly I don't really understand what that code's doing. How does the sample generator correspond to the actual code that gets tested?

thunder/tests/test_elementwise.py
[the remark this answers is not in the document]
tfogal: Indeed :-(
Good idea, I'll note that.

also, yeah, i'll move it

thunder/tests/test_elementwise.py
mruberry: Tests use the samples to call the function and verify that the output is consistent with their expectations; so if you add these cases as samples then tests for consistency with PyTorch and consistency with PyTorch's grad will automatically run on them

tfogal: Sorry, my confusion is much more granular. This function seems to yield some tuples, but at some level I'd need to add a new kind of thing to yield, and then some code somewhere (where?) to actually read the new item and create an appropriate tensor.

It's not clear if the new item should be separate from `SampleInput` or embedded in it, either.

jjsjann123 (approved):
Thanks for patching this one. It's surprising to see how `pad` handles the promotion differently from a pointwise operation.

thunder/tests/test_elementwise.py
mruberry:
```
def pad_sample_generator(op, device, dtype, requires_grad, **kwargs):
    make = partial(make_tensor, device=device, dtype=dtype, requires_grad=requires_grad)

    # shape, padding_config
    cases = (
        # 1-numel cases
        ((1, 1), ((0, 2, 0), (0, 2, 0))),
        # Other generic cases
        ((1, 3), ((0, 0, 0), (0, 0, 0))),
        ((3, 7, 5), ((-2, 1, 0), (1, 3, 0), (-1, 2, 0))),
        ((2, 2), ((1, 1, 1), (-1, 2, 0))),
        ((2, 0, 3), ((1, 0, 0), (1, 1, 2), (0, 0, 0))),
        ((7, 5), ((0, 0, 3), (-6, 2, 1))),
        ((3, 2, 5), ((-2, 1, 0), (1, -1, 0), (-1, 3, 1))),
        # Versions of above examples but with padding between elements set to 0
        ((2, 2), ((1, 1, 0), (-1, 2, 0))),
        ((2, 0, 3), ((1, 0, 0), (1, 1, 0), (0, 0, 0))),
        # See issue "PyTorch pad prim lowering handles out-of-bands negative padding incorrectly"
        # ((7, 5), ((0, 0, 0), (-6, 2, 0))),
        ((5, 7), ((0, 0, 0), (-6, 2, 0))),
        ((3, 2, 5), ((-2, 1, 0), (1, -1, 0), (-1, 3, 0))),  # negative pad in all 3 dims
    )

    for shape, padding_config in cases:
        yield SampleInput(make(shape), make_number(dtype=dtype), padding_config)
```

The tuples are templates for generating sample inputs. The generator exclusively yields `SampleInput` objects. Just add a couple lines after the current enumeration of cases:

```
...

for shape, padding_config in cases:
        yield SampleInput(make(shape), make_number(dtype=dtype), padding_config)

# Tests that values of different dtypes can be supplied
# NOTE Float values are allowed even when the tensor has an exact dtype, and the float value is 
# cast to an integer
yield SampleInput(make((2, 3), pad=pad=(1, 2), value=6.4)
yield SampleInput(make((2,), pad=(1, 2), value=1)
```

That way the two samples used in the custom tests you wrote will go through all the tests. It does switch from "golden value" testing to testing that we produce something consistent with PyTorch, but that should be fine. And it will also test the autograd.

lantiga (approved):
LGTM. Sneaky!

thunder/tests/test_elementwise.py
tfogal: Hey @mruberry, sorry for the delay, finally getting back to this.

I tried yielding a couple values like this, but unfortunately things don't work out so easily. Inside the primitive implementation `pad_meta`, we have the line https://github.com/Lightning-AI/lightning-thunder/blob/7ac4c49aca1f4282dc9dfe5ba69e343a358d1942/thunder/core/prims.py#L3045

and this ends up causing a bunch of errors such as:
```
E RuntimeError: Expected the numbertype <class 'int'>, corresponding to the dtype uint8, but found the numbertype <class 'float'>!
```

I think the primitive is actually sane (i.e. that it is pretty wild to accept a `value` that differs from the type of the underlying tensor), and so I am loathe to remove the check--it's entirely reasonable to me that this dtype difference would be handled by the frontend.

But that makes things complicated because it seems the opinfo machinery is going directly to the primitive without going through the torch executor:

```
framework.py:311:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
framework.py:529: in _fn
    return fn(*args, **kwargs)
test_ops.py:137: in test_core_vs_jax_consistency
    result = run_snippet(
framework.py:518: in run_snippet
    raise ex
framework.py:510: in run_snippet
    snippet(*args, **kwargs)  
test_ops.py:100: in snippet_jax_consistency
    thunder_result = op(*sample.args, **sample.kwargs)
../__init__.py:641: in fn_
../__init__.py:217: in cache_info_wrapper
    res = fn(*args, **kwargs)
../__init__.py:491: in get_computation_and_inputs
    jit_results: TraceResults = interpreter(
../__init__.py:205: in _general_frontend
    return thunder_general_jit(fn, args, kwargs, sharp_edges=sharp_edges, record_history=record_history)
../core/jit_ext.py:1583: in thunder_general_jit
    result = jfn(*args, **kwargs)
../core/interpreter.py:6696: in fn_
    raise e
../core/interpreter.py:6664: in fn_2
    return fn(*args, **kwargs)
../core/interpreter.py:1273: in wrapping_wrapper
    res = ufn(*uargs, **ukwargs)
../core/jit_ext.py:681: in wrapper
    return fn(*args, **kwargs)
../core/symbol.py:263: in __call__
../core/langctxs.py:132: in _fn
    result = fn(*args, **kwargs)
../core/prims.py:3046: in pad_meta
    utils.check_same_dtype(a, padding_value)
../core/utils.py:236: in check_same_dtype
    check(
```

So AFAICT our options here are:
1. Undo this change to the torch executor and instead do this as a compiler pass
2. Fiddle with the opinfo machinery such that the tests go through the torch executor first
3. Make the primitive match torch's behavior.

Thoughts?

mruberry: There are OpInfos for both `prims.pad` and `ltorch.pad` -- are you updating the sample generator for the latter?

https://github.com/Lightning-AI/lightning-thunder/blob/23da3c12840e497aaed1064dd2e585904a200e5b/thunder/tests/opinfos.py#L3752

tfogal: *sigh*, yep, I was in the wrong place. Thanks.

PR #469, opened 2024-05-29
Update link about cudnn debugging in cudnnex.py

IvanYashchuk:


t-vi (approved):
Thank you, @IvanYashchuk

PR #472, opened 2024-05-29
meta function for max_pool missed the non-pooling dimensions.

### LINKED ISSUE #365: max_pool2d：Shape propagation error for input dimensions used by resnet
## 🐛 Bug

```py
import torch
import thunder

a = torch.randn(1, 64, 112, 112).cuda().requires_grad_()
def func(a):
    return F.max_pool2d(a, 3, 2, 1, 1, False, False)  # t79: "cuda:0 f32[1, 64, 56, 56]"
cfunc = thunder.jit(func)
b = cfunc(a)

print(thunder.last_traces(cfunc)[-1].output[0]['output'].shape)
print(b.shape)
```
Outputs:
```py
(56, 56)
torch.Size([1, 64, 56, 56])
```
The output shape of the trace is wrong, but it runs successfully

Trace:
```py
import torch

@torch.no_grad()
@no_autocast
def augmented_forward_fn(a):
  # a: "cuda:0 f32[1, 64, 112, 112]"
  (t0, t1) = max_pool2d_with_indices(a, 3, 2, 1, 1, False)
  return {'output': t0, 'flat_args': [a], 'flat_output': (t0,)}, ((a, t1), (False, 3, 2, 1, 1))
```

cc @apaz-cli

jjsjann123:
## What does this PR do?

Fixes #365 

- [ ] Adding test for meta function.

### DIFF
--- a/thunder/executors/torchex.py
+++ b/thunder/executors/torchex.py
@@ -1323,7 +1323,7 @@ def _max_pool_with_indices_helper(
             )
             return seq[i]
 
-    out_sizes = []
+    out_sizes = list(a.shape[:-ndim])
     for i in range(ndim):
         in_ = a.shape[i - ndim]  # i - ndim is the i-th spatial dimension
         kernel_ = get_maybe_ith_entry("kernel_size", kernel_size, i)


lantiga (approved):
Thanks! Should we augment the test too?

jjsjann123:
> Thanks! Should we augment the test too?

Yes we definitely should. @mruberry mentioned the same in https://github.com/Lightning-AI/lightning-thunder/issues/365#issuecomment-2108606263

This is meant to be a quick fix for @kiya00 so I didn't properly look at that part. Thanks for the quick merge :bow:

Issue #257, opened 2024-03-05
Weight tying + FSDP = nvfuser internal error 


carmocca:
### To Reproduce

Code:

```python
import os
import torch
import torch.distributed as tdist
import thunder
from thunder.tests.lit_gpt_model import GPT, Config

if __name__ == "__main__":
    tdist.init_process_group(backend="nccl")
    LOCAL_RANK = int(os.environ["LOCAL_RANK"])
    device = torch.device("cuda", LOCAL_RANK)
    torch.set_default_device(device)

    config = Config(block_size=256, padded_vocab_size=32000, n_layer=6, n_head=6, head_size=48, n_embd=288, rotary_percentage=1.0, parallel_residual=False, bias=False, _norm_class='RMSNorm', _mlp_class='LLaMAMLP', intermediate_size=768)
    with device:
        model = GPT(config)

    model.transformer.wte.weight = model.lm_head.weight

    model = thunder.distributed.fsdp(model)
    model = thunder.jit(model)

    input_ids = torch.randint(1, 30010, (128, 256), dtype=torch.long, device=device)
    logits = model(input_ids)
    print(logits.shape)
```

Run with:

```shell
torchrun --nproc-per-node 2 --local-ranks-filter 0 repro.py
```

Nvfuser repro:

```python
import torch
from nvfuser import FusionDefinition, DataType

def nvfuser_fusion_id0(fd : FusionDefinition) -> None :
    T0 = fd.define_tensor(shape=[-1, -1, -1], contiguity=[None, None, True], dtype=DataType.Float, is_cpu=False, stride_order=[2, 1, 0])
    T1 = fd.define_tensor(shape=[-1, -1, -1], contiguity=[True, True, True], dtype=DataType.Float, is_cpu=False, stride_order=[2, 1, 0])
    T2 = fd.ops.mul(T1, T1)
    T3 = fd.ops.sum(T2, dims=[2], keepdim=False, dtype=DataType.Null)
    S4 = fd.define_scalar(128, dtype=DataType.Int)
    S5 = fd.define_scalar(256, dtype=DataType.Int)
    S6 = fd.define_scalar(1, dtype=DataType.Int)
    V7 = fd.define_vector([S4, S5, S6], dtype=DataType.Int)
    T8 = fd.ops.broadcast_in_dim(T3, shape=V7, broadcast_dims=[0, 1])
    S9 = fd.define_scalar(288.000, dtype=DataType.Double)
[...]

PR #473, opened 2024-05-29
Fix FSDP jit(fsdp(...)) when sharing weights.

kshitij12345:
This PR fixes first problem from the 2 mentioned in https://github.com/Lightning-AI/lightning-thunder/issues/257#issuecomment-2133070264

Will fix `fsdp(jit(...))` in a seperate PR.

### DIFF
--- a/thunder/tests/distributed/test_ddp.py
+++ b/thunder/tests/distributed/test_ddp.py
@@ -1044,8 +1044,7 @@ class CompileDDPTest(DataParallelTestCase):
 
         model.fc1.weight = model.fc2.weight
 
-        model = thunder.distributed.fsdp(model)
-        model = thunder.jit(model, executors=["torch"])
+        model = thunder.jit(thunder.distributed.fsdp(model), executors=["torch"])
 
         x = torch.randn(4, 16)
         output = model(x)
--- a/thunder/tests/distributed/test_ddp.py
+++ b/thunder/tests/distributed/test_ddp.py
@@ -1021,6 +1021,52 @@ class CompileDDPTest(DataParallelTestCase):
         with thunder.ThunderModule.no_sync(model):
             fwd_loss(model, x)
 
+    @pytest.mark.skipif(torch.cuda.device_count() < 2, reason="Requires 2 devices")
+    def test_fsdp_weight_sharing(self):
+        # This test is to verify that weight sharing works with fsdp.
+        # NOTE: Currently we end up creating 2 copies of shared weight during execution.
+        #       This should be fixed and we should update this test to check for that.
+        device = torch.device("cuda", self.rank)
+        torch.set_default_device(device)
+
+        class Model(torch.nn.Module):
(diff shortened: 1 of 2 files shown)


thunder/tests/distributed/test_ddp.py
```diff
@@ -1021,6 +1021,51 @@ def fwd_loss(m, x):
...
+
+        model.fc1.weight = model.fc2.weight
+
+        model = thunder.distributed.fsdp(thunder.jit(model, executors=["torch"]))
+
+        x = torch.randn(4, 16)
+        output = model(x)
+
+        expected_shape = (4, 16)
+        assert output.shape == expected_shape, f"{output.shape=} - {expected_shape=}"
```
t-vi, line 1067: I think things get sharded twice here (but with separate results) and we should avoid this, right?

kshitij12345: For `fsdp(jit(model))` case, we do shard the parameter twice but we shard the same original parameter from user passed module (i.e. we don't reshard the already sharded parameter). So, both copies actually hold the correct data. Ideally we should have only one copy. I am planning to look into this in a separate PR for both `fsdp(jit(model))` and `jit(fsdp(model))`. 

More details at https://github.com/Lightning-AI/lightning-thunder/issues/257#issuecomment-2133070264

We can verify this with the below snippet.

```python
import os
import torch
import thunder
import thunder.distributed

if __name__ == "__main__":

    class Model(torch.nn.Module):
        def __init__(self) -> None:
            super().__init__()
            self.fc1 = torch.nn.Linear(16, 16, bias=False)
            self.fc2 = torch.nn.Linear(16, 16, bias=False)

        def forward(self, x):
            return self.fc1(x) + self.fc2(x)

    with device:
        model = Model()
        input_t = torch.randn(4, 16)

    model.fc1.weight = model.fc2.weight
    # if LOCAL_RANK == 0:
    #     print(model)
    # model = thunder.distributed.fsdp(model)
    # model = thunder.jit(model, executors=["torch"])

    # Verify that tensor data is same in both weights.
    torch.testing.assert_close(model._overrides["fc1.weight"], model._overrides["fc2.weight"])

    out = model(input_t)

    if LOCAL_RANK == 0:
        trace = thunder.last_traces(model)[-1]
        with open("weight_sharing_trace.py", 'w') as f:
            f.write(str(trace))

```

t-vi: Seems good! Thank you!

kshitij12345: Gradients for `fsdp(jit(...))` are incorrect, I am planning to fix `fsdp(jit(...))` in a seperate PR in a day or so.

t-vi: That's the "should avoid this", probably, as we would want the shared param to get the sum of the gradients.

t-vi (approved):
Thank you @kshitij12345

PR #480, opened 2024-05-29
Add LIGHTNING_BENCH_TRACE environment variable.

tfogal:
When set, this will dump the trace for the given benchmark[s]. This was a feature request from a colleague debugging why a trace was slower than expected.

We'd talked about this, and there were some unresolved questions/thoughts:

* is printing the trace to a terminal actually the thing we want? A trace alone doesn't let the network be replayed, unfortunately
* should we add some sugar to an approach like the above? Like what if the actual pytest benchmark can be imported and asked for its function and inputs, so practitioners don't have to go the benchmark itself?
* should we add a notebook about debugging benchmarks to the `dev_tutorials` directory?

To elaborate on the second option, should we do support something like:

```python
from thunder.benchmarks.targets import foo_bench

fn = foo_bench.fn
args, kwargs = foo_bench.make_inputs()

fn(*args, **kwargs)

thunder.last_traces(fn)
```

And maybe even more sugar, like

```python
print(foo_bench.source_location)
```

and that prints a clickable link to the source for the benchmark, or the script the benchmark pulls from, so people can see and modify the code?

### DIFF
--- a/thunder/benchmarks/targets.py
+++ b/thunder/benchmarks/targets.py
@@ -1,6 +1,7 @@
 from collections.abc import Callable
 from functools import partial, wraps
 from collections.abc import Sequence
+import os
 
 from lightning_utilities.core.imports import package_available
 
@@ -166,13 +167,18 @@ def thunder_fwd(b: Benchmark, compile_fn: Callable):
 
         @wraps(cfn)
         def wrapper(*args):
-            return cfn(args)
+            result = cfn(args)
+            if os.getenv("LIGHTNING_BENCH_TRACE") is not None:
+                print("thunder trace:", thunder.last_traces(cfn)[-1])
+            return result
 
         return wrapper
 
     @wraps(cfn)
     def wrapper(*args, **kwargs):
         result = cfn(*args, **kwargs)
+        if os.getenv("LIGHTNING_BENCH_TRACE") is not None:
+            print("thunder trace:", thunder.last_traces(cfn)[-1])
         return result
 
     return wrapper
@@ -213,6 +219,8 @@ def thunder_fwd_bwd(b: Benchmark, compile_fn: Callable):
             clear_grads(module)
             result = cfn(args)
             result.backward(torch.ones_like(result))
+            if os.getenv("LIGHTNING_BENCH_TRACE") is not None:
+                print("thunder trace:", thunder.last_traces(cfn)[-1])
             return result
 
         return wrapper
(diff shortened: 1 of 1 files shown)


t-vi:
Is there a reason to not call it THUNDER_BENCH_TRACE btw?

IvanYashchuk:
> To elaborate on the second option, should we do support something like:
```py
from thunder.benchmarks.targets import foo_bench

fn = foo_bench.fn
args, kwargs = foo_bench.make_inputs()

fn(*args, **kwargs)

thunder.last_traces(fn)
```

I think we should this and we do, he's real runnable code:
```py
from thunder.benchmarks import LitGPTSDPABenchmark
import thunder

foo_bench = LitGPTSDPABenchmark("Llama-3-8B", batchdims=(2,))
fn = thunder.jit(foo_bench.fn())
args, kwargs = foo_bench.make_batch()

fn(*args, **kwargs)

thunder.last_traces(fn)

**Comment by IvanYashchuk** [comment]
> And maybe even more sugar, like
`print(foo_bench.source_location)`
and that prints a clickable link to the source for the benchmark, or the script the benchmark pulls from, so people can see and modify the code?

It's certainly a useful feature for some people, but there are specialized tools just for searching the source location. I think we can prevent the growth of complexity in Thunder simply by suggesting using these tools.
Python has a built-in library for inspecting the source code of live objects (https://docs.python.org/3/library/inspect.html#retrieving-source-code). Many IDEs have a search feature to quickly find the source location of any object definition within a repo (e.g. https://code.visualstudio.com/docs/editor/codebasics?ref=arctype.com#_search-across-files).

File: thunder/benchmarks/targets.py [hunk 1]
```diff
@@ -166,13 +167,18 @@ def thunder_fwd(b: Benchmark, compile_fn: Callable):
 
         @wraps(cfn)
         def wrapper(*args):
-            return cfn(args)
+            result = cfn(args)
+            if os.getenv("LIGHTNING_BENCH_TRACE") is not None:
```
IvanYashchuk, line 171: Is the problem we're trying to solve that people would like to see the trace but don't know where to insert a breakpoint?

tfogal: Yep! Or, well, "insert a dump" more like; I doubt there's much pdb use for this situation. Generally when an executor's underlying library causes a failure, the developer sees "test_foo" failed, and the devs' first question is, "well, what is Thunder sending to my executor?"

tfogal:
I encourage discussion to continue here, but as per Luca's guidance I am going to close as I do not intend to make time for a 2nd spin in the next ~week.

thunder/benchmarks/targets.py
IvanYashchuk: I see. To make it universally available for all tests and interactive scripts I think the best place to put this dump would be here:
https://github.com/Lightning-AI/lightning-thunder/blob/9107a3d8e9bb32345ddc5bf1df5883bc8e4ac458/thunder/__init__.py#L595

tfogal: Yeah, that's a *much* better solution! Will update it to this when I can

PR #484, opened 2024-05-29
python executor should be able to handle NumberProxy

jjsjann123:
extend the support from binary op to unary/cast

cherry-picked small fixes from #250


jjsjann123:
@kiya00 is running issue with this in resnet. I'm pulling in the fix to unblock her.

lantiga (approved):
Looks good, stamped!

PR #488, opened 2024-05-30
fix a comment--never closed parenthesis

tfogal:


crcrpar (approved):
would you happen to have something in your mind to keep this as draft?

tfogal:
> would you happen to have something in your mind to keep this as draft?

i just try to make a habit to keep things as draft until CI passes, so that I don't ping/bother people incessantly before I've figured out if things even make sense.

given that this change is so tiny that logic probably doesn't make sense, but 🤷 habit's habit.

t-vi (approved):
Thank you @tfogal  @crcrpar

PR #479, opened 2024-05-29
Update split QKV + RoPE benchmark

IvanYashchuk:
This PR splits `test_llama2_qkv_split_rope_7b_train` into separate forward and backward benchmarks and adds a parametrization over litgpt configs with smaller batch size than before (it was 32, now it's 1, 2, ...).

I hope this will help simplify analysis for this portion of the network (ref https://github.com/Lightning-AI/lightning-thunder/issues/256).

Example runs for Llama 2 7B on DGX-H100:
```py
pytest thunder/benchmarks/targets.py -k "test_litgpt_qkv_split_rope_train_forward[Llama-2-7b-hf" --benchmark-group-by='param:config,param:bs' --benchmark-columns='min,max,mean,stddev,median'
------------------------------------------------------------------------ benchmark 'config=Llama-2-7b-hf bs=1': 6 tests -----------------------------------------------------------------------
Name (time in us)                                                                                  Min                 Max                Mean             StdDev              Median          
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_litgpt_qkv_split_rope_train_forward[Llama-2-7b-hf-bs1-torch.compile]                      93.1070 (1.0)      149.2300 (1.0)       97.1205 (1.0)       9.4027 (1.0)       94.9830 (1.0)    
test_litgpt_qkv_split_rope_train_forward[Llama-2-7b-hf-bs1-torch+apex]                        121.2551 (1.30)     400.8880 (2.69)     131.5780 (1.35)     44.6493 (4.75)     122.5070 (1.29)   
test_litgpt_qkv_split_rope_train_forward[Llama-2-7b-hf-bs1-thunder+nvfuser+torch.compile]     135.2079 (1.45)     298.7320 (2.00)     147.0512 (1.51)     26.1107 (2.78)     139.8915 (1.47)   
test_litgpt_qkv_split_rope_train_forward[Llama-2-7b-hf-bs1-torch.compile+apex]                176.3180 (1.89)     271.9470 (1.82)     186.1000 (1.92)     15.3009 (1.63)     182.9585 (1.93)   
test_litgpt_qkv_split_rope_train_forward[Llama-2-7b-hf-bs1-thunder]                           368.9330 (3.96)     782.9840 (5.25)     404.0743 (4.16)     68.9241 (7.33)     382.7575 (4.03)   
[...]

PR #489, opened 2024-05-30
Add "important configs" to the benchmarks/targets.py as the default list for parametrized benchmarks

IvanYashchuk:
This PR adds a new global variable in the script to define the set of important configs to track by default. Optionally it's possible to turn on all other configs with `THUNDER_BENCH_RUN_ALL_CONFIGS=1`.

This script is used in our nightly pipelines to track performance regressions. It takes about 12 minutes to run all benchmarks for all different models. It's probably not needed to run all of the configs nightly, only some of the most important ones.

A follow-up to https://github.com/Lightning-AI/lightning-thunder/pull/479.

cc @crcrpar

### DIFF
--- a/thunder/benchmarks/targets.py
+++ b/thunder/benchmarks/targets.py
@@ -560,13 +560,7 @@ sdpa_executors_ids = (
 )
 @pytest.mark.parametrize(
     "config,",
-    (
-        "Llama-2-7b-hf",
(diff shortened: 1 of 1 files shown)


lantiga (approved):
Looks good to me, let's land this after #493 lands

PR #491, opened 2024-05-30
Use PyTorch Benchmarking Timer by default with pytest-benchmark

IvanYashchuk:
Let's change the default pytest-benchmark timer to use the one from PyTorch benchmarking utilities. We don't need "wrap_for_benchmark" with explicit CUDA device sync anymore when using the PyTorch timer.
This PR also forces pytest-benchmark to always use warmup by default.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,6 +31,7 @@ addopts = [
     "--color=yes",
     "--disable-pytest-warnings",
     "--benchmark-timer=torch.utils.benchmark.utils.timer.timer",
+    "--benchmark-warmup=on",
 ]
 markers = [
     "standalone: mark a test as standalone",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,6 +30,7 @@ addopts = [
     "--doctest-modules",
     "--color=yes",
     "--disable-pytest-warnings",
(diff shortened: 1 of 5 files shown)


IvanYashchuk:
Real CI failure: 
```
ERROR: usage: __main__.py [options] [file_or_dir] [file_or_dir] [...]
__main__.py: error: unrecognized arguments: --benchmark-timer=torch.utils.benchmark.utils.timer.timer --benchmark-warmup=on
```

lantiga (approved):
LGTM!

crcrpar:
just fyi: with this pr merged, more specifically the change in `.pyproject.toml`, it seems that `pytest-benchmark` is a requirement even for a test, not a benchmark. It's listed in https://github.com/Lightning-AI/lightning-thunder/blob/0c6e38cf9a67aba5210d37fa6235a91efad069ca/requirements/test.txt#L3 though

IvanYashchuk:
Yes, pytest complains about these new default options when pytest-benchmark is not installed even for regular tests.

t-vi:
PyTest-Benchmark 4.0 is even in the stock Debian, so it seems like an easy dependency.
PyTest 8.2 seems not compatible with some of our test magic, though.

PR #493, opened 2024-05-30
ci fix: use torch.compiler.is_compiling

kshitij12345:
Checking if this fixes the nightly CI


lantiga:
Thank you @kshitij12345, beat me to it

lantiga:
Builds on PyTorch stable are rightfully failing :-) 
We need to condition with the PyTorch version, or (maybe quicker) by checking whether `torch.dynamo` as the `is_compiling` attribute.

lantiga (approved):

PR #494, opened 2024-05-30
Let pytest-benchmark figure out required number of rounds

IvanYashchuk:
**Stacked upon https://github.com/Lightning-AI/lightning-thunder/pull/491**

Instead of using the prescribed number of iterations let pytest-benchmark choose the appropriate number of rounds according to https://pytest-benchmark.readthedocs.io/en/latest/calibration.html

cc @crcrpar


t-vi (approved):
Thank you @IvanYashchuk

PR #495, opened 2024-05-30
Add `div` operator to python executor

rdspring1:
I observed `RuntimeError: Failed to find an executor for bound symbol bsym=f31 = prims.div(f3, 16.0)  # f31: "float 0.03125"` in #260. This PR fixes this issue by adding `div` operator to the python executor.

### DIFF
--- a/thunder/executors/pythonex.py
+++ b/thunder/executors/pythonex.py
@@ -202,10 +202,18 @@ def _elementwise_unary_checker(x: NumberLike | TensorProxy) -> bool:
     return isinstance(x, (Number, NumberProxy))
 
 
-# Maps exact inputs to truncation division
-def _div_prim_impl(a: Number, b: Number) -> torch.Tensor:
-    if isinstance(a, (int, bool)) and isinstance(b, (int, bool)):
-        return a // b
+def _div_prim_impl(a: Number, b: Number) -> Number:
+    if dtypes.is_exact_dtype(type(a)) and dtypes.is_exact_dtype(type(b)):
+        if (a >= 0) != (b >= 0) and a % b:
+            # This implementation follows c-style integer division, which is
+            # truncation division. When the quotient is negative and not evenly
+            # divisible, special handling is necessary to round values to zero.
+            return a // b + 1
+        else:
+            # Python uses floor division for integers, which rounds values to
+            # negative infinity.
+            return a // b
+
     return a / b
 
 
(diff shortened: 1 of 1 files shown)


rdspring1:
Do we want the python executor operator to have the same implementation as the prim number function?

See https://github.com/Lightning-AI/lightning-thunder/blob/main/thunder/core/prims.py#L2345-L2353

lantiga:
It should have the same semantics IMO

thunder/executors/pythonex.py
```diff
@@ -202,6 +202,13 @@ def _elementwise_unary_checker(x: NumberLike | TensorProxy) -> bool:
     return isinstance(x, (Number, NumberProxy))
 
 
+# Maps exact inputs to truncation division
+def _div_prim_impl(a: Number, b: Number) -> torch.Tensor:
+    if isinstance(a, (int, bool)) and isinstance(b, (int, bool)):
+        return a // b
```
mruberry, line 208: c-style integer division is truncation division, not floor division, so I think the implementation needs to be like this:

```
def _div_numbers(a: Number, b: Number) -> Number:
    if dtypes.is_exact_dtype(type(a)) and dtypes.is_exact_dtype(type(b)):
        # Accounts for rounding towards zero instead of flooring
        if (a >= 0) != (b >= 0) and a % b:
            return a // b + 1
        else:
            return a // b

    return a / b
```

mruberry (commented):
Thanks @rdspring1! I made a note about the numerics. I think the different kinds of division are really interesting!

thunder/executors/pythonex.py
rdspring1: I updated the python executor implementation to match the prim number function.

IvanYashchuk: @mruberry, why do we need to follow C-style division? There's going to be a discrepancy in the results computed from the Thunder-jitted function and its Python original.

mruberry: The Python "division" operation is true division and we map it to the appropriate primitive here:

https://github.com/Lightning-AI/lightning-thunder/blob/63b01d58af1e7b11acb8e60ad1694e59814eb90e/thunder/core/proxies.py#L721

The `div` prim implements C-style division. We can rename it to `cdiv` if that would be clearer. If we do that, we probably shouldn't name a primitive `div`, just as Python no longer recommends setting `__div__` in favor of the explicit true division and floor division dunders.

mruberry (approved):
Cool! It's true that the name `div` for a primitive is confusing (what kind of division are we referring to?) so maybe a separate PR should rename `div` to be `cdiv`

PR #496, opened 2024-05-31
drop standalone testing in favour of plain pytest-3 for distributed

t-vi:
cc @borda


lantiga:
Didn’t we do this to avoid resource exhaustion on the CI machines? Let’s see how this looks like on CI

t-vi:
We don't run anything in parallel in the distributed, so I think it should be fine.

lantiga (approved):

PR #509, opened 2024-06-03
Remove trace dump for debug from test_tensor_parallel.py

crcrpar:


t-vi (approved):
Thanks for cleaning this up and I'm glad I'm not the only one who has that from time to time. :)

Issue #499, opened 2024-05-31
TransformerEngine API changed and caused test failure `AttributeError: 'TELinear' object has no attribute 'fp8_weight_shapes'`


xwang233:
## 🐛 Bug

Unit test failures:

```
00:42:57 FAILED thunder/tests/test_transformer_engine_executor.py::test_te_linear_forward_backward[None] - AttributeError: 'TELinear' object has no attribute 'fp8_weight_shapes'
00:42:57 FAILED thunder/tests/test_transformer_engine_executor.py::test_te_linear_forward_backward[fp8_recipe1] - AttributeError: 'TELinear' object has no attribute 'fp8_weight_shapes'
00:42:57 FAILED thunder/tests/test_transformer_engine_executor.py::test_te_linear_forward_backward_multiple_iteration[None] - AttributeError: 'TELinear' object has no attribute 'fp8_weight_shapes'
00:42:57 FAILED thunder/tests/test_transformer_engine_executor.py::test_te_linear_forward_backward_multiple_iteration[fp8_recipe1] - AttributeError: 'TELinear' object has no attribute 'fp8_weight_shapes'
00:42:57 FAILED thunder/tests/test_transformer_engine_executor.py::test_te_with_autocast - AttributeError: 'TELinear' object has no attribute 'fp8_weight_shapes'
```

### To Reproduce

thunder unit tests

#### Code sample

thunder unit tests

### Expected behavior

test pass

### Environment

pjnl-20240531

- torch_commit: https://github.com/pytorch/pytorch/commit/705346bf8db09175f48b2c8323d29e46c5f7c58b
- thunder_commit: https://github.com/Lightning-AI/lightning-thunder/commit/27158e62a19e144a3081be9507c81084b702c58e
- transformer_engine_commit: https://github.com/NVIDIA/TransformerEngine/commit/e8a17d1e8241536b4717f273e77ec4a59b649581
- nvfuser_commit: https://github.com/nvidia/fuser/commit/2310aec5c49aef3ee1806ac2b33b5d28dc27aedd

### Additional context

N/A

cc @tfogal @kshitij12345

Issue #196, opened 2024-04-16
[ci] : We should add a CI flow with TransformerEngine installed so that we can run the relevant tests.


kshitij12345:
We don't have a flow for testing TransformerEngine (TE) executor. 

It would be great to have a CI flow with TE installed to be able to run relevant tests so that we can catch the breakages early. It can also be enabled only before merge or with a github comment.

**NOTE**: TE needs to be built from source at the moment. (@xwang233 knows how to set this up in docker).

cc: @Borda

Borda:
@kshitij12345 could you please share a reference to TE and eventually how it needs to be installed?

IvanYashchuk:
Here are the instructions: https://docs.nvidia.com/deeplearning/transformer-engine/user-guide/installation.html#installation-from-source

Borda:
ok, will check it later this week

kshitij12345:
As highlighted by @carmocca (thanks!), our CIs run on 3090 which don't have FP8 support. We should pursue this once the CI have GPUs with compute capability 8.9 or higher.

GPU Compute Capability Ref: https://developer.nvidia.com/cuda-gpus
TE Check for FP8 support: https://github.com/NVIDIA/TransformerEngine/blob/f85553ea369da15fd726ab279818e415be48a228/transformer_engine/pytorch/fp8.py#L23-L34

Borda:
Cc @t-vi @lantiga so far so I know, there is plan to change used GPU for CI

lantiga:
#209 for reference

IvanYashchuk:
TransformerEngine is now available in PyPI: https://pypi.org/project/transformer-engine/:
`pip install transformer_engine[pytorch]`

kshitij12345:
Wanted to check if upgrading the GPUs in CI is still being planned ([ref](https://github.com/Lightning-AI/lightning-thunder/issues/196#issuecomment-2077605099)). This would help us run TE tests in CI to avoid silent regressions.

Example - https://github.com/Lightning-AI/lightning-thunder/issues/1624, https://github.com/Lightning-AI/lightning-thunder/pull/1626

kshitij12345:
Another example - https://github.com/Lightning-AI/lightning-thunder/pull/1690

t-vi:
The most likely scenario to my mind would be to run the TE tests on an ADA GPU via the lightning sdk.
(@lantiga)

kshitij12345:
Any update on this? 

[...]

PR #515, opened 2024-06-03
TE: Support v1.8 - current main

kshitij12345:
Fixes https://github.com/Lightning-AI/lightning-thunder/issues/499

Verified that the tests pass on v1.6 (stable), v1.7 and v1.8 (current main).

Tested with `test_transformer_engine_executor.py` and `test_ddp.py -k transformer`


kshitij12345:
cc: @IvanYashchuk

t-vi (approved):
Thank you @kshitij12345

kshitij12345:
> We don't need to do this in this PR, but after merging, should we bump the -latest to TE 1.8?

The CI doesn't have a job for TE - it is tracked in https://github.com/Lightning-AI/lightning-thunder/issues/196

Issue #471, opened 2024-05-29
Dynamic shape needs to be modeled in trace


jjsjann123:
## 🐛 Bug

dynamic shape computation inside prim_meta is opaque in trace. (see [comment](https://github.com/Lightning-AI/lightning-thunder/issues/262#issuecomment-2127729530))

This would give us a `TensorProxy` with a shape which contains magic `NumberProxy` without its operations.
e.g. the output TensorProxy below has shape `[8, 16, i42]`. However, the trace does not model how i42 is produced.
`# t43 = prims.slice_prim(t_0, [0, 0, i9], [8, 16, 32], [1, 1, 1])  # t43: "cuda:0 f32[8, 16, [IntegerProxy name=i42, value=22]]"`

Currently this triggers an assert in unpack_inputs during prologue trace construction. commit 775aa7405b8f0e9539326a069522d247c6363cd8
```
import thunder

def foo(a, dim0, dim1):
    dim2 = a.size(2) - dim0 - dim1
    x = torch.split(a, (dim0, dim1, dim2), dim=2)
    dim3 = x[2].size(2) + dim2
    return x, dim3

import torch

dtype = torch.float32

size0 = 8
size1 = 16
size2 = 32
a = torch.randn(size0, size1, size2, device="cuda").to(dtype=dtype)

dim0=2
dim1=8

a_ref = a.detach()
out_ref = foo(a_ref, dim0, dim1)

jfoo = thunder.jit(foo)

out = jfoo(a, dim0, dim1)
```

hits an error:
```
Traceback (most recent call last):
  File "/volume/thunder_scalar.py", line 35, in <module>
    out = jfoo(a, dim0, dim1)
  File "/opt/pytorch/lightning-thunder/thunder/__init__.py", line 617, in fn_
    cache_entry, inps, pro_to_epi = get_computation_and_inputs(*args, **kwargs)
  File "/opt/pytorch/lightning-thunder/thunder/__init__.py", line 202, in cache_info_wrapper
    res = fn(*args, **kwargs)
  File "/opt/pytorch/lightning-thunder/thunder/__init__.py", line 473, in get_computation_and_inputs
    jit_results: TraceResults = interpreter(
  File "/opt/pytorch/lightning-thunder/thunder/__init__.py", line 190, in _general_frontend
    return thunder_general_jit(fn, args, kwargs, sharp_edges=sharp_edges, record_history=record_history)
[...]

Issue #262, opened 2024-04-24
Dynamic constraints and NumberProxies


jjsjann123:
## 🚀 Feature

We'd like to have thunder.jit support dynamic constraints so we are not going to bake every number in a program as compile time constant. This should allow us re-use some compiled program to avoid endless recompilation with dynamic shape.

### Pitch

We'd want to:
- [x] keep Scalar input (including tensor properties like shape) as NumberProxy in the trace and feed them into compute trace as inputs.
- [x] [Partially Finished] have prologue trace to insert checks to ensure program safety.

prototyping PRs:

- [x] removing Number inheritance from derivatives of NumberProxy #286 
- [x] allow `symbolic values` cache option: #518 
- [ ] ~`symbolic values` cache option by default: #250~ shelving for priority adjustment

Issues:

- [x] number proxy is no number: #272
- [ ] ~executor specific caching rule: #263~ crossing out since this is not necessary
- [x] prim should be able to insert constraints to bake in static numbers: #463
- [ ] dynamic shape needs to be modeled in trace: #471 
- [ ] ~utils.check on NumberProxy needs to be sanitized in prim/clang/torch.~ crossing out since we might not need it in the short term. Will still follow up with a separate issue.
- [x] NumberProxy handling in grad transform: #541 (will be addressed in #244)
- [x] NumberProxy inconsistency is introduced by grad transform #541

### Progress

~NumberProxy inheritance PR merged #286.~

~Currently working on enabling caching option `symbolic values` to allow trace handling dynamic scalar inputs (WIP in PR #250). This is going to be the next milestone.~ shelving this milestone for now. Since enabling symbolic values as the default cache policy cannot be easily done without dynamic shape support in TensorProxy. Which is our immediate goal.

jjsjann123:
As my current prototype in #250, I realized that `torch/__init__.py` has a few complexities:

1. value based control flow: https://github.com/Lightning-AI/lightning-thunder/blob/82185e3a55d5b3f0bea8a7366d74a275dbe34acd/thunder/torch/__init__.py#L953
2. there's also semantic checks: https://github.com/Lightning-AI/lightning-thunder/blob/82185e3a55d5b3f0bea8a7366d74a275dbe34acd/thunder/torch/__init__.py#L752-L753

[AIs] We need to allow language-based constraints to be injected and placed in prologue trace.

[...]

PR #518, opened 2024-06-04
Enable symbolic values

jjsjann123:
## What does this PR do?

* Removing assert on `symbolic values` as the cache option for general jit. We'll keep NumberProxy as dynamic input to the trace and accept runtime value for computation.
* Rejecting reduction for nvfuser when reduction axes is dynamic.
* Adding python test to verify cache hit with changing runtime value.

### Note:

There are lots of sharp corners with NumberProxy. It's our goal to make `symbolic values` as the default cache option. See the umbrella issue #262 
Dynamic shaped tensor is not enabled in this PR. We are baking all tensor shapes in as static. This issue is tracked in #471

### DIFF
--- a/thunder/core/jit_ext.py
+++ b/thunder/core/jit_ext.py
@@ -615,7 +615,6 @@ class GeneralJitCtx(MinimalCtx):
                 else:
                     self.add_constraint((clang.check_number_type_and_value, p, uvalue))
             elif co is CACHE_OPTIONS.SYMBOLIC_VALUES:
-                # TODO: establish guarding logic
                 if p is not uvalue:
                     value.register_proxy(p)
             elif co not in (CACHE_OPTIONS.SAME_INPUT, CACHE_OPTIONS.NO_CACHING):
@@ -1584,7 +1583,6 @@ def thunder_general_jit(
         record_history=record_history,
     )
 
-    # NOTE(jiej): numbers are baked in as constant here vvv
     with general_jit_ctx(ctx):
         with tracectx(computation_trace):
             result = jfn(*args, **kwargs)
@@ -1620,7 +1618,6 @@ def thunder_general_jit(
     else:
         epilogue_trace = None
 
-    # NOTE(jiej): prologue trace is produced here vvv
     pro_to_comp_proxies, pro_to_epi_proxies = unpack_inputs(
         ctx, prologue_trace, pro_to_comp, pro_to_epi, args, kwargs, has_epilogue=epilogue_trace is not None
     )
--- a/thunder/core/jit_ext.py
+++ b/thunder/core/jit_ext.py
@@ -20,8 +20,7 @@ from io import StringIO
 import inspect
 import time
 
-from thunder.core.compile_data import compile_data_and_stats, get_cache_option, using_symbolic_values, get_compile_data
-from thunder.core.symbol import bsym_header
+from thunder.core.compile_data import compile_data_and_stats, get_cache_option, get_compile_data
 import thunder.clang as clang
 import thunder.core.transforms
 
@@ -595,6 +594,9 @@ class GeneralJitCtx(MinimalCtx):
             co: CACHE_OPTIONS = get_cache_option()
             if co is CACHE_OPTIONS.CONSTANT_VALUES:
                 self.add_constraint((clang.check_tensor_shape_and_metadata, p_orig))
+            elif co is CACHE_OPTIONS.SYMBOLIC_VALUES:
+                # TODO: establish guarding logic to allow non-broadcast shape change
+                self.add_constraint((clang.check_tensor_shape_and_metadata, p_orig))
             elif co not in (CACHE_OPTIONS.SAME_INPUT, CACHE_OPTIONS.NO_CACHING):
                 raise NotImplementedError(f"Unsupported cache option {co}")
             return p
@@ -612,6 +614,10 @@ class GeneralJitCtx(MinimalCtx):
                     self.add_constraint((clang.check_string_value, p, uvalue))
                 else:
                     self.add_constraint((clang.check_number_type_and_value, p, uvalue))
+            elif co is CACHE_OPTIONS.SYMBOLIC_VALUES:
+                # TODO: establish guarding logic
+                if p is not uvalue:
+                    value.register_proxy(p)
             elif co not in (CACHE_OPTIONS.SAME_INPUT, CACHE_OPTIONS.NO_CACHING):
                 raise NotImplementedError(f"Unsupported cache option {co}")
             return p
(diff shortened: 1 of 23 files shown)


jjsjann123:
CI cleared here. Let's revert more changes.

thunder/tests/test_examine_memory.py
```diff
@@ -150,7 +150,7 @@ def test_nanogpt_block(executor, device, dtype):
         expected_return_calculated_mem = get_return_memory(fw_extrace.bound_symbols[-1]) - 4 * 2 * 1024 * 768
         assert expected_return_calculated_mem == sum(fw_alloc_mem[1].values())
 
-        assert bw_alloc_mem[0] == 361783296
+        assert bw_alloc_mem[0] == 361881600
```
jjsjann123, line 153: This change is necessary because we are rejecting sum here, since it's taking the symbolic proxy as reduction axes and being rejected by nvfuser.

thunder/core/proxies.py
```diff
@@ -5,6 +5,8 @@
 from typing import Type, Optional, Any, Tuple, List, Union
 from collections.abc import Callable
 from collections.abc import Sequence
+
+# from functools import reduce, partial, wraps
```
lantiga, line 9:
```suggestion
```

thunder/core/proxies.py
```diff
@@ -1041,13 +1047,8 @@ def _infer_tensor_properties(
...
     )
 
-    # Extracts actual values for shape
-    # TODO RC1 Enable this
-    if using_symbolic_values():
-        raise NotImplementedError(
-            f"Trying to construct a tensor proxy while using symbolic values, but this is not yet supported"
-        )
-
+    # dynamic shape not yet enabled, otherwise, the bake in should be guarded.
```
lantiga, line 1050: Can we add a condition that will break when we add dynamic shapes, so we don't forget to guard this when it's time?

lantiga (approved):
LGTM!

thunder/tests/test_examine_memory.py
tfogal: That's a little concerning. What else might we be kicking out of an nvFuser region that previously got inserted?

Your experience here makes me think the old behavior (which AFAIK is baking the value into the graph as a constant and recompiling the graph should the value change) should be the default, and we should use a compile option to turn this on. At least for a little while.

thunder/tests/test_jit_general.py
```diff
@@ -787,6 +787,9 @@ def test_tom_overrides_proxy(device):
     ("cpu", "cuda"),
 )
 def test_cache_symbolic_values(device):
+    if device == "cuda" and not torch.cuda.is_available():
```
tfogal, line 790: nit: i doubt it matters, but might consider testing if "cuda" is a substr of device or something, so that e.g. `cuda:0` still does the right thing

thunder/core/jit_ext.py
```diff
@@ -1546,7 +1551,7 @@ def thunder_general_jit(
         )
 
     co: CACHE_OPTIONS = get_cache_option()
-    if co not in {CACHE_OPTIONS.CONSTANT_VALUES, CACHE_OPTIONS.NO_CACHING}:
+    if co not in {CACHE_OPTIONS.CONSTANT_VALUES, CACHE_OPTIONS.NO_CACHING, CACHE_OPTIONS.SYMBOLIC_VALUES}:
         raise NotImplementedError(f"Only constant constraints is supported")
```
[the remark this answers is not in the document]
t-vi: `SAME_INPUT` means ASSUME `SAME_INPUT`.

thunder/tests/test_jit_general.py
t-vi: but this is from the parametrize in 787...

thunder/core/jit_ext.py
```diff
@@ -612,6 +614,9 @@ def proxify(self, value: WrappedValue) -> Any:
                     self.add_constraint((clang.check_string_value, p, uvalue))
                 else:
                     self.add_constraint((clang.check_number_type_and_value, p, uvalue))
+            elif co is CACHE_OPTIONS.SYMBOLIC_VALUES:
+                if p is not uvalue:
```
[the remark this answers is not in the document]
t-vi: In general, `proxy` might return the original value (e.g. for str even with this patch, I think).

t-vi (approved):
Thank you @jjsjann123 @tfogal @lantiga 

Looks great!

thunder/tests/test_jit_general.py
jjsjann123: That's fair. though we did provide the device string in the parameter above. :) 
`    ("cpu", "cuda"),`

thunder/tests/test_examine_memory.py
jjsjann123: > Your experience here makes me think the old behavior (which AFAIK is baking the value into the graph as a constant and recompiling the graph should the value change) should be the default

That's a reasonable argument. We certainly can leave that as the default.

> and we should use a compile option to turn this on. At least for a little while.

This is effectively done today with `cache` option in `thunder.jit`.

> What else might we be kicking out of an nvFuser region that previously got inserted?

I would argue that the updated nvFuser behavior is the right thing to do. i.e. if we see a NumberProxy on a place where we cannot handle it, we should kick it out. You could argue that nvfuser should have updated its support on that and that's a conversation I'm happy to continue.

thunder/core/jit_ext.py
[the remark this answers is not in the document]
jjsjann123: I think we are saying that we still don't want to allow `SAME_INPUT`.

I'll update the exception string. Thanks to @tfogal for catching this.

thunder/tests/test_examine_memory.py
tfogal: > I would argue that the updated nvFuser behavior is the right thing to do.

Sorry, I didn't mean to imply that it wasn't!

But I also want to make sure we're not going to see a bunch of perf regressions after this PR. Maybe that's not a real concern? If it is, an alternate idea might be to update the lit-gpt benchmarks to set the cache option to the old behavior

jjsjann123: > But I also want to make sure we're not going to see a bunch of perf regressions after this PR. Maybe that's not a real concern?

Right now `NumberProxy` requires explicit `opt-in` via specifying cache policy to `symbolic values`. We are not changing the default behavior so we should be fine.

One exception is that grad transform (as well as other transforms) might introduce `NumberProxy`, but it's not doing that for reduction axes right now. https://github.com/Lightning-AI/lightning-thunder/blob/7ac4c49aca1f4282dc9dfe5ba69e343a358d1942/thunder/core/transforms.py#L1168-L1178

I'll be surprised if reduction axes are given as runtime inputs. Looks at the `_normalize` in `torch/__init__.py`, which only uses tensor rank to compute the reduction axes  https://github.com/Lightning-AI/lightning-thunder/blob/7ac4c49aca1f4282dc9dfe5ba69e343a358d1942/thunder/torch/__init__.py#L2872-L2873

tfogal: Ahh, I had misunderstood then. sounds great, thank you!

PR #520, opened 2024-06-04
[distributed][Tensor Parallelism] Fix output shape of tensor-parallel input/output synchronize

crcrpar:


t-vi (approved):
We really need to test meta more... Thank you @crcrpar

PR #521, opened 2024-06-04
[distributed][Tensor Parallelism] Remove pair of column-wise linear postprocessing and row-wise linear preprocessing

crcrpar:
## What does this PR do?

Automatically finds and removes pairs of postprocessing of column-wise linear and preprocessing of row-wise linear from a computation trace.

If there exists a pair of column-wise linear and row-wise linear, currently a trace would have column-wise linear postprocessing (= AllGather of the outputs in their last dimension) and row-wise linear preprocessing (= Local chunking of the inputs in their last dimension).
If we only have elementwise ops between the two ops, this pair of procesisngs are just redundant.


t-vi (approved):
Cool! Thank you @crcrpar

PR #530, opened 2024-06-05
add normalize

### LINKED ISSUE #334: Implement torch.nn.functional.normalize
### 🚀 Feature

Implement [`torch.nn.functional.normalize`](https://pytorch.org/docs/stable/generated/torch.nn.functional.normalize.html)

### Motivation

NeMo's [`MegatronClipModel`](https://github.com/NVIDIA/NeMo/blob/main/examples/multimodal/vision_language_foundation/clip/megatron_clip_pretrain.py)

cc @tfogal

k223kim:
<details>

</details>

## What does this PR do?

Fixes #334 .

## PR review

## Did you have fun?

⚡️


thunder/torch/__init__.py
```diff
@@ -2842,6 +2842,19 @@ def _normalize(a: TensorProxy, /, norm_dims, eps: Number) -> tuple[TensorLike, T
     return out, mean, rstd
 
 
+@torchsymbol(torch.nn.functional.normalize, is_method=True)
+def normalize(
+    a: TensorProxy, /, p: float = 2, dim: int | Sequence[int] = 1, eps: float = 1e-12, out: None | TensorProxy = None
+) -> TensorProxy:
+    utils.check(out is None, lambda: "normalize: out is not None which is currently unsupported", NotImplementedError)
+    denom = pow(a, p)
```
t-vi, line 2850: This only works for finite p, PyTorch also takes inf. Could you add `p == math.inf` + tests for maybe p=1, 2, 4, inf , please?

t-vi (commented):
Great job, there is a corner case in p=inf that might be worth handling.

thunder/torch/__init__.py
```diff
@@ -2842,6 +2842,19 @@ def _normalize(a: TensorProxy, /, norm_dims, eps: Number) -> tuple[TensorLike, T
     return out, mean, rstd
 
 
+@torchsymbol(torch.nn.functional.normalize, is_method=True)
+def normalize(
+    a: TensorProxy, /, p: float = 2, dim: int | Sequence[int] = 1, eps: float = 1e-12, out: None | TensorProxy = None
```
mruberry, line 2847: nit: `p: float = 2.` so that the default for `p` is actually a float

thunder/torch/__init__.py
mruberry: nit: `a**p` instead of `pow` is probably more pythonic

thunder/tests/opinfos.py
```diff
@@ -5873,6 +5873,37 @@ def tensor_constructor_error_generator(op, device, **kwargs):
 linear_algebra_ops = []
 
 
+def normalize_sample_generator(op, device, dtype, requires_grad, **kwargs):
+    make = partial(make_tensor, device=device, dtype=dtype, requires_grad=requires_grad)
+    # input shape
+    cases = (
+        (4, 4),
+        (1024, 1024),
```
mruberry, line 5881: This is a great case, but probably too big for a "sample" (we have a separate concept of "reference inputs" for larger inputs, for example:

https://github.com/Lightning-AI/lightning-thunder/blob/067f15aae47ad71229732ca6c35a5d190135e48c/thunder/tests/opinfos.py#L7017

)

The "samples" are typically smaller inputs so they can be used in more tests without taking too much CI time. Maybe `(32, 32)` would be OK?

thunder/tests/opinfos.py
```diff
@@ -5873,6 +5873,37 @@ def tensor_constructor_error_generator(op, device, **kwargs):
...
+        yield SampleInput(make(case))
+
+
+normalize_opinfo = OpInfo(
+    ltorch.normalize,
+    sample_input_generator=normalize_sample_generator,
+    torch_reference=torch.nn.functional.normalize,
+    dtypes=(datatypes.floating, datatypes.complexfloating),
+    test_directives=(
+        # The low precision floating point types sometimes fail
```
mruberry, line 5895: Nice special case

mruberry:
Awesome start, @k223kim!

thunder/torch/__init__.py
k223kim: Oh, I wasn't aware of `inf`. Thanks for catching! Will test these!
+ Thanks for suggesting a pythonic approach Mark!

thunder/tests/opinfos.py
k223kim: Yeah absolutely! Thanks for explaining the differences between reference inputs and samples.

thunder/torch/__init__.py
k223kim: Oh, my bad. Thanks for catching this! 🙌

t-vi:
I think the failing checks are (mostly) real.

k223kim:
@t-vi Yes, sorry. I think there is an issue with my implementation. Let me take a closer look and update you on this!

k223kim:
Hi @t-vi Tom! Just updated my code with the requested test cases. I will add another PR to handle different `dims`! (cc. @mruberry )

thunder/torch/__init__.py
```diff
@@ -2845,6 +2845,39 @@ def _normalize(a: TensorProxy, /, norm_dims, eps: Number) -> tuple[TensorLike, T
...
+        dim = utils.canonicalize_dim(a.ndim, dim)
+        a_ = clang.maybe_convert_to_dtype(a, computation_dtype)
+        is_p_even = p % 2.0 == 0
+        if not (is_p_even and dtypes.is_float_dtype(a.dtype)):
+            a_ = abs(a_)
+        denom = a_**p
+        denom = sum(denom, dim=dim, keepdim=True)
+        denom = denom ** (1.0 / p)
+        denom = clamp(denom, min=eps)
+        denom = expand_as(denom, a)
```
t-vi, line 2875: I think the clamp and the expand_as should be part of the common path (i.e. outside the else).

thunder/torch/__init__.py
```diff
@@ -2845,6 +2845,39 @@ def _normalize(a: TensorProxy, /, norm_dims, eps: Number) -> tuple[TensorLike, T
...
+) -> TensorProxy:
+    utils.check(
+        dtypes.is_float_dtype(a.dtype) or dtypes.is_complex_dtype(a.dtype),
+        lambda: f"normalize: Expected a floating point or complext tensor as input. Got {a.dtype}",
+        TypeError,
+    )
+    utils.check(out is None, lambda: "normalize: out is not None which is currently unsupported", NotImplementedError)
+    computation_dtype, result_dtype = _reduction_dtypes(a, REDUCTION_OUTPUT_TYPE_KIND.COMPLEX_TO_FLOAT, a.dtype)
+    if p == 0.0:
+        return a / sum(ne(a, 0.0), dim=dim, keepdim=True)
```
t-vi, line 2860: Would this work?
```suggestion
        denom = sum(a != 0.0, dim=dim, keepdim=True)
```

t-vi (approved):
Two tiny nits and we're ready to go. :)

thunder/torch/__init__.py
k223kim: Hmm after thinking about this again, you are right. Let me pull those out to the common path! Thank you!

thunder/torch/__init__.py
k223kim: Yup! I don't see any problem with this. It's also more consistent with the other part so let me accept this! Thank you again!

t-vi:
One last thing: I think we want to increase the allowed difference to `rtol=1e-5` or test with a larger eps. Normalization can be expected to be numerically touchy.
```
FAILED thunder/tests/test_grad.py::test_vjp_correctness_normalize_nvfuser_cuda_float64 - AssertionError: Scalars are not close!

Expected 13.858696783947043 but got 13.858595215142994.
Absolute difference: 0.00010156880404821322 (up to 1e-05 allowed)
Relative difference: 7.328885654375779e-06 (up to 1.3e-06 allowed)
```

thunder/tests/opinfos.py
```diff
@@ -5853,6 +5853,41 @@ def tensor_constructor_error_generator(op, device, **kwargs):
...
+        DecorateInfo(
+            custom_comparator(partial(assert_close, atol=1e-3, rtol=1e-1)),
+            "test_core_vs_torch_consistency",
+            dtypes=(datatypes.bfloat16, datatypes.float16),
+            devicetypes=(devices.DeviceType.CPU, devices.DeviceType.CUDA),
+        ),
```
t-vi, line 5885:
```suggestion
        ),
        DecorateInfo(
            custom_comparator(partial(assert_close, atol=1e-4, rtol=1e-5)),
            "test_vjp_correctness",
        ),
```

thunder/torch/__init__.py
```diff
@@ -2844,6 +2844,39 @@ def _normalize(a: TensorProxy, /, norm_dims, eps: Number) -> tuple[TensorLike, T
...
+def normalize(
+    a: TensorProxy, /, p: float = 2.0, dim: int | Sequence[int] = 1, eps: float = 1e-12, out: None | TensorProxy = None
+) -> TensorProxy:
+    utils.check(
+        dtypes.is_float_dtype(a.dtype) or dtypes.is_complex_dtype(a.dtype),
+        lambda: f"normalize: Expected a floating point or complext tensor as input. Got {a.dtype}",
+        TypeError,
+    )
+    utils.check(out is None, lambda: "normalize: out is not None which is currently unsupported", NotImplementedError)
+    computation_dtype, result_dtype = _reduction_dtypes(a, REDUCTION_OUTPUT_TYPE_KIND.COMPLEX_TO_FLOAT, a.dtype)
```
mruberry, line 2857: Is this really complex to float? I thought normalizing a complex tensor also produced a complex tensor?

k223kim: Yeah, normalizing a complex tensor produces a complex tensor. And the current implementation of thunder.normalize also outputs a complex due to:
```
out = a / denom
```
Most of the implementation was inspired by `vector_norm` from PyTorch which I expect should be implemented in Thunder. However, I can get rid of this for now as it does not have much effect. What do you think?

k223kim: Eventually to handle different `dim` cases, I think I should implement `vector_norm` and `matrix_norm` like PyTorch. I can do this in the next PR if it is an acceptable approach!

mruberry: `vector_norm` and `matrix_norm` would be great to implement! I see what you mean now about the dtype stuff with complex. No need to change it!

mruberry:
The test failure is related, the vjp tests just need their tolerance tightened:

```
FAILED thunder/tests/test_grad.py::test_vjp_correctness_normalize_nvfuser_cuda_float64 - AssertionError: Scalars are not close!

Expected 13.261265786378505 but got 13.261199947853306.
Absolute difference: 6.583852519881361e-05 (up to 1e-05 allowed)
Relative difference: 4.964724051186771e-06 (up to 1.3e-06 allowed)
```

Maybe 1e-03 for both values would be OK for now?

t-vi:
> The test failure is related, the vjp tests just need their tolerance tightened:

This needs an update on the vjp test itself to use the tolerance (which is something we tried earlier and didn't work out of the box, so we tried with larger eps)

thunder/torch/__init__.py
```diff
@@ -2844,6 +2844,39 @@ def _normalize(a: TensorProxy, /, norm_dims, eps: Number) -> tuple[TensorLike, T
...
+        denom = amin(abs(a), dim=dim, keepdim=True)
+    else:
+        dim = utils.canonicalize_dim(a.ndim, dim)
+        a_ = clang.maybe_convert_to_dtype(a, computation_dtype)
+        is_p_even = p % 2.0 == 0
+        if not (is_p_even and dtypes.is_float_dtype(a.dtype)):
+            a_ = abs(a_)
+        denom = a_**p
+        denom = sum(denom, dim=dim, keepdim=True)
+        denom = denom ** (1.0 / p)
```
nikitaved, line 2872: Fingers crossed we do not suffer numerical issues :)

thunder/torch/__init__.py
```diff
@@ -2844,6 +2844,39 @@ def _normalize(a: TensorProxy, /, norm_dims, eps: Number) -> tuple[TensorLike, T
...
+    elif p == float("inf"):
+        denom = amax(abs(a), dim=dim, keepdim=True)
+    elif p == -float("inf"):
+        denom = amin(abs(a), dim=dim, keepdim=True)
+    else:
+        dim = utils.canonicalize_dim(a.ndim, dim)
+        a_ = clang.maybe_convert_to_dtype(a, computation_dtype)
+        is_p_even = p % 2.0 == 0
+        if not (is_p_even and dtypes.is_float_dtype(a.dtype)):
+            a_ = abs(a_)
```
nikitaved, line 2869: nit: probably should be `a <- abs(a) if (complex or not is_p_even)`?

thunder/torch/__init__.py
k223kim: Oh, I should handle that. Thanks for catching!

thunder/torch/__init__.py
k223kim: Yeah I can definitely do that! Is there a reason why we would prefer one or the other? Just curious :)

thunder/torch/__init__.py
k223kim: What do you mean by numerical issues? Are you considering cases where `p` is too small or too big?

thunder/torch/__init__.py
nikitaved: Well, because for complex inputs we always have to call abs.

thunder/torch/__init__.py
nikitaved: Well, you can compute these things in a multitude of ways. There is a difference between `|x_i|^p / ||x||_p` and `(x_i / ||x||_p)^p`. The latter could be sometimes better.

nikitaved: Just a generic statement :) It is possible to deploy different scaling techniques to increase numerical stability, but these are data-dependent.

nikitaved: BTW, we could also check what PyTorch is doing and whether it is any sophisticated. If so, we can dispatch there if needed.

thunder/torch/__init__.py
k223kim: Oh I think my question was, aren't the following two the same?
`if (complex or not is_p_even)` and `if not (is_p_even and float)`

thunder/torch/__init__.py
k223kim: Thanks for the detailed comment and suggestion! When observing PyTorch's implementation [here](https://github.com/pytorch/pytorch/blob/f9dae86222aaf15ea085c7774da70781bae46ff9/torch/_refs/linalg/__init__.py#L160C1-L160C114), it doesn't seem it is doing something sophisticated with the `p` value (for PyTorch, it'll be the `ord` value). Let me know if you have a different suggestion regarding this numerical issue.

thunder/torch/__init__.py
k223kim: However, just incase I am missing anything, let me update the code :)

thunder/tests/opinfos.py
```diff
@@ -5886,6 +5886,13 @@ def normalize_sample_generator(op, device, dtype, requires_grad, **kwargs):
             custom_comparator(partial(assert_close, atol=1e-3, rtol=1e-3)),
             "test_vjp_correctness",
         ),
+        # TODO Investigate the low precision difference
+        DecorateInfo(
+            custom_comparator(partial(assert_close, atol=1e-1, rtol=1e-1)),
+            "test_phantom_grad_vs_torch_consistency",
```
t-vi, line 5892: This depends on #598 , right?

k223kim: No, in my understanding, #598 is only related to `test_vjp_correctness` and `test_jvp_correctness`!

k223kim: Well, now the CI test that is failing is related to #598 😅

thunder/torch/__init__.py
nikitaved: Thanks for checking it out! But it was not actionable, just a thought :)

thunder/torch/__init__.py
nikitaved: If you grep over `is_float_type` you could see that it does not contain complex types.

k223kim: Oh, I see the point. Thanks for the detailed explanation. I naively thought `not float` means `complext` 😅 thank you so much for the help!

k223kim:
Hey all! The current CI testing is failing with the following error: 
```
FAILED thunder/tests/test_grad.py::test_vjp_correctness_normalize_torch_cpu_float64 - AssertionError: Scalars are not close!

Expected 13.56249856229506 but got 13.562559772296954.
Absolute difference: 6.121000189374115e-05 (up to 1e-05 allowed)
Relative difference: 4.513180341556706e-06 (up to 1.3e-06 allowed)
```
And I believe this can be resolved by #598! Other than that, I think this PR is good to go! Please let me know if I could improve more :)

PR #536, opened 2024-06-05
Add pad tests

tfogal:
As per #459's review comments.


t-vi (approved):
Who can resist tests. Thank you for following up with this @tfogal

PR #537, opened 2024-06-05
Minor cleanup for PR #260

rdspring1:
This PR is a minor follow-up to #260. It renames `channels` to `class` for `cross_entropy` and `nll_loss` references. It also removes duplicated gradient functions.


t-vi (approved):
Thank you @rdspring1

PR #538, opened 2024-06-05
Added kwargs to bound symbol rhs hash

### LINKED ISSUE #397: CSE pass should work on symbols with keyword arguments
### Motivation

Currently, if kwargs is of nonzero length then all bound symbols are set to be not equal to each other:
https://github.com/Lightning-AI/lightning-thunder/blob/33b36ec33442e4ae921b8913d8bc728d4451b596/thunder/core/symbol.py#L672-L673
This needs to be changed for CSE to work properly.

Here's an example where CSE fails to eliminate computation because there's a keyword argument:
```py
import torch
import thunder

@thunder.jit
def func(x):
    t1 = thunder.prims.var(x, (0, 1), correction=1)
    t2 = thunder.prims.var(x, (0, 1), correction=1)
    t3 = thunder.prims.add(t1, t2)
    return t3

x = torch.randn(512, 512, device="cuda")
out = func(x)
print(thunder.last_traces(func)[-1])
```
execution trace:
```py
@torch.no_grad()
@no_autocast()
def func(x):
  # x: "cuda:0 f32[512, 512]" 
  [t4] = nvFusion0(x)
    # t1 = prims.var(x, (0, 1), correction=1)  # t1: "cuda:0 f32[]"
    # t3 = prims.var(x, (0, 1), correction=1)  # t3: "cuda:0 f32[]"
    # t4 = prims.add(t1, t3)  # t4: "cuda:0 f32[]"
  del x
  return t4
```

One key data structure is a "frozen dict", dictionary that is hashable and immutable, it was added in https://github.com/Lightning-AI/lightning-thunder/commit/388a1a8edca65bc277a5f6b200d6c128556f0823.

cc @apaz-cli

riccardofelluga:
<details>
- [ x] Was this discussed/approved via a Github issue? (no need for typos and docs improvements)
- [ x] Did you write any new necessary tests?
</details>

## What does this PR do?
Fixes #397 by hashing the kwargs of a bound symbol together with args and symbol. 
## PR review
I've refactored the  `BoundSymbolRHS` dataclass in the same way `BoundSymbol` is with the `__hash__` dunder that calls `_hash` cached property. This drops the frozen property that `BoundSymbolRHS` had before but I think this should be addressed in another PR.
However, a problem that was not there before was the wrong elimination of the inputs as they differ only by the output name and they equal in symbol, args and kwargs. As a workaround, I added the `sym.name == 'unpack_trivial'` to identify them, however advice on how to solve this problem differently is very welcome!
I feel that a `BoundSymbol` refactor might soon be needed.
## Did you have fun?
Oh yes! this one was interesting as I think there is still room for improvement for the `BoundSymbol` dataclass. Since we consider the args and kwargs to be immutable we could enforce this contraint even stronger by freezing the dataclass. That would also help with hashing and `==` operator.

### DIFF
--- a/thunder/core/symbol.py
+++ b/thunder/core/symbol.py
@@ -662,7 +662,7 @@ class BoundSymbolRHS:
     @functools.cached_property
     def _hash(self) -> int:
         # TODO: Find a better way to identify inputs by id instead of hash.
-        if self.parent.sym.name == 'unpack_trivial':
+        if self.parent.sym.name == "unpack_trivial":
             return id(self)
         try:
             return hash((self.parent.sym, self.parent._var_args, self._frozen_kwargs))
@@ -677,4 +677,8 @@ class BoundSymbolRHS:
             return False
         if self.parent is other.parent:
             return True
-        return (self.parent.sym, self.parent._var_args, self._frozen_kwargs) == (other.parent.sym, other.parent._var_args, other._frozen_kwargs)
+        return (self.parent.sym, self.parent._var_args, self._frozen_kwargs) == (
+            other.parent.sym,
+            other.parent._var_args,
+            other._frozen_kwargs,
+        )
--- a/thunder/core/symbol.py
+++ b/thunder/core/symbol.py
@@ -21,6 +21,7 @@ from thunder.core.pytree import tree_flatten, tree_unflatten, tree_map
 import thunder.core.dtypes as dtypes
 import thunder.core.devices as devices
 from thunder.core.proxies import Proxy, NumberProxy, variableify, CollectionProxy
+from thunder.core.utils import FrozenDict
 
 from thunder.core.trace import (
     get_tracectx,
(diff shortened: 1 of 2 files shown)


t-vi (approved):
Awesome! Thank you @riccardofelluga

IvanYashchuk:
> However, a problem that was not there before was the wrong elimination of the inputs as they differ only by the output name and they equal in symbol, args and kwargs. As a workaround, I added the sym.name == 'unpack_trivial' to identify them

Could you please elaborate and provide a code example? There's a special tag for DCE to skip things, we should maybe add the same for CSE. Please create an issue for this.

IvanYashchuk:
> I've refactored the BoundSymbolRHS dataclass in the same way BoundSymbol is with the __hash__ dunder that calls _hash cached property. This drops the frozen property that BoundSymbolRHS had before but I think this should be addressed in another PR.

BoundSymbol is not the best example of data classes with custom hashing. BoundSymbolRHS should be frozen as before this PR. There's a special `field` annotation (https://docs.python.org/3/library/dataclasses.html#dataclasses.field) to specify which fields should participate in hashing.

IvanYashchuk:
> I feel that a BoundSymbol refactor might soon be needed.

What do you have in mind? Put your ideas into an issue

Issue #533, opened 2024-06-05
Add `post_optimization_transform` to `thunder.jit`


kshitij12345:
Support to add transforms which will be applied after the `transform_for_execution/optimization` pass is done. This will help enable transforms to help with profiling (like adding nvtx markers on the execution trace), adding debug prints or checks, etc

One question, what should be the signature of this transform?
Option 1: `def transform(trace)` - Takes single trace at a time and it is passed `forward` and `backward` computation trace one by one.
Option 2: `def transform(fwd_trace, bwd_trace)` - Takes both `forward` and `backward` traces (`backward` trace could be None) and applies the transform to both. This would allow the transform to consider both traces if it wants to or apply independently.

cc: @IvanYashchuk @t-vi

crcrpar:
I'd suggest defining interfaces for both cases so that thunder can tell which case a user wants. An example would look as follows 

```
class AdditionalTransform(ABC):
    @abstractmethod
    def __call__(self, trace: TraceCtx) -> TraceCtx:
        ...

class FwdBwdAdditionalTransform(ABC):
    @abstractmethod
    def __call__(self, fwd_trace: TraceCtx, bwd_trace: TraceCtx) -> tuple[TraceCtx, TraceCtx]:
        ...
```

We could define another for the existing `add_transform` as well

IvanYashchuk:
> One question, what should be the signature of this transform?

My preference is option 1 because I'm not convinced that backward trace would continue to exist in the current form. I plan to change it and having `transform(trace)` would help me isolate the changes.

Issue #483, opened 2024-05-29
load/save_state_dict hooks for early transforms


t-vi:
Currently, we do a number of things in the distributed early transforms:

- instantiate things on the meta device on the target,
- shard parameters etc. into overrides,
- register early_transform to process the traces.

Looking at the module lifetime, it would be awesome to have a mechanism load/save state dicts targeted at the original model.
Similar to the transformations, this should be composable across several transforms.
For efficiency reasons, this should be able to process one parameter (or those of one submodule?) at a time.

This would likely also help with the memory issue #478 because we could then leave things on meta until we need it (at the end of the transforms) if we need several steps to merge things.

So the goal would be to be able to do
```
# m on meta dev
cm = fsdp(quantize(thunder.jit(m)))
cm.load_state_dict(state_dict_of_m)
```

crcrpar:
nit: isn't it `fsdp(quantize(thunder.jit(m)))`?

t-vi:
Discussing with Masaki: We'll make the early transforms a class with methods to achieve the various transformations and also giving a good repository of the details of the transformation.

t-vi:
This has been achieved.

PR #542, opened 2024-06-06
thunder.jit: add post_optimization_transform

kshitij12345:
Fixes: #533


t-vi:
When we make this switch, let us use a different method than `__call__` on these, e.g. `transform_traces` or so.
Then we could have the transforming of the modules (for early transforms) in `transform_module`, and #483 can add it's methods.

thunder/__init__.py
```diff
@@ -1,5 +1,5 @@
 from functools import wraps
-from typing import Any
+from typing import Any, List
```
crcrpar, line 2:
```suggestion
from typing import Any
```

crcrpar (commented):
this covers what I have in my mind. great

thunder/__init__.py
```diff
@@ -592,6 +599,17 @@ def get_computation_and_inputs(*args, **kwargs):
 
             if not compile_options.get("disable_inplace_copy_check", False):
                 thunder.core.transform_common._inplace_copy_sanity_check(computation_trc)
+
+            for transform in post_optimization_transforms:
+                # NOTE: `backward_trc` could be None.
```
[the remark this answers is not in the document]
kshitij12345: I think this is fine, example - I have some transformation for profiling, it is ok to not have `backward_trc` as I maybe only interested in profiling the forward trace for now.

thunder/tests/test_jit_general.py
```diff
@@ -813,3 +813,45 @@ def foo(a, scalar):
...
+                        bsyms.append(comment_bsym)
+
+                    bsyms.append(bsym)
+
+                commented_trace.bound_symbols = bsyms
+                return commented_trace
+
+            return _comment_add_op_transform(fwd_trc), (
+                _comment_add_op_transform(bwd_trc) if bwd_trc is not None else None
+            )
```
IvanYashchuk, line 842: I think the transform should take just one trace and "if backward is not None" should be in infra code not in transform code.

kshitij12345: Makes sense, also linking the relevant [comment](https://github.com/Lightning-AI/lightning-thunder/issues/533#issuecomment-2152894973) from the issue here

PR #543, opened 2024-06-06
Simplify `BoundSymbolRHS` and enforce immutability

riccardofelluga:
<details>

</details>

## What does this PR do?
Simplifies on #397 and PR #538. In reply to [this comment](https://github.com/Lightning-AI/lightning-thunder/pull/538#issuecomment-2150837852), there is the need to enforce immutability to the `BoundSymbolRHS` object. To do so there are two alternatives, one is by freezing the dataclass and another is to create a `BoundSymbolRHS` as a NamedTuple which is sintax sugar for [namedtuple](https://docs.python.org/3/library/collections.html#collections.namedtuple). I decided to go for the latter as it greatly simplifies the code and takes care of hash and eq operator automatically. To do so however there is the need to transform the args and kwargs into tuples which is carried out by using the very fast tree_flatten from optree. 

Furthermore, once the issue with `unpack_trivial` is resolved, those methods can be removed and we will have a nice and tidy namedtuple.

The reasoning behind this change is the enforcement of immutability.

## PR review
cc. @IvanYashchuk

### DIFF
--- a/thunder/core/symbol.py
+++ b/thunder/core/symbol.py
@@ -7,7 +7,6 @@ from contextvars import ContextVar
 from contextlib import contextmanager
 from itertools import chain
 from types import ModuleType
-from typing import NamedTuple
 
 from dataclasses import dataclass, field
 from typing import Any, Dict, Optional, List, Type, Tuple, TYPE_CHECKING
@@ -526,9 +525,10 @@ class BoundSymbol(BoundSymbolInterface):
 
         return (self.sym, self._var_args, self._var_output) == (other.sym, other._var_args, other._var_output)
 
-    def rhs(self):
-        hashable_args = tuple(map(make_hashable, self._var_args))
-        return BoundSymbolRHS(self.sym, hashable_args, FrozenDict(self._var_kwargs))
+    def rhs(self) -> BoundSymbolRHS:
+        hashable_args = make_hashable(self._var_args)
+        hashable_kwargs = make_hashable(self._var_kwargs)
+        return BoundSymbolRHS(self.sym, hashable_args, hashable_kwargs)
 
     # TODO Document contexts
     def import_ctx(self):
--- a/thunder/core/symbol.py
+++ b/thunder/core/symbol.py
@@ -7,6 +7,7 @@ from contextvars import ContextVar
 from contextlib import contextmanager
 from itertools import chain
 from types import ModuleType
+from typing import NamedTuple
 
 from dataclasses import dataclass, field
 from typing import Any, Dict, Optional, List, Type, Tuple, TYPE_CHECKING
@@ -21,7 +22,6 @@ from thunder.core.pytree import tree_flatten, tree_unflatten, tree_map
 import thunder.core.dtypes as dtypes
 import thunder.core.devices as devices
 from thunder.core.proxies import Proxy, NumberProxy, variableify, CollectionProxy
-from thunder.core.utils import FrozenDict
 
 from thunder.core.trace import (
     get_tracectx,
@@ -527,7 +527,7 @@ class BoundSymbol(BoundSymbolInterface):
         return (self.sym, self._var_args, self._var_output) == (other.sym, other._var_args, other._var_output)
 
     def rhs(self):
-        return BoundSymbolRHS(self)
+        return BoundSymbolRHS(self.sym, tuple(tree_flatten(self._var_args)[0]), tuple(tree_flatten(self._var_kwargs)[0]))
 
     # TODO Document contexts
     def import_ctx(self):
@@ -650,35 +650,18 @@ def has_tags(bsym: BoundSymbol, tags: set[OpTags]) -> bool:
 
 # Wrapper class that hashes and equates only the right hand side of a BoundSymbol for CSE.
 # That is to say, its symbol, args, and kwargs, but not its output.
-@dataclass
-class BoundSymbolRHS:
-    parent: BoundSymbol
-    _frozen_kwargs: FrozenDict
-
-    def __init__(self, parent: BoundSymbol) -> None:
-        self.parent = parent
-        self._frozen_kwargs = FrozenDict(parent._var_kwargs)
-
-    @functools.cached_property
-    def _hash(self) -> int:
-        # TODO: Find a better way to identify inputs by id instead of hash.
-        if self.parent.sym.name == "unpack_trivial":
-            return id(self)
-        try:
-            return hash((self.parent.sym, self.parent._var_args, self._frozen_kwargs))
-        except:
-            return id(self)
+# NOTE: once the unpack_trivial is fixed this can be changed to a default immmutable namedtuple.
+class BoundSymbolRHS(NamedTuple):
+    sym: Symbol
+    args: tuple[Any]
+    kwargs: tuple[Any]
 
     def __hash__(self) -> int:
-        return self._hash
+        # TODO: Fix unpack_trivial usage such that the name is kept in the kwargs.
+        if self.sym.name == "unpack_trivial":
+            return id(self)
 
-    def __eq__(self, other: BoundSymbolRHS) -> bool:
-        if not isinstance(other, BoundSymbolRHS):
-            return False
-        if self.parent is other.parent:
-            return True
-        return (self.parent.sym, self.parent._var_args, self._frozen_kwargs) == (
-            other.parent.sym,
-            other.parent._var_args,
-            other._frozen_kwargs,
-        )
+        return hash((self.sym, self.args, self.kwargs))
+
+    def __eq__(self, other) -> bool:
+        return isinstance(other, self.__class__) and tuple(self) == tuple(other)
--- a/thunder/core/symbol.py
+++ b/thunder/core/symbol.py
@@ -527,7 +527,9 @@ class BoundSymbol(BoundSymbolInterface):
         return (self.sym, self._var_args, self._var_output) == (other.sym, other._var_args, other._var_output)
 
     def rhs(self):
-        return BoundSymbolRHS(self.sym, tuple(tree_flatten(self._var_args)[0]), tuple(tree_flatten(self._var_kwargs)[0]))
+        return BoundSymbolRHS(
+            self.sym, tuple(tree_flatten(self._var_args)[0]), tuple(tree_flatten(self._var_kwargs)[0])
+        )
 
     # TODO Document contexts
     def import_ctx(self):
--- a/thunder/core/symbol.py
+++ b/thunder/core/symbol.py
@@ -18,7 +18,7 @@ import thunder.core.baseutils as baseutils
 import thunder.core.codeutils as codeutils
 from thunder.core.codeutils import Printable, Positions
 from thunder.core.baseutils import BoundSymbolInterface, ProxyInterface
-from thunder.core.utils import FrozenDict
+from thunder.core.utils import FrozenDict, make_hashable
 from thunder.core.pytree import tree_flatten, tree_unflatten, tree_map
 import thunder.core.dtypes as dtypes
 import thunder.core.devices as devices
@@ -527,17 +527,7 @@ class BoundSymbol(BoundSymbolInterface):
         return (self.sym, self._var_args, self._var_output) == (other.sym, other._var_args, other._var_output)
 
     def rhs(self):
-        def make_hashable(x: Any) -> tuple | FrozenDict:
-            if isinstance(x, dict):
-                return FrozenDict(x)
-            if isinstance(x, (list, tuple)):
-                return tuple(map(make_hashable, x))
-            if isinstance(x, slice):
-                return id(x)
-            return x
-
         hashable_args = tuple(map(make_hashable, self._var_args))
-
         return BoundSymbolRHS(self.sym, hashable_args, FrozenDict(self._var_kwargs))
 
     # TODO Document contexts
--- a/thunder/core/symbol.py
+++ b/thunder/core/symbol.py
@@ -530,8 +530,10 @@ class BoundSymbol(BoundSymbolInterface):
         def make_hashable(x: Any) -> tuple | FrozenDict:
             if isinstance(x, dict):
                 return FrozenDict(x)
-            if isinstance(x, list):
-                return tuple(x)
+            if isinstance(x, (list, tuple)):
+                return tuple(map(make_hashable, x))
+            if isinstance(x, slice):
+                return id(x)
             return x
 
         hashable_args = tuple(map(make_hashable, self._var_args))
--- a/thunder/core/symbol.py
+++ b/thunder/core/symbol.py
@@ -18,6 +18,7 @@ import thunder.core.baseutils as baseutils
 import thunder.core.codeutils as codeutils
 from thunder.core.codeutils import Printable, Positions
 from thunder.core.baseutils import BoundSymbolInterface, ProxyInterface
+from thunder.core.utils import FrozenDict
 from thunder.core.pytree import tree_flatten, tree_unflatten, tree_map
 import thunder.core.dtypes as dtypes
 import thunder.core.devices as devices
@@ -526,9 +527,16 @@ class BoundSymbol(BoundSymbolInterface):
         return (self.sym, self._var_args, self._var_output) == (other.sym, other._var_args, other._var_output)
 
     def rhs(self):
-        return BoundSymbolRHS(
-            self.sym, tuple(tree_flatten(self._var_args)[0]), tuple(tree_flatten(self._var_kwargs)[0])
-        )
+        def make_hashable(x: Any) -> tuple | FrozenDict:
+            if isinstance(x, dict):
+                return FrozenDict(x)
+            if isinstance(x, list):
+                return tuple(x)
+            return x
+
+        hashable_args = tuple(map(make_hashable, self._var_args))
+
+        return BoundSymbolRHS(self.sym, hashable_args, FrozenDict(self._var_kwargs))
 
     # TODO Document contexts
     def import_ctx(self):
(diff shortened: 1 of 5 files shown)


thunder/core/symbol.py
```diff
@@ -650,35 +652,18 @@ def has_tags(bsym: BoundSymbol, tags: set[OpTags]) -> bool:
 
 # Wrapper class that hashes and equates only the right hand side of a BoundSymbol for CSE.
 # That is to say, its symbol, args, and kwargs, but not its output.
-@dataclass
-class BoundSymbolRHS:
```
IvanYashchuk, line 654: Frozen dataclass and namedtuples are very similar. Dataclasses are used in other parts of the project and I don't think there's a need to change it here. Can you please explain your motivation and what limitations in dataclasses do you see?

thunder/core/symbol.py
```diff
@@ -527,7 +527,9 @@ def __eq__(self, other):
         return (self.sym, self._var_args, self._var_output) == (other.sym, other._var_args, other._var_output)
 
     def rhs(self):
-        return BoundSymbolRHS(self)
+        return BoundSymbolRHS(
+            self.sym, tuple(tree_flatten(self._var_args)[0]), tuple(tree_flatten(self._var_kwargs)[0])
```
IvanYashchuk, line 531: `tree_flatten(self._var_kwargs)` removes information about the keys of the dictionary. I don't think it's correct to consider `{correction: 1}` and `{dim: 1}` as equal.

thunder/core/symbol.py
```diff
@@ -21,7 +22,6 @@
 import thunder.core.dtypes as dtypes
 import thunder.core.devices as devices
 from thunder.core.proxies import Proxy, NumberProxy, variableify, CollectionProxy
-from thunder.core.utils import FrozenDict
```
IvanYashchuk, line 24: Why do you suggest that FrozenDict has no use here? How do you deal with keyword arguments names?

thunder/core/symbol.py
```diff
@@ -650,35 +652,18 @@ def has_tags(bsym: BoundSymbol, tags: set[OpTags]) -> bool:
...
 # That is to say, its symbol, args, and kwargs, but not its output.
-@dataclass
-class BoundSymbolRHS:
-    parent: BoundSymbol
-    _frozen_kwargs: FrozenDict
-
-    def __init__(self, parent: BoundSymbol) -> None:
-        self.parent = parent
-        self._frozen_kwargs = FrozenDict(parent._var_kwargs)
+# NOTE: once the unpack_trivial is fixed this can be changed to a default immmutable namedtuple.
```
IvanYashchuk, line 655: Let's fix this unpack_trivial problem first before introducing any changes to `BoundSymbolRHS`.

IvanYashchuk (changes requested):

thunder/core/symbol.py
riccardofelluga: True! My main idea behind using a tuple based data structure is its immutable property. `NamedTuple` is very similar to dataclass as noted in the original [PEP](https://peps.python.org/pep-0557):
>  Data Classes can be thought of as “mutable namedtuples with defaults”. 

In our case we use the dataclass with the `frozen` tag set to true which make them immutable and even closer to `NamedTuple`. However, tuples offer the advantage of having a small memory footprint. The only practical limitation I see in using a frozen dataclass is the hack required to change a value whereas namedtuple has a convenient method for it. In any case I recognize that this is a weak point as if we want a mutable dataclass then we shouldn't declare it as frozen and hack around.

With that said, I recognize that the difference in our use case is minimal, and if it is better for readability to use a frozen dataclass instead, it's ok for me

thunder/core/symbol.py
riccardofelluga: Woops, I missed that! Thanks for pointing that out 🙏

thunder/core/symbol.py
riccardofelluga: I thought I could get away with flattening and keeping the tuple, however that's not the best idea as you pointed out in earlier comment.

thunder/core/symbol.py
riccardofelluga: I agree! I'll open a PR for that fix

riccardofelluga:
Opened #565, when that closes I'll update this PR with the discussed changes.

jjsjann123:
> Does this have any impact on dynamic shape support? CC @jjsjann123

IIUC, we are only freezing instances in the namedtuple (or dataclasses as Ivan suggested). But a proxy would remain a proxy even after frozen.

`rhs()` method is only used in CSE transformation? I understand it's to prevent accidental mutation. So it shouldn't impact any other transformation since it's not in a trace. Sounds right? @riccardofelluga

riccardofelluga:
> `rhs()` method is only used in CSE transformation? I understand it's to prevent accidental mutation. So it shouldn't impact any other transformation since it's not in a trace. Sounds right? @riccardofelluga

Yes exactly, this PR shouldn't impact that

t-vi:
@riccardofelluga I think this is something effected by the patch: 

```
FAILED thunder/tests/test_examine_memory.py::test_view_ops_torch_cuda_float32 - assert 144 == 128
FAILED thunder/tests/test_examine_memory.py::test_view_ops_nvfuser_cuda_float32 - AssertionError: assert 16 == 32
 +  where 16 = sum(dict_values([0, 112, 0, 16, -16, -96]))
 +    where dict_values([0, 112, 0, 16, -16, -96]) = <built-in method values of dict object at 0x7f64ec0a3680>()
 +      where <built-in method values of dict object at 0x7f64ec0a3680> = {'clear_mutable_collection ': 0, 'del t2': -16, 'del t3': -96, 'nvFusion0 t69': 16, ...}.values
 +  and   32 = get_return_memory(return (t69, t60))
```

thunder/core/utils.py
```diff
@@ -772,7 +772,14 @@ def __repr__(self) -> str:
...
     def __hash__(self) -> int:
-        return hash(frozenset(self.items()))
+        def make_hashable(item) -> tuple:
+            k, v = item
+            if isinstance(v, list):
+                return (k, tuple(v))
+            return item
+
+        hashable_items = map(make_hashable, self.items())
+        return hash(frozenset(hashable_items))
```
IvanYashchuk: Items that are put into the FrozenDict should already be hashable similarly to how `frozenset` doesn't accept unhashable types:
```py
In [1]: frozenset(("str", 1, ["list"]))
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[1], line 1
----> 1 frozenset(("str", 1, ["list"]))

TypeError: unhashable type: 'list'
```

thunder/core/symbol.py
```diff
@@ -526,9 +527,16 @@ def __eq__(self, other):
...
-            self.sym, tuple(tree_flatten(self._var_args)[0]), tuple(tree_flatten(self._var_kwargs)[0])
-        )
+        def make_hashable(x: Any) -> tuple | FrozenDict:
+            if isinstance(x, dict):
+                return FrozenDict(x)
+            if isinstance(x, list):
+                return tuple(x)
+            return x
+
+        hashable_args = tuple(map(make_hashable, self._var_args))
```
IvanYashchuk, line 530: What if there's an argument like `[[0, 0], [1,]]`? Nested sequences are used for example in `prims.pad`: https://github.com/Lightning-AI/lightning-thunder/blob/6bc5752373c085a284413d84f420371212a9b996/thunder/core/prims.py#L3051
One possible solution is to make it recursive.

thunder/core/symbol.py
```diff
@@ -526,9 +527,16 @@ def __eq__(self, other):
...
+
+        return BoundSymbolRHS(self.sym, hashable_args, FrozenDict(self._var_kwargs))
```
IvanYashchuk, line 531: It's better to convert the values of the dictionary to hashable objects before passing them to FrozenDict.

thunder/core/symbol.py
```diff
@@ -526,9 +527,16 @@ def __eq__(self, other):
         return (self.sym, self._var_args, self._var_output) == (other.sym, other._var_args, other._var_output)
 
     def rhs(self):
-        return BoundSymbolRHS(
-            self.sym, tuple(tree_flatten(self._var_args)[0]), tuple(tree_flatten(self._var_kwargs)[0])
-        )
+        def make_hashable(x: Any) -> tuple | FrozenDict:
```
IvanYashchuk: This function looks more like "maybe_make_hashable" as there's a fallback path that just returns the input that may or may not be hashable?

IvanYashchuk (changes requested):

thunder/core/symbol.py
riccardofelluga: On the latest commit it's already recursive and the snippet:
```python
tuple(map(make_hashable, [[0, 0], [1,]]))
```
generates `((0, 0), (1,))` which, being a tuple of tuples, hashable.

Maybe the code should be more explicit but in general, `make_hashable` recurses inside the lists here:
https://github.com/Lightning-AI/lightning-thunder/blob/e409cec8d03c7e8b7de9ada451501054da7d6eee/thunder/core/utils.py#L1130-L1131

thunder/core/utils.py
riccardofelluga: While this is not the latest commit I do agree with the suggestion of making `FrozenDict` more explicit. It makes sense for it to throw a `TypeError` instead of starting a recursion. I will update such that `make_hashable` is called in the `rhs()` method before creating the `FrozenDict`. This requires to modify the recursive fn too(totally ok), as of now I was banking on the fact that if the dict happens to not have un-hashable values then the `make_hashable` wouldn't have been called.

thunder/core/symbol.py
riccardofelluga: I don't know if that is still the case in the latest revision

riccardofelluga:
> @riccardofelluga I think this is something effected by the patch:
> 
> ```
> FAILED thunder/tests/test_examine_memory.py::test_view_ops_torch_cuda_float32 - assert 144 == 128
> FAILED thunder/tests/test_examine_memory.py::test_view_ops_nvfuser_cuda_float32 - AssertionError: assert 16 == 32
>  +  where 16 = sum(dict_values([0, 112, 0, 16, -16, -96]))
>  +    where dict_values([0, 112, 0, 16, -16, -96]) = <built-in method values of dict object at 0x7f64ec0a3680>()
>  +      where <built-in method values of dict object at 0x7f64ec0a3680> = {'clear_mutable_collection ': 0, 'del t2': -16, 'del t3': -96, 'nvFusion0 t69': 16, ...}.values
>  +  and   32 = get_return_memory(return (t69, t60))
> ```

I am investigating this a bit but it really looks like this is failing because we are actually saving memory! 😁 

image

thunder/tests/test_examine_memory.py
```diff
@@ -71,7 +71,7 @@ def bar(a, b):  # [4] [2,2]
     bw_traces = thunder.last_backward_traces(cbar)
     bw_extrace = bw_traces[-1]
     max_mem_bw = get_alloc_memory(bw_extrace)
-    assert max_mem_bw[0] == 144
+    assert max_mem_bw[0] == 128
```
IvanYashchuk, line 74: How did the trace change?

thunder/core/utils.py
```diff
@@ -1123,3 +1124,13 @@ def partition(pred, iterable):
...
     return itertools.filterfalse(pred, t1), filter(pred, t2)
+
+
+def make_hashable(item: Any, /) -> tuple:
+    if isinstance(item, (list, tuple)):
+        return tuple(map(make_hashable, item))
+    if isinstance(item, dict):
+        return FrozenDict(item)
+    if isinstance(item, slice):
+        return id(item)
```
IvanYashchuk, line 1135: What function uses slice objects as inputs? What do you think about checking if the object is hashable or not and taking its id?

thunder/core/utils.py
```diff
@@ -1123,3 +1124,13 @@ def partition(pred, iterable):
     # partition(is_odd, range(10)) --> 0 2 4 6 8   and  1 3 5 7 9
     t1, t2 = itertools.tee(iterable)
     return itertools.filterfalse(pred, t1), filter(pred, t2)
+
+
+def make_hashable(item: Any, /) -> tuple:
+    if isinstance(item, (list, tuple)):
```
IvanYashchuk, line 1130: Is there anything useful in collections.abc that would cover lists, tuples, and all other similar objects?

thunder/core/symbol.py
IvanYashchuk: Thanks, GitHub probably showed me a different version for some reason.

thunder/tests/test_examine_memory.py
riccardofelluga: We go from this:
```python
@torch.no_grad()
@no_autocast
  # cotangents: "Collection"
  t6, t7, = cotangents
  del cotangents
  [t58, t69] = nvFusion0(t6, t7)
    # t64 = prims.sum(t7, (0, 1))  # t64: "cuda:0 f32[4]"
    # t65 = prims.broadcast_in_dim(t64, [1, 1, 4], [2])  # t65: "cuda:0 f32[1, 1, 4]"
    # t66 = prims.add(t6, t65)  # t66: "cuda:0 f32[1, 1, 4]"
    # t67 = prims.sum(t66, (0, 1))  # t67: "cuda:0 f32[4]"
    # t68 = prims.broadcast_in_dim(t67, [1, 4], [1])  # t68: "cuda:0 f32[1, 4]"
    # t69 = prims.sum(t68, (0,))  # t69: "cuda:0 f32[4]"
    # t54 = prims.broadcast_in_dim(t64, [1, 1, 4], [2])  # t54: "cuda:0 f32[1, 1, 4]"
    # t55 = prims.add(t6, t54)  # t55: "cuda:0 f32[1, 1, 4]"
    # t56 = prims.sum(t55, (0, 1))  # t56: "cuda:0 f32[4]"
    # t57 = prims.broadcast_in_dim(t56, [1, 4], [1])  # t57: "cuda:0 f32[1, 4]"
    # t58 = prims.sum(t57, (0,))  # t58: "cuda:0 f32[4]"
  del t6, t7
  t60 = torch.reshape(t58, (2, 2))  # t60: "cuda:0 f32[2, 2]"
    # t60 = ltorch.reshape(t58, (2, 2))  # t60: "cuda:0 f32[2, 2]"
      # t60 = prims.reshape(t58, (2, 2))  # t60: "cuda:0 f32[2, 2]"
  del t58
  return (t69, t60)
```

to this: 
```python
@torch.no_grad()
@no_autocast
  # cotangents: "Collection"
  t6, t7, = cotangents
  del cotangents
  [t69] = nvFusion0(t6, t7)
    # t64 = prims.sum(t7, (0, 1))  # t64: "cuda:0 f32[4]"
    # t65 = prims.broadcast_in_dim(t64, [1, 1, 4], [2])  # t65: "cuda:0 f32[1, 1, 4]"
    # t66 = prims.add(t6, t65)  # t66: "cuda:0 f32[1, 1, 4]"
    # t67 = prims.sum(t66, (0, 1))  # t67: "cuda:0 f32[4]"
    # t68 = prims.broadcast_in_dim(t67, [1, 4], [1])  # t68: "cuda:0 f32[1, 4]"
    # t69 = prims.sum(t68, (0,))  # t69: "cuda:0 f32[4]"
  del t6, t7
  t60 = torch.reshape(t69, (2, 2))  # t60: "cuda:0 f32[2, 2]"
    # t60 = ltorch.reshape(t69, (2, 2))  # t60: "cuda:0 f32[2, 2]"
      # t60 = prims.reshape(t69, (2, 2))  # t60: "cuda:0 f32[2, 2]"
  return (t69, t60)
```
Thanks @kiya00 for the discussion on this test and the traces!

thunder/core/utils.py
riccardofelluga: Yeah I was thinking the same, as of now also some other libraries check for specific types before deciding is things are hashable or not.  However, I am looking out for types that can help with that if there are any and I think that would be the cleaner option

thunder/core/utils.py
riccardofelluga: Actively looking out for that actually! I would be such a neater solution, do you have any pointer?

t-vi: Sequence

thunder/core/symbol.py
```diff
@@ -526,9 +525,10 @@ def __eq__(self, other):
 
         return (self.sym, self._var_args, self._var_output) == (other.sym, other._var_args, other._var_output)
 
-    def rhs(self):
-        hashable_args = tuple(map(make_hashable, self._var_args))
-        return BoundSymbolRHS(self.sym, hashable_args, FrozenDict(self._var_kwargs))
+    def rhs(self) -> BoundSymbolRHS:
```
riccardofelluga, line 528: This could now be made `@functools.cached_property`!

thunder/core/symbol.py
```diff
@@ -525,8 +525,11 @@ def __eq__(self, other):
 
         return (self.sym, self._var_args, self._var_output) == (other.sym, other._var_args, other._var_output)
 
-    def rhs(self):
-        return BoundSymbolRHS(self)
+    @functools.cached_property
```
IvanYashchuk, line 528: This should be just `@property`. BoundSymbol itself is not immutable and having cached_property is unsafe. Take a look at this simplified example:
```py
import functools

@dataclass
class Test:
    test_attr: int

    @functools.cached_property
    def test(self):
        return self.test_attr

test = Test(2)
print(test.test)
test.test_attr = 3.0
print(test.test)
assert not test.test == 2
```

thunder/core/symbol.py
```diff
@@ -649,35 +652,8 @@ def has_tags(bsym: BoundSymbol, tags: set[OpTags]) -> bool:
...
-            return False
-        if self.parent is other.parent:
-            return True
-        return (self.parent.sym, self.parent._var_args, self._frozen_kwargs) == (
-            other.parent.sym,
-            other.parent._var_args,
-            other._frozen_kwargs,
-        )
+    sym: Symbol
+    args: tuple[Any]
```
IvanYashchuk, line 658: We don't do any type checking but would this be more accurate?
```suggestion
    args: tuple[typing.Hashable]
```

IvanYashchuk (approved):
Great cleanup and improvement! It's nice to see immediate memory usage improvements in the dedicated test (thunder/tests/test_examine_memory.py)!

thunder/core/symbol.py
IvanYashchuk: Nevermind, we have other similar usages already with https://github.com/Lightning-AI/lightning-thunder/blob/8309fc015c1ea9d684337cdcf95138524cd8645c/thunder/core/symbol.py#L436-L437

BoundSymbol should be just made a frozen dataclass.

t-vi (approved):

PR #546, opened 2024-06-06
Minimize weight range for `cross_entropy` opinfo test

rdspring1:
This PR limits the range for the `cross_entropy` weight argument for test stability. The range is inherited from `nll_loss` opinfo test.


t-vi (approved):
Thank you @rdspring1

PR #550, opened 2024-06-07
use transform_trace(s) on Transform classes instead of __call__

t-vi:
building on @kshitij12345 's #542 

The next step is to include the transformation of the module as a method on the transform.

### DIFF
--- a/thunder/__init__.py
+++ b/thunder/__init__.py
@@ -606,7 +606,7 @@ def jit(
                 computation_trc = transform.transform_trace(computation_trc, executors_list=cd.executors_list)
                 extraces.append(computation_trc)
                 if backward_trc is not None:
-                    backward_trc = transform(backward_trc, executors_list=cd.executors_list)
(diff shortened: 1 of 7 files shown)


kshitij12345 (approved):
Thank you for the follow-up update, LGTM

IvanYashchuk:
Why do you prefer `transform_trace` to `__call__`? Does it mean in the future whoever wants to introduce a transform would need to create a subclass instead of `function(trace) -> trace`?

t-vi:
@IvanYashchuk so with the typecheck added in #542, we would have that, yes.

At least for early transforms, we have
- modify the traces,
- modify the module,
- process checkpoints for load/save,
and also (in principle) "add the transform to a thunder.jit'ed function/module", all of which might be candidates for `__call__`. 

To my mind, a "bare" callback has less information about the intuition here, which is why it seems a good idea.
That said, I'm happy to reconsider if we arrive at the conclusion that it's too much of a hassle.

PR #552, opened 2024-06-07
update amp.autocast call for torch deprecation

t-vi:


t-vi:
I'm emergency merging this to fix CI.

Issue #551, opened 2024-06-07
CSE should always modify subsymbols


IvanYashchuk:
## 🐛 Bug

There's an interesting error reported in https://github.com/Lightning-AI/lightning-thunder/issues/292 about missing key value in some dict. Let's discuss in details what's going on there:
```py
  File "/home/carlos/lightning-thunder/thunder/executors/nvfuserex_impl.py", line 273, in create_fd
    translate_bound_symbol(bsym)
  File "/home/carlos/lightning-thunder/thunder/executors/nvfuserex_impl.py", line 263, in translate_bound_symbol
    nvresults = translator(*bsym.args, **bsym.kwargs, fd=fd, lc_to_nv_map=lc_to_nv_map)
  File "/home/carlos/lightning-thunder/thunder/executors/nvfuserex_impl.py", line 1786, in mul
    nvb = getnv(b, fd, lc_to_nv_map)
  File "/home/carlos/lightning-thunder/thunder/executors/nvfuserex_impl.py", line 115, in getnv
    return lc_to_nv_map[x]
  File "/home/carlos/lightning-thunder/thunder/core/utils.py", line 907, in __getitem__
    return self._dict[key_]
KeyError: 't4030'
```

Jumping through a few steps of narrowing down where the problem is (walking up and down in pdb with breakpoints in fusion pass and materialization led here), consider the following code snippet together with the suggested breakpoint
```py
import thunder
from thunder import prims
from thunder import torch as ltorch
from thunder import dtypes
import torch

def func(t126):
    t127 = prims.convert_element_type(t126, dtypes.float32)  # t127: "cuda:0 f32[5, 5, 10240]"
    t128 = prims.convert_element_type(t126, dtypes.float32)  # t128: "cuda:0 f32[5, 5, 10240]"
    t129 = ltorch.mul(t127, t128)  # t129: "cuda:0 f32[5, 5, 10240]"
      # t129 = prims.mul(t127, t128)  # t129: "cuda:0 f32[5, 5, 10240]"
    return t129

t126 = torch.full((5, 5, 10240), 1.0, dtype=torch.bfloat16, device="cuda")  # t126: "cuda:0 bf16[5, 5, 10240]"

jfunc = thunder.jit(func, executors=["torchcompile", "nvfuser"])
out = jfunc(t126)
print(thunder.last_traces(jfunc)[-1])
```
```diff
diff --git a/thunder/executors/torch_compile.py b/thunder/executors/torch_compile.py
[...]

Issue #292, opened 2024-04-29
Model phi-2 doesn't work with Thunder+Inductor compilation


mpatel31415:
## 🐛 Bug

With newest version of Docker image (tested on 2024-04-28 ) training with thunder.jit with additional inductor executor gives KeyError – 't4030'.

### To Reproduce

Minimal reproducer: https://github.com/Lightning-AI/lightning-thunder/issues/292#issuecomment-2085245555

Before each testing each compilation method I restarted the container:
```
mkdir -p output
docker run --pull=always --gpus all --ipc=host --ulimit memlock=-1 --ulimit stack=67108864  -v $PWD/output:/output -it INTERNAL_IMAGE:pjnl-20240427
```
**Thunder inductor concat**
* Error:

> An error occurred: KeyError – 't4030'

* Command: 
```
python /opt/pytorch/lightning-thunder/thunder/benchmarks/benchmark_litgpt.py --model_name phi-2 --compile  "thunder_inductor_cat"
```

**Thunder**
* Works fine
* Command: 
```
python /opt/pytorch/lightning-thunder/thunder/benchmarks/benchmark_litgpt.py --model_name phi-2 --compile  "thunder"
```

**Inductor**
* Works fine
* Command:
```
 python /opt/pytorch/lightning-thunder/thunder/benchmarks/benchmark_litgpt.py --model_name phi-2 --compile  "inductor"
```

### Expected behavior

If we can run model in Eager mode, with Thunder + Pytorch default compilation we should be able to use it for Thunder + Inductor as well.

### Environment

As in the Docker image. This results come from single A100. 
nvidia-smi output:
![image](https://github.com/Lightning-AI/lightning-thunder/assets/149149379/9d9d3972-02a4-4991-9a88-43a92c9a7aa1)

cc @apaz-cli

carmocca:
I can reproduce this. The error stacktrace is:

```python
  File "/home/carlos/lightning-thunder/thunder/benchmarks/benchmark_litgpt.py", line 357, in train
    loss.backward()
  File "/home/carlos/nightly-env/lib/python3.10/site-packages/torch/_tensor.py", line 523, in backward
    torch.autograd.backward(
  File "/home/carlos/nightly-env/lib/python3.10/site-packages/torch/autograd/__init__.py", line 267, in backward
    _engine_run_backward(
[...]

PR #553, opened 2024-06-07
Modify subsymbols in cse_single_bsym; unblocks phi-2 with inductor_cat+nvfuser executors

IvanYashchuk:
There was an unfortunate interaction between the CSE pass and the "flattening" of symbols which is done in nvFuser fusion pass. CSE pass was changing only the highest level of the symbol information leaving its subsymbols untouched, nvFuser fusion pass then extracted the subsymbols that were not correct anymore and tried converting to FusionDefinition leading to an error reported in #292.

Fixes https://github.com/Lightning-AI/lightning-thunder/issues/292, fixes https://github.com/Lightning-AI/lightning-thunder/issues/551.

cc @apaz-cli

### DIFF
--- a/thunder/core/transform_common.py
+++ b/thunder/core/transform_common.py
@@ -218,7 +218,6 @@ def cse_single_bsym(
         swap_map=redundant_map,
         skip_inputs=False,
         skip_output=True,
-        skip_subsymbols=True,
     )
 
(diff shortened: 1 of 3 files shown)


t-vi (approved):
Supergood. Great find! Thank you @IvanYashchuk

PR #554, opened 2024-06-07
Automatic registration of torch operators using FakeTensor

### LINKED ISSUE #811: Automatic registration of unknown PyTorch operations
## 🚀 Feature
Currently, if we want to support an operator that only has Torch executor as a backend, we have to manually register the symbol and implementation. While it's not complicated, it would be nice to have an automatic fallback mechanism to save work.

### Motivation

We want to move faster enabling the execution of new models. A common blocker is lack of support for certain PyTorch operations.

### Pitch

Once the unsupported torch operator is detected during the interpretation, it’s automatically registered using the torch backend

kiya00:
## What does this PR do?
Design doc: https://docs.google.com/document/d/1BVr_5co6GDj3wC7NrWq0ckfsbgX5l6w1dOSFBi2c-Hc/edit?usp=sharing

This PR did some experiments to use FakeTensor for our meta function to allow automatic fallback to torch operators when the op is not supposed to have a fusion backend.

Currently it can work for the example test cases(the sinc, hsplit, normalize, adaptive_avg_pool3d are ops currently not added in thunder ):
```py
import torch
import thunder
from torch import nn

def func(a, out):
    t1 = torch.nn.functional.adaptive_avg_pool3d(a, output_size=out)
    return torch.nn.functional.normalize(t1)

def func1(a):
    x = torch.sinc(a)
    t = torch.arange(16.0).reshape(4,4)
    return torch.hsplit(t, 2), x


a = torch.randn((1, 64, 8, 9, 10)).cuda()
jfunc = thunder.jit(func)
out = jfunc(a, (5, 7, 9))
res = func(a, (5, 7, 9))
torch.testing.assert_close(out, res)

inp = torch.rand(2,2).cuda()
jfunc1 = thunder.jit(func1)
out = jfunc1(inp)
expected = func1(inp)
torch.testing.assert_close(out, expected)
```

```py
import torch
import thunder
def func(a):
    x = torch.sinc(a)
    x1,x2= torch.hsplit(x, 2)
    return x1+x2

def func1(a, out):
    t1 = torch.nn.functional.adaptive_avg_pool3d(a, output_size=out)
    return torch.nn.functional.normalize(t1)


inp = torch.rand(4,4).cuda().requires_grad_()
inp_1 = inp.detach().clone().requires_grad_()


jfunc = thunder.jit(func)
out = jfunc(inp)
out.sum().backward()
print(thunder.last_traces(jfunc)[-1])
print(thunder.last_backward_traces(jfunc)[-1])


expected = func(inp_1)
expected.sum().backward()

torch.testing.assert_close(out, expected)
torch.testing.assert_close(inp.grad, inp_1.grad)


a = torch.randn((1, 64, 8, 9, 10)).cuda().requires_grad_()
a1 = a.detach().clone().requires_grad_()

res = func1(a1, (5, 7, 9))
res.sum().backward()

jfunc = thunder.jit(func1)
out = jfunc(a, (5, 7, 9))
out.sum().backward()
print(thunder.last_traces(jfunc)[-1])
print(thunder.last_backward_traces(jfunc)[-1])

torch.testing.assert_close(out, res)
torch.testing.assert_close(a.grad, a1.grad)
```

Fixes #811 

cc: @IvanYashchuk


kiya00:
currently output trace is like:
```
@torch.no_grad()
@no_autocast
def augmented_forward_fn(a):
  # a: "cuda:0 f32[1, 64, 8, 9, 10]"
  t0 = torch.nn.functional.adaptive_avg_pool3d(a, output_size=(5, 7, 9))  # t0: "cuda:0 f32[1, 64, 5, 7, 9]"
  t1 = torch.nn.functional.normalize(t0)  # t1: "cuda:0 f32[1, 64, 5, 7, 9]"
  return {'output': t1, 'flat_args': [a], 'flat_output': (t1,)}, ((a, t0), (5, 7, 9))
import torch

@torch.no_grad()
@no_autocast
  # cotangents: "Collection"
  C0, C1, = saved_for_backward
  del saved_for_backward
  t2, = cotangents
  del cotangents
  a, t0, = C0
  clear_mutable_collection(C0)
  del C0
  i0, i1, i2, = C1
  clear_mutable_collection(C1)
  del C1
  (t14,) = normalize_vjp({'inputs': ((t0,), {}), 'func': _function_0}, t2)
  del t0, t2
  (t15,) = adaptive_avg_pool3d_vjp({'inputs': ((a,), {'output_size': (i0, i1, i2)}), 'func': _function_1}, t14)
  del a, i0, i1, i2, t14
  return (t15,)

```

t-vi:
First off, great job, @kiya00 !

One fundamental question I would have about the design is: given that we are talking about a finite set of operators here, would it be more transparent to autogenerate the code for them and keep them a file? That would have us need to keep up / update for new stuff PyTorch introduces, but in general, it would be less magic. When we implement something in a more detailed way, it would be removed from the autogeneration to manual. 

I'm imagining that we could/would still use the bulk of your code, just the time of registration would move.

WDYT?

thunder/torch/__init__.py
```diff
@@ -4072,6 +4072,234 @@ def avg_pool3d(
...
+
+    augmented_forward_impls[sym.id] = augmented_forward_adaptor(torchfn, op)
+
+    def _vjp_impl(residules, *gs) -> torch.Tensor:
+        assert isinstance(residules, dict)
+        inp_args = residules["inputs"][0]
+        inp_kwargs = residules["inputs"][1]
+        func = residules["torch_func"]
+
+        wrapped_func, diff_args = _make_differentiable_wrapper(func, *inp_args, **inp_kwargs)
```
kshitij12345, line 4152: I was just wondering if we need to pass `func` from augmented forward to the `_vjp_impl` here. Instead can't `_vjp_impl` just capture the `torchfn` and use it?

kiya00: yes, I'll let it capture, thanks~

thunder/torch/__init__.py
```diff
@@ -5359,6 +5359,227 @@ def wait(slf) -> None:
...
+
+
+def backward_adaptor():
+    def wrapper(*args):
+        saved_for_backward: dict = args[0]
+        grad_output = args[1:]
+        torch_func = saved_for_backward["torch_func"]
+        inps = saved_for_backward["inputs"]
+        inp_args = inps[0]
+        inp_kwargs = inps[1]
```
IvanYashchuk, line 5518:
```suggestion
        args, kwargs = saved_for_backward["inputs"]
```

thunder/torch/__init__.py
```diff
@@ -5359,6 +5359,227 @@ def wait(slf) -> None:
...
+        return VJPDual(primal, residuals)
+
+    return wrapper
+
+
+def backward_adaptor():
+    def wrapper(*args):
+        saved_for_backward: dict = args[0]
+        grad_output = args[1:]
+        torch_func = saved_for_backward["torch_func"]
```
IvanYashchuk, line 5515: This could be an input to `backward_adaptor`.

thunder/torch/__init__.py
```diff
@@ -5331,6 +5331,215 @@ def wait(slf) -> None:
...
+
+
+def register_default_torch_op(torchfn: Callable, fn_meta: Callable, m):
+    _fn = langctx(Languages.TORCH)(fn_meta)
+    sym = Symbol(
+        name=torchfn.__name__,
+        meta=_fn,
+        id=f"{m.__name__}.{torchfn.__name__}",
+    )
+    _torch_to_thunder_function_map[torchfn] = sym
```
kiya00, line 5378: Do we use another map `_auto_registered_thunder_torch_ops` to record this and modify the `examine` to show which are auto registered and might have a problem?

thunder/torch/__init__.py
```diff
@@ -5331,6 +5331,215 @@ def wait(slf) -> None:
         utils.check(False, lambda: "torch.distributed is not available")
 
 
+#
+# The automatically registered torch operators
+#
+
+
+def _is_differentiable(arg: Any):
```
crcrpar, line 5339:
```suggestion
def _is_differentiable(arg: Any) -> bool:
```

thunder/tests/test_auto_register_torchops.py
```diff
@@ -0,0 +1,197 @@
...
+    import thunder.torch.default_torch_ops as ops
+    from torch.testing._internal.common_methods_invocations import op_db
+
+    name2func = {}
+    [name2func.setdefault(v.__name__, v) for v in ops.torch_auto_registered_ops[torch]]
+    [name2func.setdefault(f"nn.functional.{v.__name__}", v) for v in ops.torch_auto_registered_ops[torch.nn.functional]]
+    # Use the sample input from torch.xx to test torch.tensor.xx
+    [name2func.setdefault(f"Tensor.{v.__name__}", v) for v in ops.torch_auto_registered_ops[torch.Tensor]]
+    print(f"auto reg ops: {len(name2func)}")
+    op_infos = [opinfo for opinfo in op_db if opinfo.name in name2func.keys()]
```
crcrpar, line 29:
```suggestion
    op_infos = [opinfo for opinfo in op_db if opinfo.name in name2func]
```

thunder/tests/test_auto_register_torchops.py
```diff
@@ -0,0 +1,197 @@
...
+
+    import thunder.torch.default_torch_ops as ops
+    from torch.testing._internal.common_methods_invocations import op_db
+
+    name2func = {}
+    [name2func.setdefault(v.__name__, v) for v in ops.torch_auto_registered_ops[torch]]
+    [name2func.setdefault(f"nn.functional.{v.__name__}", v) for v in ops.torch_auto_registered_ops[torch.nn.functional]]
+    # Use the sample input from torch.xx to test torch.tensor.xx
+    [name2func.setdefault(f"Tensor.{v.__name__}", v) for v in ops.torch_auto_registered_ops[torch.Tensor]]
+    print(f"auto reg ops: {len(name2func)}")
```
crcrpar, line 28:
```suggestion
```

thunder/tests/test_auto_register_torchops.py
```diff
@@ -0,0 +1,197 @@
...
+            continue
+
+        # No cuda backend support
+        if op_info.name in ("nonzero_static",) and device == "cuda":
+            continue
+        # RuntimeError: Calling torch.linalg.cholesky on a CPU tensor requires compiling PyTorch with LAPACK.
+        if op_info.name in ("cholesky",) and device == "cpu":
+            continue
+        # (t6, t7, t8, t9, t10, t11) = torch.histogramdd(t_0, (1, 1, 1, 1, 1), weight=None, density=False)
+        # ValueError: not enough values to unpack (expected 6, got 2)
```
crcrpar, line 46: could you tell me what these lines mean? I couldn't get my head around

thunder/tests/test_auto_register_torchops.py
```diff
@@ -0,0 +1,197 @@
...
+                    else:
+                        fwd_trc = thunder.last_traces(jfun)[-1]
+                        assert any(
+                            bsym.sym.name.endswith(op_info.name.split(".")[-1]) and not bsym.subsymbols
+                            for bsym in fwd_trc.bound_symbols
+                        )
+            else:
+                suc += 1
+
+    print(f"total number of ops with sample input: {total}, success: {suc}, needs manual registeration: {cnt}")
```
crcrpar, line 97: would it be feasible to convert this into an assert?

thunder/tests/test_auto_register_torchops.py
```diff
@@ -0,0 +1,197 @@
...
+    get_opinfo("adaptive_avg_pool2d"),
+    get_opinfo("max_pool2d"),
+]
+disable_opinfos = skip_ops_nanogpt + skip_ops_alexnet
+tmp1 = dict(thunder.core.jit_ext._general_jit_lookaside_map)
+list(tmp1.pop(k.torch_reference, None) for k in disable_opinfos)
+tmp2 = dict(thunder.torch._torch_to_thunder_function_map)
+list(tmp2.pop(k.torch_reference, None) for k in disable_opinfos)
+tmp3 = dict(thunder.core.jit_ext._minimal_lookaside_map)
+list(tmp3.pop(k.torch_reference, None) for k in disable_opinfos)
```
crcrpar, line 119: could you tell me what these `tmp` variables and list comprehensions are doing?

thunder/torch/__init__.py
```diff
@@ -5331,6 +5331,215 @@ def wait(slf) -> None:
...
+    elif isinstance(inp, (TupleProxy, ListProxy, DictProxy)):
+        return inp._value
+    elif isinstance(inp, (torch.dtype, torch.device, torch.Size, torch.memory_format, str, bool, int, float, complex)):
+        return inp
+    elif isinstance(inp, devices.Device):
+        return to_torch_device(inp)
+    elif isinstance(inp, dtypes.dtype):
+        return to_torch_dtype(inp)
+    else:
+        from builtins import type
```
crcrpar, line 5435: Q: would this import be strongly required?

thunder/torch/__init__.py
```diff
@@ -5331,6 +5331,215 @@ def wait(slf) -> None:
...
+
+    return wrapper
+
+
+def meta_adaptor(torch_func: Callable):
+    def wrapper(*args, **kwargs):
+        from thunder.core.pytree import tree_flatten, tree_unflatten
+        from torch._subclasses.fake_tensor import FakeTensorMode
+
+        if kwargs.get("inplace", False) == True:
```
crcrpar, line 5520:
```suggestion
        if kwargs.get("inplace", False):
```

crcrpar, line 5520: pytorch functions could have `out` kwarg, wouldn't it be better to check and error out for this kwarg?
e.g. https://pytorch.org/docs/stable/generated/torch.add.html

thunder/torch/__init__.py
```diff
@@ -5331,6 +5331,215 @@ def wait(slf) -> None:
...
+
+def register_default_torch_ops():
+    from thunder.torch import default_torch_ops
+
+    for m, fns in default_torch_ops.torch_auto_registered_ops.items():
+        for fn in fns:
+            register_default_torch_op(fn, meta_adaptor(fn), m)
+
+
+def register_default_torch_op(torchfn: Callable, fn_meta: Callable, m):
```
crcrpar, line 5371: what would be the expected behavior if `torchfn` is an in-place op whose name has trailing underscore like `torch.sin_`?

thunder/tests/test_auto_register_torchops.py
kiya00: After checking I find this operator is not supported by the vjp, it reports TypeError when cpu-train and the above error when cpu-inference, I'll remove this from the op list

thunder/tests/test_auto_register_torchops.py
kiya00: This is for mocking the ops I want to replace with auto registration, the ops are registered in these global maps, so I tried to remove them in the tmp maps and mock the original ones with the tmp maps.

thunder/torch/__init__.py
kiya00: Great point, thanks~ I'll add some checks

thunder/torch/__init__.py
crcrpar: ah I see. I have no nice alternatives out of my mind

thunder/torch/__init__.py
crcrpar: understood

thunder/torch/__init__.py
mruberry: Something for us to think about in the future:

In general, we don't like using names that conflict with Python builtins, but this isn't always possible. If there are multiple `type` functions, then we should probably note the confusion and try to explicitly disambiguate them by calling the fully qualified name of each function, like `builtins.type`. We don't need to do anything about this today, it's just an idea for how we write maintainable and readable code in the future.

thunder/torch/__init__.py
kiya00: Sure, I'll add that~

thunder/torch/__init__.py
```diff
@@ -5331,6 +5331,230 @@ def wait(slf) -> None:
...
+    from thunder.executors.torchex import _always_executable, ex
+
+    op = ex.register_operator(torchfn.__name__, module=m, meta=fn_meta)
+    ex.register_implementation(sym, op, checker=_always_executable)
+
+    from thunder.core.transforms import augmented_forward_impls, backward_impls
+
+    augmented_forward_impls[sym.id] = augmented_forward_adaptor(op)
+
+    def _vjp_impl(residules, *gs) -> torch.Tensor:
```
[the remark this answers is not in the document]
kiya00: It seems the impl func is not always near the metafunction, I thought adding another layer of wrapper just to move the impl near the metafunc might be unnecessary, what do you think?

thunder/torch/__init__.py
```diff
@@ -5331,6 +5331,230 @@ def wait(slf) -> None:
...
+        return tree_unflatten(outs, inp_spec)
+
+    bwd_op = ex.register_operator(torchfn.__name__ + "_vjp", meta=backward_adaptor(torchfn), fn=_vjp_impl)
+    ex.register_implementation(bwd_op.id, bwd_op, checker=_always_executable)
+    backward_impls[sym.id] = bwd_op
+
+
+def _get_fake_arg(inp: Any, fake_mode):
+    from thunder.core.devices import to_torch_device
+    from thunder.core.proxies import TupleProxy, ListProxy, DictProxy
```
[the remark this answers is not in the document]
kiya00: Do you mean move it to the top of this file? I thought only this func uses them and put them here https://github.com/Lightning-AI/lightning-thunder/blob/3c315b4ca14ada9159c4350f9fb058321a29f7b9/thunder/torch/__init__.py#L30-L36

thunder/torch/__init__.py
```diff
@@ -5331,6 +5331,230 @@ def wait(slf) -> None:
...
+
+
+def backward_adaptor(torch_func: Callable):
+    def wrapper(saved_for_backward, *grad_output):
+        inp_args, inp_kwargs = saved_for_backward
+        from thunder.core.pytree import tree_flatten, tree_unflatten
+        from builtins import sum
+
+        flat_args, spec = tree_flatten(inp_args)
+        if sum(tree_map(_is_differentiable, grad_output)) == 0:
```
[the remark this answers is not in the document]
kiya00: Thanks for catching that! It seems to be already flattened in `backward_pass`, I'll use the map as you suggested

thunder/torch/__init__.py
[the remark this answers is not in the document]
kshitij12345: I don't have a strong preference to move them closer but here is why I think it would be nice to have them together.
The function bodies look similar and if we need an update then it is likely that both need to be updated - so having them close would probably remind this sooner. (But again not a strong preference, so feel free to resolve this if you think it would make the code harder to follow)

thunder/torch/__init__.py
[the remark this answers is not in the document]
kshitij12345: As per pep8 guidelines - import should be on top - I think one of the exception is when they would lead to cyclical dependency (or maybe more) but since we already import `thunder.core` in this file, I think it would be better to move them at the beginning of the file.

https://peps.python.org/pep-0008/#imports

Another discussion on stackoverflow - https://stackoverflow.com/questions/128478/should-import-statements-always-be-at-the-top-of-a-module

thunder/torch/__init__.py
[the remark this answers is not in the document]
kiya00: Yeah, you convinced me :), "if we need an update then it is likely that both need to be updated", a simple wrapper func can't hurt, I'll update it

thunder/tests/test_auto_register_torchops.py
```diff
@@ -0,0 +1,187 @@
...
+    def test_nanogpt_block(self):
+        import thunder.tests.nanogpt_model as nanogpt_model
+
+        for op in skip_ops_nanogpt:
+            if op.name == "gelu":
+                register_default_torch_op(op.torch_reference, meta_adaptor(op.torch_reference), torch)
+            else:
+                register_default_torch_op(op.torch_reference, meta_adaptor(op.torch_reference), torch.nn.functional)
+            self._tmp_update_jit_lookup(op.torch_reference)
+        tdtype = torch.float32  # thunder.torch.to_torch_dtype(dtype)
```
crcrpar, line 150:
```suggestion
        tdtype = torch.float32
```

thunder/torch/__init__.py
```diff
@@ -5331,6 +5336,220 @@ def wait(slf) -> None:
...
+    return wrapper, differentiable_args
+
+
+def register_default_torch_ops():
+    from thunder.torch import default_torch_ops
+
+    for m, fns in default_torch_ops.torch_auto_registered_ops.items():
+        for fn in fns:
+            # Ensure no inplace op in the list
+            assert not fn.__name__.endswith("_")
```
crcrpar, line 5372: maybe raising an exception?
```suggestion
            utils.check(
                fn.__name__.endswith("_"),
                lambda: f"Automatic registration does not support in-place op of {fn.__name__}, please manually register it",
            )
```

thunder/tests/test_auto_register_torchops.py
```diff
@@ -0,0 +1,187 @@
...
+                        trc = thunder.last_backward_traces(jfun)[-1]
+                        fwd_trc = thunder.last_traces(jfun)[-1]
+                        # skip if it is not differentiable
+                        outs = fwd_trc.output[0]["output"]
+                        outs = outs if isinstance(outs, tuple) else (outs,)
+                        if all(not thunder.core.dtypes.is_inexact_dtype(o.dtype) for o in outs):
+                            continue
+                        vjp_op_name = f"{op_info.name.split('.')[-1]}_vjp"
+                        if op_info.name == "mm":
+                            assert any(bsym.sym.name.endswith(vjp_op_name) for bsym in trc.bound_symbols)
```
crcrpar, line 77: could you spell out the name of `bsym.sym`, instead of `.endswith(vjp_op_name)`?
as it's purely a special case for "mm", I'd say it's okay to hard code

kiya00: It's a bit tricky here, for testing `torch.Tensor.mm` and `torch.mm`, both of their op_info.name is `mm`, and because the `torch.mm` and `torch.spmm` are aliases, they are actually the same callable object, so in the `_torch_to_thunder_function_map` they map to the same symbol name `torch.spmm`. So here for `torch.Tensor.mm` the backward symbol name is `mm_vjp` and for `torch.mm` is the `spmm_vjp`, so it's checked with endswith(`mm_vjp`)
Currently we do not distinguish alias, the ops are just listed in the file alphabetically, so the last one that appeared is the name shown in the trace, I think it does not matter which name is used since they are alias.

thunder/torch/__init__.py
```diff
@@ -5331,6 +5336,223 @@ def wait(slf) -> None:
...
+
+    op = ex.register_operator(torchfn.__name__, module=m, meta=fn_meta)
+    ex.register_implementation(sym, op, checker=_always_executable)
+
+    from thunder.core.transforms import augmented_forward_impls, backward_impls
+
+    augmented_forward_impls[sym.id] = augmented_forward_adaptor(op)
+
+    def _vjp_impl_wrapper(residules, *gs):
+        return _vjp_impl(torchfn, residules, *gs)
```
IvanYashchuk, line 5397: Alternatively `functools.partial` can be used here:
```suggestion
    _vjp_impl_wrapper = partial(_vjp_impl, torchfn)
```

thunder/torch/__init__.py
```diff
@@ -5331,6 +5336,223 @@ def wait(slf) -> None:
...
+    wrapped_func, diff_args = _make_differentiable_wrapper(torchfn, *inp_args, **inp_kwargs)
+    _, vjp_outs = torch.autograd.functional.vjp(wrapped_func, diff_args, v=gs)
+    flat_inp_args, inp_spec = tree_flatten(inp_args)
+    iter_vjp_out = iter(vjp_outs if isinstance(vjp_outs, Sequence) else (vjp_outs,))
+    vjp_outs = tuple(next(iter_vjp_out) if _is_differentiable(arg) else None for arg in flat_inp_args)
+    return tree_unflatten(vjp_outs, inp_spec)
+
+
+def meta_adaptor(torch_func: Callable):
+    def wrapper(*args, **kwargs):
```
IvanYashchuk, line 5515: It's not a wrapper because this function doesn't wrap `torch_func` it constructs a meta function for `torch_func`.
```suggestion
    def meta_func(*args, **kwargs):
```

thunder/torch/__init__.py
```diff
@@ -5331,6 +5336,223 @@ def wait(slf) -> None:
...
+            return inp.value
+    elif isinstance(inp, TensorProxy):
+        return fake_mode.from_tensor(
+            torch.empty(
+                inp.shape,
+                dtype=_thunder_to_torch_dtype_map[inp.dtype],
+                requires_grad=inp.requires_grad,
+                device=devices.to_torch_device(inp.device),
+            )
+        )
```
IvanYashchuk, line 5420: This would unnecessarily reserve device memory that would be freed almost immediately if the fake mode wasn't active. `_get_fake_arg` is called inside the active fake mode, is "from_tensor" necessary? Can we simply return the result of `torch.empty`?

thunder/torch/__init__.py
```diff
@@ -5331,6 +5336,223 @@ def wait(slf) -> None:
...
+        for fn in fns:
+            # Ensure no inplace op in the list
+            utils.check(
+                not fn.__name__.endswith("_"),
+                lambda: f"Automatic registration does not support in-place op of {m.__name__}.{fn.__name__}, please manually register it",
+            )
+            register_default_torch_op(fn, meta_adaptor(fn), m)
+
+
+def register_default_torch_op(torchfn: Callable, fn_meta: Callable, m):
```
IvanYashchuk, line 5379: This function is used as `register_default_torch_op(fn, meta_adaptor(fn))` everywhere. Let's hide the meta adaptor then inside this function. Is there any benefit of exposing the meta adaptor function?
```suggestion
def register_default_torch_op(torchfn: Callable, m):
    fn_meta = meta_adaptor(fn)
```

thunder/torch/__init__.py
```diff
@@ -5331,6 +5336,223 @@ def wait(slf) -> None:
...
+    elif isinstance(inp, torch.dtype):
+        return _torch_to_thunder_dtype_map[inp]
+    elif isinstance(inp, (str, bool, int, float, complex)):
+        return inp
+    else:
+        raise NotImplementedError(f"Unsupported type: {builtins.type(inp)}")
+
+
+def augmented_forward_adaptor(sym_op: Callable):
+    def wrapper(*args, **kwargs):
```
IvanYashchuk, line 5461:
```suggestion
    def augmented_forward(*args, **kwargs):
```

thunder/torch/__init__.py
```diff
@@ -5331,6 +5336,223 @@ def wait(slf) -> None:
...
+        out = sym_op(*args, **kwargs)
+        primal = out if isinstance(out, tuple) else (out,)
+        residuals = ((args, kwargs),)
+        return VJPDual(primal, residuals)
+
+    return wrapper
+
+
+def backward_adaptor(torch_func: Callable):
+    def wrapper(saved_for_backward, *grad_output):
```
IvanYashchuk, line 5473:
```suggestion
    def backward(saved_for_backward, *grad_output):
```

thunder/torch/__init__.py
```diff
@@ -5331,6 +5336,223 @@ def wait(slf) -> None:
...
+
+
+def augmented_forward_adaptor(sym_op: Callable):
+    def wrapper(*args, **kwargs):
+        from thunder.core.transforms import VJPDual
+
+        out = sym_op(*args, **kwargs)
+        primal = out if isinstance(out, tuple) else (out,)
+        residuals = ((args, kwargs),)
+        return VJPDual(primal, residuals)
```
IvanYashchuk, line 5467: The function signature of `wrapper` in `backward_adaptor` uses the `saved_for_backward` name for this variable. Let's use the same variable name across the functions for readability.
```suggestion
        saved_for_backward = ((args, kwargs),)
        return VJPDual(primal, saved_for_backward)
```

thunder/torch/__init__.py
```diff
@@ -5331,6 +5336,223 @@ def wait(slf) -> None:
...
+        return _torch_to_thunder_dtype_map[inp]
+    elif isinstance(inp, (str, bool, int, float, complex)):
+        return inp
+    else:
+        raise NotImplementedError(f"Unsupported type: {builtins.type(inp)}")
+
+
+def augmented_forward_adaptor(sym_op: Callable):
+    def wrapper(*args, **kwargs):
+        from thunder.core.transforms import VJPDual
```
IvanYashchuk, line 5462: Returning a plain tuple should also work, I hope this dual container is not confusing.

thunder/torch/__init__.py
```diff
@@ -5331,6 +5336,223 @@ def wait(slf) -> None:
...
+        return tree_unflatten(fake_vjp, spec)
+
+    return wrapper
+
+
+def _vjp_impl(torchfn, residules, *gs):
+    inp_args, inp_kwargs = residules
+
+    wrapped_func, diff_args = _make_differentiable_wrapper(torchfn, *inp_args, **inp_kwargs)
+    _, vjp_outs = torch.autograd.functional.vjp(wrapped_func, diff_args, v=gs)
```
IvanYashchuk, line 5507: A follow-up would be to compute the training-forward during forward that would implicitly save the required for backward tensors and the corresponding mini autograd graph and invoke `torch.autograd.grad` here instead of `vjp` to have about the same runtime performance as PyTorch eager.

thunder/torch/__init__.py
kiya00: Yes, the original intent of passing in fake_mode is to make `_get_fake_arg` a stand along func that can be used outside of the context manager, but since we only use it in side of the active fake mode, I can change it to return the result of torch.empty.

crcrpar (approved):
what an awesome work

thunder/torch/__init__.py
```diff
@@ -5331,6 +5336,222 @@ def wait(slf) -> None:
...
+    augmented_forward_impls[sym.id] = augmented_forward_adaptor(op)
+
+    _vjp_impl_wrapper = partial(_vjp_impl, torchfn)
+
+    bwd_op = ex.register_operator(torchfn.__name__ + "_vjp", meta=backward_adaptor(torchfn), fn=_vjp_impl_wrapper)
+    ex.register_implementation(bwd_op.id, bwd_op, checker=_always_executable)
+    backward_impls[sym.id] = bwd_op
+
+
+# Note this function should be used in side the fake mode context manager
```
kshitij12345, line 5404:
```suggestion
# Note this function should be used inside the fake mode context manager
```

thunder/tests/test_auto_register_torchops.py
```diff
@@ -0,0 +1,187 @@
...
+    name2func = {}
+    [name2func.setdefault(v.__name__, v) for v in ops.torch_auto_registered_ops[torch]]
+    [name2func.setdefault(f"nn.functional.{v.__name__}", v) for v in ops.torch_auto_registered_ops[torch.nn.functional]]
+    # Use the sample input from torch.xx to test torch.tensor.xx
+    [name2func.setdefault(f"Tensor.{v.__name__}", v) for v in ops.torch_auto_registered_ops[torch.Tensor]]
+    op_infos = [opinfo for opinfo in op_db if opinfo.name in name2func]
+    total = 0
+    cnt = 0
+    suc = 0
+    for op_info in op_infos:
```
kshitij12345, line 33: I was wondering if we can parametrize the test over opinfos. I think it would make the test quite easy to read and we won't need the exception handling as we would have the test per op.

We can just move the code which filters opinfo outside into a function and this function would provide values for parametrization.

```python
opinfos = get_filtered_opinfos_for_test()

@pytest.mark.parametrize("opinfo,", opinfos, ids=list(map(lambda opinfo: opinfo.name), opinfos))
```

Wdyt?

thunder/tests/test_auto_register_torchops.py
```diff
@@ -0,0 +1,187 @@
...
+    from torch.testing._internal.common_device_type import skipCPUIfNoLapack
+
+    name2func = {}
+    [name2func.setdefault(v.__name__, v) for v in ops.torch_auto_registered_ops[torch]]
+    [name2func.setdefault(f"nn.functional.{v.__name__}", v) for v in ops.torch_auto_registered_ops[torch.nn.functional]]
+    # Use the sample input from torch.xx to test torch.tensor.xx
+    [name2func.setdefault(f"Tensor.{v.__name__}", v) for v in ops.torch_auto_registered_ops[torch.Tensor]]
+    op_infos = [opinfo for opinfo in op_db if opinfo.name in name2func]
+    total = 0
+    cnt = 0
```
kshitij12345, line 31: Should this be renamed to `exception_cnt`?

thunder/tests/test_auto_register_torchops.py
```diff
@@ -0,0 +1,187 @@
...
+    get_opinfo("scaled_dot_product_attention"),
+]
+skip_ops_alexnet = [
+    get_opinfo("conv2d"),
+    get_opinfo("linear"),
+    get_opinfo("adaptive_avg_pool2d"),
+    get_opinfo("max_pool2d"),
+]
+disable_opinfos = skip_ops_nanogpt + skip_ops_alexnet
+tmp1 = dict(thunder.core.jit_ext._general_jit_lookaside_map)
```
kshitij12345, line 104: I think it would be nicer to rename it as `tmp_lookaside_map` and similar for `tmp2` and `tmp3`.

kshitij12345 (approved):
The PR looks really great! Just had a couple of suggestions, thank you @kiya00!

thunder/tests/test_auto_register_torchops.py
kiya00: Hi @kshitij12345 , thanks, it's updated by using parameterized opinfo

kiya00:
Hi @t-vi , I found the CI failure seems unrelated to this PR, and can only reproduce it locally by running `pytest thunder/tests/test_inplace_functionalization.py thunder/tests/test_nvfuser_remat.py::test_find_cut_dropout_nvfuser_cuda_None` together (Separate execution succeeds).

t-vi (approved):
Thank you @kiya00 @kshitij12345 @crcrpar @IvanYashchuk @riccardofelluga @mruberry

Issue #555, opened 2024-06-07
jit : if torch.device is an input to jitted fn, it's value is not checked in prologue.


kshitij12345:
**Repro**:
```python
import thunder
import torch

def foo(x, device):
    return x.to(device)

jfoo = thunder.jit(foo)
x = torch.randn(4,)
print(jfoo(x, torch.device('cuda')).device)  # output : cuda:0
print(jfoo(x, torch.device('cpu')).device)  # output : cuda:0
print("****")
print(thunder.last_prologue_traces(jfoo)[-1])
print()
print(thunder.last_traces(jfoo)[-1])

```

Output for both execution of `jfoo` is `cuda:0`  while it should be `cuda:0` and `cpu`.

**Cause**: 

It can be seen in the prologue trace that we don't verify the value of `device`. So during the second call to `jfoo`, we retrieve the previous computation trace (which has baked in the previous value of device).

Prologue Trace:
```python
# Constructed by Transform for execution (took 1 milliseconds)
import torch
from thunder.executors.torchex import no_autocast

@torch.no_grad()
@no_autocast
def prologue(*args, **kwargs):
  # args: "Any"
  check_len(args, 2)
    # prims.check_len(args, 2)
  # kwargs: "Any"
  check_len(kwargs, 0)
    # prims.check_len(kwargs, 0)
  x: "cpu f32[4]" = args[0]
  check_tensor_metadata(x, (4,), 'cpu', torch.float32, False)
    # prims.check_tensor_metadata(x, (4,), 'cpu', torch.float32, False)
  cache_info: "Any" = thunder._get_cache_info()
  cache_info_is_autocast_enabled: "bool False" = cache_info['is_autocast_enabled']
  check_number_type_and_value(cache_info_is_autocast_enabled, False)
    # prims.check_number_type_and_value(cache_info_is_autocast_enabled, False)
  cache_info_no_grad_sync: "bool False" = cache_info['no_grad_sync']
  check_number_type_and_value(cache_info_no_grad_sync, False)
    # prims.check_number_type_and_value(cache_info_no_grad_sync, False)
  return (x,)
```

Computation Trace
```python
# Constructed by Delete Last Used (took 0 milliseconds)
from torch import Tensor
import torch
from thunder.executors.torchex import no_autocast

@torch.no_grad()
@no_autocast
def computation(x):
  # x: "cpu f32[4]"
[...]

PR #556, opened 2024-06-07
jit: add constraint for torch.device as input to jitted fn

kshitij12345:
Fixes: #555 

This PR adds constraints to the prologue trace to check the passed torch.device to jitted function matches the previously traced torch.device.

**NOTE**: We add this constraint even for `SYMBOLIC_VALUES` caching option, as the computation trace may have operations which are supported only on certain devices, so we shouldn't be reusing that trace.

Example:
```python
import thunder
import torch

def foo(x, device):
    return x.to(device)

jfoo = thunder.jit(foo)
x = torch.randn(3, 3)
print(jfoo(x, torch.device('cuda')).device)  # output : cuda:0
print("****")
print(thunder.last_prologue_traces(jfoo)[-1])
```

Prologue Trace (after the PR):
```python
# Constructed by Transform for execution (took 1 milliseconds)
import torch

@torch.no_grad()
@no_autocast
  # args: "Any"
  check_len(args, 2)
    # prims.check_len(args, 2)
  # kwargs: "Any"
  check_len(kwargs, 0)
    # prims.check_len(kwargs, 0)
  x: "cpu f32[3, 3]" = args[0]
  check_tensor_metadata(x, (3, 3), 'cpu', torch.float32, False)
    # prims.check_tensor_metadata(x, (3, 3), 'cpu', torch.float32, False)

  # THIS CHECK IS ADDED BY THE PR
  p0: "<class 'torch.device'>" = args[1]
  check_literal_like(p0, torch.device("cuda"))
    # prims.check_literal_like(p0, torch.device("cuda"))

  check_number_type_and_value(cache_info_is_autocast_enabled, False)
    # prims.check_number_type_and_value(cache_info_is_autocast_enabled, False)
  check_number_type_and_value(cache_info_no_grad_sync, False)
    # prims.check_number_type_and_value(cache_info_no_grad_sync, False)
  return (x,)
```

### DIFF
--- a/thunder/core/jit_ext.py
+++ b/thunder/core/jit_ext.py
@@ -56,6 +56,7 @@ from thunder.core.proxies import (
     DistParallelType,
     proxy,
     Proxy,
+    AnyProxy,
     NumberProxy,
     StringProxy,
     TensorProxy,
@@ -560,6 +561,17 @@ class GeneralJitCtx(MinimalCtx):
         # avoid double registration by skipping if value has a registered proxy.
         if isinstance(uvalue, Proxy) or value.original_value is not value.nothing:
             return uvalue
+        elif isinstance(uvalue, torch.device):
+            co: CACHE_OPTIONS = get_cache_option()
+            p: AnyProxy = proxy(uvalue, history=value.provenance)
+            if co in (CACHE_OPTIONS.CONSTANT_VALUES, CACHE_OPTIONS.SYMBOLIC_VALUES):
+                # NOTE: Even with SYMBOLIC_VALUES, we want to strictly constraint the device as
+                # the computation trace may utilize device specific executors.
+                self.add_constraint((clang.check_literal_like, p, uvalue))
+            elif co in (CACHE_OPTIONS.SAME_INPUT,):
+                raise NotImplementedError(f"Unsupported cache option {co}")
+            else:  # co is CACHE_OPTIONS.NO_CACHING
+                pass
         elif isinstance(uvalue, torch.Tensor):
             # we always want to proxy torch.Tensor, even const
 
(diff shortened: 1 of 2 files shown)


thunder/core/jit_ext.py
```diff
@@ -895,7 +907,7 @@ def proxy_recursion(v):
             v.track_items()
             need_proxy = any(proxy_recursion(i) for i in v.item_wrappers)
         else:
-            need_proxy = isinstance(v.value, torch.Tensor)
+            need_proxy = isinstance(v.value, torch.Tensor) or isinstance(v.value, torch.device)
```
t-vi, line 910: I might be wrong, but could we just add the check but not proxy just like for `int` / `str`?

kshitij12345: It works (locally verified on the test). I have applied it in https://github.com/Lightning-AI/lightning-thunder/pull/556/commits/2073cb41384bcdae00c0baeaddbfbd95dd251ea7. Is this what you had in mind? Thanks!

thunder/core/jit_ext.py
```diff
@@ -560,6 +561,17 @@ def proxify(self, value: WrappedValue) -> Any:
         # avoid double registration by skipping if value has a registered proxy.
         if isinstance(uvalue, Proxy) or value.original_value is not value.nothing:
             return uvalue
+        elif isinstance(uvalue, torch.device):
+            co: CACHE_OPTIONS = get_cache_option()
+            p: AnyProxy = proxy(uvalue, history=value.provenance)
+            if co in (CACHE_OPTIONS.CONSTANT_VALUES, CACHE_OPTIONS.SYMBOLIC_VALUES):
+                # NOTE: Even with SYMBOLIC_VALUES, we want to strictly constraint the device as
+                # the computation trace may utilize device specific executors.
+                self.add_constraint((clang.check_literal_like, p, uvalue))
```
kshitij12345, line 570: cc: @jjsjann123 for opinion and sanity check.

jjsjann123: :clap:

jjsjann123 (approved):
LGTM

PR #559, opened 2024-06-07
Ignore checkpoint_activations for Thunder in benchmark_litgpt.py

IvanYashchuk:
We need to see the real error from Thunder and not from the artificial barrier.

cc @crcrpar


crcrpar (approved):

t-vi (approved):
Thank you @IvanYashchuk @crcrpar

PR #562, opened 2024-06-08
Added lookside for torch.compile

### LINKED ISSUE #186: Add a lookaside for `torch.compile` and investigate interpreter failure
Earlier, there was confusion about how to use `torch.compile` with thunder. And so, someone tried to jit a model with a call to `torch.compile` in it. This did not work, and the result is still an interpreter stack trace spew. Ideally it should fail more gracefully. 

Reproduce with the following:
```py
import torch
import thunder

def model(a, b, c):
  return a @ b + c

def jit_me(x):
  cfn = torch.compile(model)
  return cfn(x)

jcfn = thunder.jit(jit_me)

x = torch.randn(2, 2)
y = torch.randn(2, 2)
z = torch.randn(2, 2)
res = jcfn(x, y, z)
```

Two things should be done. First, to provide a better error message, you can add a lookaside for `torch.compile`. The convention is to use `do_raise()` for exceptions thrown from inside the code, and `raise` for interpreter errors. This is an interpreter error, so you would just `raise` an exception with an informative error message about how to use the full `torch.compile` executor. Which is still in the works. So, a placeholder would suffice, at least until #140 is merged.

Second, it would be nice to understand what the stack trace means. If there's a bug in the interpreter, we would like to know about it. We know that `torch.compile` does plenty of trickery that we can't model. But it would be nice to know exactly where we're getting tripped up, and why.

cc @apaz-cli

Rana-KV:
Added a lookaside function `torch_compile_lookaside` in `thunder/core/jit_ext.py` to handle cases where `torch.compile` is used within a function intended for JIT compilation.

<details>

- [X] Was this discussed/approved via a Github issue? (no need for typos and docs improvements) - Yes
- [X] Did you read the [contributor guideline](https://github.com/Lightning-AI/pytorch-lightning/blob/main/.github/CONTRIBUTING.md), Pull Request section? - Yes
- [X] Did you make sure to update the docs? - No
- [X] Did you write any new necessary tests? - No

</details>

## What does this PR do?

Fixes #186 

- Added a lookaside function `torch_compile_lookaside` in `thunder/core/jit_ext.py` to handle cases where `torch.compile` is used within a function intended for JIT compilation.
- The lookaside raises a `NotImplementedError` with a clear message, guiding users to remove the `torch.compile` call or apply it outside the function.
- Ensured the function is decorated with `@general_jit_lookaside(torch.compile)` to integrate it properly with Thunder's JIT infrastructure.

## PR review

## Did you have fun?

Sure, did!

### DIFF
--- a/thunder/core/jit_ext.py
+++ b/thunder/core/jit_ext.py
@@ -774,6 +774,12 @@ def _general_jit_setattr_lookaside(obj: Any, name: str, value: Any):
         return res
     return res
 
+@general_jit_lookaside(torch.compile)
+def torch_compile_lookaside(*args, **kwargs):
+    return do_raise(NotImplementedError(
+        "Using torch.compile within a function to be JIT-compiled by Thunder is not supported. "
+        "Please remove the call to torch.compile or apply it outside the function."
+    ))
 
 # TODO Expand on this
 @interpreter_needs_wrap
--- a/thunder/core/jit_ext.py
+++ b/thunder/core/jit_ext.py
@@ -774,12 +774,16 @@ def _general_jit_setattr_lookaside(obj: Any, name: str, value: Any):
         return res
     return res
 
+
 @general_jit_lookaside(torch.compile)
 def torch_compile_lookaside(*args, **kwargs):
-    return do_raise(NotImplementedError(
-        "Using torch.compile within a function to be JIT-compiled by Thunder is not supported. "
-        "Please remove the call to torch.compile or apply it outside the function."
-    ))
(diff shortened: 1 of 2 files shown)


t-vi:
Hi @Rana-KV Thank you! Looks good, but could you add a test in test_jit_general.py, please?

Rana-KV:
hi @t-vi, Thanks for the review. I have updated the testcase.

t-vi (approved):
Thank you @Rana-KV

PR #565, opened 2024-06-10
Fix `unpack_trivial` usage in interpreter

riccardofelluga:
<details>

</details>

## What does this PR do?

This PR adds the name argument when using `prims.unpack_trivial` as it is key for the later CSE pass to understand unique inputs.

## PR review

This is linked to #543 and the other CSE issue #397.

### DIFF
--- a/thunder/core/jit_ext.py
+++ b/thunder/core/jit_ext.py
@@ -1500,12 +1500,12 @@ def process_recorded_modifications(ctx, epilogue_trace):
 
 def bind_inputs(name, trace, input_vars, input_proxies):
     # Unpacks inputs into the computation trace
-    # TODO This currently does the unpacks at the end of he trace, then moves them to the beginning, there's
+    # TODO This currently does the unpacks at the end of the trace, then moves them to the beginning, there's
     #   almost certainly a more elegant way to do this
     with tracectx(trace):
         p: Proxy
         for p in input_proxies:
-            prims.unpack_trivial(p)
+            prims.unpack_trivial(p, name=p.name)
 
     bsyms = trace.bound_symbols
     trace.bound_symbols = bsyms[-len(input_proxies) :] + bsyms[: -len(input_proxies)]
--- a/thunder/core/prims.py
+++ b/thunder/core/prims.py
@@ -634,6 +634,7 @@ def unpack_trivial_impl(x: Any, /, *, name: str | None = None) -> Any:
 
 
 def unpack_trivial_meta(x: Any, /, *, name: str | None = None) -> Any:
+    utils.check(name is not None, lambda: "Expected name argmument to not be None")
     return _collectify(x, name=name)
 
 
--- a/thunder/functional.py
+++ b/thunder/functional.py
@@ -325,7 +325,7 @@ def _eager_unpacking_interpreter(
         for name, x in si.args:
             p: Proxy
             p, _ = _eager_unpack(x, name, co=co)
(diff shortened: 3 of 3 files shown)


thunder/core/jit_ext.py
```diff
@@ -1500,12 +1500,12 @@ def process_recorded_modifications(ctx, epilogue_trace):
...
 def bind_inputs(name, trace, input_vars, input_proxies):
     # Unpacks inputs into the computation trace
-    # TODO This currently does the unpacks at the end of he trace, then moves them to the beginning, there's
+    # TODO This currently does the unpacks at the end of the trace, then moves them to the beginning, there's
     #   almost certainly a more elegant way to do this
     with tracectx(trace):
         p: Proxy
         for p in input_proxies:
-            prims.unpack_trivial(p)
+            prims.unpack_trivial(p, name=p.name)
```
t-vi, line 1508: Would it make sense to fail when name is not given?

riccardofelluga: It might! Because without name it's hard to identify the input at later stages. Should the check mechanism be located in the implementation of the prim?

t-vi: Anything that errors with the call to the prim in the backtrace... I think bind_postprocess might be a good location.

IvanYashchuk: `bind_postprocess` is a callback that is called after the bound symbol is created. The backtrace might not be that nice, I think Riccardo's idea of putting the check in the implementation or meta of the prim would work better and more consistent with all other checks in the codebase, I don't think we have any errors put into bind_postprocess today.

IvanYashchuk: That's a very interesting bind_postprocess:
https://github.com/Lightning-AI/lightning-thunder/blob/20aa0bd51cfbf9c9786bf12b492268cc5b721c79/thunder/core/prims.py#L659-L661
We call this Symbol with some proxies that are then put in the BoundSymbol as output and its args are cleared. What would be a less special way to do this?

IvanYashchuk: Why do we need unpack_trivial at all?

t-vi: > bind_postprocess is a callback that is called after the bound symbol is created. The backtrace might not be that nice, I think Riccardo's idea of putting the check in the implementation or meta of the prim would work better and more consistent with all other checks in the codebase, I don't think we have any errors put into bind_postprocess today.

True! Thank you.

> that's a very interesting bind_postprocess:
maybe with requiring the name, we could drop this and just create it with empty inputs.

> Why do we need unpack_trivial at all?

We use `unpack_trivial` to let every proxy have a creator. I'm not attached to it but it seems a reasonable invariant.

t-vi: But so for this patch, should we add a check to the meta and merge?

t-vi (approved):

thunder/core/prims.py
```diff
@@ -634,6 +634,7 @@ def unpack_trivial_impl(x: Any, /, *, name: str | None = None) -> Any:
 
 
 def unpack_trivial_meta(x: Any, /, *, name: str | None = None) -> Any:
+    utils.check(name is not None, lambda: "Expected name argmument to not be None")
```
IvanYashchuk, line 637: The type hint should be updated as well and the default value removed `name: str | None = None` -> `name: str`

PR #571, opened 2024-06-11
Fix undefined variables

kshitij12345:


t-vi (approved):
Oops, than you @kshitij12345

PR #572, opened 2024-06-11
Added `torch.baddbmm` operator

### LINKED ISSUE #326: Implement _VariableFunctionsClass.baddbmm of torch
## 🚀 Feature

Implement [`torch.baddbmm`](https://pytorch.org/docs/stable/generated/torch.baddbmm.html)

### Motivation

NeMo [Vision Transformer](https://github.com/NVIDIA/NeMo/blob/main/examples/vision/vision_transformer/megatron_vit_classification_pretrain.py)

cc @apaz-cli @tfogal

riccardofelluga:
<details>

</details>

## What does this PR do?

This PR works towards supporting NeMo models, in particular fixes #326. I still think that the version mentioned in the issue is the inplace one, however this op is needed.

## Did you have fun?

One more op! 🚀

cc @apaz-cli @tfogal

### DIFF
--- a/thunder/tests/opinfos.py
+++ b/thunder/tests/opinfos.py
@@ -6156,25 +6156,49 @@ nn_ops = []
 def baddbmm_sample_generator(op, device, dtype, requires_grad, **kwargs):
     make = partial(make_tensor, device=device, dtype=dtype, requires_grad=requires_grad)
     cases = (
-        ((0, 0, 0), (0, 0, 0), (0, 0, 0)),
-        ((3, 0, 5), (3, 0, 0), (3, 0, 5)),
-        ((0, 5, 6), (0, 5, 0), (0, 0, 6)),
-        ((3, 5, 6), (3, 5, 0), (3, 0, 6)),
-        ((3, 5, 6), (3, 5, 8), (3, 8, 6)),
+        ((0, 0, 0), (0, 0, 0), (0, 0, 0), 0.0, None),
+        ((3, 0, 5), (3, 0, 0), (3, 0, 5), 0.0, None),
+        ((0, 5, 6), (0, 5, 0), (0, 0, 6), 0.0, None),
+        ((3, 5, 6), (3, 5, 0), (3, 0, 6), 0.0, None),
+        ((3, 5, 6), (3, 5, 8), (3, 8, 6), 0.0, None),
+        ((3, 5, 6), (3, 5, 8), (3, 8, 6), 0.25, float("inf")),
+        ((3, 5, 6), (3, 5, 8), (3, 8, 6), 0.25, float("-inf")),
+        ((3, 5, 6), (3, 5, 8), (3, 8, 6), 0.25, float("nan")),
     )
 
-    constants_cases = (
+    int_constants_cases = ((2, 2), (1, 0), (0, 1))
+
+    float_constants_cases = (
         (2.0, 2.0),
         (1.0, 0.0),
         (0.0, 1.0),
+        (float("inf"), 2.0),
+        (2.0, float("inf")),
+        (float("inf"), float("inf")),
+        (float("-inf"), 2.0),
+        (2.0, float("-inf")),
+        (float("-inf"), float("-inf")),
+        (float("nan"), 2.0),
+        (2.0, float("nan")),
+        (float("nan"), float("nan")),
     )
 
-    for shape_in, shape_batch1, shape_batch2 in cases:
-        for alpha, beta in constants_cases:
-            if isinstance(to_dtype(dtype), datatypes.exact):
-                alpha, beta = int(alpha), int(beta)
+    for shape_in, shape_batch1, shape_batch2, singularity_amount, singularity_value in cases:
+        a = make(shape_in)
+        b1 = make(shape_batch1)
+        b2 = make(shape_batch2)
 
-            yield SampleInput(make(shape_in), make(shape_batch1), make(shape_batch2), alpha=alpha, beta=beta)
+        if isinstance(to_dtype(dtype), datatypes.exact):
+            for alpha, beta in int_constants_cases:
+                yield SampleInput(make(shape_in), make(shape_batch1), make(shape_batch2), alpha=alpha, beta=beta)
+        else:
+            if singularity_value is not None:
+                a = replace_random_percentage(a, singularity_value, singularity_amount)
+                b1 = replace_random_percentage(b1, singularity_value, singularity_amount)
+                b2 = replace_random_percentage(b2, singularity_value, singularity_amount)
+
+            for alpha, beta in float_constants_cases:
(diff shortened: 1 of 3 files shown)


thunder/torch/__init__.py
```diff
@@ -3041,6 +3041,26 @@ def batch_norm(
...
+    utils.check_same_dtype(a, b1, b2)
+    utils.check(b1.ndim == 3, lambda: f"batch1 must be a 3D tensor, found {b1.ndim} instead.")
+    utils.check(b2.ndim == 3, lambda: f"batch2 must be a 3D tensor, found {b2.ndim} instead.")
+
+    if a.dtype not in dtypes.inexact_dtypes:
+        utils.check_type(beta, int)
+        utils.check_type(alpha, int)
+
+    t0 = matmul(b1, b2)
+    t1 = mul(alpha, t0)
```
nikitaved, line 3059: no need to `mul` for the same reason.

thunder/torch/__init__.py
```diff
@@ -3041,6 +3041,26 @@ def batch_norm(
 #
 
 
+@torchsymbol(torch.baddbmm, is_method=True)
+def baddbmm(
+    a: TensorLike, b1: TensorLike, b2: TensorLike, *, beta: float = 1.0, alpha: float = 1.0, out: TensorLike = None
+) -> TensorLike:
+    utils.check(out is None, lambda: "Non-None out is not supported", NotImplementedError)
+
+    utils.check_same_dtype(a, b1, b2)
```
nikitaved, line 3050: We will probably do shortcuts below, so we should probably check that all things are on the same device.

thunder/tests/opinfos.py
```diff
@@ -6153,6 +6153,63 @@ def tensor_1d_sample_generator(op, device, dtype, requires_grad, **kwargs):
 nn_ops = []
 
 
+def baddbmm_sample_generator(op, device, dtype, requires_grad, **kwargs):
+    make = partial(make_tensor, device=device, dtype=dtype, requires_grad=requires_grad)
+    cases = (
+        ((0, 0, 0), (0, 0, 0), (0, 0, 0)),
+        ((3, 0, 5), (3, 0, 0), (3, 0, 5)),
+        ((0, 5, 6), (0, 5, 0), (0, 0, 6)),
+        ((3, 5, 6), (3, 5, 0), (3, 0, 6)),
```
nikitaved, line 6162: Let's add cases where the matmul part actually produces something other than zero

thunder/torch/__init__.py
```diff
@@ -3041,6 +3041,26 @@ def batch_norm(
...
+    utils.check(b1.ndim == 3, lambda: f"batch1 must be a 3D tensor, found {b1.ndim} instead.")
+    utils.check(b2.ndim == 3, lambda: f"batch2 must be a 3D tensor, found {b2.ndim} instead.")
+
+    if a.dtype not in dtypes.inexact_dtypes:
+        utils.check_type(beta, int)
+        utils.check_type(alpha, int)
```
nikitaved, line 3056: Could you please elaborate on these lines?

thunder/torch/__init__.py
```diff
@@ -3041,6 +3041,26 @@ def batch_norm(
...
+
+    t0 = matmul(b1, b2)
```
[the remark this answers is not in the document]
riccardofelluga: That's true, however I've noticed that it broke the gradient computation. Let me see what I can do

thunder/torch/__init__.py
```diff
@@ -3041,6 +3041,26 @@ def batch_norm(
 #
 
 
+@torchsymbol(torch.baddbmm, is_method=True)
+def baddbmm(
+    a: TensorLike, b1: TensorLike, b2: TensorLike, *, beta: float = 1.0, alpha: float = 1.0, out: TensorLike = None
```
mruberry, line 3046: `, /, *,` let's not allow binding to names like `a`, `b1` and `b2`

thunder/torch/__init__.py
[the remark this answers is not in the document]
mruberry: Curious what's going on with that; also fyi @IvanYashchuk

thunder/torch/__init__.py
mruberry: nit: `alpha * t0`

thunder/torch/__init__.py
```diff
@@ -3041,6 +3041,26 @@ def batch_norm(
...
+    utils.check(b2.ndim == 3, lambda: f"batch2 must be a 3D tensor, found {b2.ndim} instead.")
+
+    if a.dtype not in dtypes.inexact_dtypes:
+        utils.check_type(beta, int)
+        utils.check_type(alpha, int)
+
+    t0 = matmul(b1, b2)
+    t1 = mul(alpha, t0)
+
+    return add(t1, a, alpha=beta)
```
mruberry, line 3061: nit: `t1 + (a * beta)`

mruberry:
Looks really good so far, @riccardofelluga! Just some minor tweaks

thunder/torch/__init__.py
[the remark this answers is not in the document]
tfogal: it is odd to break gradient...

but, i would think the matmul call itself would do this optimization if possible, and it'll do so with fewer instructions. If we want to do this, I'd recommend doing it at a higher level, via (say) a constant propagation pass, which could remove the op entirely and then potentially DCE some tensors.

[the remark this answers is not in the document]
mruberry: You're absolutely right that the logic could be handled by an optimization pass, @tfogal. I don't think we currently perform numeric simplifications. 

Also, is it true that we don't need to perform this matmul if `alpha` is zero? What if one or both of the tensors contained NaNs?

thunder/tests/opinfos.py
```diff
@@ -6152,6 +6152,72 @@ def tensor_1d_sample_generator(op, device, dtype, requires_grad, **kwargs):
...
+        ((3, 5, 6), (3, 5, 8), (3, 8, 6)),
+    )
+
+    constants_cases = (
+        (2.0, 2.0),
+        (1.0, 0.0),
+        (0.0, 1.0),
+    )
```
IvanYashchuk, line 6169: `float('inf')`, `float('-inf')`, `float('nan')` are also important to check besides 0.0, 1.0, and 2.0. Could you please add them?

riccardofelluga:
I've added the singularities in the testing. However, I reverted back to the code without the early exits because doing so would require writing the backward implementation for those computation paths. At the moment I think we can live with it and then in the future we might want to implement some optimization pass that does this for us.  

cc @nikitaved @mruberry @tfogal

mruberry (approved):

tfogal (approved):

Issue #327, opened 2024-05-01
Implement torch.device 


athitten:
### 🚀 Feature

Implement [`torch.device`](https://pytorch.org/docs/stable/tensor_attributes.html#torch.device:~:text=float_tensor%20*%3D%20complex_float_tensor-,torch.device,-CLASS)

### Motivation

NeMo [Vision Transformer](https://github.com/NVIDIA/NeMo/blob/main/examples/vision/vision_transformer/megatron_vit_classification_pretrain.py)

cc @tfogal

tfogal:
would like to discuss at triage review: can we just say `cuda` for everything? what about tensors from the outside (i.e. input tensors) that are on `cpu` or even things like `meta`?

mruberry:
triage review — if the jit could annotate this function as constexpr and detect that the input is constexpr it could just call it

Issue #463, opened 2024-05-26
Constraints to insert static numbers 


jjsjann123:
## 🚀 Feature

prims operations should be able to insert contraints on prologue trace.

### Motivation

Issue first brought up here: https://github.com/Lightning-AI/lightning-thunder/issues/262#issuecomment-2127729530

1. problem arising from slice -> dynamic shape support?

Looking at the computation trace here for a `slice_prim`:
```
    # t18 = prims.slice_prim(a, [0, 0, 0], [8, 16, i8], [1, 1, 1])  # t18: "cuda:0 f32[8, 16, [IntegerProxy name=i17, value=2]]"
```
We notice that the output TensorProxy has a shape of `[8, 16, [IntegerProxy name=i17, value=2]]`, where the last dimension with size `i17` is not produced in the trace. This is because prim has arithmetic operation defined inside meta functions, as defined here:
https://github.com/Lightning-AI/lightning-thunder/blob/b7154dc77b5bf8a02bcbc2e56310c62fd75808ff/thunder/core/prims.py#L3102

Ideally we would want to expose that logic in the trace in order to support dynamic shape, we should lift the shape logic to clang instead of meta. (See note 1 in alternatives)

2. operations that would to insert static constraints

operations should be able to insert constraints by marking certain NumberProxy/TensorProxy dimensions as static shaped. e.g. axes args for a reduction operation could be a dynamic NumberProxy, which would instead want to bake in as static.

This would also allow us to temporarily resolve issue regarding prim operation which we have not yet added dynamic shape support.

In the slice_prim example above, we could have marked `i8` as a static and adding a constraint as `i8 == 2`. This would allow us to have slice working without proper dynamic shape support. Which would be nice to keep #262 moving forward.

### Pitches

We can use `check_number_type_and_value` prim for the check. The challenge here is to propagate the proper check into prologue trace.

Taking the slice example above, the operation would want to `check_number_type_and_value(i8, pyval(i8))`. But there's not guarantee that it's legit operation in prologue trace, which would require i8 to be a direct input. So we would want to project `static_constraints` from `i8` to its producers, which could be `NumberProxy` as well as `TensorProxy`'s attributes.

### Alternatives

note 1:
(Alternatively we can think about having logic to trace through prim_meta functions.)

jjsjann123:
Another example coming from https://github.com/Lightning-AI/lightning-thunder/pull/575#issuecomment-2161257731
```
import thunder
import torch

[...]

PR #575, opened 2024-06-11
Support torch.device

kshitij12345:
Fixes: https://github.com/Lightning-AI/lightning-thunder/issues/327

This PR enables using `torch.device` in the jitted function. For this PR, we bake the device into the trace rather than creating a symbol for `torch.device`.

Example:
```python
import thunder
import torch

def foo(dev, idx, t):
    return torch.ones(1, device=torch.device(dev, idx)) + torch.ones(1, device=t.device) * 2

jfoo = thunder.jit(foo)
# jfoo = thunder.jit(foo, cache=thunder.CACHE_OPTIONS.SYMBOLIC_VALUES)
print(jfoo("cuda", 0, torch.randn(3, device='cuda')))

# print(thunder.last_prologue_traces(jfoo)[-1])

```

Trace
```
import torch

@torch.no_grad()
@no_autocast
def computation():
  [t5] = nvFusion0()
    # t2 = prims.full((1,), 1, device=devices.Device("cuda:0"), dtype=dtypes.int64_)  # t2: "cuda:0 i64[1]"
    # result = prims.mul(t2, 2)  # result: "cuda:0 i64[1]"
    # t5 = prims.add(t2, result)  # t5: "cuda:0 i64[1]"
  return t5
```

--------

**Supporting `SYMBOLIC_VALUES` (in future)** - 
Above approach of baking device into the trace works fine with the current caching scheme but this will lead to issue with SYMBOLIC_VALUES caching mechanism (below snippet will fail with `SYMBOLIC_VALUES`). I think one way to support this will be to make `torch.device` actually a symbol which can take `int` or `IntegerProxy` and return a device proxy for consumption. However, I am not sure if this is the best approach. @jjsjann123 would love to know your thoughts and suggestions.

```python
import thunder
import torch

def foo(dev, idx):
    return torch.ones(1, device=torch.device(dev, idx))

# jfoo = thunder.jit(foo)  # works
jfoo = thunder.jit(foo, cache=thunder.CACHE_OPTIONS.SYMBOLIC_VALUES)
print(jfoo("cuda", 0))  # fails as our current implementation can't handle `IntegerProxy`.
print(jfoo("cuda", 1))

# print(thunder.last_prologue_traces(jfoo)[-1])
# print(thunder.last_traces(jfoo)[-1])
```

--------

**Testing** - 

Since, `torch.device` is not aTensor Operation (i.e. a function that takes or returns a Tensor) and also since this is not implemented as a Symbol in trace, I haven't used the OpInfo mechanism but instead written the tests in `test_core.py`.


jjsjann123:
I'm interesting in seeing how the trace is constructed with `symbolic values`

> I think one way to support this will be to make torch.device actually a symbol which can take int or IntegerProxy and return a device proxy for consumption. However, I am not sure if this is the best approach. @jjsjann123 would love to know your thoughts and suggestions.

I'm not seeing use in having device as dynamic input. I haven't really seeing deivce changes across program. I think it's safe to just bake them in as-is. We probably  cc'ing @tfogal @t-vi  since they have seen a lot more models than I do.

I think we can just bake them in as constant. which I thought is already the case when they are given as inputs: https://github.com/Lightning-AI/lightning-thunder/blob/067f15aae47ad71229732ca6c35a5d190135e48c/thunder/core/jit_ext.py#L570

So I think this is yet another case where we need to have `torch/__init__.py::device` inserting a constraint.

tfogal:
> I'm interesting in seeing how the trace is constructed with `symbolic values`
> 
> > I think one way to support this will be to make torch.device actually a symbol which can take int or IntegerProxy and return a device proxy for consumption. However, I am not sure if this is the best approach. @jjsjann123 would love to know your thoughts and suggestions.
> 
> I'm not seeing use in having device as dynamic input. I haven't really seeing deivce changes across program. I think it's safe to just bake them in as-is. We probably cc'ing @tfogal @t-vi since they have seen a lot more models than I do.

I agree; AFAICT when devs want to change devices, they take some existing thing and `to()` it to the new device. At that point, whether they invoke `device` on the pre-`to()`'d thing or the `to()`'s result, the device is effectively a constant.

Even when the device is an argument, as it is in the `foo` here, over the long run I'd expect thunder would want to do a different kind of compile for different devices. That is, you'd probably want an entirely different cost model if invoked as `jfoo("cpu", ...)` rather than `jfoo("cuda:0", ...)`

tfogal (approved):

kshitij12345:
> I'm interesting in seeing how the trace is constructed with symbolic values

Snippet
```python
import thunder
import torch

def foo(dev, idx):
    return torch.ones(1, device=torch.device(dev, idx))

# jfoo = thunder.jit(foo)  # works
jfoo = thunder.jit(foo, cache=thunder.CACHE_OPTIONS.SYMBOLIC_VALUES)
print(jfoo("cuda", 0))  # fails as our current implementation can't handle `IntegerProxy`.
print(jfoo("cuda", 1))

# print(thunder.last_prologue_traces(jfoo)[-1])
# print(thunder.last_traces(jfoo)[-1])
```

Currently, the snippet fails with error -> `ValueError: [IntegerProxy name=i1, value=0] had an unexpected type <class 'thunder.core.proxies.IntegerProxy'>. Supported types are <class 'int'>` as this implementation of `torch.device` doesn't handle `IntegerProxy`.

Since, this snippet is now tracked in #463, we can fix this separately in another PR.

thunder/tests/test_core.py
```diff
@@ -2656,3 +2656,73 @@ def foo_torch(x):
...
+            expected = foo(input)
+            assert actual.device == expected.device
+
+    # Test with str input
+    device_strs = ("cpu", "cuda", "cuda:0", "meta")
+
+    def foo1(dev):
+        # If we return the device here, thunder.jit version will return `thunder.device`
+        # while eager will return `torch.device`
+        # https://github.com/Lightning-AI/lightning-thunder/issues/573
```
mruberry, line 2681: Great comment and nice link to the issue

thunder/torch/__init__.py
```diff
@@ -4371,6 +4371,33 @@ def softmax(a: TensorLike, dim: int, dtype: None | dtypeLike = None, _stacklevel
     return _softmax(a, dim=dim, dtype=dtype)
 
 
+def torch_device(device_or_str: DeviceLike, /, index: int | None = None) -> devices.Device:
+    if isinstance(device_or_str, (devices.Device, torch.device)):
+        # PyTorch behavior:
+        # >>> torch.device(torch.device("cuda"), 0)
+        # TypeError: device(): argument 'type' (position 1) must be str, not torch.device
```
mruberry, line 4378: Nice detailed handling of input types

thunder/torch/__init__.py
```diff
@@ -4371,6 +4371,33 @@ def softmax(a: TensorLike, dim: int, dtype: None | dtypeLike = None, _stacklevel
...
+        utils.check(index is None, lambda: f"device(): `index` is only allowed when `device` is a `str`.")
+        return to_device(device_or_str)
+
+    # NOTE: device_or_str is `str`
```
mruberry, line 4382: Should this function explicitly check for this?

thunder/torch/__init__.py
```diff
@@ -4371,6 +4371,33 @@ def softmax(a: TensorLike, dim: int, dtype: None | dtypeLike = None, _stacklevel
...
+    if index is not None:
+        # PyTorch behavior:
+        # >>> torch.device("cuda:0", 0)
+        # RuntimeError: type (string) must not include an index because index was passed explicitly: cuda:0
+        has_device_idx = len(device_or_str.split(":")) > 1
+        utils.check(
+            not has_device_idx,
+            lambda: f"device string must not include an index because index was passed explicitly: {device_or_str}",
```
mruberry, line 4390: Nice special case again

thunder/torch/__init__.py
```diff
@@ -4371,6 +4371,33 @@ def softmax(a: TensorLike, dim: int, dtype: None | dtypeLike = None, _stacklevel
...
+        )
+
+    return devices.Device(device_or_str, index)
+
+
+# We don't use @torchsymbol as we don't want `torch.device()` to appear in trace as a symbol.
```
mruberry, line 4396: Great comment

thunder/torch/__init__.py
```diff
@@ -4371,6 +4371,33 @@ def softmax(a: TensorLike, dim: int, dtype: None | dtypeLike = None, _stacklevel
...
+# Because of this, we need to manually register the implementation.
+_torch_to_thunder_function_map[torch.device] = torch_device
```
mruberry, line 4398: We should think about adding a layer of indirection here, like a `register_function()` -- but not for this PR

mruberry (approved):
Awesome! Great looking ahead to consider how symbolic values may need to be constrained when passed to `torch.device()` calls

mruberry:
@t-vi this will probably require some help merging, looks like there are some unrelated sdpa test failures

Issue #145, opened 2024-04-09
Handling inplace through SSA


t-vi:
This issue is to facilitate discussion of inplace handling, namely the "big" solution of having a static single assignment (SSA) representation.

For any handling of inplace, we want to make certain that two things are achieved:
- we don't want to take shortcuts that complicate passes by introducing the need to detect obstacles to optimizations, because it would harm usability and extensibility of Thunder.
-  we don't want to create ad-hoc band-aids to get things working that we would need to regress on later to introduce more proper handling because developing in the open more or less means no regressions.

Some thoughts from video/chat discussions:

About the problem:
- The key difficulty in SSA is that we would need to keep track of which tensors get modified by an inplace update (i.e. which
 have memory that is to be updated), so we would need to know about views (the fancy term is *alias analysis*),
- this is difficult for some things in PyTorch (i.e. `reshape`),
- "assuming the worst" works to some extend.

Solution considerations:
- Likely we would want inplace updates to have all affected tensors as outputs.
- on inputs we would need to check for aliases as part of the prologue (maybe with a separate "assume aliasing is the OK" cache mode or sorts later),
- operations need to know if their output is a view of their inputs (difficult for `reshape`, easy for most others),
- initially, we would only check if tensors share storage,
- likely the translation could be done in the interpretation phase,
- we would need to have versioning / disambiguation of versions for tensor proxies during this, but not when we have the SSA.

Later versions could refine the alias analysis as needed.

@tfogal @mruberry @IvanYashchuk

mruberry:
One thing I've struggled with in these discussions is how does static single assignment form address the challenges of inplace operations? Doesn't Thunder's IR essentially already have the SSA property (I think there are some cases where we model operations which do nothing as the equivalent of `x = x`, but I don't think they're a problem)?

mruberry:
> Well, we do have SSA form in the Thunder IR until we allow inplace operations, which seems to be one of the things people want to do. The transformation to SSA is just there to make sure we don't run into correctness issues.

OK; I guess I think about this as "let's handle inplace operations while preserving our SSA form and relying on dataflow to determine the order of operations"?

t-vi:
Exactly, or even "let's preserve our SSA form and the fact that dataflow describes the order of operations (and admissable reorderings) even when we want inplace".

mruberry:
Maybe we can to be too conservative with operations like `reshape` and for ambiguous cases act like they (might have) created a view?

[...]

PR #584, opened 2024-06-12
Functionalize in-place ops

crcrpar:
## What does this PR do?

Express in-place ops using their out-of-place counterpart with following `prims.copy_`. e.g. `a.add_(b)` as `prims.copy_(prims.add(a, b), a)`, then let the added transform functionalize the trace by removing the trailing copy amd updating the signature. 
Let’s say we have `t.exp_()` in a script and `t` is used afterwards, thunder translates it into `t_out = ltorch.exp_(t)`. This bound symbol has two sub bound symbols: `t0 = ltorch.exp(t)` and `t_out = prims.copy_(t0, t)`. The functionalization removes the copy, and replaces `ltorch.exp_(t)` with `t0 = ltorch.exp(t)` and `t` uses after `exp_` with `t0`.

The covered ops are ones that either (a) have in-place variant such as `torch.exp` and `torch.add` or (b) have `inplace` as one of their args such as `torch.nn.functional.relu`.
 
So this would not cover the entire #145 broadly, nor take aliases into considerations.

### DIFF
--- a/thunder/__init__.py
+++ b/thunder/__init__.py
@@ -510,10 +510,8 @@ def jit(
             prologue_traces = [prologue_trc]
             computation_traces = [computation_trc]
             if not compile_options.get("skip_inplace_functionalization", False):
-                computation_trc = functionalize_inplace_ops(
-                    computation_trace=computation_trc,
-                    computation_traces=computation_traces,
-                )
+                computation_traces.extend(functionalize_inplace_ops(computation_trace=computation_trc))
+                computation_trc = computation_traces[-1]
 
             if epilogue_trc is not None:
                 epilogue_traces = [epilogue_trc]
--- a/thunder/__init__.py
+++ b/thunder/__init__.py
@@ -510,10 +510,8 @@ def jit(
             prologue_traces = [prologue_trc]
             computation_traces = [computation_trc]
             if not compile_options.get("skip_inplace_functionalization", False):
-                print(f"# Before functionalization\n{computation_trc}")
                 computation_traces.extend(functionalize_inplace_ops(computation_trace=computation_trc))
                 computation_trc = computation_traces[-1]
-                print(f"# After functionalization\n{computation_trc}")
 
             if epilogue_trc is not None:
                 epilogue_traces = [epilogue_trc]
--- a/thunder/__init__.py
+++ b/thunder/__init__.py
@@ -33,7 +33,13 @@ from thunder import functional as functional
 import thunder.core.prims as prims
 import thunder.core.dtypes as dtypes
 import thunder.core.devices as devices
-from thunder.core.transform_common import dce, EarlyTransform, AdditionalTransform, PostOptimizationTransform
+from thunder.core.transform_common import (
+    dce,
+    EarlyTransform,
+    AdditionalTransform,
+    PostOptimizationTransform,
+    functionalize_inplace_ops,
+)
 from thunder.common import (
     CompileData,
     CompileStats,
@@ -503,6 +509,11 @@ def jit(
 
             prologue_traces = [prologue_trc]
             computation_traces = [computation_trc]
+            if not compile_options.get("skip_inplace_functionalization", False):
+                computation_trc = functionalize_inplace_ops(
+                    computation_trace=computation_trc,
+                    computation_traces=computation_traces,
+                )
 
             if epilogue_trc is not None:
                 epilogue_traces = [epilogue_trc]
--- a/thunder/__init__.py
+++ b/thunder/__init__.py
@@ -510,8 +510,10 @@ def jit(
             prologue_traces = [prologue_trc]
             computation_traces = [computation_trc]
             if not compile_options.get("skip_inplace_functionalization", False):
+                print(f"# Before functionalization\n{computation_trc}")
                 computation_traces.extend(functionalize_inplace_ops(computation_trace=computation_trc))
                 computation_trc = computation_traces[-1]
+                print(f"# After functionalization\n{computation_trc}")
 
             if epilogue_trc is not None:
                 epilogue_traces = [epilogue_trc]
--- a/thunder/core/jit_ext.py
+++ b/thunder/core/jit_ext.py
@@ -1744,6 +1744,12 @@ def thunder_general_jit(
 
 
 def functionalize_inplace_ops(computation_trace: TraceCtx) -> TraceCtx:
+    """Functionalize in-place ops in ``computation_trace``.
+
+    This is a two-step function. The first step is to use the output of :func:`thunder.core.prims.copy_`'s
+    outputs as trace's outputs. The second step is to remove :func:`thunder.core.prims.copy_`
+    from :attr:`~thunder.core.symbol.Boundsymbol.subsymbols` if it's the last symbol.
+    """
     from thunder.core import utils
     from thunder.core.proxies import ProxyInterface
     from thunder.core.symbol import VariableInterface
@@ -1758,8 +1764,7 @@ def functionalize_inplace_ops(computation_trace: TraceCtx) -> TraceCtx:
     if not any(is_inplace(bsym) for bsym in computation_trace.bound_symbols):
         return computation_trace
 
-    # Step 1: don't return tensors who have their consumers in any way, i.e., return
-    # the tensors returned from `prims.copy_` as possible.
+    # Step 1: return the tensors returned from `prims.copy_` as possible not the args for clarity.
     bsym: BoundSymbol
     swap_map: dict[VariableInterface, ProxyInterface] = {}
     bsyms: list[BoundSymbol] = []
--- a/thunder/core/jit_ext.py
+++ b/thunder/core/jit_ext.py
@@ -67,7 +67,7 @@ from thunder.core.proxies import (
     unvariableify,
     is_proxy_name_available,
 )
-from thunder.core.trace import set_tracectx, reset_tracectx, tracectx, from_trace, TraceProvenance
+from thunder.core.trace import set_tracectx, reset_tracectx, tracectx, from_trace
 from thunder.core.interpreter import (
     InterpreterLogItem,
     interpret,
--- a/thunder/core/jit_ext.py
+++ b/thunder/core/jit_ext.py
@@ -1576,14 +1576,29 @@ def process_recorded_modifications(ctx, epilogue_trace):
 
 
 def bind_inputs(name, trace, input_vars, input_proxies):
+    from thunder.core import utils
+
     # Unpacks inputs into the computation trace
     # TODO This currently does the unpacks at the end of the trace, then moves them to the beginning, there's
     #   almost certainly a more elegant way to do this
-    with tracectx(trace):
-        p: Proxy
-        for p in input_proxies:
-            prims.unpack_trivial(p, name=p.name)
+    num_orig_bsyms = len(trace.bound_symbols)
 
+    # note(crcrpar): The path looks neat but it does not work for a trace
+    # if it's :func:`thunder.core.jit_ext.functionalize_inplace_ops`'d, :shrug:
+    # with tracectx(trace):
+    #     p: Proxy
+    #     for p in input_proxies:
+    #         prims.unpack_trivial(p)
+
+    for p in input_proxies:
+        bsym = prims.unpack_trivial.bind(p, output=p)
+        trace.add_bound_symbol(bsym)
+
+    num_bsyms = len(trace.bound_symbols)
+    utils.check(
+        num_bsyms == len(input_proxies) + num_orig_bsyms,
+        lambda: f"{num_bsyms=} expected to be {len(input_proxies)=} + {num_orig_bsyms}",
+    )
     bsyms = trace.bound_symbols
     trace.bound_symbols = bsyms[-len(input_proxies) :] + bsyms[: -len(input_proxies)]
 
--- a/thunder/core/jit_ext.py
+++ b/thunder/core/jit_ext.py
@@ -67,7 +67,7 @@ from thunder.core.proxies import (
     unvariableify,
     is_proxy_name_available,
 )
-from thunder.core.trace import set_tracectx, reset_tracectx, tracectx, from_trace
+from thunder.core.trace import set_tracectx, reset_tracectx, tracectx, from_trace, TraceProvenance
 from thunder.core.interpreter import (
     InterpreterLogItem,
     interpret,
@@ -1734,21 +1734,28 @@ def functionalize_inplace_ops(computation_trace: TraceCtx) -> TraceCtx:
     from thunder.core.symbol import VariableInterface
     from thunder.core.symbol import variableify
 
-    swap_map: dict[VariableInterface, ProxyInterface] = {}
-    bound_symbols: list[BoundSymbol] = []
-    print(computation_trace)
+    def is_inplace(bsym: BoundSymbol) -> bool:
+        if (isinstance(bsym.sym.id, str) and bsym.sym.id.endswith("_")) and bsym.subsymbols:
+            return bsym.subsymbols[-1].sym.id == prims.PrimIDs.COPY_
+        else:
+            return False
 
+    if not any(is_inplace(bsym) for bsym in computation_trace.bound_symbols):
+        return computation_trace
+
+    # Step 1: don't return tensors who have their consumers in any way, i.e., return
+    # the tensors returned from `prims.copy_` as possible.
     bsym: BoundSymbol
-    found_copy_in_subsymbols = False
+    swap_map: dict[VariableInterface, ProxyInterface] = {}
+    bsyms: list[BoundSymbol] = []
     for bsym in computation_trace.bound_symbols:
         new_bsym = bsym.from_bsym_swap_proxies(swap_map)
 
-        # This means the symbol of bsym is in-place.
-        if not (isinstance(bsym.sym.id, str) and bsym.sym.id.endswith("_")):
-            bound_symbols.append(new_bsym)
+        # in-place ops has `prims.copy_` as the last subsymbol.
+        if not is_inplace(new_bsym):
+            bsyms.append(new_bsym)
             continue
 
-        found_copy_in_subsymbols = True
         sub_bsyms: list[BoundSymbol] = bsym.subsymbols
         utils.check(sub_bsyms, lambda: f"{bsym.sym.id=} expected to have subsymbols but {bsym.subsymbols=}")
         copy_bsym = sub_bsyms[-1]
@@ -1756,31 +1763,58 @@ def functionalize_inplace_ops(computation_trace: TraceCtx) -> TraceCtx:
             copy_bsym.sym.id == prims.PrimIDs.COPY_,
             lambda: f"bsym.subsymbols[-1] expected to be {prims.PrimIDs.COPY_} but {copy_bsym.sym.id=}",
         )
-        functional_out = copy_bsym.flat_proxy_args[0]
-        copy_dst = copy_bsym.flat_proxy_outs[0]
-        swap_map[variableify(copy_dst)] = functional_out
+        copy_out = copy_bsym.flat_proxy_outs[0]
+        copy_dst = copy_bsym.flat_proxy_args[1]
+        swap_map[variableify(copy_dst)] = copy_out
         # Remove `prims.copy_`
-        new_bsym.subsymbols = new_bsym.subsymbols[:-1]
-        new_bsym = new_bsym.from_bsym_swap_proxies(swap_map, skip_subsymbols=True)
+        new_bsym = new_bsym.from_bsym_swap_proxies(swap_map, skip_inputs=True, skip_subsymbols=True)
+        bsyms.append(new_bsym)
 
+    intermediate_trace = from_trace(computation_trace)
+    intermediate_trace.bound_symbols = bsyms[:]
+    del bsyms
+
+    # Step 2: Remove `prims.copy_` if it's the last one of `bsym.subsymbols`.
+    bsym_inplace_to_functional = {}
+    swap_map.clear()
+    new_bsyms: list[BoundSymbol] = []
+    for bsym in intermediate_trace.bound_symbols:
+        new_bsym = bsym.from_bsym_swap_proxies(swap_map)
+
+        # in-place ops has `prims.copy_` as the last subsymbol.
+        if not is_inplace(new_bsym):
+            new_bsyms.append(new_bsym)
+            continue
         functional_sym_name = new_bsym.sym.id.split(".")[-1][:-1]
         utils.check(
             hasattr(thunder.torch, functional_sym_name),
             lambda: f"{functional_sym_name}, out-of-place impl of {bsym.sym.id=} not found in `thunder.torch` namespace",
         )
+        sub_bsyms: list[BoundSymbol] = bsym.subsymbols
+        utils.check(sub_bsyms, lambda: f"{bsym.sym.id=} expected to have subsymbols but {bsym.subsymbols=}")
+        copy_bsym = sub_bsyms[-1]
+        utils.check(
+            copy_bsym.sym.id == prims.PrimIDs.COPY_,
+            lambda: f"bsym.subsymbols[-1] expected to be {prims.PrimIDs.COPY_} but {copy_bsym.sym.id=}",
+        )
+        swap_map[variableify(copy_bsym.flat_proxy_outs[0])] = copy_bsym.flat_proxy_args[0]
+        new_bsym.subsymbols = new_bsym.subsymbols[:-1]
+        new_bsym = new_bsym.from_bsym_swap_proxies(swap_map)
         functional_sym: Symbol = getattr(thunder.torch, functional_sym_name)
-        new_bsym = functional_sym.bind(
+        new_functional_bsym = functional_sym.bind(
             *new_bsym.args,
             **new_bsym.kwargs,
             output=new_bsym.output,
             subsymbols=new_bsym.subsymbols,
             _call_ctx=new_bsym._call_ctx,
         )
-        bound_symbols.append(new_bsym)
-
-    if not found_copy_in_subsymbols:
-        return computation_trace
+        new_bsyms.append(new_functional_bsym)
+        bsym_inplace_to_functional[new_bsym] = new_functional_bsym
 
     functionalized_computation_trace = from_trace(computation_trace)
-    functionalized_computation_trace.bound_symbols = bound_symbols
+    functionalized_computation_trace.bound_symbols = new_bsyms
+    functionalized_computation_trace.set_provenance(TraceProvenance("Functionalize in-place ops"))
+    # note(crcrpar): I kind of want to do the following two.
+    # functionalized_computation_trace._provenance.swap_map = swap_map
+    # functionalized_computation_trace._provenance.bsym_inplace_to_functional = bsym_inplace_to_functional
     return functionalized_computation_trace
--- a/thunder/core/jit_ext.py
+++ b/thunder/core/jit_ext.py
@@ -1678,8 +1678,6 @@ def thunder_general_jit(
             process_recorded_modifications(ctx, epilogue_trace)
             last_interpreter_log = jfn._last_interpreter_log
 
-    computation_trace = functionalize_inplace_ops(computation_trace)
-
     pro_to_comp, computation_intermediates = get_computation_inputs_and_intermediates(computation_trace)
     epilogue_inputs, _ = get_computation_inputs_and_intermediates(epilogue_trace)
 
@@ -1741,100 +1739,3 @@ def thunder_general_jit(
         )
 
     return TraceResults(prologue_trace, computation_trace, epilogue_trace, last_interpreter_log)
-
-
-def functionalize_inplace_ops(computation_trace: TraceCtx) -> TraceCtx:
-    """Functionalize in-place ops in ``computation_trace``.
-
-    This is a two-step function. The first step is to use the output of :func:`thunder.core.prims.copy_`'s
-    outputs as trace's outputs. The second step is to remove :func:`thunder.core.prims.copy_`
-    from :attr:`~thunder.core.symbol.Boundsymbol.subsymbols` if it's the last symbol.
-    """
-    from thunder.core import utils
-    from thunder.core.proxies import ProxyInterface
-    from thunder.core.symbol import VariableInterface
-    from thunder.core.symbol import variableify
-
-    def is_inplace(bsym: BoundSymbol) -> bool:
-        if (isinstance(bsym.sym.id, str) and bsym.sym.id.endswith("_")) and bsym.subsymbols:
-            return bsym.subsymbols[-1].sym.id == prims.PrimIDs.COPY_
-        else:
-            return False
-
-    if not any(is_inplace(bsym) for bsym in computation_trace.bound_symbols):
-        return computation_trace
-
-    # Step 1: return the tensors returned from `prims.copy_` as possible not the args for clarity.
-    bsym: BoundSymbol
-    swap_map: dict[VariableInterface, ProxyInterface] = {}
-    bsyms: list[BoundSymbol] = []
-    for bsym in computation_trace.bound_symbols:
-        new_bsym = bsym.from_bsym_swap_proxies(swap_map)
-
-        # in-place ops has `prims.copy_` as the last subsymbol.
-        if not is_inplace(new_bsym):
-            bsyms.append(new_bsym)
-            continue
-
-        sub_bsyms: list[BoundSymbol] = bsym.subsymbols
-        utils.check(sub_bsyms, lambda: f"{bsym.sym.id=} expected to have subsymbols but {bsym.subsymbols=}")
-        copy_bsym = sub_bsyms[-1]
-        utils.check(
-            copy_bsym.sym.id == prims.PrimIDs.COPY_,
-            lambda: f"bsym.subsymbols[-1] expected to be {prims.PrimIDs.COPY_} but {copy_bsym.sym.id=}",
-        )
-        copy_out = copy_bsym.flat_proxy_outs[0]
-        copy_dst = copy_bsym.flat_proxy_args[1]
-        swap_map[variableify(copy_dst)] = copy_out
-        # Remove `prims.copy_`
-        new_bsym = new_bsym.from_bsym_swap_proxies(swap_map, skip_inputs=True, skip_subsymbols=True)
-        bsyms.append(new_bsym)
-
-    intermediate_trace = from_trace(computation_trace)
-    intermediate_trace.bound_symbols = bsyms[:]
-    del bsyms
-
-    # Step 2: Remove `prims.copy_` if it's the last one of `bsym.subsymbols`.
-    bsym_inplace_to_functional = {}
-    swap_map.clear()
-    new_bsyms: list[BoundSymbol] = []
-    for bsym in intermediate_trace.bound_symbols:
-        new_bsym = bsym.from_bsym_swap_proxies(swap_map)
-
-        # in-place ops has `prims.copy_` as the last subsymbol.
-        if not is_inplace(new_bsym):
-            new_bsyms.append(new_bsym)
-            continue
-        functional_sym_name = new_bsym.sym.id.split(".")[-1][:-1]
-        utils.check(
-            hasattr(thunder.torch, functional_sym_name),
-            lambda: f"{functional_sym_name}, out-of-place impl of {bsym.sym.id=} not found in `thunder.torch` namespace",
-        )
-        sub_bsyms: list[BoundSymbol] = bsym.subsymbols
-        utils.check(sub_bsyms, lambda: f"{bsym.sym.id=} expected to have subsymbols but {bsym.subsymbols=}")
-        copy_bsym = sub_bsyms[-1]
-        utils.check(
-            copy_bsym.sym.id == prims.PrimIDs.COPY_,
-            lambda: f"bsym.subsymbols[-1] expected to be {prims.PrimIDs.COPY_} but {copy_bsym.sym.id=}",
-        )
-        swap_map[variableify(copy_bsym.flat_proxy_outs[0])] = copy_bsym.flat_proxy_args[0]
-        new_bsym.subsymbols = new_bsym.subsymbols[:-1]
-        new_bsym = new_bsym.from_bsym_swap_proxies(swap_map)
-        functional_sym: Symbol = getattr(thunder.torch, functional_sym_name)
-        new_functional_bsym = functional_sym.bind(
-            *new_bsym.args,
-            **new_bsym.kwargs,
-            output=new_bsym.output,
-            subsymbols=new_bsym.subsymbols,
-            _call_ctx=new_bsym._call_ctx,
-        )
-        new_bsyms.append(new_functional_bsym)
-        bsym_inplace_to_functional[new_bsym] = new_functional_bsym
-
-    functionalized_computation_trace = from_trace(computation_trace)
-    functionalized_computation_trace.bound_symbols = new_bsyms
-    functionalized_computation_trace.set_provenance(TraceProvenance("Functionalize in-place ops"))
-    # note(crcrpar): I kind of want to do the following two.
-    # functionalized_computation_trace._provenance.swap_map = swap_map
-    # functionalized_computation_trace._provenance.bsym_inplace_to_functional = bsym_inplace_to_functional
-    return functionalized_computation_trace
--- a/thunder/core/jit_ext.py
+++ b/thunder/core/jit_ext.py
@@ -1588,7 +1588,7 @@ def bind_inputs(name, trace, input_vars, input_proxies):
     # with tracectx(trace):
     #     p: Proxy
     #     for p in input_proxies:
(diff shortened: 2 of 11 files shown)


thunder/core/proxies.py
```diff
@@ -1413,6 +1441,13 @@ def __rtruediv__(self, other):
         method = resolve_method("true_divide", other, self)
         return method(other, self)
 
+    def __itruediv__(self, other):
+        return self.div_(other)
+
+    def div_(self, other):
```
crcrpar, line 1447: should have a kwarg of `rounding_mode`

t-vi:
Awesome stuff. The fundamental questions I have about this are

- Does this give silent errors? What do we do about this
- If we do work on #145 (and we want to shortly), how will we manage the transition?

crcrpar:
> * Does this give silent errors? What do we do about this

could you give me some examples of silent errors? I'm not quite following what they would be.

> * If we do work on [Handling inplace through SSA #145](https://github.com/Lightning-AI/lightning-thunder/issues/145) (and we want to shortly), how will we manage the transition?

"the transition" of what? The better coverage by supporting in-place ops?
Anyway I'd like to give it a try as well. I just am not quite clear about what would look better even in the short term

t-vi:
> could you give me some examples of silent errors? I'm not quite following what they would be.

Essentially things that rely on aliases (`x = thunder.zeros(1, 1); x.diag().fill_(1.0)`)

> "the transition" of what? The better coverage by supporting in-place ops?

So I'm imagining that people will use this as soon as we have it and won't like us to regress on the support for their model.
Relative to #145, I wonder whether this is an alternative long-term solution or whether can we evolve it to a long-term solution or how else this fits into a long-term plan.

thunder/core/jit_ext.py
```diff
@@ -1608,6 +1623,8 @@ def thunder_general_jit(
             process_recorded_modifications(ctx, epilogue_trace)
             last_interpreter_log = jfn._last_interpreter_log
 
+    computation_trace = functionalize_inplace_ops(computation_trace)
+
```
IvanYashchuk, line 1627: I think it's useful to apply functionalization where possible, let's make this step more discoverable and also add an option to disable it by putting it into the `thunder.jit` function.

thunder/core/transform_common.py
```diff
@@ -363,3 +367,101 @@ class PostOptimizationTransform(Transform, ABC):
...
+    outputs as trace's outputs. The second step is to remove :func:`thunder.core.prims.copy_`
+    from :attr:`~thunder.core.symbol.Boundsymbol.subsymbols` if it's the last symbol.
+    """
+    from thunder.core import utils
+    from thunder.core.proxies import ProxyInterface
+    from thunder.core.symbol import VariableInterface
+    import thunder.torch
+
+    def is_inplace(bsym: BoundSymbol) -> bool:
+        if (isinstance(bsym.sym.id, str) and bsym.sym.id.endswith("_")) and bsym.subsymbols:
```
mruberry, line 385: We should probably add a tag for an operation being inplace instead of relying on parsing the id

mruberry:
I think I'm a little confused. This allows for operations like `add_` to appear in traces, and then a pass converts those operations to something like

```
c = add(a, b)
copy_(a, c)
```

?

What happens to the `copy_` operations?

In general I think we need a design review of how to handle inplace operations before merging PRs related to them.

thunder/core/transform_common.py
```diff
@@ -363,3 +367,101 @@ class PostOptimizationTransform(Transform, ABC):
...
+        new_bsym = bsym.from_bsym_swap_proxies(swap_map)
+
+        # in-place ops has `prims.copy_` as the last subsymbol.
+        if not is_inplace(new_bsym):
+            bsyms.append(new_bsym)
+            continue
+
+        sub_bsyms: list[BoundSymbol] = bsym.subsymbols
+        utils.check(sub_bsyms, lambda: f"{bsym.sym.id=} expected to have subsymbols but {bsym.subsymbols=}")
+        copy_bsym = sub_bsyms[-1]
```
jjsjann123, line 407: Similarly, this assumption doesn't hold for batchnorm?

thunder/core/transform_common.py
jjsjann123: this is going to miss `copy_` in subsymbols.

i.e. thunder/torch/__init__.py::_batch_norm uses prims.copy_ in the middle of the op and won't be identified as inplace op by this

thunder/core/transform_common.py
jjsjann123: But that might not be a bad thing. for what I know, we cannot handle batchnorm running stats update with functionalization anyway. So maybe we would want to keep it as-is.

jjsjann123 (changes requested):
One thing we are missing out here is that, in functionalization pass, we should identify input sources.

i.e. if inplace update is applied on non-intermediate buffer created in the trace, we shouldn't functionalize it yet.

If you try to run a batch_norm with your example, you'll notice that `num_batches_tracked`'s update gets functionalized. So that's silent wrong result. We should instead throw a loud error for that case.

jjsjann123:
> I think I'm a little confused. This allows for operations like `add_` to appear in traces, and then a pass converts those operations to something like
> 
> ```
> c = add(a, b)
> copy_(a, c)
> ```
> 
> ?
> 
> What happens to the `copy_` operations?
> 
> In general I think we need a design review of how to handle inplace operations before merging PRs related to them.

I think @crcrpar is only handling inplace on intermediate that can be functionalized away. Which is already help (i.e. inplace activation in resnet)

jjsjann123:
The example I was referring to in my comment

```
import thunder
import torch

val = 5

def foo(flag):
    return flag

mod = torch.nn.modules.BatchNorm2d(4)
#mod.track_running_stats = None
mod.cuda()

jfoo = thunder.jit(mod)
 
a = torch.randn(2, 4, 5, 5, device="cuda")
 
print(mod.running_mean) 
print(mod.num_batches_tracked)
print(jfoo(a))  # tensor([1.])
print(mod.running_mean)
print(mod.num_batches_tracked)
 
orig_trace = thunder.last_traces(jfoo)[0]
traces = thunder.last_traces(jfoo)
 
print(f"===\n{traces=}")
```

I'm not asking you to necessarily support it in this PR, but we should error out instead of silent wrong result.

crcrpar:
> If you try to run a batch_norm with your example, you'll notice that num_batches_tracked's update gets functionalized. So that's silent wrong result. We should instead throw a loud error for that case.

To this point, https://github.com/Lightning-AI/lightning-thunder/pull/584/commits/8889120e4542d4ac649b1efb7aa18c71f2a3f34b should be helpful:

```python
import torch
import torch.nn as nn

import thunder


def main():
    num_features = 128
    m = nn.BatchNorm2d(num_features).cuda()
    m_ref = nn.BatchNorm2d(num_features).cuda()
    m_ref.load_state_dict(m.state_dict())
    jm = thunder.jit(m)

    x = torch.randn((4, num_features, 32, 32), device="cuda")
    y: thunder.core.module.ThunderModule = jm(x)
    x_ref = x.clone().detach()
    y_ref = m_ref(x_ref)

    print(thunder.last_traces(jm)[-1])

    torch.testing.assert_close(y, y_ref)

    print(f"{m.num_batches_tracked=}")
    print(f"{jm.get_buffer('num_batches_tracked') = }")


if __name__ == "__main__":
    main()
```

```
import torch
from torch import Tensor

@torch.no_grad()
@no_autocast
def augmented_forward_fn(input, t_bias, t_num_batches_tracked, t_running_mean, t_running_var, t_weight):
  # input: "cuda:0 f32[4, 128, 32, 32]"
  # t_bias: "cuda:0 f32[128]"
  # t_num_batches_tracked: "cuda:0 i64[]"
  # t_running_mean: "cuda:0 f32[128]"
  # t_running_var: "cuda:0 f32[128]"
  # t_weight: "cuda:0 f32[128]"
  t31 = torch.reshape(t_weight, (1, 128, 1, 1))  # t31: "cuda:0 f32[1, 128, 1, 1]"
    # t31 = ltorch.reshape(t_weight, (1, 128, 1, 1))  # t31: "cuda:0 f32[1, 128, 1, 1]"
      # t31 = prims.reshape(t_weight, (1, 128, 1, 1))  # t31: "cuda:0 f32[1, 128, 1, 1]"
  t32 = Tensor.expand(t31, (4, 128, 32, 32))  # t32: "cuda:0 f32[4, 128, 32, 32]"
    # t32 = ltorch.expand(t31, (4, 128, 32, 32))  # t32: "cuda:0 f32[4, 128, 32, 32]"
      # t32 = prims.broadcast_in_dim(t31, (4, 128, 32, 32), (0, 1, 2, 3))  # t32: "cuda:0 f32[4, 128, 32, 32]"
  del t31
  t39 = torch.reshape(t_bias, (1, 128, 1, 1))  # t39: "cuda:0 f32[1, 128, 1, 1]"
    # t39 = ltorch.reshape(t_bias, (1, 128, 1, 1))  # t39: "cuda:0 f32[1, 128, 1, 1]"
      # t39 = prims.reshape(t_bias, (1, 128, 1, 1))  # t39: "cuda:0 f32[1, 128, 1, 1]"
  t40 = Tensor.expand(t39, (4, 128, 32, 32))  # t40: "cuda:0 f32[4, 128, 32, 32]"
    # t40 = ltorch.expand(t39, (4, 128, 32, 32))  # t40: "cuda:0 f32[4, 128, 32, 32]"
      # t40 = prims.broadcast_in_dim(t39, (4, 128, 32, 32), (0, 1, 2, 3))  # t40: "cuda:0 f32[4, 128, 32, 32]"
  del t39
  [t18, t41] = nvFusion0(input, t32, t40, t_num_batches_tracked, t_running_mean, t_running_var)
    # t0 = prims.add(t_num_batches_tracked, 1)  # t0: "cuda:0 i64[]"
    # prims.copy_(t0, t_num_batches_tracked)
    # (t4, t5) = prims.var_mean(input, (0, 2, 3), correction=0)
    # t6 = prims.add(t4, 1e-05)  # t6: "cuda:0 f32[128]"
    # t7 = prims.rsqrt(t6)  # t7: "cuda:0 f32[128]"
    # t8 = prims.reshape(t7, (1, 128, 1, 1))  # t8: "cuda:0 f32[1, 128, 1, 1]"
    # t14 = prims.reshape(t5, (1, 128, 1, 1))  # t14: "cuda:0 f32[1, 128, 1, 1]"
    # t15 = prims.broadcast_in_dim(t14, (4, 128, 32, 32), (0, 1, 2, 3))  # t15: "cuda:0 f32[4, 128, 32, 32]"
    # t16 = prims.sub(input, t15)  # t16: "cuda:0 f32[4, 128, 32, 32]"
    # t17 = prims.broadcast_in_dim(t8, (4, 128, 32, 32), (0, 1, 2, 3))  # t17: "cuda:0 f32[4, 128, 32, 32]"
    # t18 = prims.mul(t16, t17)  # t18: "cuda:0 f32[4, 128, 32, 32]"
    # t19 = prims.mul(0.9, t_running_mean)  # t19: "cuda:0 f32[128]"
    # t20 = prims.mul(0.1, t5)  # t20: "cuda:0 f32[128]"
    # t21 = prims.add(t19, t20)  # t21: "cuda:0 f32[128]"
    # prims.copy_(t21, t_running_mean)
    # t23 = prims.mul(t4, 1.0002442002442002)  # t23: "cuda:0 f32[128]"
    # t25 = prims.mul(0.9, t_running_var)  # t25: "cuda:0 f32[128]"
    # t27 = prims.mul(0.1, t23)  # t27: "cuda:0 f32[128]"
    # t28 = prims.add(t25, t27)  # t28: "cuda:0 f32[128]"
    # prims.copy_(t28, t_running_var)
    # t33 = prims.mul(t18, t32)  # t33: "cuda:0 f32[4, 128, 32, 32]"
    # t41 = prims.add(t33, t40)  # t41: "cuda:0 f32[4, 128, 32, 32]"
  del t32, t40
  return {'output': t41, 'flat_args': [input, t_bias, t_num_batches_tracked, t_running_mean, t_running_var, t_weight], 'flat_output': (t41,)}, ((t18,), ())
m.num_batches_tracked=tensor(1, device='cuda:0')
jm.get_buffer('num_batches_tracked') = tensor(1, device='cuda:0')
```

thunder/core/transform_common.py
crcrpar: > Similarly, this assumption doesn't hold for batchnorm?

if batchnorm's `torchsymbol` exists and it doesn't have `OpTags.IN_PLACE` then it's free from the possibility of functionalization.

crcrpar:
> I think I'm a little confused. This allows for operations like `add_` to appear in traces, and then a pass converts those operations to something like
> 
> ```
> c = add(a, b)
> copy_(a, c)
> ```
> 
> ?
> 
> What happens to the `copy_` operations?
> 
> In general I think we need a design review of how to handle inplace operations before merging PRs related to them.

`ltorch.add_(a, b)` is expressed with two subsymbols, they are `t_something = add(a, b)` and `copy_(t_something, a)` then the latter is removed by functionalization.

IvanYashchuk:
> What happens to the copy_ operations? (https://github.com/Lightning-AI/lightning-thunder/pull/584#issuecomment-2168185877)

If it's in the user script it's not traced today and it's not traced in this PR. The only way to introduce `copy_` is from within Thunder.

thunder/core/transform_common.py
IvanYashchuk: I agree, `prims.copy_` from subsymbols of batch norm should not be functionalized. It's our way to insert in-place copy in a controlled setting. Also it shouldn't functionalized because the tensors that are modified in-place are inputs to the compiled function. Currently, it's achieved by not marking batchnorm's torchsymbol with the IN_PLACE tag.
Many operations can be tagged as in-place, but they should be functionalized only if they act on tensors that were produced within the compiled function, is this implemented in the current version of the pull request?

IvanYashchuk:
According to https://github.com/Lightning-AI/lightning-thunder/pull/584#issuecomment-2169124164 `num_batches_tracked` of BatchNorm is correctly updated in-place with an nvFuser region. I hope nvFuser's fusion performance is not suddenly destroyed by this. We've got one benchmark case to try it https://github.com/Lightning-AI/lightning-thunder/blob/21a222b180009616a4cc48176958b4506894a330/thunder/benchmarks/targets.py#L199

thunder/torch/__init__.py
IvanYashchuk, line 1: All of the changes to this file look good. There's no other obvious way we can support reading in-place PyTorch operations from user code with Thunder's Python Interpreter.
Independent of how the in-place operations are treated later we need to get them into the initial trace first.

thunder/tests/opinfos.py
```diff
@@ -1278,16 +1278,10 @@ def _abs_torch(x: torch.Tensor | Number):
 elementwise_unary_ops.append(signbit_opinfo)
 
 
-def silu_error_generator(op, device, dtype=torch.float32, **kwargs):
-    a = make_tensor((), dtype=dtype, device=device)
-    yield (SampleInput(a, inplace=True), NotImplementedError, "Thunder only supports silu with inplace=False")
```
IvanYashchuk, line 1283: Nice, so we record silu with inplace=True. In the subsymbols we use `prims.copy_` and then functionalization pass effectively changes `silu(x, inplace=True)` to just `silu(x)`.

thunder/core/jit_ext.py
```diff
@@ -1521,14 +1521,29 @@ def process_recorded_modifications(ctx, epilogue_trace):
...
     # TODO This currently does the unpacks at the end of the trace, then moves them to the beginning, there's
     #   almost certainly a more elegant way to do this
-    with tracectx(trace):
-        p: Proxy
-        for p in input_proxies:
-            prims.unpack_trivial(p, name=p.name)
-
+    num_orig_bsyms = len(trace.bound_symbols)
+
+    # note(crcrpar): The path looks neat but it does not work for a trace
```
IvanYashchuk, line 1531: Where is this called? Is it still needed now that functionalization is moved into `thunder.jit` to be applied after the interpreter?

thunder/core/jit_ext.py
```diff
@@ -1521,14 +1521,29 @@ def process_recorded_modifications(ctx, epilogue_trace):
...
+    # if it's :func:`thunder.core.jit_ext.functionalize_inplace_ops`'d, :shrug:
```
IvanYashchuk, line 1532:
```suggestion
    # if it's :func:`thunder.core.transform_common.functionalize_inplace_ops`'d, :shrug:
```

thunder/core/transform_common.py
```diff
@@ -363,3 +368,105 @@ class PostOptimizationTransform(Transform, ABC):
...
+    For example, :func:`thunder.torch.add_` is represented as a :class:`thunder.core.symbol.BoundSymbol`
+    whose `subsymbols` are :func:`thunder.torch.add` and :func:`thunder.core.prims.copy_`. This function
+    replaces it with a :class:`~thunder.core.symbol.BoundSymbol` of :func:`~thunder.torch.add`.
+
+    Note that functionalization is not applied, if any of an in-place op's arguments are
+    ``computation_trace.args`` or ``computation_trace.kwargs``.
+    """
+    import thunder.torch
+
+    def is_inplace(bsym: BoundSymbol) -> bool:
```
IvanYashchuk, line 388: To me "is_functionalizable" is a better name because there are more conditions than just in-place?

thunder/core/transform_common.py
```diff
@@ -363,3 +368,105 @@ class PostOptimizationTransform(Transform, ABC):
     @abstractmethod
     def transform_trace(self, computation_trace: Trace, **kwargs):
         pass
+
+
+def functionalize_inplace_ops(computation_trace: Trace) -> list[Trace]:
+    """Functionalize in-place ops in ``computation_trace``.
+
+    In thunder, an in-place is an out-of-place or functional op followed by :func:`~thunder.core.prims.copy_`.
+    This function replaces such in-place ops with out-of-place ops.
```
IvanYashchuk, line 377: "... only if the in-place argument is intermediate to the trace", right?

thunder/core/transform_common.py
```diff
@@ -363,3 +368,105 @@ class PostOptimizationTransform(Transform, ABC):
...
+        return []
+
+    # Step 1: return the tensors returned from `prims.copy_` as possible not the args for clarity.
+    bsym: BoundSymbol
+    swap_map: dict[VariableInterface, ProxyInterface] = {}
+    bsyms: list[BoundSymbol] = []
+    for bsym in computation_trace.bound_symbols:
+        new_bsym = bsym.from_bsym_swap_proxies(swap_map)
+
+        # in-place ops has `prims.copy_` as the last subsymbol.
```
IvanYashchuk, line 406:
```suggestion
        # in-place functionalizable ops has `prims.copy_` as the last subsymbol.
```

thunder/core/transform_common.py
IvanYashchuk: Oh, I see later "functionalization is not applied, if any of an in-place op's arguments are ``computation_trace.args`` or ``computation_trace.kwargs``."

IvanYashchuk (approved):
This PR does two things at once to demonstrate the usability of in-place tracing and generating functional code:

1. In-place operations like `abs_`, `add_`, etc. are added to `thunder.torch` so that Thunder's Python Interpreter can recognize corresponding PyTorch operations and put them into the initial Thunder trace.
2. Transform in-place operations on intermediates into out-of-place variants. This is an operator-level transform, there's no interaction between ops. In-place on views is handled in a separate pull request (https://github.com/Lightning-AI/lightning-thunder/pull/597).

We need part 1, an alternative to part 2 could be freezing the order of in-place operations and let the PyTorch executor to execute them. There could be other good alternatives and the part 2 can easily be disabled if needed in the future.

thunder/torch/__init__.py
crcrpar: noticed that we might want to add some distributed calls such as `torch.distributed.all_reduce`, `torch.distributed.all_gather_into_tensor`, and `torch.distributed.reduce_scatter_tensor` later.

thunder/core/jit_ext.py
crcrpar: no, great catch. reverted the changes in this file

thunder/core/transform_common.py
jjsjann123: If we already have commit 8889120 that's guarding the functionalization of inplace update on inputs, is there anything else we needed to special case for batch norm?

thunder/core/transform_common.py
```diff
@@ -363,3 +368,105 @@ class PostOptimizationTransform(Transform, ABC):
...
+        copy_return = copy_bsym.flat_proxy_outs[0]
+        copy_from = copy_bsym.flat_proxy_args[0]
+        copy_to = copy_bsym.flat_proxy_args[1]
+        if copy_to in trace_args_set:
+            new_bsyms.append(new_bsym)
+        else:
+            swap_map[variableify(copy_return)] = copy_from
+            new_bsym.subsymbols = new_bsym.subsymbols[:-1]
+            new_bsym = new_bsym.from_bsym_swap_proxies(swap_map)
+            functional_sym: Symbol = getattr(thunder.torch, functional_sym_name)
```
jjsjann123, line 455: The inconsistency here looks bad.

It's fine to choose to only functionalize `torch.xxx_` to `torch.xxx`. But the `is_functionalizable` thing should have the same logic.

thunder/core/transform_common.py
```diff
@@ -363,3 +368,105 @@ class PostOptimizationTransform(Transform, ABC):
...
+
+    For example, :func:`thunder.torch.add_` is represented as a :class:`thunder.core.symbol.BoundSymbol`
+    whose `subsymbols` are :func:`thunder.torch.add` and :func:`thunder.core.prims.copy_`. This function
+    replaces it with a :class:`~thunder.core.symbol.BoundSymbol` of :func:`~thunder.torch.add`.
+    """
+    import thunder.torch
+
+    def is_functionalizable(bsym: BoundSymbol) -> bool:
+        """Has `OpTags.IN_PLACE` and its args are NOT ``computation_trace.args`` nor ``computation_trace.kwargs``."""
+        return (
```
jjsjann123, line 389: I don't think we care about having `IN-PLACE` tag or not here. since the logic below for replacing doesn't take any consideration like that.

I feel the logic here should just check for `torch.xxx_` and see if there is a `torch.xxx`

thunder/core/transform_common.py
```diff
@@ -363,3 +368,105 @@ class PostOptimizationTransform(Transform, ABC):
...
+            bsym.sym.tags
+            and prims.OpTags.IN_PLACE in bsym.sym.tags
+            and bsym.subsymbols
+            and bsym.subsymbols[-1].sym.id == prims.PrimIDs.COPY_
```
jjsjann123, line 393: we certainly should drop the subsymbols check. This is irrelevant from how this PR is handling functionalization.

thunder/core/transform_common.py
```diff
@@ -363,3 +368,105 @@ class PostOptimizationTransform(Transform, ABC):
...
+    swap_map.clear()
+    new_bsyms: list[BoundSymbol] = []
+    for bsym in intermediate_trace.bound_symbols:
+        new_bsym = bsym.from_bsym_swap_proxies(swap_map)
+
+        if not is_functionalizable(new_bsym):
+            new_bsyms.append(new_bsym)
+            continue
+        functional_sym_name = new_bsym.sym.id.split(".")[-1][:-1]
+        check(
```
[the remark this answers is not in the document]
jjsjann123: If `is_functionalizable` checks for an existing mapping. we shouldn't need this check here.

thunder/core/transform_common.py
crcrpar: > This is irrelevant from how this PR is handling functionalization.

why is it irrelevant? Currently in-place bsyms have out-of-place and copy as their subsymbols so I think it fair to check the last sub bound symbol is copy.

crcrpar: how can we tell an appropriate output tensor proxy if a bsym doesn't have a copy_ as its last sub bsym, while avoiding having a lot of new tensor proxy names in a trace?

crcrpar:
I now think `OpTags.IN_PLACE` would not work quite well given that there are some functions that takes `inplace` as their argument such as `torch.relu` whose `ltorch` def is https://github.com/Lightning-AI/lightning-thunder/blob/8309fc015c1ea9d684337cdcf95138524cd8645c/thunder/torch/__init__.py#L1458-L1462 which I cannot pass this tag separately

thunder/core/transform_common.py
jjsjann123: oops. sorry I read your implementation wrong earlier... I thought we are doing a blind `torch.xxx_` to `torch.xxx` replacement but that's not the case. You actually are only looking at the last subsymbol and replacing that one entry only.

jjsjann123: That feels a bit restricted... But a first step is still better then nothing and I'll stop nitpicking on that.

jjsjann123:
> I now think `OpTags.IN_PLACE` would not work quite well given that there are some functions that takes `inplace` as their argument such as `torch.relu` whose `ltorch` def is
> 
> https://github.com/Lightning-AI/lightning-thunder/blob/8309fc015c1ea9d684337cdcf95138524cd8645c/thunder/torch/__init__.py#L1458-L1462
> 
> which I cannot pass this tag separately

The `IN_PLACE` tag is used more like a `MAYBE_INPLACE` in the implementation. Maybe switching to that?

thunder/core/transform_common.py
crcrpar: I don't think so given the change of [test_batch_norm_running_stats](https://github.com/Lightning-AI/lightning-thunder/pull/584/files#diff-147ca1270ed52defb8a1ad70ef9c0fb88b989be45bdaedd938fc98befa452226)

thunder/core/transform_common.py
crcrpar: > That feels a bit restricted... But a first step is still better then nothing and I'll stop nitpicking on that.

how would it be a bit restricted compared to a blind `torch.foo_` to `torch.foo` replacement?

jjsjann123: no that's not what I meant.

By `That feels a bit restricted`, I'm referring to the alternative that we functionalize directly at the subsymbol `prim.copy_` level. But again we don't have to do that in this PR.

thunder/core/transform_common.py
crcrpar: now here's `_inplace_to_out_of_place`

jjsjann123 (approved):
My earlier concern has been addressed. Stamping.

thunder/tests/test_inplace_functionalization.py
```diff
@@ -0,0 +1,131 @@
...
+                yield SampleInput(*(list(sample.args) + [True]), **sample.kwargs)
+            else:
+                # silu treats `inplace` as a kwarg
+                # ref: https://github.com/Lightning-AI/lightning-thunder/commit/335d84c89
+                new_kwargs = {"inplace": True}
+                if sample.kwargs:
+                    new_kwargs.update(sample.kwargs)
+                yield SampleInput(*sample.args, **new_kwargs)
+
+    return f
```
crcrpar, line 31: rel: https://github.com/Lightning-AI/lightning-thunder/pull/615

thunder/core/prims.py
```diff
@@ -275,6 +275,7 @@ class OpTags(Enum):
     DEVICE_SYNC_OP = auto()
     # Labels operations that should not be removed by the dead code elimination (DCE) pass
     DONT_DCE = auto()
+    IN_PLACE = auto()
```
nikitaved, line 278: Looks like `IN_PLACE` is not actually used anymore?

thunder/core/transform_common.py
nikitaved: I see no usage of `IN_PLACE` anymore

nikitaved: nit: The name `is_functionalizable` was a bit confusing. It seems it checks whether a symbol was already "functionalized".

thunder/core/transform_common.py
```diff
@@ -363,3 +368,107 @@ class PostOptimizationTransform(Transform, ABC):
...
+
+    def is_functionalizable(bsym: BoundSymbol) -> bool:
+        """Has `OpTags.IN_PLACE` and its args are NOT ``computation_trace.args`` nor ``computation_trace.kwargs``."""
+        return (
+            bsym.sym in thunder.torch._inplace_to_out_of_place
+            and bsym.subsymbols
+            and bsym.subsymbols[-1].sym.id == prims.PrimIDs.COPY_
+        )
+
+    if not any(is_functionalizable(bsym) for bsym in computation_trace.bound_symbols):
```
nikitaved, line 395: nit: `not any` -> `all`

thunder/core/transform_common.py
```diff
@@ -363,3 +368,107 @@ class PostOptimizationTransform(Transform, ABC):
...
+    ``computation_trace.args`` or ``computation_trace.kwargs``.
+
+    For example, :func:`thunder.torch.add_` is represented as a :class:`thunder.core.symbol.BoundSymbol`
+    whose `subsymbols` are :func:`thunder.torch.add` and :func:`thunder.core.prims.copy_`. This function
+    replaces it with a :class:`~thunder.core.symbol.BoundSymbol` of :func:`~thunder.torch.add`.
+    """
+    import thunder.torch
+
+    def is_functionalizable(bsym: BoundSymbol) -> bool:
+        """Has `OpTags.IN_PLACE` and its args are NOT ``computation_trace.args`` nor ``computation_trace.kwargs``."""
```
nikitaved, line 388: I see no usage of `IN_PLACE` anymore.
nit: The name `is_functionalizable` is a bit confusing. It seems it checks whether a symbol has already been "functionalized".

thunder/core/transform_common.py
```diff
@@ -363,3 +368,107 @@ class PostOptimizationTransform(Transform, ABC):
...
+        copy_dst = copy_bsym.flat_proxy_args[1]
+        swap_map[variableify(copy_dst)] = copy_out
+        # make sure an in-place bsym returns `prims.copy_` output
+        new_bsym = new_bsym.from_bsym_swap_proxies(swap_map, skip_inputs=True, skip_subsymbols=True)
+        bsyms.append(new_bsym)
+
+    intermediate_trace = from_trace(computation_trace)
+    intermediate_trace.bound_symbols = bsyms[:]
+    intermediate_trace.set_provenance(TraceProvenance("Intermediate trace of `functionalize_inplace_ops`"))
+    del bsyms
```
nikitaved, line 421: nit: can't we just do `intermediate_tensors.bound_symbols = bsyms`?

crcrpar: yes. I didn't want to reuse `bsyms`

thunder/core/transform_common.py
crcrpar: yes, at https://github.com/Lightning-AI/lightning-thunder/pull/584/files#diff-6f6d1d14b9e8ec7b29e0af14d1c6c621317f36739fd2e8fc3f16ad2387929cc7R443

thunder/core/transform_common.py
```diff
@@ -363,3 +368,107 @@ class PostOptimizationTransform(Transform, ABC):
...
+        copy_bsym = bsym.subsymbols[-1]
+        copy_return = copy_bsym.flat_proxy_outs[0]
+        copy_from = copy_bsym.flat_proxy_args[0]
+        copy_to = copy_bsym.flat_proxy_args[1]
+        if copy_to in trace_args_set:
+            new_bsyms.append(new_bsym)
+        else:
+            swap_map[variableify(copy_return)] = copy_from
+            new_bsym.subsymbols = new_bsym.subsymbols[:-1]
+            new_bsym = new_bsym.from_bsym_swap_proxies(swap_map)
```
nikitaved, line 448: This logic looks similar to what step 1 is doing. Couldn't they be merged?

thunder/core/transform_common.py
```diff
@@ -363,3 +368,107 @@ class PostOptimizationTransform(Transform, ABC):
...
+
+            functional_sym: Symbol
+            optional_inplace_arg_index: int
+            functional_sym, optional_inplace_arg_index = thunder.torch._inplace_to_out_of_place[new_bsym.sym]
+
+            flat_args, flat_args_spec = tree_flatten((new_bsym.args, new_bsym.kwargs))
+            if optional_inplace_arg_index > -1:
+                flat_args[optional_inplace_arg_index] = False
```
nikitaved, line 456: This probably needs a comment.

thunder/core/transform_common.py
```diff
@@ -363,3 +368,107 @@ class PostOptimizationTransform(Transform, ABC):
...
+            args, kwargs = tree_unflatten(flat_args, flat_args_spec)
+            new_functional_bsym = functional_sym.bind(
+                *args,
+                **kwargs,
+                output=new_bsym.output,
+                subsymbols=new_bsym.subsymbols,
+                _call_ctx=new_bsym._call_ctx,
+            )
+            new_bsyms.append(new_functional_bsym)
+            bsym_inplace_to_functional[new_bsym] = new_functional_bsym
```
nikitaved, line 466: `bsym_inplace_to_functional` is never read from?

thunder/core/langctxs.py
```diff
@@ -72,7 +72,7 @@ def resolve_method(id: Any, *args, **kwargs) -> None | Callable:
         # ctx.get_method throws an AttributeError when the context does not have the requested attribute, except
         # for the prims language context, which always throws a ValueError
         method: Callable = ctx.get_method(id, *args, **kwargs)
-    except (AttributeError, ValueError) as e:
+    except (AttributeError, ValueError):
```
t-vi, line 75: Great catch

thunder/__init__.py
```diff
@@ -503,6 +509,9 @@ def get_computation_and_inputs(*args, **kwargs):
 
             prologue_traces = [prologue_trc]
             computation_traces = [computation_trc]
+            if not compile_options.get("skip_inplace_functionalization", False):
+                computation_traces.extend(functionalize_inplace_ops(computation_trace=computation_trc))
+                computation_trc = computation_traces[-1]
```
t-vi, line 514: Longer term, I wonder if this should be a "default transform", but maybe it is important that this goes first and so it is tricky with the timing.

thunder/core/transform_common.py
```diff
@@ -363,3 +368,107 @@ class PostOptimizationTransform(Transform, ABC):
...
         pass
+
+
+def functionalize_inplace_ops(computation_trace: Trace) -> list[Trace]:
+    """Functionalize in-place ops in ``computation_trace``.
+
+    In thunder, an in-place is an out-of-place or functional op followed by :func:`~thunder.core.prims.copy_`.
+    This function replaces such in-place ops with out-of-place ops.
+    Note that functionalization is not applied, if any of an in-place op's arguments are
+    ``computation_trace.args`` or ``computation_trace.kwargs``.
```
t-vi, line 379: I wonder what we should do in these cases, though, warn, error?

thunder/core/transform_common.py
```diff
@@ -363,3 +368,107 @@ class PostOptimizationTransform(Transform, ABC):
...
+        copy_bsym = bsym.subsymbols[-1]
+        copy_out = copy_bsym.flat_proxy_outs[0]
+        copy_dst = copy_bsym.flat_proxy_args[1]
+        swap_map[variableify(copy_dst)] = copy_out
+        # make sure an in-place bsym returns `prims.copy_` output
+        new_bsym = new_bsym.from_bsym_swap_proxies(swap_map, skip_inputs=True, skip_subsymbols=True)
+        bsyms.append(new_bsym)
+
+    intermediate_trace = from_trace(computation_trace)
+    intermediate_trace.bound_symbols = bsyms[:]
```
t-vi, line 419: Nit: I don't think we strictly need the copy here.

thunder/tests/test_inplace_functionalization.py
```diff
@@ -0,0 +1,124 @@
+from __future__ import annotations
```
t-vi, line 1: Great to have tests for it!

thunder/core/transform_common.py
t-vi: but `not any(foo, ...` has much fewer ops than `all(not foo`...

t-vi (approved):
Supergood! We'll need to make things safer, but this is an awesome start.
Thank you @IvanYashchuk @jjsjann123 @nikitaved @mruberry for your reviews and comments.

thunder/core/transform_common.py
crcrpar: it seems that BatchNorm's `num_batches_tracked` tensor update is expressed as `ltorch.add_(num_batches_tracked, 1)` and the tensor is an arg. so this makes sense to me. also, if one or more of args & kwargs are updated in an in-place manner, then I guess there's some intention so I'm not inclined to ban such cases

thunder/torch/__init__.py
```diff
@@ -1241,96 +1249,191 @@ def abs(a: NumberLike | TensorLike, /) -> Number | TensorLike:
     return clang.abs(a)
 
 
+@torchsymbol(torch.Tensor.abs_, is_method=True, tags=(prims.OpTags.IN_PLACE,))
+def abs_(a: NumberLike | TensorLike, /) -> Number | TensorLike:
+    return prims.copy_(abs(a), a)
```
nikitaved, line 1254: nit: maybe adding a decorator for an out-of-place op to register an in-place counterpart could be cleaner if realizable?

thunder/tests/test_inplace_functionalization.py
```diff
@@ -0,0 +1,124 @@
...
+_functional_to_inplace: dict[Callable, Callable] = {
+    functional: inplace for inplace, (functional, index) in _inplace_to_out_of_place.items() if index == -1
+}
+_functional_to_functional_with_inplace_arg: dict[Callable, tuple[Callable, int]] = {
+    functional: (inplace, index) for inplace, (functional, index) in _inplace_to_out_of_place.items() if index >= 0
+}
+_inplace_opinfos: list[OpInfo] = []
+for op in opinfos:
+    if not (op.op in _functional_to_inplace or op.op in _functional_to_functional_with_inplace_arg):
+        continue
```
nikitaved, line 48: nit: maybe in some future we would like to have a flag to test in-place ops in opinfos for out-of-place OpInfos?

thunder/tests/test_inplace_functionalization.py
```diff
@@ -0,0 +1,124 @@
...
+        )
+        _inplace_opinfos.append(inplace_opinfo)
+    # in-place ops whose name ends with `_`
+    if op.op in _functional_to_inplace:
+        inplace_op = _functional_to_inplace[op.op]
+        inplace_opinfo = OpInfo(
+            inplace_op,
+            sample_input_generator=(
+                op.sample_input_generator if op.name != "masked_fill" else inplace_masked_fill_sample_generator
+            ),
```
nikitaved, line 66: nit: seems this motivates the comment from above. We could set `test_in_place=False` for `masked_fill` and create a separate OpInfo for the in-place variant.

thunder/tests/test_inplace_functionalization.py
```diff
@@ -0,0 +1,124 @@
...
+        t = args[idx] + 1.0
+        args[idx] = t
+
+        self.torch_func(*args, **kwargs)
+        return t
+
+
+@ops(_inplace_opinfos, supported_dtypes=(dtypes.float32,))
+def test_functionalization(op: OpInfo, device: str, dtype: dtypes.dtype, executor, _):
+    import thunder
```
nikitaved, line 95: Seems like we might be missing some test with a sequence of in-places ops?

crcrpar: > Seems like we might be missing some test with a sequence of in-places ops?

yes, this pr doesn't have such tests.

> Just to be sure that the bsym replacement logic is sound across multiple symbols, not just a single one. Or is it not for this PR?

Locally I've been using the following snippet so I hope the functionalization works. I just didn't have a clear picture of designing tests with a sequence of in-place ops.

```python
import torch
import thunder


def f(a: torch.Tensor, b: torch.Tensor) -> tuple[torch.Tensor, torch.Tensor]:
    a += b
    c = torch.exp(a)
    d = torch.tanh(b)

    c += d

    d.div_(a)

    e = c + d
    f = torch.nn.functional.relu(e, inplace=True)
    g = a + b
    return f, c, d, torch.relu_(g)


def main():
    a, b = [torch.randn((2, 2), device="cuda", requires_grad=False) for _ in range(2)]
    a_, b_ = a.clone().detach(), b.clone().detach()
    jit_f = thunder.jit(f, executors=[thunder.executors.get_torch_executor()])
    c, d, e, g = jit_f(a, b)
    c_, d_, e_, g_ = f(a_, b_)

    traces = thunder.last_traces(jit_f)
    print(traces[-1])

    torch.testing.assert_close(d, d_)
    torch.testing.assert_close(c, c_)
    torch.testing.assert_close(e, e_)
    torch.testing.assert_close(g, g_)


if __name__ == "__main__":
    main()
```

nikitaved: We could add some handpicked tests for sure. We can always throw some line duplication for numerically more stable ops :)

apaz-cli:
Glorious.

Issue #332, opened 2024-05-01
Implement _set_grad_enabled of torch._C


athitten:
### 🚀 Feature

Implement _set_grad_enabled of torch._C

### Motivation

NeMo [Vision Transformer](https://github.com/NVIDIA/NeMo/blob/main/examples/vision/vision_transformer/megatron_vit_classification_pretrain.py)

cc @tfogal

tfogal:
triage team: looking to understand if this is high-effort or low-effort (honestly, looking for that on all NeMo things, this one's just particularly out of my depth).

mruberry:
triage review — implementing this operation is probably out of scope for thunder, and the model may need to be revised to target thunder

IvanYashchuk:
[`torch.autograd.grad_mode.set_grad_enabled`](https://pytorch.org/docs/stable/generated/torch.autograd.grad_mode.set_grad_enabled.html) is PyTorch's Autograd concept and is out of scope for Thunder. Options for "supporting" it is to make Thunder's JIT interpreter ignore it and not raise an error (if there's an error today).

tfogal:
> model may need to be revised to target thunder

I'm not sure "revise the model" is going to be a reasonable in the general case; already seen this in GPT, CLIPEncoder, and a vision transformer from NeMo.

There is an error today. "Ignore it" seems like a good approach to at least make progress on identifying issues... will we still produce the same computation as torch if we do that?

kshitij12345:
> "Ignore it" seems like a good approach to at least make progress on identifying issues... will we still produce the same computation as torch if we do that?

Using `no_grad`/`enabled_grad` now works without an error on `main` branch. We will still produce the same forward computation but we will hold onto intermediates (and be able to compute gradients) if one of the input has requires_grad=True even if `no_grad` was enabled. I think we should add a warning when these operations will be ignored when used with `thunder.jit`.

```python
import torch
import thunder

def foo(x):   
    # with torch.enable_grad():
    #     x =  x + 1
    with torch.no_grad():
        return x + 1

jfoo = thunder.jit(foo)
x = torch.randn(3, requires_grad=True)
o = jfoo(x)
o.sum().backward()
print(x.grad)

# print(thunder.last_traces(jfoo))

```

PR #589, opened 2024-06-13
Warn when torch._C._set_grad_enabled is called

kshitij12345:
Fixes: https://github.com/Lightning-AI/lightning-thunder/issues/332

Currently, on main, when `_set_grad_enabled` is called via `torch.enable_grad()` or `torch.no_grad()`, it is silently ignored. thunder decides whether to compute gradients based on input setting `requires_grad=True`. This PR adds a warning stating the same so that the user knows `torch.{enable_grad/no_grad}` will not have any effect when used with `thunder.jit`.

```python
import torch
import thunder

def foo(x):   
    with torch.no_grad():
        return x + 1

jfoo = thunder.jit(foo)
x = torch.randn(3, requires_grad=True)
o = jfoo(x)
o.sum().backward()
print(x.grad)  # grad exists

# print(thunder.last_traces(jfoo))

# # Constructed by Delete Last Used (took 0 milliseconds)
# import torch
# from thunder.executors.torchex import no_autocast

# @torch.no_grad()
# @no_autocast
# def augmented_forward_fn(x):
#   # x: "cpu f32[3]"
#   t0 = torch.add(x, 1.0)  # t0: "cpu f32[3]"
#     # t0 = ltorch.add(x, 1.0, alpha=None)  # t0: "cpu f32[3]"
#       # t0 = prims.add(x, 1.0)  # t0: "cpu f32[3]"
#   return {'output': t0, 'flat_args': [x], 'flat_output': (t0,)}, ((), ())
```

### DIFF
--- a/thunder/torch/__init__.py
+++ b/thunder/torch/__init__.py
@@ -150,6 +150,14 @@ class torchsymbol:
         return sym
 
 
+# This is function maps an implementation for `torch` operation without creating a Symbol.
+# So, the registered implementation will not show up in trace as a Symbol (but will get inlined).
+# This is helpful if we want to support a torch operation and bake it's output directly into the trace.
+# See `clone` and `torch.device` for example.
+def register_function(torchfn, thunderfn_impl):
+    _torch_to_thunder_function_map[torchfn] = thunderfn_impl
+
+
 #
 # Tensor properties
 #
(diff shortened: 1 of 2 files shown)


thunder/torch/__init__.py
```diff
@@ -4394,7 +4402,14 @@ def torch_device(device_or_str: DeviceLike, /, index: int | None = None) -> devi
 
 # We don't use @torchsymbol as we don't want `torch.device()` to appear in trace as a symbol.
 # Because of this, we need to manually register the implementation.
-_torch_to_thunder_function_map[torch.device] = torch_device
+register_function(torch.device, torch_device)
+
+
+def _set_grad_enabled_with_warning(enabled: bool) -> None:
```
mruberry, line 4408: A warning seems like a good start

mruberry (approved):

PR #591, opened 2024-06-13
TE: fix the placement of bwd_sync symbol in trace

kshitij12345:
> I find that the te_sync_fp8_meta_bwd seems to appear after the return in the bwd trace

As observed by @kiya00, `te_sync_fp8_meta_bwd` appears after `return` in the backward trace and hence doesn't do anything. This PR fixes this by moving it before the return.

As to why this worked even when it was placed incorrectly is because - we called `FP8GlobalStateManager.is_first_fp8_module()` in the `__init__` of the `TELinear` which happens during the compilation phase and not during the execution (where the `fp8_autocast` context manager is active). This flag is set and reset only under `fp8_autocast` manager (see the reference below). So, we didn't really consume the token and relied on TE to automatically introduce this sync.

https://github.com/Lightning-AI/lightning-thunder/blob/69e80f0a094376576a39306f62b9c510138e41fa/thunder/executors/transformer_engineex.py#L185

Also, to reiterate the reason we want to control these sync ourselves is because reordering operations can lead to failure if we rely on TE to automatically sync. See https://github.com/Lightning-AI/lightning-thunder/pull/80 for more context

Reference to where `fp8_autocast` sets the `IS_FIRST_FP8_MODULE` flag on entering the ctx_manager:
https://github.com/NVIDIA/TransformerEngine/blob/b5a7c9f95e995236afae301366ec433c73b52690/transformer_engine/pytorch/fp8.py#L383-L411

-----

Tested on TE v1.6 (stable), v1.7 and v1.8 (main)

### DIFF
--- a/thunder/tests/distributed/test_ddp.py
+++ b/thunder/tests/distributed/test_ddp.py
@@ -1548,13 +1548,19 @@ def _test_ddp_transformer_engine_llama_sanity(input_data):
             out = jit_model(x, y).sum()
             out.backward()
 
-        fwd_exec_trace = thunder.last_traces(jit_model)[-1]
         bwd_exec_trace = thunder.last_backward_traces(jit_model)[-1]
+
+        # Last symbol of the trace should be `return`
+        return_sym_idx = len(bwd_exec_trace.bound_symbols) - 1
+        assert thunder.core.prims.PrimIDs.RETURN == bwd_exec_trace.bound_symbols[return_sym_idx].sym.id
 
         # Verify that the symbol to sync backward
         # fp8 metadata is present in backward trace.
-        for bsym in reversed(bwd_exec_trace.bound_symbols):
+        for idx, bsym in enumerate(bwd_exec_trace.bound_symbols):
             if bsym.sym.id == te_sync_fp8_meta_bwd.id:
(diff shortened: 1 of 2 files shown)


thunder/tests/distributed/test_ddp.py
```diff
@@ -1548,13 +1548,19 @@ def _test_ddp_transformer_engine_llama_sanity(input_data):
...
+        assert thunder.core.prims.PrimIDs.RETURN == bwd_exec_trace.bound_symbols[return_sym_idx].sym.id
+
         # Verify that the symbol to sync backward
         # fp8 metadata is present in backward trace.
-        for bsym in reversed(bwd_exec_trace.bound_symbols):
+        for idx, bsym in enumerate(bwd_exec_trace.bound_symbols):
             if bsym.sym.id == te_sync_fp8_meta_bwd.id:
+                # Verify that `te_sync_fp8_meta_bwd` is before the last symbol of the trace
+                # which is `return`
+                assert idx < return_sym_idx
```
IvanYashchuk, line 1563: Thank you for updating the test!

IvanYashchuk (approved):

t-vi (approved):

Issue #557, opened 2024-06-07
TransformerEngine: Communication and Computation is not overlapped in backward pass with FSDP Zero3


kshitij12345:
Repro -
Generate the profile with 
`nsys profile -c cudaProfilerApi torchrun --nproc_per_node=8 --nnodes=1 thunder/benchmarks/benchmark_litgpt.py --return_metrics_as_json=True --json_path=/tmp/benchmark_litgpt_data.json --distributed_mode=fsdp --shard_mode=zero3 --model_name=Llama-2-7b-hf --micro_batch_size=1 --compile=thunder_inductor_cat_transformerengine_cudnn --nsys_enabled=True --dynamic=False`

In the image below, it can be seen that once backward pass begins, there is almost no overlap between comms and compute (there should be some overlap in reduce scatters and computations).
![te_backward_compute](https://github.com/Lightning-AI/lightning-thunder/assets/19503980/3e52ee4f-0eec-48d7-b235-2654fe9aafb9)

Zoomed out timeline:
![te_backward_compute_2](https://github.com/Lightning-AI/lightning-thunder/assets/19503980/ead42bcb-2b1d-48a8-8aee-c03d2b29145b)

If `TE` executors is removed then most of the computation and communication is overlapping (see below)
![image](https://github.com/Lightning-AI/lightning-thunder/assets/19503980/beef5bb6-4262-480c-96d5-8d991196d766)

-----------
~~My first guess is that the blocking `all_reduce` called on `fp8_autocast_exit` could be the reason.~~ -> This is incorrect.

Blocking AllReduce: https://github.com/NVIDIA/TransformerEngine/blob/0edf30b87159e82048b5f248e4b379aebb8f364a/transformer_engine/pytorch/fp8.py#L302-L313
fp8_autocast_exit: https://github.com/NVIDIA/TransformerEngine/blob/0edf30b87159e82048b5f248e4b379aebb8f364a/transformer_engine/pytorch/fp8.py#L420-L421

kshitij12345:
Finding - Turns out that the passes `sort_communication_ops` and `limit_in_flight_allgathers` move the reduce_scatters to the end when TE is present.

kshitij12345:
Applying the patch - to disable comms sorting on bw_trace, we can see overlap in comms and compute stream as well as slightly improved timings. Will investigate why all comms are moved to the last when `TE` executors is present.

cc: @kiya00 - if you have any idea as to why this happens.

```patch
[...]

PR #592, opened 2024-06-13
Sort allgathers according to consumer order, reduce scatter according to producer order

### LINKED ISSUE #574: Find a way to properly sort the communication operators for zero2/zero3
### Background
There are problems with the positioning of the communication and computation operators, causing them not to overlap well.

Currently we have 2 functions that rely on topological sorting and the weights we specify to sort the communication and computation operators

https://github.com/Lightning-AI/lightning-thunder/blob/067f15aae47ad71229732ca6c35a5d190135e48c/thunder/distributed/utils.py#L54

bottom-up sorting, relying on the order of the outputs, can have problems with the position of reduce_scatter, e.g. in issue #557 

https://github.com/Lightning-AI/lightning-thunder/blob/067f15aae47ad71229732ca6c35a5d190135e48c/thunder/distributed/utils.py#L112

top-down sorting, relying on the order of the inputs, can have problems with the position of all_gathers, e.g. in issue https://github.com/Lightning-AI/lightning-thunder/issues/277

### Expectation:

For zero2: 
sort the allgathers to their consumer order and list them at the beginning of the trace, the corresponding waits are right before the consumers (maximum distance between allgathers and waits).
put the reduce/reduce_scatter just after its producer and corresponding wait just before its consumer (maximum distance between reduce and wait)

For zero3:
same as zero2, then apply the appropriate limit on in-flight allgathers to balance memory usage.

cc: @IvanYashchuk @crcrpar @kshitij12345 

cc @carmocca @crcrpar

kiya00:
<details>

</details>

## What does this PR do?

Fixes #574.

This PR suggests a way to sort the communication ops. Previously we used one function to sort the allgather and reduce_scatter, but ran into some problems when dealing with topological equal nodes (when we sort allgather and reduce_scatter in one pass, if top-down, it results in allgather relying on the order of input params, if bottom-up, it results in reduce_scatter might accumulate before wait)

In this PR we have 2 sorting function:

- sort allgather: bottom-up topological sorting, sort the all_gather_prim_impl and its wait nodes according to the consumer order, and put all_gather_prim_impl just before wait.

- sort reduce_scatter: top-down topological sorting, sort the reduce/reduce_scatter and its wait node according to the producer order, and maximum the distance between reduce/reduce_scatter and wait.

**But since our reduce_scatter is already in the right place without the sorting, we need to think about whether this PR is really necessary**


t-vi:
> But since our reduce_scatter is already in the right place without the sorting, we need to think about whether this PR is really necessary

There is small memory increase. Is that expected?

@kiya00 @crcrpar curious to hear your thoughts.

IvanYashchuk:
The issue description has the following expectation:
> sort the allgathers to their consumer order and list them at the beginning of the trace, the corresponding waits are right before the consumers (maximum distance between allgathers and waits).

but the description of this pull request says that allgather sorting puts the allgather op right before the wait op so the communication wouldn't be overlapping with computation. Is the pull request description accurate?

kiya00:
currently the single sort_allgather puts the allgather op right before the wait op, and the number in limit_in_flight_allgathers (INT_MAX, 3) will get the allgather to proper position for zero2/3. It can be changed to "sort the allgathers to their consumer order and list them at the beginning of the trace" for sort_allgathers and the limit_in_flight_allgathers  for zero2 could be removed.

kiya00:
After some discussion with @IvanYashchuk , I think we can use the fix suggested by @kshitij12345 , the reduce_scatters before sorting are already in the right position, it's enough to solve the #557 . I think we could consider using this PR if we can not rely on the original order one day. But for now I'll close it.  
```
diff --git a/thunder/distributed/utils.py b/thunder/distributed/utils.py
index 7e0e81a..3ff3337 100644
--- a/thunder/distributed/utils.py
+++ b/thunder/distributed/utils.py
@@ -84,8 +84,9 @@ def sort_communication_ops(execution_trace):
                 case wait_prim_impl.id | unpack_for_fsdp_prim_impl.id:
                     return len(order_in_trace)
                 case reduce_scatter_prim_impl.id | all_reduce_prim_impl.id:
-                    # Prefer larger communication ops over smaller ones
-                    return -node.bsym.args[0].numel
+                    # We want to keep the `reduce` close to it's producer
+                    # (which is close to the original place in the trace).
+                    return order_in_trace[node.bsym]
                 case all_gather_prim_impl.id:
                     return len(order_in_trace) + order_in_trace[node.bsym]
                 case _: 
```

PR #596, opened 2024-06-14
Restore usage of CUDAGraphExecutor with compiled backward function

IvanYashchuk:
We had CUDA Graphs being applied to the backward function before (here's a snapshot from Feb):
https://github.com/Lightning-AI/lightning-thunder/blob/91a45a84f322a029f6fa536371789dcd277dec52/thunder/executors/torch_autograd.py#L169-L174
and this PR restores this functionality. The test already exists:
https://github.com/Lightning-AI/lightning-thunder/blob/6bc5752373c085a284413d84f420371212a9b996/thunder/tests/test_networks.py#L97

### DIFF
--- a/thunder/executors/torch_autograd.py
+++ b/thunder/executors/torch_autograd.py
@@ -10,6 +10,7 @@ from thunder.core.pytree import tree_flatten
 from thunder.core.symbol import BoundSymbol
 from thunder.core.trace import TraceCtx, from_trace, set_tracectx, reset_tracectx
 from thunder.core.transform_common import replace_redundant_inputs
+from thunder.cudagraphs import CUDAGraphExecutor
 
 if TYPE_CHECKING:
     from thunder.core.trace import VariableInterface
@@ -93,6 +94,9 @@ class ThunderFunction(torch.autograd.Function):
         # reference to the saved tensors in the context
         ctx.maybe_clear_saved_tensors()  # Delete the reference to all saved tensors in the context
         grads = ctx.compiled_backward([saved_tensors_list, ctx.saved_other], args)
+
+        if isinstance(ctx.compiled_backward, CUDAGraphExecutor):
+            saved_tensors_list.clear()
 
         # Inside the compiled backward we must clear the saved_tensors_list
         assert not saved_tensors_list, "saved_tensors_list must be empty after calling compiled_backward"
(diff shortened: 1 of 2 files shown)


thunder/executors/torch_autograd.py
```diff
@@ -94,6 +95,9 @@ def backward(ctx, *args):
         ctx.maybe_clear_saved_tensors()  # Delete the reference to all saved tensors in the context
         grads = ctx.compiled_backward([saved_tensors_list, ctx.saved_other], args)
 
+        if isinstance(ctx.compiled_backward, CUDAGraphExecutor):
+            saved_tensors_list.clear()
```
nikitaved, line 99: Fine for now, but I am working on fixing this bit.

nikitaved: BTW, why is `ctx.saved_other` is not treated similarly?

IvanYashchuk: It's not treated similarly because we care only about Tensor objects and CUDA memory usage and `ctx.saved_other` is most likely a tuple, not a list, to prevent accidental changes. We need to remove all possible references to Tensor objects as soon as possible during backward to not increase peak memory usage while computing the result of backward. `ctx.saved_other` includes all other objects that are not Tensors but still required to compute the backward function.
This memory optimization doesn't apply to CUDA Graphs because simple usage of CUDA Graphs expects static memory region to be available and used for all saved tensors from forward to backward, we can't free this piece of memory.
Clearing the list in this if statement doesn't have any effect on memory, it's only to skip the assert in the next line. I'd like to keep the assert because it ensures correct logic execution inside the generated backward Python code.

nikitaved: Luckily, there is a solution that does not require doing so explicitly in this part of the code. I believe it should be agnostic to such things

nikitaved: And we can still safe on things which are not static?

IvanYashchuk: Could you please create an issue for discussion summarizing what is problematic today and how you plan to address it?

nikitaved: I will show in the PR. But the idea is very much what we discussed together. Having a storage that returns proxy objects with references to the storage with the ability to "clean"  memory once del is called upon them.

t-vi (approved):
Seems like a good basis for @nikitaved 's improvements. I wonder if there should be a test somewhere to make sure we don't break it again.

nikitaved:
Not sure about improvements, thought :) Just trying to potentially generalize things a bit, and maybe even future-proof them...

nikitaved:
@t-vi , yes, I plan to extend a test with reproductions. Currently we run things only once, while constructing the graph. We should run it multiple times to test whether the cached graph is fetched and whether we, indeed, do not remove memory where it should be present.

PR #597, opened 2024-06-14
Partially support in-place ops and tensor aliases

crcrpar:
when
- the original of views is not ``trace.args`` nor ``trace.kwargs``
- thr original of views has no other consumers than ops creating the
  views
- views have the same number of elements as the original

then replace the use of the originals in the later bsyms with the
reshaped views that have been in-place updated.

### DIFF
--- a/thunder/__init__.py
+++ b/thunder/__init__.py
@@ -510,9 +510,13 @@ def jit(
 
             prologue_traces = [prologue_trc]
             computation_traces = [computation_trc]
-            check_inplace_to_views(computation_trc)
+            orig_to_view_swap_map = check_inplace_to_views(computation_trc)
             if not compile_options.get("skip_inplace_functionalization", False):
-                computation_traces.extend(functionalize_inplace_ops(computation_trace=computation_trc))
+                computation_traces.extend(
+                    functionalize_inplace_ops(
+                        computation_trace=computation_trc, orig_to_view_swap_map=orig_to_view_swap_map
+                    )
+                )
                 computation_trc = computation_traces[-1]
 
             if epilogue_trc is not None:
--- a/thunder/core/transform_common.py
+++ b/thunder/core/transform_common.py
@@ -573,9 +573,6 @@ def functionalize_inplace_ops(
             new_bsyms.append(new_functional_bsym)
             bsym_inplace_to_functional[new_bsym] = new_functional_bsym
 
-            in_tensor = new_functional_bsym.flat_proxy_args[0]
-            prod_bsym: BoundSymbol = producer_map[in_tensor]
-
     functionalized_computation_trace = from_trace(computation_trace)
     functionalized_computation_trace.bound_symbols = new_bsyms
     functionalized_computation_trace.set_provenance(TraceProvenance("Functionalize in-place ops"))
--- a/thunder/core/transform_common.py
+++ b/thunder/core/transform_common.py
@@ -419,13 +419,8 @@ def check_inplace_to_views(computation_trace: Trace) -> None:
         for in_tensor in filter(lambda p: isinstance(p, TensorProxy), bsym.flat_proxy_args):
             prod_bsym: BoundSymbol = producer_bsyms[in_tensor]
             utils.check(
-                not has_tags(prod_bsym, {prims.OpTags.SHAPE_OP}) or prod_bsym.sym in allowed_ltorch_ops,
-                lambda: f"in-place op to view tensors is not allowed but `{bsym.sym.id}` takes `{prod_bsym.sym.id}` output `{in_tensor}`",
-                NotImplementedError,
-            )
-            utils.check(
-                prod_bsym.sym != ltorch.contiguous,
-                lambda: f"in-place op to `torch.Tensor.contiguous` output is not allowed but `{bsym.sym.id}` takes `{prod_bsym.sym.id}` output `{in_tensor}`",
+                prod_bsym.sym not in ltorch._syms_returning_runtime_dependently_views,
+                lambda: f"in-place op of `{bsym.sym.id}` to `{prod_bsym.sym.id}` output `{in_tensor}` is not supported",
                 NotImplementedError,
             )
 
--- a/thunder/core/transform_common.py
+++ b/thunder/core/transform_common.py
@@ -412,6 +412,8 @@ def check_inplace_to_views(computation_trace: Trace) -> dict[VariableInterface, 
     for bsym in filter(lambda b: has_tag(b, prims.OpTags.IN_PLACE), computation_trace.bound_symbols):
         in_tensor: TensorProxy = list(filter(lambda p: isinstance(p, TensorProxy), bsym.flat_proxy_args))[0]
 
+        if in_tensor in trace_args_set:
+            continue
         prod_bsym: BoundSymbol = producer_bsyms[in_tensor]
         orig_tensor = prod_bsym.flat_proxy_args[0]
         consumer_of_orig_tensor = consumers[orig_tensor]
--- a/thunder/core/transform_common.py
+++ b/thunder/core/transform_common.py
@@ -9,10 +9,10 @@ from functools import partial
 import thunder
 import thunder.core.prims as prims
 from thunder.core.baseutils import BoundSymbolInterface
-from thunder.core.proxies import Proxy, variableify, Variable, TensorProxy
+from thunder.core.proxies import Proxy, variableify, Variable, TensorProxy, unvariableify
 from thunder.core.pytree import tree_flatten, tree_map, tree_unflatten
 from thunder.core.symbol import BoundSymbol, BoundSymbolRHS, has_tags
-from thunder.core.trace import from_trace, TraceProvenance, TraceCtx as Trace
+from thunder.core.trace import from_trace, TraceProvenance, TraceCtx as Trace, tracectx
 from thunder.core.utils import ProxyDict, producers, check
 
 if TYPE_CHECKING:
@@ -388,12 +388,17 @@ class PostOptimizationTransform(Transform, ABC):
         pass
 
 
-def check_inplace_to_views(computation_trace: Trace) -> None:
-    """Error out if ``computation_trace`` has any in-place op of `torch.reshape`'s output."""
+def check_inplace_to_views(computation_trace: Trace) -> dict[VariableInterface, TensorProxy]:
+    """Error out if in-place op that outputs of different number of elements from the input and the input has other consumers."""
     from thunder.core import utils
     import thunder.torch as ltorch
 
     producer_bsyms = producers(computation_trace)
+    trace_args_set = ProxyDict()
+    for a in filter(
+        lambda a: isinstance(a, TensorProxy), tree_flatten((computation_trace.args, computation_trace.kwargs))[0]
+    ):
+        trace_args_set[a] = a
 
     # note(crcrpar): Why not using :func:`~thunder.core.symbol.has_tags`?
     # Because it looks into `.sym.tags` of the input bsym and its subsymbols,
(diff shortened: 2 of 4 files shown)


IvanYashchuk:
Are you attempting to generate a trace similar to this one?
```py
def f(a: torch.Tensor, b: torch.Tensor) -> tuple[torch.Tensor, torch.Tensor]:
    c = torch.exp(a)
    d = torch.tanh(b)
    e = c.flatten()
    e += d.flatten() # new_e = e + d.flatten(); new_c = reshape(new_e, c.shape);
    d.div_(a) # new_d = d.div(a)
    return c, d # new_c, new_d 
```

crcrpar:
> Are you attempting to generate a trace similar to this one?
> 
> ```python
> def f(a: torch.Tensor, b: torch.Tensor) -> tuple[torch.Tensor, torch.Tensor]:
>     c = torch.exp(a)
>     d = torch.tanh(b)
>     e = c.flatten()
>     e += d.flatten() # new_e = e + d.flatten(); new_c = reshape(new_e, c.shape);
>     d.div_(a) # new_d = d.div(a)
>     return c, d # new_c, new_d 
> ```

yes

crcrpar:
nvfuser executor doesn't seem to love this logic though (as of ad1f2f9)

```
  File "/home/mkozuki/ghq/github.com/Lightning-AI/lightning-thunder/snippet.py", line 50, in <module>
    main()
  File "/home/mkozuki/ghq/github.com/Lightning-AI/lightning-thunder/snippet.py", line 21, in main
    c, d, e = jit_f(a, b)
  File "/home/mkozuki/ghq/github.com/Lightning-AI/lightning-thunder/thunder/__init__.py", line 667, in fn_
  File "/home/mkozuki/ghq/github.com/Lightning-AI/lightning-thunder/thunder/__init__.py", line 223, in cache_info_wrapper
    res = fn(*args, **kwargs)
  File "/home/mkozuki/ghq/github.com/Lightning-AI/lightning-thunder/thunder/__init__.py", line 610, in get_computation_and_inputs
    thunder.core.transform_common._inplace_copy_sanity_check(computation_trc)
  File "/home/mkozuki/ghq/github.com/Lightning-AI/lightning-thunder/thunder/core/transform_common.py", line 84, in _inplace_copy_sanity_check
    check(copy_to_arg, "'copy_to' argument")
  File "/home/mkozuki/ghq/github.com/Lightning-AI/lightning-thunder/thunder/core/transform_common.py", line 79, in check
    raise NotImplementedError(
NotImplementedError: t8 = prims.reshape(e, (2, 2))  # t8: "cuda:0 f32[2, 2]" trying to use e (the 'copy_to' argument of 'prims.copy_') as input, which is not safe. There is a risk of accessing the wrong memory. If you are sure you don't want to use this check, it can be disabled by setting `disable_inplace_copy_check=True` in `thunder.jit`.
```

thunder/core/symbol.py
```diff
@@ -268,6 +268,21 @@ def __call__(self, *args, **kwargs):
...
+
+            if OpTags.SHAPE_OP in self.tags:
+                flat_result, _ = tree_flatten(result)
+                check(
+                    len(flat_result) == 1,
+                    lambda: f"{self.id=} has {self.tags} but {len(flat_result)} results, not 1",
+                )
+                result_t: TensorProxy = flat_result[0]
+                check_type(result_t, TensorProxy)
+                result_t.set_view_of(tree_flatten((args, kwargs))[0][0])
```
mruberry, line 284: This isn't correct. The result of a shape operation is not always a view of the original tensor

mruberry:
Let's prioritize a design review for inplace operations, and before the review let's share a document with some cases and examples. I'm not really sure how this would address the challenges of inplace operations

jjsjann123:
> nvfuser executor doesn't seem to love this logic though (as of [ad1f2f9](https://github.com/Lightning-AI/lightning-thunder/commit/ad1f2f923dad905edf808aef47db24c551805313))
> 
> ```
> Traceback (most recent call last):
>   File "/home/mkozuki/ghq/github.com/Lightning-AI/lightning-thunder/snippet.py", line 50, in <module>
>     main()
>   File "/home/mkozuki/ghq/github.com/Lightning-AI/lightning-thunder/snippet.py", line 21, in main
>     c, d, e = jit_f(a, b)
>   File "/home/mkozuki/ghq/github.com/Lightning-AI/lightning-thunder/thunder/__init__.py", line 667, in fn_
>     cache_entry, inps, pro_to_epi = get_computation_and_inputs(*args, **kwargs)
>   File "/home/mkozuki/ghq/github.com/Lightning-AI/lightning-thunder/thunder/__init__.py", line 223, in cache_info_wrapper
>     res = fn(*args, **kwargs)
>   File "/home/mkozuki/ghq/github.com/Lightning-AI/lightning-thunder/thunder/__init__.py", line 610, in get_computation_and_inputs
>     thunder.core.transform_common._inplace_copy_sanity_check(computation_trc)
>   File "/home/mkozuki/ghq/github.com/Lightning-AI/lightning-thunder/thunder/core/transform_common.py", line 84, in _inplace_copy_sanity_check
>     check(copy_to_arg, "'copy_to' argument")
>   File "/home/mkozuki/ghq/github.com/Lightning-AI/lightning-thunder/thunder/core/transform_common.py", line 79, in check
>     raise NotImplementedError(
> NotImplementedError: t8 = prims.reshape(e, (2, 2))  # t8: "cuda:0 f32[2, 2]" trying to use e (the 'copy_to' argument of 'prims.copy_') as input, which is not safe. There is a risk of accessing the wrong memory. If you are sure you don't want to use this check, it can be disabled by setting `disable_inplace_copy_check=True` in `thunder.jit`.
> ```

The warning is added explicitly (by @kiya00 IIRC).

nvfuser's support of inplace update is limited at this moment. We don't guarantee order of operations in generated kernel, which means if you try to use a tensor that has been updated inside a fusion, you might ended up with a silent result. Hence we decided to throw a warning here.

We can explore WAR in thunder for this. i.e. we can break the fusion region, which ensures the use after update happens in a follow up kernel, so there's no chance for nvfuser to change the memory access sequence.
But this also requires thunder to ensure the order of the regions, which I don't think we have any restriction in as of today.

jjsjann123:
> > Are you attempting to generate a trace similar to this one?
> > ```python
> > def f(a: torch.Tensor, b: torch.Tensor) -> tuple[torch.Tensor, torch.Tensor]:
> >     c = torch.exp(a)
> >     d = torch.tanh(b)
> >     e = c.flatten()
> >     e += d.flatten() # new_e = e + d.flatten(); new_c = reshape(new_e, c.shape);
> >     d.div_(a) # new_d = d.div(a)
> >     return c, d # new_c, new_d 
> > ```
> 
> yes

I think we should clear up our target first.
In the example here, the original program could be functionalized as a whole.

```
def f(a: torch.Tensor, b: torch.Tensor) -> tuple[torch.Tensor, torch.Tensor]:
    c = torch.exp(a)
    d = torch.tanh(b)

    e = c.flatten()
    e += d.flatten()

    d.div_(a)
    return c, d, e
```

Hence the resulting trace *should NOT* show any `copy_` within. Since we are already calling out of place math ops, the memory save from doing inplace is gone.
Blindly following the semantics here with `copy_` seems to be meaninglessly adding complexity to follow up mutation passes.

thunder/core/symbol.py
crcrpar: you're right but it'd require strides information to tell if the output is a view or a copy

crcrpar:
> I take it as that, the trace we are showing in PR description is just a side effect from that our in-place pass isn't functionalizing all the corner cases yet?!

what would you think is a better approach propagating the in-place op on a view?

thunder/core/symbol.py
IvanYashchuk: > The result of a shape operation is not always a view of the original tensor

This is true only for three operations `reshape`, `reshape_as` and `flatten`.
PyTorch documentation says that users shouldn't rely on the result of `reshape` being a view or copy, the same is translated to `flatten` because `reshape` is used to implement it. In Thunder, `reshape` should be assumed to always produce a copy, if `reshape+in_place_op` is traced from the user code there can be a warning under a debug flag but not in the normal run.

There's also `contiguous` that might return the input itself. It might be better to error out for `contiguous+in_place_op` at first for the user to resolve the ambiguity.

mruberry: Refusing to execute programs with ambiguous inplace semantics seems totally fine, but we should consider that approach in a design review. There may be more operations than contiguous that detect when a no-op occurs and just return their input, too, so we should think what the semantics of those operations are.

IvanYashchuk: > There may be more operations than contiguous that detect when a no-op occurs and just return their input, too, so we should think what the semantics of those operations are.

What are these operations?

mruberry: `to` variations, we could have `a + 0` return `a`, `abs` on dtypes without negative values, `isfinite` on exact dtypes, ...

IvanYashchuk: > `to` variations, we could have `a + 0` return `a`, `abs` on dtypes without negative values, `isfinite` on exact dtypes, ...

These examples are resolvable at Thunder tracing time (except for zero addition) and eliminated by DCE today.

mruberry: Sure, but we should be clear that our design, then, is that if an operation decides to no-op it does not create a new storage (even if it otherwise would), and inplace operations on the result of the operation will affect the input's storage, too

jjsjann123:
> > I take it as that, the trace we are showing in PR description is just a side effect from that our in-place pass isn't functionalizing all the corner cases yet?!
> 
> what would you think is a better approach propagating the in-place op on a view? The trace tries to propagate the in-place update on a view into the original tensor

I'm not saying what you are doing here for `view` is not right.

I'm just pointing out this is a case showing that our in-place pass is missing optimization opportunities here, where an in-place update on an intermediate tensor isn't being optimized away.

nikitaved:
I get it why PyTorch might need `version` for eager mode. But can't we just get all this from the trace if needed?

thunder/core/proxies.py
```diff
@@ -1277,6 +1294,31 @@ def distparallel_type(self):
...
+    def version(self):
+        return self._version
+
+    @property
+    def view_of(self):
+        return self._view_of
+
+    @property
+    def is_view(self) -> bool:
+        return self.view_of is not None
```
nikitaved, line 1307: Why do we have to store these as properties? Seems like `version` is more reminiscent of the eager mode. Looks like all these versions and view_of can be read from a trace and packed into a datastructure of a UnionFind type...

crcrpar: Yes this should be simplified. I think I drastically change the implementation following #630 and #633. I also think of using views rather than original tensors to avoid may copies based on @jjsjann123 advice

thunder/__init__.py
```diff
@@ -510,9 +510,13 @@ def get_computation_and_inputs(*args, **kwargs):
 
             prologue_traces = [prologue_trc]
             computation_traces = [computation_trc]
-            check_inplace_to_views(computation_trc)
+            orig_to_view_swap_map = check_inplace_to_views(computation_trc)
```
t-vi, line 513: I'm not totally happy about having this in `thunder/__init__.py`. Maybe we could have a function `check_and_functionalize_...` to which we 
 pass `compile_options.get("skip_inplace_functionalization", False)` ?

thunder/core/transform_common.py
```diff
@@ -476,26 +516,36 @@ def is_functionalizable(bsym: BoundSymbol) -> bool:
         bsyms.append(new_bsym)
 
     intermediate_trace = from_trace(computation_trace)
-    intermediate_trace.bound_symbols = bsyms[:]
+    intermediate_trace.bound_symbols = bsyms
     intermediate_trace.set_provenance(TraceProvenance("Intermediate trace of `functionalize_inplace_ops`"))
-    del bsyms
```
t-vi, line 481: Maybe we keep the deletion? That would make it clear that it's now owned by `intermediate_trace`.

thunder/core/transform_common.py
```diff
@@ -388,49 +388,70 @@ def transform_trace(self, computation_trace: Trace, **kwargs):
...
     # Because it looks into `.sym.tags` of the input bsym and its subsymbols,
     # thus even `ltorch.batch_norm` is regarded as `prims.OpTags.IN_PLACE`.
     def has_tag(bsym: BoundSymbol, tag: prims.OpTags) -> bool:
         return bsym.sym.tags and tag in bsym.sym.tags
 
-    # note(crcrpar): Following ops would not look like a `prims.OpTags.SHAPE_OP`
-    # especially with respect to the relationship between the input and the output
-    # but some of their sub boundsymbols are. Therefore `thunder.core.symbol.gather_tags` gives it to them.
-    allowed_ltorch_ops = {
-        ltorch.batch_norm,
```
t-vi, line 408: Great to have this in a more principled way.

t-vi (commented):
Looks great, thank you @crcrpar

t-vi (approved):

t-vi:
Let's move additions to a follow-up.

PR #598, opened 2024-06-14
use provided comp in test_grad.py

k223kim:
<details>

</details>

## What does this PR do?

This is not directly related to any issue. However, this was needed for #530. 
Also, I am not sure if there should be a test case for this. Please let me know if so!

## PR review

## Did you have fun?

⚡️


thunder/tests/test_grad.py
```diff
@@ -327,7 +328,8 @@ def check_vjp(f, *primals, executor="torch", atol=1e-5, rtol=1.3e-6):
     if J_u_v.isnan().any():
         # TODO: find a better way to handle NaNs in finite differences
         return  # skip this sample
-    torch.testing.assert_close(J_u_v, u_J_star_v, atol=atol, rtol=rtol, check_device=False)
+    partial_comp = partial(comp, atol=atol, rtol=rtol, check_device=False)
```
t-vi, line 331: I do wonder if atol and rtol and check_device should be applied to custom comparison functions, or whether, in fact, the custom comparison function could replace passing non-default atol/rtol (because that's what we are using it for after all).

k223kim: Observing the most use-cases for custom comparators, it seems like we should have atol, rtol, and check_device. If not, I think opinfos.py should be modified accordingly. Are you expecting custom comparators to be used in a different way?

t-vi: I would expect the custom comparator to do these things with it's own values and not feed atol/rtol to it (and I'm not sure whether the current PR would override the atol/rtol in the custom comparator's signature).
I was hoping to use the `test_ops` for comparison, I think there is this TODO: 
https://github.com/Lightning-AI/lightning-thunder/blob/21a222b180009616a4cc48176958b4506894a330/thunder/tests/test_ops.py#L52C1-L52C2

k223kim: I see what you mean! Thanks for the explanation. I think what you said (`todo these things with it's own values and not feed atol/rtol to it`) makes more sense. Let me update this!

k223kim:
FYI, I have updated `type_as_sample` in `opinfos.py` because it was failing a CI test. I passed atol/rtol values that were passed for `test_vjp_correctness` earlier.

thunder/tests/test_grad.py
```diff
@@ -282,7 +282,7 @@ def _dot(x, y):
     return sum([torch.dot(a.ravel().type(torch.float64), b.ravel().type(torch.float64)) for a, b in zip(x, y)])
 
 
-def check_vjp(f, *primals, executor="torch", atol=1e-5, rtol=1.3e-6):
+def check_vjp(f, *primals, comp, executor="torch"):
```
t-vi, line 285: Do you know what the default comparison function has as tolerances?

t-vi:
```
   if rtol is None and atol is None:
        rtol, atol = default_tolerances(
            actual,
            expected,
            dtype_precisions={
                torch.float16: (1e-3, 1e-3),
                torch.float32: (1e-4, 1e-5),
                torch.float64: (1e-5, 1e-8),
            },
        )

```
I wonder if we need to make the default comp depend on `template.__name__` to give `test_vjp_correctness` the more generous limits by default here (even more pretty, but maybe overkill would be to annotate the template with the desired default tolerances).

https://github.com/Lightning-AI/lightning-thunder/blob/c155ca0f9ee2720c4235e729a38bd2bd817635d9/thunder/tests/framework.py#L315

k223kim: I could definitely do that. When taking a closer look at `test_vjp_correctness` in `opinfos.py`, I can see how most of them are skipped due to its flakiness and speed. Do you think having a more generous limit would resolve this issue? If so, I think there is a point to customize the tolerances for `test_vjp_correctness`. Even if that is not the case, out of 50 skipped `test_vjp_correctness`, I think around 1/5 of them are related to the tolerance issue. Do you think that is a enough portion to consider a more generous tolerance? Additionally, after a more generous tolerance is given, I think we have to check the ones that we have skipped to see if it passes now. Let me know your thoughts!

t-vi: Yes, it would be neat to get more test coverage, but to my mind, we can merge it if it works. There seems something here (the `Tensor.to` method apparently, I have not yet looked why):
```
FAILED thunder/tests/test_grad.py::test_vjp_correctness_to_nvfuser_cuda_float64 - AssertionError: The values for attribute 'device' do not match: cpu != cuda:0.
FAILED thunder/tests/test_grad.py::test_vjp_correctness_to_torch_cuda_float64 - AssertionError: The values for attribute 'device' do not match: cpu != cuda:0.
```

t-vi: This is because the finite differences of the inputs will be on the original device and the output on the target, so the right thing seems to be amending the `to_opinfo` like this:
```
to_opinfo = OpInfo(
    ltorch.to,
    sample_input_generator=to_sample_generator,
    torch_reference=torch.Tensor.to,
    test_directives=(
        DecorateInfo(
            custom_comparator(lambda a, b: assert_close(a, b.to(a.device))),
            "test_vjp_correctness",
        ),
    ),
)
data_movement_ops.append(to_opinfo)
```
Should we add that and merge?

k223kim: Thanks so much for looking into this Tom! I originally thought that this could be something related to the custom comparator itself, but it seems like it is more relevant to the nature of `Tensor.to` (I wonder how this was passed before hand). Let me update this!

k223kim: Oh, to answer my own question, we passed `check_device=False` for `test_vjp_correctness` earlier!

thunder/tests/opinfos.py
```diff
@@ -2924,6 +2924,12 @@ def to_sample_generator(op, device, dtype, requires_grad, **kwargs):
     ltorch.to,
     sample_input_generator=to_sample_generator,
     torch_reference=torch.Tensor.to,
+    test_directives=(
+        DecorateInfo(
+            custom_comparator(lambda a, b: assert_close(a, b.to(a.device))),
```
t-vi, line 2929:
```suggestion
            custom_comparator(partial(assert_close, check_device=False)),
```

t-vi (approved):
Thank you @k223kim

Issue #338, opened 2024-05-01
Implement _VariableFunctionsClass.max of torch


athitten:
### 🚀 Feature

Implement [`torch.max`](https://pytorch.org/docs/stable/generated/torch.max.html)

### Motivation

NeMo's [MegatronNeVaModel](https://github.com/NVIDIA/NeMo/blob/main/examples/multimodal/multimodal_llm/neva/neva_pretrain.py)

cc @tfogal

PR #599, opened 2024-06-14
Add torch.max

kshitij12345:
Fixes: https://github.com/Lightning-AI/lightning-thunder/issues/338

`torch.max` is an interesting operator with multiple overloads and different return types based on the overload.
Ref: https://pytorch.org/docs/stable/generated/torch.max.html

In this PR, we implement `torch.max` as decomposition of `amax`, `argmax` and `maximum` depending on the input.

Also, we add grad formula for `maximum` and also `argmax` (which is no gradients for input).

**Problem**: However, using `amax` leads to a difference in gradient computation.

As noted in the PyTorch [docs](https://pytorch.org/docs/stable/generated/torch.amax.html), 
> amax/amin evenly distributes gradient between equal values, while max(dim)/min(dim) propagates gradient only to a single index in the source tensor.

I am planning to tackle this seperately in a follow-up but wanted to know if we should add a new primitive or if there are other ways around the same.

If this PR looks good, will also follow-up with `torch.min`  (and relevant updates) seperately.

### DIFF
--- a/thunder/core/transforms.py
+++ b/thunder/core/transforms.py
@@ -1364,10 +1364,11 @@ def _maximum_grad(a: TensorProxy, b: TensorProxy, /):
     g = get_grad(fwd)
 
     # Compute sub-gradient if `a == b`
-    a_grad = prims.where(a == b, g / 2, 0.0)
-    b_grad = prims.where(a == b, g / 2, 0.0)
-    a_grad = prims.where(a > b, g, a_grad)
-    b_grad = prims.where(b > a, g, b_grad)
+    # NOTE: We evenly distribute the gradient where the values are equal.
+    vals_equal_grad = prims.where(a == b, g / 2, 0.0)
+
+    a_grad = prims.where(a > b, g, vals_equal_grad)
+    b_grad = prims.where(b > a, g, vals_equal_grad)
 
     put_grad(a, a_grad)
     put_grad(b, b_grad)
--- a/thunder/core/transforms.py
+++ b/thunder/core/transforms.py
@@ -1357,6 +1357,27 @@ def _embedding_prim_grad(
 
 register_grad(pids.EMBEDDING, _embedding_prim_grad)
 
+
+def _maximum_grad(a: TensorProxy, b: TensorProxy, /):
+    fwd = prims.maximum(a, b)
+
+    g = get_grad(fwd)
+
+    a_grad = prims.where(a == b, g / 2, 0.0)
+    b_grad = prims.where(a == b, g / 2, 0.0)
+    a_grad = prims.where(a > b, g, a_grad)
+    b_grad = prims.where(b > a, g, b_grad)
+
+    put_grad(a, a_grad)
+    put_grad(b, b_grad)
+    return fwd
+
+
+register_grad(pids.MAXIMUM, _maximum_grad)
+
+# This operation creates no grad associations
+register_grad(pids.ARGMAX, prims.argmax)
+
 #
 # Phantom grad transform helpers
 #
--- a/thunder/core/transforms.py
+++ b/thunder/core/transforms.py
@@ -1365,7 +1365,7 @@ def _maximum_grad(a: TensorProxy, b: TensorProxy, /):
 
     # Compute sub-gradient if `a == b`
     # NOTE: We evenly distribute the gradient where the values are equal.
-    vals_equal_grad = prims.where(a == b, g / 2, 0.0)
+    vals_equal_grad = prims.where(a == b, g / 2, g)
 
     a_grad = prims.where(a > b, g, vals_equal_grad)
     b_grad = prims.where(b > a, g, vals_equal_grad)
--- a/thunder/core/transforms.py
+++ b/thunder/core/transforms.py
@@ -1363,12 +1363,22 @@ def _maximum_grad(a: TensorProxy, b: TensorProxy, /):
 
     g = get_grad(fwd)
 
+    # NOTE: NaN propagation if either `a` or `b` is a NaN, then both elements receive `g` as gradient.
+    # This is because comparison in presence of NaN is False (except for Not Equal which is always True).
+    # Eg. Here `a` = NaN and `b` = 42
+    # sub_g = where(NaN == 42 i.e. False, g / 2, g)  # sub_grad = g
+    # grad_a = where(NaN < 42 i.e. False, 0., sub_g)  # grad_a = sub_g = g
+    # grad_b = where(42 < NaN i.e. False, 0., sub_g)  # grad_b = sub_g = g
+    # NOTE: If `g` is `NaN` then it will be propagated as the gradient of max element between a and b
+    # and if both are equal, then as we evenly distributing the gradients, `NaN` will propagate through
+    # the gradients of both `a` and `b`.
+
     # Compute sub-gradient if `a == b`
     # NOTE: We evenly distribute the gradient where the values are equal.
-    vals_equal_grad = prims.where(a == b, g / 2, g)
+    sub_grad = prims.where(a == b, g / 2, g)
 
-    a_grad = prims.where(a > b, g, vals_equal_grad)
-    b_grad = prims.where(b > a, g, vals_equal_grad)
+    a_grad = prims.where(a < b, 0.0, sub_grad)
+    b_grad = prims.where(b < a, 0.0, sub_grad)
 
     put_grad(a, a_grad)
     put_grad(b, b_grad)
--- a/thunder/core/transforms.py
+++ b/thunder/core/transforms.py
@@ -1363,6 +1363,7 @@ def _maximum_grad(a: TensorProxy, b: TensorProxy, /):
 
     g = get_grad(fwd)
 
+    # Compute sub-gradient if `a == b`
     a_grad = prims.where(a == b, g / 2, 0.0)
     b_grad = prims.where(a == b, g / 2, 0.0)
     a_grad = prims.where(a > b, g, a_grad)
--- a/thunder/tests/opinfos.py
+++ b/thunder/tests/opinfos.py
@@ -5000,6 +5000,15 @@ def max_sample_generator(op, device, dtype, requires_grad, **kwargs):
     def make_with_extremal_value(shape, extremal, percentage=0.5):
         return replace_random_percentage(make(shape), extremal, percentage=percentage)
 
+    # NOTE: Gradient Computation with multiple max values
+    # Currently, if there are multiple `max` values
+    # `torch` eager - gradients max(dim) propagates gradient only to a single index in the source tensor
+    # `thunder` - gradients are distributed evenly.
+    # So, we use the function below to create tensor with unique values.
+    def make_unique_t(shape):
+        # Add two random inputs, so it is unlikely to have same value across the tensor elements.
+        return make(shape) + make(shape)
+
     # shape, dim, keepdim
     cases = (
         ((2, 2, 3), 1, True),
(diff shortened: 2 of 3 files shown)


thunder/core/transforms.py
```diff
@@ -1357,6 +1357,27 @@ def _embedding_prim_grad(
...
+
+def _maximum_grad(a: TensorProxy, b: TensorProxy, /):
+    fwd = prims.maximum(a, b)
+
+    g = get_grad(fwd)
+
+    a_grad = prims.where(a == b, g / 2, 0.0)
+    b_grad = prims.where(a == b, g / 2, 0.0)
+    a_grad = prims.where(a > b, g, a_grad)
+    b_grad = prims.where(b > a, g, b_grad)
```
nikitaved, line 1369: Took me a bit to realize that what we are computing is a subgradient :)

nikitaved: nit: Maybe we should isolate `g / 2`? Or is it optimized out from the trace?

thunder/tests/opinfos.py
```diff
@@ -4992,6 +4993,52 @@ def amax_amin_sample_generator(op, device, dtype, requires_grad, **kwargs):
...
 
+def max_sample_generator(op, device, dtype, requires_grad, **kwargs):
+    # For grad test stability it's better to use wider range of values
+    make = partial(make_tensor, device=device, dtype=dtype, requires_grad=requires_grad, low=-1000, high=1000)
+
+    # shape, dim, keepdim
+    cases = (
+        ((2, 2, 3), 1, True),
+        ((2, 3, 1), 0, False),
+    )
```
nikitaved, line 5004: How does it work with NaNs, infs?

thunder/core/transforms.py
kshitij12345: We just have one `g/2` in the final trace.

```python
import torch
import thunder

def foo(*args):
    return torch.max(*args)

jfoo = thunder.jit(foo)

x = torch.randn(3, 3, requires_grad=True)

def _test(*args):
    expected = foo(*args)
    actual = jfoo(*args)
    torch.testing.assert_close(actual, expected)

# _test(x)
# _test(x, 0)
# _test(x, 0, True)
_test(x, x + 1)

print(thunder.last_backward_traces(jfoo)[-1])

```

Backward Trace
```python
import torch

@torch.no_grad()
@no_autocast
  # cotangents: "Collection"
  C0, _, = saved_for_backward
  del saved_for_backward
  t1, = cotangents
  del cotangents
  t_0, t_1, = C0
  clear_mutable_collection(C0)
  del C0
  t12 = torch.true_divide(t1, 2)  # t12: "cpu f32[3, 3]"
    # t12 = ltorch.true_divide(t1, 2)  # t12: "cpu f32[3, 3]"
      # _ = prims.convert_element_type(2, float)
      # t12 = prims.div(t1, 2.0)  # t12: "cpu f32[3, 3]"
  t11 = torch.eq(t_0, t_1)  # t11: "cpu b8[3, 3]"
    # t11 = ltorch.eq(t_0, t_1)  # t11: "cpu b8[3, 3]"
      # t11 = prims.eq(t_0, t_1)  # t11: "cpu b8[3, 3]"
  t16 = torch.gt(t_1, t_0)  # t16: "cpu b8[3, 3]"
    # t16 = ltorch.gt(t_1, t_0)  # t16: "cpu b8[3, 3]"
      # t16 = prims.gt(t_1, t_0)  # t16: "cpu b8[3, 3]"
  t14 = torch.gt(t_0, t_1)  # t14: "cpu b8[3, 3]"
    # t14 = ltorch.gt(t_0, t_1)  # t14: "cpu b8[3, 3]"
      # t14 = prims.gt(t_0, t_1)  # t14: "cpu b8[3, 3]"
  del t_0, t_1
  t13 = torch.where(t11, t12, 0.0)  # t13: "cpu f32[3, 3]"
    # t13 = ltorch.where(t11, t12, 0.0)  # t13: "cpu f32[3, 3]"
      # t13 = prims.where(t11, t12, 0.0)  # t13: "cpu f32[3, 3]"
  del t11, t12
  t17 = torch.where(t16, t1, t13)  # t17: "cpu f32[3, 3]"
    # t17 = ltorch.where(t16, t1, t13)  # t17: "cpu f32[3, 3]"
      # t17 = prims.where(t16, t1, t13)  # t17: "cpu f32[3, 3]"
  del t16
  t15 = torch.where(t14, t1, t13)  # t15: "cpu f32[3, 3]"
    # t15 = ltorch.where(t14, t1, t13)  # t15: "cpu f32[3, 3]"
      # t15 = prims.where(t14, t1, t13)  # t15: "cpu f32[3, 3]"
  del t14, t1, t13
  return (t15, t17)
```

thunder/tests/opinfos.py
kshitij12345: It always returns the first occurence.

```python
import torch
import thunder

def foo(*args):
    return torch.max(*args)

jfoo = thunder.jit(foo)

for dev in ("cpu", "cuda"):
    x1 = torch.tensor([float("nan"), float("nan")])
    x2 = torch.tensor([float("inf"), float("inf")])

    def _test(*args):
        expected = foo(*args)
        actual = jfoo(*args)
        torch.testing.assert_close(actual, expected, equal_nan=True)

    for x in (x1, x2):
        _test(x)
        # Test that same indices are returned
        _test(x, 0)
        # Test that same indices are returned
        _test(x, 0, True)
        _test(x, x)

```

However, to add test with multiple repeating values - we will have to add a fix for this problem from the PR desc

> Problem: However, using amax leads to a difference in gradient computation.

> amax/amin evenly distributes gradient between equal values, while max(dim)/in(dim) propagates gradient only to a single index in the source tensor.

thunder/tests/opinfos.py
```diff
@@ -5007,6 +5010,14 @@ def max_sample_generator(op, device, dtype, requires_grad, **kwargs):
         # overload: torch_max(a: TensorLike, /) -> TensorLike
         yield SampleInput(make(shape))
 
+        if not requires_grad and dtype.is_floating_point:
+            # Currently, if there are multiple `max` values
```
mruberry, line 5014: Nice note

thunder/core/transforms.py
```diff
@@ -1357,6 +1357,28 @@ def _embedding_prim_grad(
...
+
+    put_grad(a, a_grad)
+    put_grad(b, b_grad)
+    return fwd
+
+
+register_grad(pids.MAXIMUM, _maximum_grad)
+
+# This operation creates no grad associations
```
mruberry, line 1379: Nice note

thunder/core/transforms.py
```diff
@@ -1357,6 +1357,28 @@ def _embedding_prim_grad(
 
 register_grad(pids.EMBEDDING, _embedding_prim_grad)
 
+
+def _maximum_grad(a: TensorProxy, b: TensorProxy, /):
+    fwd = prims.maximum(a, b)
+
+    g = get_grad(fwd)
+
+    # Compute sub-gradient if `a == b`
```
mruberry, line 1366: Elaborating a little on this comment might be nice -- about how we want to split the grad in this case

mruberry: What happens if an element in `a` or `b` is `NaN`? What happens if the same element of `a` and `b` is a `NaN`?

thunder/tests/opinfos.py
mruberry: Let's definitely expand the comment to describe what happens in this case!

thunder/core/transforms.py
```diff
@@ -1357,6 +1357,28 @@ def _embedding_prim_grad(
...
+    a_grad = prims.where(a == b, g / 2, 0.0)
+    b_grad = prims.where(a == b, g / 2, 0.0)
```
mruberry, line 1368: If `g` is `NaN` somewhere, then what should the gradient for `a` and `b` be in that location?

thunder/core/transforms.py
```diff
@@ -1357,6 +1357,28 @@ def _embedding_prim_grad(
...
 register_grad(pids.EMBEDDING, _embedding_prim_grad)
 
+
+def _maximum_grad(a: TensorProxy, b: TensorProxy, /):
+    fwd = prims.maximum(a, b)
+
+    g = get_grad(fwd)
+
+    # Compute sub-gradient if `a == b`
+    a_grad = prims.where(a == b, g / 2, 0.0)
```
mruberry, line 1367: Nit: the result of `a == b` and `g / 2` could be explicitly saved for reuse (a CSE pass might do it for you, but it's still nice to be explicit when it's readable)

thunder/tests/opinfos.py
```diff
@@ -4992,6 +4993,71 @@ def amax_amin_sample_generator(op, device, dtype, requires_grad, **kwargs):
...
+    yield (SampleInput(make(3, 3), keepdim=True), RuntimeError, err_msg)
+
+
+max_opinfo = OpInfo(
+    ltorch.torch_max,
+    supports_grad=True,
+    sample_input_generator=max_sample_generator,
+    error_input_generator=max_error_generator,
+    torch_reference=torch.max,
+    # Complex numbers are unordered
```
mruberry, line 5054: Nice note

thunder/torch/__init__.py
```diff
@@ -2044,6 +2045,46 @@ def amin(a, /, dim=None, keepdim: bool = False):
     )
 
 
+# NOTE: Using name `torch_max` to avoid conflict with Python's `max`
```
mruberry, line 2048: Great comment; nice avoidance of the builtin

thunder/torch/__init__.py
```diff
@@ -2044,6 +2045,46 @@ def amin(a, /, dim=None, keepdim: bool = False):
...
+        b = dim
+        return maximum(a, b)
+
+    if dim is None:
+        # overload - torch_max(a: TensorLike, /) -> TensorLike
+        utils.check(not keepdim, lambda: "keepdim=True is invalid for torch.max(a) overload.")
+        dim = list(range(a.ndim))
+        return amax(a, dim, keepdim)
+
+    # overload - torch_max(a: TensorLike, /, dim: int | tuple[int], keepdim: bool = False) -> TensorLike, TensorLike
```
mruberry, line 2081: The handling of these cases is very good; let's elaborate in each comment to specify what that overload conceptually does. For the first two overloads this is very clear, but for this last overload it might be confusing what's happening

mruberry (commented):
Great start, @kshitij12345! I just have a few comments and suggestions; nothing too big. 

@IvanYashchuk may want to take a look as well

thunder/core/transforms.py
kshitij12345: If `NaN` is somewhere in `g`, then it would just propagate depending on whether to which tensor had the max value or in case the values were equal then it would propagate through gradients of both tensors. I dont think there is much that can be done if NaN is already present in `g`

```python
>>> t = torch.tensor(1., requires_grad=True)
>>> s = torch.tensor(2., requires_grad=True)
>>> torch.max(t, s).backward(torch.tensor(float("nan")))
>>> t.grad
tensor(0.)
>>> s.grad
tensor(nan)
>>> 
>>> t = torch.tensor(1., requires_grad=True)
>>> s = torch.tensor(1., requires_grad=True)
>>> torch.max(t, s).backward(torch.tensor(float("nan")))
>>> t.grad
tensor(nan)
>>> s.grad
tensor(nan)
```

mruberry: Sounds great -- would you just add that to the comment and explain how the logic enforces this?

thunder/core/transforms.py
kshitij12345: As `NaN > anything = False` and `NaN == NaN = False` , so even if one element happens to be `NaN`, gradient will propate to both elements

```python
sub_g = where(NaN == 42 i.e. False, g / 2, g)  # sub_grad = g

grad_a = where(NaN > 42 i.e. False, g, sub_g)  # grad_a = sub_g = g
grad_b = where(42 > NaN i.e. False, g, sub_g)  # grad_b = sub_g = g
```

kshitij12345:
Test failures are real, will have a look.

thunder/core/transforms.py
mruberry: Great! Let's update the comment to explain these cases

kshitij12345: Have updated the comment and addressed other comments. This should be ready for another round. Thanks

thunder/core/transforms.py
t-vi: I think it might be prettier to not rely on CSE for it, but seems OK for now.

thunder/tests/opinfos.py
t-vi: Follow-up seems like a good plan. Thank you.

PR #611, opened 2024-06-17
`numpy` version requirement fix

nikitaved:
Fix for numpy 2.0 version.


t-vi:
The versioning is fine (I did add that when numpy 2 first acted up), we need to fix the GPU docker recipe to fix this.

PR #613, opened 2024-06-17
Allowing static constraint in torch/__init__.py

jjsjann123:
## What does this PR do?

Fixes issues in comment https://github.com/Lightning-AI/lightning-thunder/issues/463#issuecomment-2161258626.

This PR includes:
1. Adding `prims.sink` with a DONT_DCE tag to explicitly mark dependency of computation trace on inputs. This allows prologue trace to inject static constraint logic on NumberProxy inputs, which would otherwise not show up in computation trace.
2. Adding constraint to device index in `torch_device`.


thunder/executors/pythonex.py
```diff
@@ -365,5 +365,11 @@ def _elementwise_binary_checker(a: NumberLike | TensorProxy, b: NumberLike | Ten
 ex.register_implementation(prims.sub, sub, checker=_elementwise_binary_checker)
 ex.register_implementation(prims.div, div, checker=_elementwise_binary_checker)
 
+def _sink(*args, **kwargs):
+    return
+
+sink = ex.register_operator("sink", like=prims.sink, fn=_sink)
+ex.register_implementation(prims.sink, sink, checker=_always_executable)
```
jjsjann123, line 374: I added a no-op to have pythonex claim the sink prim.

Alternatively we can just remove this node after prologue is constructed. I'm open to suggestions here.

jjsjann123:
Note that the target branch here is `static_constraint`. So let's not merge this PR yet.

jjsjann123:
failure is not related to this PR. I'll just patch the test with an explicit dtype.

thunder/executors/pythonex.py
t-vi: I think it is fine as is. Specifically, I would avoid removing early because with early transforms, it'll be funny...

t-vi (approved):
Thank you @jjsjann123

t-vi (approved):
Thank you @jjsjann123

thunder/core/prims.py
```diff
@@ -261,6 +261,8 @@ class PrimIDs(Enum):
     # Memory access methods
     ITEM = auto()
     COPY_ = auto()
+    #
```
crcrpar, line 264: @jjsjann123 Could we have a brief in-file comment here?

jjsjann123: oops, let me do that.

PR #616, opened 2024-06-18
test: no Nones in saved_tensors_list

nikitaved:
Testing whether indeed we do not have Nones in the `saved_tensors_list` across all our CIs.


t-vi:
So now that the test has run, do we need this?
It would be cool to leave things that are not to be looked at for merging in draft state (but preferably closing them after they have done what they need, too).

nikitaved:
Closing as it proves the point.

PR #617, opened 2024-06-18
fix numpy version for tests

nikitaved:
Could be a temporary solution before we enable full NumPy 2 support.


t-vi (approved):
Thank you, @nikitaved

Issue #582, opened 2024-06-12
NotImplementedError: requires_grad=True is not yet supported within thunder.compile


mpatel31415:
## 🐛 Bug

There is not implemented error when running models:

Llama-3-70B
falcon-180B
longchat-13b-16k
CodeLlama-34b-hf
vicuna-7b-v1.5-16k
Mixtral-8x7B-v0.1

with `thunder_cudnn` and `thunder_inductor_cat_cudnn`

```
An error occurred: NotImplementedError \xe2\x80\x93 requires_grad=True is not yet supported within thunder.compile
[rank0]: NotImplementedError: requires_grad=True is not yet supported within thunder.compile"
```

### To Reproduce

(example for one model)
```
mkdir -p output
docker run --pull=always --gpus all --ipc=host --ulimit memlock=-1 --ulimit stack=67108864  -v $PWD/output:/output -it INTERNAL_IMAGE:pjnl-20240607
```

Run in the container:
```
torchrun --nproc-per-node=8 /opt/pytorch/lightning-thunder/thunder/benchmarks/benchmark_litgpt.py --model_name longchat-13b-16k --compile thunder_cudnn --distributed_mode fsdp --shard_mode zero2 
```
### Expected behavior

The model should run or we should get OOM error. 

### Environment

As in the Docker image

tfogal:
cc @kshitij12345 this sounds related to your [recent comment](https://github.com/Lightning-AI/lightning-thunder/issues/332#issuecomment-2165045081) on #332.

kshitij12345:
Hi,

I am unable to repro this with the mentioned `pjnl-20240607` and the latest `pjnl-20240613`

In the container, I ran the command as mentioned above and it failed with OOM (on both images)
```
torchrun --nproc-per-node=8 /opt/pytorch/lightning-thunder/thunder/benchmarks/benchmark_litgpt.py --model_name longchat-13b-16k --compile thunder_cudnn --distributed_mode fsdp --shard_mode zero2
```

I also tried to run the same model with fewer layers with the following command and it worked fine (on both images)
```
torchrun --nproc-per-node=8 /opt/pytorch/lightning-thunder/thunder/benchmarks/benchmark_litgpt.py --model_name longchat-13b-16k --compile thunder_cudnn --distributed_mode fsdp --shard_mode zero2 --n_layer=15
```

Output of second command:
```
iter 40: loss 4.6250, iter time: 1200.55ms, t: 16384
[...]

PR #620, opened 2024-06-18
Update error messages.

tfogal:
As Wojciech points out in #582, the error message here was never updated after the switch of entry points.

Reported-by: Wojciech Prazuch


t-vi (approved):
Thank you @tfogal and @wprazuch

PR #622, opened 2024-06-18
Handles non-contiguous input strides

### LINKED ISSUE #583: CUDA error: CUDA_ERROR_ILLEGAL_ADDRESS failed when training falcon-7b
## 🐛 Bug

There is an error when training falcon-7b model with thunder_cudnn.

### To Reproduce
Start a docker container:
```
mkdir -p output
docker run --pull=always --gpus all --ipc=host --ulimit memlock=-1 --ulimit stack=67108864  -v $PWD/output:/output -it INTERNAL_IMAGE:pjnl-20240607
```
Run inside container:

```
python /opt/pytorch/lightning-thunder/thunder/benchmarks/benchmark_litgpt.py --model_name falcon-7b --compile thunder_cudnn --micro_batch_size 1
```

### Expected behavior

We should be able to run the training.

### Environment

As in the docker image, tested on H100.

### Additional context

The same issue is visible for distributed training with DDP and FSDP (zero2).

vedaanta:
<details>

</details>

## What does this PR do?
Earlier non-contiguous input strides were being overwritten during actual execution. With this change, correct strides from torch tensor will be passed to make cudnn graphs.

The main change in this PR is both fwd and bwd cudnn graph makers take in explicit input strides.
- If in execution phase, directly use torch.Tensor.stride()
- If in checker phase, assume row major strides and check cudnn support.

`CudnnTensorAttributes`  has been removed from cudnnex. It is a vestige from earlier caching method, which can be seen in the stale cudnn_layernorm_ex.py.

Fixes #583 .

## PR review

## Did you have fun?


thunder/executors/cudnnex.py
```diff
@@ -124,23 +116,26 @@ def __setitem__(self, key, value):
 _cudnnex_cache = CudnnexLRUCache(maxlen=1024)
 
 
-def _make_cudnn_sdpa_forward_graph(query, key, value, attn_mask, dropout_p, is_causal):
+def _make_cudnn_sdpa_forward_graph(
+    query, key, value, attn_mask, dropout_p, is_causal, query_stride, key_stride, value_stride
```
vedaanta, line 120: The main change here is that both fwd and bwd graph take in explicit input strides.

t-vi (approved):
Thank you @vedaanta 
Seems good, do we have a small enough repro to test this with by chance?

tfogal (approved):
> Seems good, do we have a small enough repro to test this with by chance?

There's a test there now! Thanks, Vedaanta.

vedaanta:
- [ ] Wait for cudnn v1.5.2. It will fix the pipeline failures.

tfogal:
> * [ ]  Wait for cudnn v1.5.2. It will fix the pipeline failures.

I assume you mean cudnn-fe?
Could we conditionalize the code (or just the test?) on that version? I worry about always requiring specific versions of packages, as I've seen other projects become very difficult to install due to long dependency chains with such requirements.

vedaanta:
- [ ] Wait for confirmation that these changes do solve #583

vedaanta:
> Is it reasonable to conditionalize the code (or just the test?) on 1.5.2? I worry about always requiring specific versions of packages, as I've seen other projects become very difficult to install due to long dependency chains with such requirements.
@tfogal 
I bumped up the minimum required cudnn-fe version to 1.5.2. cudnn-fe is meant to be forward and backward compatible with cudnn backend. Plus, cudnn-fe itself is backward compatible.
So in case users have say cudnn-fe 1.3, they can easily move to 1.5.2 and it will work fine. (albeit some newer APIs will not work as expected.)

t-vi:
Thank you all!

PR #624, opened 2024-06-19
[inplace][distributed] support `torch.distributed` common ops

crcrpar:
This implements in-place all_reduce, all_gather_into_tensor, and reduce_scatter_tensor.

The implementations call out-of-place ones and `prims.copy_` output into the user prepared buffer.

When `async_op=True`, a script would have an idiom like
```python
handle = dist.all_gahter_into_tensor(tensor_out, tensor_in, async_op=True)
...
handle.wait()
```
but thunder ignores `handle.wait()` since thunder has multiple passes/transforms that move `dist_prims.wait(future)` around.
So under the hood, if `async_op=True`, we call `dist_prims.wait` before `prims.copy_`.
This means that `handle` is a `TensorProxy` object in a trace, and this PR adds `TensorProxy.wait` as a no-op to avoid NotImplementedError from `handle.wait`.

### DIFF
--- a/thunder/tests/distributed/test_ddp.py
+++ b/thunder/tests/distributed/test_ddp.py
@@ -262,17 +262,8 @@ class CompileDDPTest(DataParallelTestCase):
 
         for async_op in (True, False):
             expected = foo(a, b, process_group, async_op, dim)
-
-            if not (async_op and inplace):
-                actual = cfoo(a, b, process_group, async_op, dim)
-
-                self.assertEqual(actual, expected)
-            else:
-                with self.assertRaisesRegex(
-                    NotImplementedError,
-                    re.escape("`torch.distributed.all_gather_into_tensor` with async_op=True is not supported"),
-                ):
-                    cfoo(a, b, process_group, async_op, dim)
+            actual = cfoo(a, b, process_group, async_op, dim)
+            self.assertEqual(actual, expected)
 
     @common_utils.parametrize("executor", tuple(executors_map.keys()))
     def test_broadcast(self, executor):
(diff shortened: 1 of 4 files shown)


thunder/torch/__init__.py
```diff
@@ -5064,6 +5066,28 @@ def all_gather(
...
+        async_op: bool = False,
+    ) -> TensorLike:
+        utils.check(
+            not async_op,
+            lambda: f"`torch.distributed.all_gather_into_tensor` with {async_op=} is not supported",
+            NotImplementedError,
+        )
+        group = group if group is not None else torch.distributed.new_group()
+        out = dist_prims.all_gather(input_tensor, group, async_op, dim=None, output_tensor=output_tensor)
+        return prims.copy_(out, output_tensor)
```
IvanYashchuk, line 5089: What do you think about also using `wait` in this decomposition? Is this additional information in the trace useful?
```suggestion
        async_op = True
        out = dist_prims.all_gather(input_tensor, group, async_op, dim=None, output_tensor=output_tensor)
        out = dist_prims.wait(out)
        return prims.copy_(out, output_tensor)
```

thunder/tests/distributed/test_ddp.py
```diff
@@ -189,16 +190,23 @@ def lc_foo(
         process_group = c10d.new_group()
 
         # NOTE Preprocessing is disabled because we call thunder.torch operations directly
```
t-vi, line 192: I think this NOTE could go, too, right?

t-vi (approved):

PR #626, opened 2024-06-20
recursive error fix

### LINKED ISSUE #461: Recursion error in transformer module with NeMo Stable Diffusion
### 🐛 Bug

NeMo's Stable Diffusion uses [CLIPTextModel](https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/multimodal/modules/stable_diffusion/encoders/modules.py#L455) from HuggingFace transformers. Using thunder.jit with the CLIPTextModel is causing a RecursionError.

### To Reproduce

1. Add the following lines to transformers/models/clip/modeling_clip.py [here](https://github.com/huggingface/transformers/blob/v4.41.2/src/transformers/models/clip/modeling_clip.py#L668) in the location where transformers is installed in your container
```python
        ## thunder.jit
        self.embeddings = thunder.jit(self.embeddings)
        self.encoder = thunder.jit(self.encoder)
        self.final_layer_norm = thunder.jit(self.final_layer_norm)
``` 
2. Run NeMo Stable Diffusion with the command below:
```python
python examples/multimodal/text_to_image/stable_diffusion/sd_train.py trainer.precision=16 trainer.num_nodes=1 trainer.devices=1 ++exp_manager.max_time_per_run=00:00:03:00 trainer.max_steps=20 model.micro_batch_size=1 model.global_batch_size=1 model.data.synthetic_data=True exp_manager.exp_dir=/workspace/TestData/multimodal/stable_diffusion_train model.inductor=False model.cond_stage_config._target_=nemo.collections.multimodal.modules.stable_diffusion.encoders.modules.FrozenCLIPEmbedder ++model.cond_stage_config.version=openai/clip-vit-large-patch14 ++model.cond_stage_config.max_length=77 ~model.cond_stage_config.restore_from_path ~model.cond_stage_config.freeze ~model.cond_stage_config.layer model.unet_config.from_pretrained=null model.first_stage_config.from_pretrained=null model.unet_config.use_flash_attention=False model.unet_config.attention_resolutions=\[1\] model.unet_config.channel_mult=\[1\]
```

Partial stack trace below:
```python
  File "/usr/local/lib/python3.10/dist-packages/torch/nn/modules/module.py", line 1541, in _call_impl
    return forward_call(*args, **kwargs)
  File "/workspace/software/lightning-thunder/thunder/core/module.py", line 49, in forward
    res = self._forward_fn(*args, **kwargs)
  File "/workspace/software/lightning-thunder/thunder/__init__.py", line 617, in fn_
  File "/workspace/software/lightning-thunder/thunder/__init__.py", line 202, in cache_info_wrapper
    res = fn(*args, **kwargs)
  File "/workspace/software/lightning-thunder/thunder/__init__.py", line 540, in get_computation_and_inputs
    autocast(computation_trc.python_callable(), dtype=autocast_thunder_dtype), *inps
  File "/workspace/software/lightning-thunder/thunder/core/trace.py", line 437, in python_callable
    python_str = self.python(**kwargs)
  File "/workspace/software/lightning-thunder/thunder/core/trace.py", line 318, in python
    import_ctx, call_ctx, object_ctx = self._gather_ctxs()
  File "/workspace/software/lightning-thunder/thunder/core/trace.py", line 281, in _gather_ctxs
    bsym_import_ctx, bsym_call_ctx, bsym_object_ctx = bsym.gather_ctxs()
  File "/workspace/software/lightning-thunder/thunder/core/symbol.py", line 580, in gather_ctxs
    return self.import_ctx(), self._get_call_ctx(), self.object_ctx()
  File "/workspace/software/lightning-thunder/thunder/core/symbol.py", line 520, in import_ctx
    self._out_printables, self._arg_printables, self._kwarg_printables  # type: ignore
  File "/workspace/software/lightning-thunder/thunder/core/symbol.py", line 472, in _out_printables
    return codeutils.to_printable(trace, self.output, import_ctx=self._import_ctx, object_ctx=self._object_ctx)
  File "/workspace/software/lightning-thunder/thunder/core/codeutils.py", line 128, in to_printable
    printables.append(to_printable(trace, f, import_ctx=import_ctx, object_ctx=object_ctx))
  File "/workspace/software/lightning-thunder/thunder/core/codeutils.py", line 128, in to_printable
    printables.append(to_printable(trace, f, import_ctx=import_ctx, object_ctx=object_ctx))
  File "/workspace/software/lightning-thunder/thunder/core/codeutils.py", line 128, in to_printable
    printables.append(to_printable(trace, f, import_ctx=import_ctx, object_ctx=object_ctx))
  [Previous line repeated 2899 more times]
  File "/workspace/software/lightning-thunder/thunder/core/codeutils.py", line 123, in to_printable
    if is_collection(x):
  File "/workspace/software/lightning-thunder/thunder/core/baseutils.py", line 153, in is_collection
    return isinstance(x, collections.abc.Collection) and not isinstance(x, (str, torch.Tensor, np.ndarray))
  File "/usr/lib/python3.10/abc.py", line 117, in __instancecheck__
    def __instancecheck__(cls, instance):
  File "/root/.vscode-server/extensions/ms-python.debugpy-2024.6.0-linux-x64/bundled/libs/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_trace_dispatch_regular.py", line 469, in __call__
    return None if event == 'call' else NO_FTRACE
RecursionError: maximum recursion depth exceeded in comparison
```
CC: @tfogal 

cc @apaz-cli @tfogal

k223kim:
<details>

</details>

## What does this PR do?

Fixes #461.

## PR review

## Did you have fun?


k223kim:
The issue is that, HF's ModelOutput inherits `dict` therefore, when calling `is_collection(x)`, it is always `True` resulting in an infinite loop within `to_printable` and `prettyprint`. Currently, I have added such condition to filter out those `x`s to avoid recursive error.
```python
        if type(x) not in (
            dict,
            list,
            str,
            int,
            bool,
            tuple,
            torch.dtype,
            float,
            dtypes.floating,
            devices.Device,
            torch.memory_format,
        ):
            return x
```
I am uncertain regarding two points:
-  if this is the best way to filter `x`s that cause this recursive error
- if `x` is a type that cause this recursive error, what should we do with it?

thunder/core/codeutils.py
```diff
@@ -121,8 +121,21 @@ def to_printable(
...
+            str,
+            int,
+            bool,
+            tuple,
+            torch.dtype,
+            float,
+            dtypes.floating,
+            devices.Device,
+            torch.memory_format,
+        ):
```
riccardofelluga, line 136: What about using [`isinstance(x, ...)`](https://docs.python.org/3.10/library/functions.html#isinstance) instead? With that you could check if the type is or inherits from a collection from [`collections.abc`](https://docs.python.org/3.10/library/collections.abc.html#module-collections.abc) 👀

thunder/core/codeutils.py
```diff
@@ -195,6 +208,20 @@ def prettyprint(
     if isinstance(x, ContextObject):
         return m(x.name)
     if is_collection(x):
+        if type(x) not in (
+            dict,
```
riccardofelluga, line 212: Same as the other check

riccardofelluga (dismissed):

thunder/core/codeutils.py
k223kim: Do you mean something along the lines of `isinstance(x, collections.abc.Container)`? If that is the case, from my understanding, this will be `True` for HF ModelOutput (`class 'transformers.modeling_outputs.BaseModelOutput'`) which is not what we want. Let me know if I misunderstood anything! :)

k223kim:
I have updated so that we check the type in `tree_flatten`. Not sure if there is a way to handle the `if` condition with only `isinstance`. (cc. @t-vi , @riccardofelluga)

t-vi (commented):
I think this is a great patch, thank you @k223kim .
However, before merging, I would want to get @kshitij12345 's input on how this interacts with #632 . Supporting dataclasses is high prio to us, so we would not want to block #632 if we decide that staying with the solution there is the way we want to go now. @kshitij12345 WDYT about not flattening the world here?

thunder/core/codeutils.py
t-vi: As @k223kim explains, we tried that, but so HF ModelOutput inherits OrderedDict (and thus dict), but cannot be handled.

riccardofelluga: Oh I see, why did we choose to go for a whitelist logic instead of blacklisting the `OrderedDict`? 
i.e. : 
``` python
if isinstance(args, OrderedDict):
    raise TypeError(f"tree_flatten of type {type(args)} is not supported.")
```

t-vi: So the idea is to reliably avoid the recursion error and present a reasonable message instead. tree_flatten has a positive list of things it works with, so that's what we did here. My only hesitation is the interaction with @kshitij12345 's proposed patch for #632, which I want to figure out today-ish.

thunder/core/pytree.py
```diff
@@ -13,7 +16,31 @@
...
+        dtypes.floating,
+        dtypes.bool_,
+        devices.Device,
+        torch.memory_format,
+        type(None),
+        slice,
+        complex,
+        type,
+        type(Ellipsis),
+    } and not isinstance(args, (ProxyInterface)):
```
kshitij12345, line 39: I think this should include `torch.Size` (which we register above for flattening) - otherwise it will start erroring.

kshitij12345:
>  Supporting dataclasses is high prio to us, so we would not want to block https://github.com/Lightning-AI/lightning-thunder/pull/632 if we decide that staying with the solution there is the way we want to go now. @kshitij12345 WDYT about not flattening the world here?

I think this fix is good and should not block #632. Thanks @k223kim @t-vi

thunder/core/pytree.py
k223kim: Thanks for catching this! Let me do that :)

t-vi (approved):
Thank you @k223kim  @kshitij12345 @riccardofelluga

lantiga (approved):

PR #630, opened 2024-06-20
Disallow in-place to view tensors

crcrpar:
## What does this PR do?

Following #584, this PR bans in-place to view tensors. There are some cases that we would be able to support but initially it'd make sense to be conservative.


t-vi (commented):
Thank you @crcrpar , very prudent.

Issue #623, opened 2024-06-19
support returning dataclasses with tensors


t-vi:
This needs to work for transformers (and thus is high-impact / high prio):

```
from dataclasses import dataclass
import torch
import thunder

@dataclass
class MyContainer:
    res: torch.Tensor
    num: int

def fn(x):
    return MyContainer(x, 1)

jfn = thunder.jit(fn)

x = torch.randn(5, 5)

print(jfn(x))
```

(somewhat related is #461 , but that is to not have infinite recursion, here we want the feature)

cc @apaz-cli

Issue #628, opened 2024-06-20
Make epilogue bits more uniform and useful


t-vi:
For making transformations easier and allowing more general return types (e.g. #623) it will be useful to enhance the epilogue uses: 
- Currently, the prologue return format depends on the presence of an epilogue (it returns either "`inputs_to_compute`" or"`(inputs_to_compute, inputs_to_epilogue)`". This leads to early transformation authors needing to consider both cases. If we move to always return also `inputs_to_epiloge` (an empty tuple if there is no epilogue), this will be more uniform. 
- Currently, the epilogue does not process the return values. For #623 we might consider constructing the dataclass instance in the epilogue, it would be good to pass the outputs of the compute trace to the epilogue and return the epilogue's return from the function (if there is an epilogue, otherwise, consider it to be an identity function).

t-vi:
fixed with #1518 and #1541

PR #632, opened 2024-06-20
jit: support dataclass as an output for jitted function

kshitij12345:
Fixes: #623 

Example Snippet:
```python
import torch
import thunder

@dataclass
class MyContainer:
    res: torch.Tensor
    num: int


def fn(x):
    return MyContainer(x, 1)


jfn = thunder.jit(fn)

x = torch.ones(3)

print(jfn(x))
print(jfn(x + 1))
print(thunder.last_traces(jfn)[-1])
```

Output
```python
MyContainer(res=tensor([1., 1., 1.]), num=1)
MyContainer(res=tensor([2., 2., 2.]), num=1)
import torch

@torch.no_grad()
@no_autocast
def computation(x):
  # x: "cpu f32[3]"
  return __main___MyContainer(res=x, num=1)
```

To get this working, we require
1. Support for representing the dataclass in Trace.
2. Support flattening the dataclass so that we can see if it is storing any proxy - this is required when we are inspecting the flat args, kwargs and output of a BoundSymbol.

Main changes - 

1. codeutils.py - Changes in this file related to how the dataclass instace will look and be represented in the trace. In this PR, the name of the class is `class.__module__ + class.__qualname__` to avoid collision if there are two classes with same name from different module (can be seen in above snippet).
3. pytree.py - We provide `tree_flatten_with_dataclass` which also flattens the dataclass so that we can see all the proxies that could be contained in a dataclass instance.
4. symbol.py - We update functions in `symbol.py` to use `tree_flatten_with_dataclass` instead of `tree_flatten` which keeps the dataclass instances opaque.


thunder/tests/test_core.py
```diff
@@ -2750,3 +2751,47 @@ def foo2(x):
...
+
+    assert dataclasses.is_dataclass(actual)
+    assert isinstance(actual, TestDataclass)
+    torch.testing.assert_close(actual.t, expected.t)
+    torch.testing.assert_close(actual.s, expected.s)
+    torch.testing.assert_close(actual.i, expected.i)
+    torch.testing.assert_close(actual.f, expected.f)
+
+
+def test_dataclass_input():
```
kshitij12345, line 2781: Not directly related to the PR. This already works.

t-vi:
Hi @kshitij12345 . Thank you for working on this.
There are two aspects:
- I'm not particularly enchanted at relying on OpTree for this and wonder if we should have a mechanism in the Python interpreter.
- I tentatively think we should have the creation of the dataclass to the epilogue rather than the compute trace (see also #628 )

thunder/tests/test_core.py
t-vi: Nice test!

t-vi:
@kshitij12345 Just to add: I think if we don't have a better solution this week and this enables running HF/NeMo models, I'd probably be looking at taking it as a stop-gap thing.

kshitij12345:
Sounds good, I am looking at this a bit more. Also, if required - I will sync offline with you once. Thanks for having a look.

t-vi (approved):
Seems like a great solution for now. Thank you @kshitij12345

thunder/core/pytree.py
```diff
@@ -1,52 +1,101 @@
...
+
+def _maybe_register_dataclass(t):
+    if dataclasses.is_dataclass(t) and t.__class__ not in _registered_dataclasses:
+        return True
+    return False
+
+
+# `tree_flatten_with_dataclass` iterates over the tree and registers functions to flatten dataclass objects present in the `tree`.
+# This is to facilitate peeking into the dataclass object to correctly get proxies when inspecting the BoundSymbols in the trace.
+def tree_flatten_with_dataclass(tree):
```
IvanYashchuk, line 86: Is there a situation where we'd prefer to use "tree_flatten_without_dataclasses"? Should "tree_flatten_with_dataclass" be just "tree_flatten"?

kshitij12345: IIRC, the `tree_flatten` usage in `codeutils.py` (used for printing/creating traces) requires `tree_flatten_without_dataclasses` as we want to handle a dataclass atomically (vs flattening it to a dict).

Ref: https://github.com/Lightning-AI/lightning-thunder/blob/a3e432f7174019b2eda85865890d5f7342a993c2/thunder/core/codeutils.py#L137-L143

kshitij12345: I think it would probably be better with dataclass construction symbol (trace would look cleaner and probably we won't need `tree_flatten_with_dataclass` as proxies will be arguments to this symbol). Will check this approach (and probably sync offline for more details). Thanks!

PR #633, opened 2024-06-21
Fix in-place to views condition

crcrpar:
## What does this PR do?

Special case batch norm and avg/max-pool in the checker as they would not return a view as their output.

Add a test that make sure that thunder can execute `torchvision.models.resnet18`

### DIFF
--- a/thunder/tests/test_inplace_functionalization.py
+++ b/thunder/tests/test_inplace_functionalization.py
@@ -149,6 +149,13 @@ def test_invalid_cases():
         jitted(a)
 
 
+# TODO(crcrpar): Investigate the numerical accuracy when `train=True` and dtype is fp32.
+# with RTX6000 Ada and CUDA 12.3, I see somewhat huge error:
+# E   AssertionError: Tensor-likes are not close!
+# E
+# E   Mismatched elements: 913 / 1000 (91.3%)
+# E   Greatest absolute difference: 0.000273287296295166 at index (0, 50) (up to 1e-05 allowed)
+# E   Greatest relative difference: 0.4177769422531128 at index (0, 727) (up to 1.3e-06 allowed)
 @requiresCUDA
 @pytest.mark.parametrize("train", (False, True))
 def test_parse_resnet18(train: bool):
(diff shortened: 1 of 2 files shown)


thunder/tests/test_inplace_functionalization.py
```diff
@@ -147,3 +148,22 @@ def f_with_contiguous(a: torch.Tensor) -> torch.Tensor:
...
+
+
+@requiresCUDA
+def test_parse_resnet18():
+    import thunder
+
+    torchvision = pytest.importorskip("torchvision")
+
+    device = torch.device("cuda")
+    dtype = torch.bfloat16 if device_supports_bf16(device) else torch.float16
```
t-vi, line 160:
```suggestion
    dtype = torch.bfloat32
```

thunder/tests/test_inplace_functionalization.py
```diff
@@ -147,3 +148,22 @@ def f_with_contiguous(a: torch.Tensor) -> torch.Tensor:
...
+    model: nn.Module = torchvision.models.resnet18().to(device=device, dtype=dtype)
+    ref_model: nn.Module = torchvision.models.resnet18().to(device=device, dtype=dtype)
+    ref_model.load_state_dict(model.state_dict())
+
+    jitted = thunder.jit(model)
+    x = torch.randn((1, 3, 224, 224), device=device, dtype=dtype)
+    # FIXME(crcrpar): Why...?
+    with pytest.raises(AssertionError, match="Tensor-likes are not close!"):
```
t-vi, line 168:
```suggestion
```

thunder/tests/test_inplace_functionalization.py
```diff
@@ -147,3 +148,22 @@ def f_with_contiguous(a: torch.Tensor) -> torch.Tensor:
...
+        torch.testing.assert_close(jitted(x), ref_model(x))
```
t-vi, line 169:
```suggestion
      torch.testing.assert_close(jitted(x), ref_model(x), atol=1e-3, rtol=1e-3)
```

thunder/tests/test_inplace_functionalization.py
```diff
@@ -147,3 +148,22 @@ def f_with_contiguous(a: torch.Tensor) -> torch.Tensor:
...
+
+    device = torch.device("cuda")
+    dtype = torch.bfloat16 if device_supports_bf16(device) else torch.float16
+    model: nn.Module = torchvision.models.resnet18().to(device=device, dtype=dtype)
+    ref_model: nn.Module = torchvision.models.resnet18().to(device=device, dtype=dtype)
+    ref_model.load_state_dict(model.state_dict())
+
+    jitted = thunder.jit(model)
+    x = torch.randn((1, 3, 224, 224), device=device, dtype=dtype)
+    # FIXME(crcrpar): Why...?
```
t-vi, line 167:
```suggestion
    # FIXME(crcrpar): Investigate numerical accuracy.
```

t-vi (approved):
Thank you @crcrpar 

Supergreat milestone running stock ResNet18! :rocket: 
I think it would be better to test fp32 + generous bounds, but in addition we should file an issue about the accuracy. WDYT?

t-vi:
I looked at the numerical accuracy a bit more and does not seem to happen with eval mode, so maybe something with the training mode batch norm.

PR #637, opened 2024-06-23
pass results through epilogue if present

### LINKED ISSUE #623: support returning dataclasses with tensors

t-vi:
Here we pass results through the epilogue if we have one.
The goal is to have a fix #623 / enhance #632 to put the construction of the dataclass into the epilogue rather than into the compute trace. This way the compute trace could always return a (flat tuple of) plain tensors / numbers.

### DIFF
--- a/thunder/__init__.py
+++ b/thunder/__init__.py
@@ -560,6 +560,7 @@ def jit(
             cs.last_prologue_execution_stop = time.time_ns()
 
             cs.last_traces = computation_traces
+            cs.last_epilogue_traces = epilogue_traces
             backward_traces = []
             cs.last_backward_traces = backward_traces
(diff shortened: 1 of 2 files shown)


t-vi:
I still need to fix the interaction with transformations adding outputs (but that's good, so we can have a principled way for that, too).

PR #638, opened 2024-06-24
Detailed `__repr__`

### LINKED ISSUE #510: Thunder object's `__repr__` should indicate what object they are (TensorProxy and others)
An older issue in the old repo was for TensorProxy, but it also affects other types. Here is the old discussion:

Currently `TensorProxy __repr__` returns just the name.
This is bad because it is obscuring error messages which include `TensorProxy` mentions.

Note that we should use `<TensorProxy {self.name}>` or so. Using `[...]` should only be for lists (please!), as it will look like lists.
Python itself uses `<{type} object ...>` as repr, I'm not attached to "object", but it generally seems a good idea to follow the example with `<...>`.

@mruberry commented: We probably are implicitly relying on `TensorProxy.__repr__` printing just the name, but we should be able to fix our reliance on that without too much code change

Similar considerations apply to other objects (e.g. dtypes).

cc @apaz-cli

k223kim:
<details>

</details>

## What does this PR do?

Fixes #510.

## PR review

## Did you have fun?

### DIFF
--- a/thunder/core/proxies.py
+++ b/thunder/core/proxies.py
@@ -133,7 +133,7 @@ class Proxy(VariableInterface, ProxyInterface):
         return self.__class__(name=name)
 
     def __repr__(self) -> str:
-        return f"{self.name}"
+        return f"<{type(self).__name__} {self.name}>"
 
     def type_string(self) -> str:
         return "Any"
@@ -1111,11 +1111,17 @@ class FutureTensorProxy(Proxy, TensorProxyInterface):
 
     @property
     def device(self):
+        # conver to torch device
+        from thunder.torch import _thunder_to_torch_dtype_map
+
         return self._device
 
     @property
     def dtype(self):
-        return self._dtype
+        from thunder.torch import _thunder_to_torch_dtype_map
+
+        print("hello")
+        return _thunder_to_torch_dtype_map[self._dtype]
 
     @property
     def true_dtype(self):
--- a/thunder/tests/test_core.py
+++ b/thunder/tests/test_core.py
@@ -1242,7 +1242,7 @@ def test_argument_of_none(executor, device, dtype):
     consumers = thunder.core.utils.consumers(trace)
     region_bsyms = trace.bound_symbols[:3]
     region = Region(producers, consumers, region_bsyms)
-    assert len(region.inputs) == 0 and sorted(str(v) for v in region.outputs) == ["t0"]
+    assert len(region.inputs) == 0 and sorted(str(v) for v in region.outputs) == ["<TensorProxy t0>"]
 
 
 # This test ensures that calls to torch functions are recorded in the trace


thunder/core/proxies.py
```diff
@@ -134,7 +134,7 @@ def replace_name(self, name: str | None = None):
         return self.__class__(name=name)
 
     def __repr__(self) -> str:
-        return f"{self.name}"
+        return f"<{type(self).__name__} {self.name}>"
```
IvanYashchuk, line 137: We can probably drop the <> brackets and include other attributes besides the name. For TensorProxies minimal valuable information is dtype and shape. Python's documentation suggests making `eval(repr(object))` work when possible, from [https://docs.python.org/3/reference/datamodel.html#object.__repr__](https://docs.python.org/3/reference/datamodel.html#object.__repr__):
> If at all possible, this should look like a valid Python expression that could be used to recreate an object with the same value (given an appropriate environment). If this is not possible, a string of the form <...some useful description...> should be returned.

k223kim: Thanks for the suggestion and information! Would something along the lines of `f"TensorProxy name=x, dtype=thunder.dtypes.float32, shape=(3,3)"` be sufficient? I think this will require a `prettyprint` update..

thunder/core/codeutils.py
```diff
@@ -243,12 +243,14 @@ def prettyprint(
         # NOTE: The `class` packagename1_MyContainer will present in `import_ctx` and passed to the compiled function.
         # This is taken care of by function `to_printable`.
         name = _generate_dataclass_class_name(x)
-        instance_repr = str(x)
-        parens_idx = instance_repr.find("(")
-        call_repr = instance_repr[
-            parens_idx:
-        ]  # only keep the construction part of the repr (as we will use our generated name)
-        return m(f"{name + call_repr}")
+        call_repr_str = ""
```
t-vi, line 246: I think it would be a tad neater to build a list and then join.
a loop of `+=` on strings isn't as bad as it sometimes is made out to be, but usually using a list to accumulate is prefered.

thunder/core/proxies.py
t-vi: I suggested the brackets, thanks Ivan for bringing this up, sounds great.

t-vi: We can also do this in a followup instead.

IvanYashchuk: Sure, it can happen in a followup. 
> Would something along the lines of f"TensorProxy name=x, dtype=thunder.dtypes.float32, shape=(3,3)" be sufficient?

The parentheses are missing, without parentheses it's not a valid Python expression: `TensorProxy(name="x", dtype=thunder.dtypes.float32, shape=(3,3))`.

k223kim: Thanks @IvanYashchuk @t-vi, let me update this by adding brackets, parentheses, dtype, and shape.

thunder/core/codeutils.py
k223kim: Thanks for the feedback. Let me update this!

lantiga (approved):
Looks great! Let's merge and layer future improvements on later PRs

thunder/distributed/tensor_parallel/common.py
```diff
@@ -249,7 +249,7 @@ def transform_traces(
                 assert param_thunder_module is thunder_module_proxy
                 if name not in self.chunked_param_name_to_layer_type:
                     a0, shape, _, *a2pp = bsym.args
-                    bsym.args = (a0, shape, str(a0.device), *a2pp)
+                    bsym.args = (a0, shape, a0.device, *a2pp)
```
t-vi, line 252:
```suggestion
                    bsym.args = (a0, shape, a0.device.device_str(), *a2pp)
```

t-vi:
Thank you @k223kim !

PR #642, opened 2024-06-24
Check number of proxy args before indexing

crcrpar:
## What does this PR do?

when an in-place op is applied to a tensor from a factory method, the producer bound symbol does not take any tensors as their args. Therefore it leads to out of range error.

### DIFF
--- a/thunder/core/transform_common.py
+++ b/thunder/core/transform_common.py
@@ -415,6 +415,9 @@ def check_inplace_to_views(computation_trace: Trace) -> dict[VariableInterface, 
         if in_tensor in trace_args_set:
             continue
         prod_bsym: BoundSymbol = producer_bsyms[in_tensor]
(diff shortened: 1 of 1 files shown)


thunder/core/transform_common.py
```diff
@@ -415,6 +415,9 @@ def has_tag(bsym: BoundSymbol, tag: prims.OpTags) -> bool:
         if in_tensor in trace_args_set:
             continue
         prod_bsym: BoundSymbol = producer_bsyms[in_tensor]
+        if not prod_bsym.flat_proxy_args:
```
t-vi, line 418: would we need to filter for tensors (to exclude numberproxies) when @jjsjann123 's work fully comes to fruition?

t-vi (approved):
Thank you @crcrpar

PR #648, opened 2024-06-25
ThunderModule - fix load_original_state_dict to work with modules with buffers

kshitij12345:
Repro: 
```python
import torch
import thunder

class Model(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.register_parameter("param", torch.nn.Parameter(torch.randn(3)))
        self.register_buffer("buffer", torch.randn(3))

    def forward(self, x):
        return x

m = Model()

thunder_module = thunder.jit(Model())


# Fails with - NameError: name 'model' is not defined
thunder_module.load_original_state_dict(m.state_dict())
```

### DIFF
--- a/thunder/core/module.py
+++ b/thunder/core/module.py
@@ -43,7 +43,7 @@ class ThunderModule(pytorch.nn.Module):
         return self._model.get_buffer(name)
 
     def set_buffer(self, name, value):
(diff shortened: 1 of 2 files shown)


t-vi (approved):
Oops. Thank you @kshitij12345

Issue #635, opened 2024-06-21
Make CudaGraph wrapping a transform (but call it explicitly)


t-vi:
Currently when running

```
def fn(x, y):
    a = torch.relu(x * y)
    return a

jfn = thunder.jit(fn)
x, y = torch.randn(2, 5, 5, device="cuda")
jfn(x, y)

trace = thunder.last_traces(jfn)[-1]
```

I get

```
# Constructed by Delete Last Used (took 0 milliseconds)
import torch
from thunder.executors.torchex import no_autocast

@torch.no_grad()
@no_autocast
def computation(x, y):
  # x: "cuda:0 f32[5, 5]"
  # y: "cuda:0 f32[5, 5]"
  [a] = TorchCompile0(x, y)
    # result = ltorch.mul(x, y)  # result: "cuda:0 f32[5, 5]"
      # result = prims.mul(x, y)  # result: "cuda:0 f32[5, 5]"
    # a = ltorch.relu(result, False)  # a: "cuda:0 f32[5, 5]"
      # t1 = ltorch.gt(result, 0)  # t1: "cuda:0 b8[5, 5]"
        # _ = prims.convert_element_type(0, float)
        # t1 = prims.gt(result, 0.0)  # t1: "cuda:0 b8[5, 5]"
      # a = ltorch.where(t1, result, 0)  # a: "cuda:0 f32[5, 5]"
        # _ = prims.convert_element_type(0, float)
        # a = prims.where(t1, result, 0.0)  # a: "cuda:0 f32[5, 5]"
  del x, y

  # /tmp/ipykernel_982252/871691319.py:2: 	    a = torch.relu(x * y)
  return a
```

Ideally, we could transform the extrace for CudaGraphs instead of wrapping:

```
# Constructed by CUDA graph transform  (took 0 milliseconds)
import torch
from thunder.executors.torchex import no_autocast

@torch.no_grad()
@no_autocast
def computation(x, y):
  # x: "cuda:0 f32[5, 5]"
  # y: "cuda:0 f32[5, 5]"
  [a] = CudaGraph0(x, y)
    # [a] = TorchCompile0(x, y)
      # result = ltorch.mul(x, y)  # result: "cuda:0 f32[5, 5]"
        # result = prims.mul(x, y)  # result: "cuda:0 f32[5, 5]"
      # a = ltorch.relu(result, False)  # a: "cuda:0 f32[5, 5]"
[...]

PR #656, opened 2024-06-26
CUDAGraphs as executor/transform/fusion pass

nikitaved:
As per title. Fixes https://github.com/Lightning-AI/lightning-thunder/issues/635.

Also, it fixes the following subtle bugs:

- `CUDAGraphExecutor` - does not properly update static buffers when the same graph is invoked on inputs with meta-data that allows to fetch cached graphs, but with different storage data. The area of concern - training and the backward pass.
- `horizontal_merge` in the fusion logic - that one, when grouping bound symbols, does not consider precedence between ops horizontally. It is not an issue with nvfusions, but it could cause issues when deciding whether to place something like `del x` after `op(x)` in a custom FusionExecutor. The fix sorts bsyms in each group wrt trace position (which is expected to be toposorted prior to any fusions) and, hence, restores the inter-/intra-bsym groups topological order.

### DIFF
--- a/thunder/executors/cudagraphex.py
+++ b/thunder/executors/cudagraphex.py
@@ -40,7 +40,7 @@ def to_arg_descriptor(*args):
 @lru_cache
 def build_cuda_graph(
     fn: Callable, args_descriptor: ArgsDescriptor, static_args_mask: tuple[bool, ...]
-    ) -> tuple[torch.cuda.CUDAGraph, Sequence[torch.Tensor | Any], Sequence[torch.Tensor | Any]]:
+) -> tuple[torch.cuda.CUDAGraph, Sequence[torch.Tensor | Any], Sequence[torch.Tensor | Any]]:
 
     def get_static_buffer(x):
         if isinstance(x, torch.Tensor):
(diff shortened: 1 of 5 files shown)


thunder/executors/cudagraphex.py
```diff
@@ -0,0 +1,246 @@
...
+            prims.PrimIDs.UNPACK_SEQUENCE,
+            prims.PrimIDs.RETURN,
+            # Data-dependent ops
+            prims.PrimIDs.ITEM,
+        }
+
+        if bsym.sym.id in do_not_fuse_sym_set:
+            return False
+
+        return True
```
t-vi, line 188: Would we need to check for fixed sizes of the proxies in input and output or ist this handled?

t-vi (approved):
Looks great overall, thank you @nikitaved .
I have one question in the comments and then I'd be all for merging it.

IvanYashchuk:
Nice! This change is needed to make my PR https://github.com/Lightning-AI/lightning-thunder/pull/214 work with CUDA Graphs correctly. Because there I try to put torch.autograd.Function.apply into the forward trace but it should be executed outside of the CUDA Graph-captured region.

lantiga (approved):
Looks great @nikitaved!
I'm imagining that in the future we could consider not special-casing use_cudagraphs but keeping it as a transform, but maybe that's overkill and it's certainly it's not for now.

t-vi:
Let's merge and fix anything that needs fixing later.

thunder/executors/cudagraphex.py
nikitaved: I wonder whether the future API is agreed upon? Does it mean that dynamic shape Tensors will contain Integer proxies in their meta-data?

PR #595, opened 2024-06-14
Remove  _create_callable usage

IvanYashchuk:
Cleaning up usage of old function that is now used only for soft-deprecated `thunder.compile`.

PR #668, opened 2024-06-27
retire the previous impl of CUDAGraphExecutor

nikitaved:
As per title. https://github.com/Lightning-AI/lightning-thunder/pull/656 implements a fusing ex instead.


thunder/common.py
```diff
@@ -672,10 +671,6 @@ def _execute_trace(
     # Constructs the Python callable
     c = extrace.python_callable()
 
-    # TODO RC1 Mark this option as experimental
-    if compile_data.use_cudagraphs:
-        c = CUDAGraphExecutor(c, num_constant_args=compile_data.num_constant_args)
-
```
nikitaved, line 678: @IvanYashchuk , from https://github.com/Lightning-AI/lightning-thunder/pull/595 I see that, ultimately, a plan is to remove `_execute_trace` as well?

t-vi (approved):
Thank you @nikitaved

thunder/common.py
IvanYashchuk: Yes, it's used only in `thunder.compile`.

IvanYashchuk (approved):

PR #670, opened 2024-06-27
CUDAGraphExecutor - add tests for the graph caching mechanism

nikitaved:
As per title. To make sure that graphs are being properly cached and fetched without redundant re-builds.


t-vi (approved):
Supergood, thank you @nikitaved

Issue #601, opened 2024-06-14
unhashable type: 'TensorProxy' error on NeVA model


tfogal:
While trying to `thunder.jit` portions of the NeMo's NeVA model, I found this `unhashable type` error. Attached is [a simplified / reduced example that requires NeMo](https://github.com/user-attachments/files/15844230/unhashable-type.py.txt).

The full traceback is:
```python
Traceback (most recent call last):
  File "$HOME/dev/nemo/repro/unhashable-type.py", line 64, in <module>
    x = fwd(input_ids)
  File "$HOME/dev/thunder/thunder/__init__.py", line 659, in fn_
    cache_entry, inps, pro_to_epi = get_computation_and_inputs(*args, **kwargs)
  File "$HOME/dev/thunder/thunder/__init__.py", line 217, in cache_info_wrapper
    res = fn(*args, **kwargs)
  File "$HOME/dev/thunder/thunder/__init__.py", line 496, in get_computation_and_inputs
    jit_results: TraceResults = interpreter(
  File "$HOME/dev/thunder/thunder/__init__.py", line 205, in _general_frontend
    return thunder_general_jit(fn, args, kwargs, sharp_edges=sharp_edges, record_history=record_history)
  File "$HOME/dev/thunder/thunder/core/jit_ext.py", line 1605, in thunder_general_jit
    result = jfn(*args, **kwargs)
  File "$HOME/dev/thunder/thunder/core/interpreter.py", line 6696, in fn_
    raise e
  File "$HOME/dev/thunder/thunder/core/interpreter.py", line 6664, in fn_2
    return fn(*args, **kwargs)
  File "$HOME/dev/thunder/thunder/core/interpreter.py", line 6061, in _impl
    return fn.__func__(fn.__self__, *args, **kwargs)
  File "$HOME/dev/nemo/nemo/collections/multimodal/models/multimodal_llm/neva/neva_model.py", line 168, in forward
    return self.replace_media_embeddings(input_ids, words_embeddings, media)
  File "$HOME/dev/nemo/nemo/collections/multimodal/models/multimodal_llm/neva/neva_model.py", line 216, in replace_media_embeddings
    media_features = self.encode_vision_x(media)  # b T F S(eq) H(idden)
  File "$HOME/dev/nemo/nemo/collections/multimodal/models/multimodal_llm/neva/neva_model.py", line 194, in encode_vision_x
    vision_x = vision_x.to(self.vision_encoder.dtype)
[...]

PR #672, opened 2024-06-27
jit: support nn.Moudle.named_parameters/buffers

kshitij12345:
Fixes: https://github.com/Lightning-AI/lightning-thunder/issues/601

As elaborated in the [comment](https://github.com/Lightning-AI/lightning-thunder/issues/601#issuecomment-2192912403), we add lookaside for `Module.named_{parameters/buffers}` (for now).


thunder/tests/test_jit_general.py
```diff
@@ -991,3 +991,59 @@ def forward(self, x):
...
+@pytest.mark.parametrize("recurse", (True, False), ids=("recurse=True", "recurse=False"))
+@pytest.mark.parametrize(
+    "remove_duplicate",
+    (
+        False,
+        pytest.param(
+            True,
+            marks=pytest.mark.xfail(
+                raises=(AssertionError),
+                reason="We can't differentiate between shared buffers and parameters",
```
kshitij12345, line 1006: Maybe there is a way for this?

t-vi: To my mind, we could record the Ids of tensors we proxy, then we could know?

thunder/core/jit_ext.py
```diff
@@ -829,6 +829,51 @@ def _general_jit_bool_lookaside(wrapped_x: Any) -> bool | INTERPRETER_SIGNALS:
 
 _general_jit_lookaside_map[bool] = _general_jit_bool_lookaside
 
+
+def _torch_module_named_member_lookaside(model, get_member_func, prefix="", recurse=True, remove_duplicate=True):
```
t-vi, line 833: This is not a `_lookaside` but an `_impl`, maybe even with adding a comment that it will be interpreted.

thunder/core/jit_ext.py
```diff
@@ -829,6 +829,51 @@ def _general_jit_bool_lookaside(wrapped_x: Any) -> bool | INTERPRETER_SIGNALS:
...
+    modules = model.named_modules(prefix=prefix, remove_duplicate=remove_duplicate) if recurse else [(prefix, model)]
+    for module_prefix, module in modules:
+        members = get_member_func(module)
+        for k, v in members.items():
+            # We use name of the proxy as id.
+            id_v = v.name
+            if id_v is None or id_v in memo:
+                continue
+            if remove_duplicate:
+                memo.add(id_v)
```
t-vi, line 844: The problem with memo is why my idea was to run `named_parameters` inside the lookaside rather than the the impl and only keep a list of names (but a deduplicated one!). Then, in an `_impl` with `_interpret_call` we can take the list of names use `m.get_parameter(n))` to get the deduped list. Or we could just make set of names in the lookaside and just return those that are in the set here (that is likely quicker).
WDYT?

thunder/core/jit_ext.py
```diff
@@ -829,6 +829,51 @@ def _general_jit_bool_lookaside(wrapped_x: Any) -> bool | INTERPRETER_SIGNALS:
...
+    def get_parameters(model):
+        return getattr(model, "_parameters")
+
+    return _interpret_call(
+        partial(_torch_module_named_member_lookaside, model=model, get_member_func=get_parameters), *args, **kwargs
+    )
+
+
+@general_jit_lookaside(torch.nn.Module.named_buffers)
+def _general_jit_named_buffer_lookaside(obj: Any, *args, **kwargs):
```
t-vi, line 864: Nit:

```suggestion
def _general_jit_named_buffers_lookaside(obj: Any, *args, **kwargs):
```

thunder/core/jit_ext.py
```diff
@@ -829,6 +829,51 @@ def _general_jit_bool_lookaside(wrapped_x: Any) -> bool | INTERPRETER_SIGNALS:
...
+            if id_v is None or id_v in memo:
+                continue
+            if remove_duplicate:
+                memo.add(id_v)
+            name = module_prefix + ("." if module_prefix else "") + k
+            yield name, v
+
+
+@general_jit_lookaside(torch.nn.Module.named_parameters)
+def _general_jit_named_parameter_lookaside(obj: Any, *args, **kwargs):
```
t-vi, line 850: Nit: Can we call the lookaside with the "s" of `named_parameters`, please?

```suggestion
def _general_jit_named_parameters_lookaside(obj: Any, *args, **kwargs):
```

thunder/tests/test_jit_general.py
t-vi: Yes, but we could get the deduplicated list inside the lookaside and then filter by that. :)

t-vi (changes requested):
@kshitij12345 Absolutely fantastic to work on this!
I think we can solve the duplication by first iterating in the lookaside (directly in Python without proxying / interpreter), keeping the names we get there, and then in the _impl filter...

tfogal:
Just wanted to comment that I tried this with NeVA (which is where #601 came from) and the run gets by where we were stopped before! 🎉
We hit another bug before the model actually runs, so this isn't an e2e signal--but a very promising start!

thunder/tests/test_jit_general.py
```diff
@@ -991,3 +991,97 @@ def forward(self, x):
...
+    m = MyModel()
+    expected = dict(m())
+
+    jm = thunder.jit(m)
+    actual = dict(jm())
+
+    torch.testing.assert_close(actual, expected)
+
+
+def test_isinstance_parameter():
```
kshitij12345, line 1042: We needed to support this as `get_parameter` has such a check.

https://github.com/pytorch/pytorch/blob/dabaebd3399a14a082da489b93312fab77f52c39/torch/nn/modules/module.py#L815-L816

t-vi: Super diligent!

t-vi (approved):
Thank you @kshitij12345

PR #679, opened 2024-06-28
Add missing interpolate() parameters

tfogal:
## What does this PR do?

Adds missing arguments for [interpolate](https://pytorch.org/docs/stable/generated/torch.nn.functional.interpolate.html).

## PR review

### DIFF
--- a/thunder/torch/__init__.py
+++ b/thunder/torch/__init__.py
@@ -4543,6 +4543,9 @@ def interpolate(
     size: None | int | Sequence[int] = None,
     scale_factor: float | Sequence[float] | None = None,
     mode: str = "nearest",
+    align_corners: bool = False,
+    recompute_scale_factor: bool | None = None,
+    antialias: bool = False
 ) -> TensorLike:
     utils.check(
         mode == "nearest",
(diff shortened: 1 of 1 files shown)


thunder/torch/__init__.py
```diff
@@ -4552,6 +4555,12 @@ def interpolate(
 
     utils.check(a.ndim >= 3, lambda: f"Expected {a.ndim=} >= 3")
     utils.check(a.numel() > 0, lambda: f"Expected {a.numel=} to be greater than 0")
+    utils.check(align_corners == None, lambda: f"'align_corners' is not yet supported.")
```
t-vi, line 4558: Nit: should this (and the following) be `exception_type=NotImplementedError` and maybe "not yet supported _in Thunder_"?

t-vi (approved):
Looks great. Seems not high risk, but do we want to add a test?

thunder/torch/__init__.py
nikitaved: From the doc: https://pytorch.org/docs/stable/generated/torch.ao.nn.quantized.functional.interpolate.html, `align_corners` has only effect when `mode = 'bilinear'`.

lantiga (approved):

nikitaved:
@tfogal , I assume having support for `linear` mode is something we would like to support eventually?

t-vi:
@nikitaved the immediate thing we want is https://github.com/Lightning-AI/lightning-thunder/issues/680

thunder/torch/__init__.py
tfogal: didn't know that was an option, neat :-)
yes, it should be. will fix in a follow-up.

tfogal:
> Looks great. Seems not high risk, but do we want to add a test?

No new functionality was added, so there's nothing _new_ to test. We could add negative tests for the errors that come up; maybe that's what you meant?

It seems the existing test machinery already populates default arguments, as the initial patch for this did not set defaults properly and this caused the existing tests to break.

tfogal:
> @tfogal , I assume having support for `linear` mode is something we would like to have eventually?

eventually, of course! As TomV points out, #680 was the priority here and this seems to be enough 🤞 .

thunder/torch/__init__.py
tfogal: #721

Issue #658, opened 2024-06-26
TransformerEngine's FP8 LayerNorm support


tfogal:
## 🚀 Feature

120 Mixology runs are failing due to:

```python
raise ValueError("LayerNorm is currently not supported by Thunder!")
```

### Additional context

```
set -e
export PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True
export NCCL_BLOCKING_WAIT=1
export TORCH_NCCL_BLOCKING_WAIT=1
export NCCL_ASYNC_ERROR_HANDLING=1
export TORCH_NCCL_ASYNC_ERROR_HANDLING=1
export NCCL_TIMEOUT=600
python -m mixology_logs.execution.main \
--nsys.enable True \
--nsys.output_path /jet/assets/recipe/-gemma-7b_-lit-gpt-pjnl_-perf-train_--eos-dgx-h100-_-bfloat16_-1_-8_--1_-train_-false_--_-thunder-cudnn_-ddp_-fp8-delayed-te_-none_-s_-lit-gpt/nsys_report \
--nsys.new_kwargs '{""--nsys_enabled"": ""True"", ""--output_dir"": ""/tmp""}' \
'{""--micro_batch_size"": ""exp_range(0, 10)""}' \
""python thunder/benchmarks/benchmark_litgpt.py \
    --max_iters 20 \
    --warmup_iters 5 \
    --output_dir /jet/logs/recipe/-gemma-7b_-lit-gpt-pjnl_-perf-train_--eos-dgx-h100-_-bfloat16_-1_-8_--1_-train_-false_--_-thunder-cudnn_-ddp_-fp8-delayed-te_-none_-s_-lit-gpt \
    --model_name Gemma-7b \
    --distributed_mode ddp \
    --shard_mode None \
    --compile thunder_cudnn \
    --checkpoint_activations False \
    --low_precision_mode fp8-delayed-te""
```

t-vi:
But what's the traceback? I don't see that error being raised anywhere in the thunder repo...

tfogal:
> I don't see that error being raised anywhere in the thunder repo...

Yes, ditto. We think this is coming from the mixology scripts.

There's an internal thread with @wprazuch; stay tuned, we'll report back here.
I can't seem to assign to @wprazuch (?) so assigning to me temporarily instead.

wprazuch:
Thanks @tfogal for notifying!
[...]

PR #689, opened 2024-07-01
Implement TransformerEngine benchmarking for lit-gpt

wprazuch:
As mentioned in the [conversation](https://github.com/Lightning-AI/lightning-thunder/issues/658), the aim of this PR is to enable benchmarking `lit-gpt` through `TransformerEngine`. 

<details>

</details>

This PR adds a new CLI argument `low_precision_mode`, which can be set and the model is executed through respective low_precision_mode accordingly. Currently, only `TransformerEngine` delayed is supported.

Fixes https://github.com/Lightning-AI/lightning-thunder/issues/658 or at least makes it possible to benchmark it.

## PR review

## Did you have fun?

### DIFF
--- a/thunder/benchmarks/benchmark_litgpt.py
+++ b/thunder/benchmarks/benchmark_litgpt.py
@@ -234,7 +234,7 @@ class Benchmark_litGPT:
 
         if self.use_te_fp8_autocast:
             te_precision = TransformerEnginePrecision(weights_dtype=torch.bfloat16, replace_layers=True)
-            self.model =  te_precision.convert_module(self.model)
+            self.model = te_precision.convert_module(self.model)
 
         # Setup the distributed algorithm choices
         self.model = self.setup_distributed(self.model)
--- a/thunder/benchmarks/benchmark_litgpt.py
+++ b/thunder/benchmarks/benchmark_litgpt.py
@@ -185,12 +185,11 @@ class Benchmark_litGPT:
                     f"[WARNING] Bucketing mode is set to {self.bucketing_mode}. --fsdp_bucket_params will be ignored."
                 )
 
-        if not transformer_engine_available and is_transformer_engine(low_precision_mode):
-            raise ImportError(
-                "Selected benchmark config is for TransformerEngine but could not import the TransformerEngine library!"
-            )
-
         if is_transformer_engine(low_precision_mode):
+            if not transformer_engine_available:
+                raise ImportError(
+                    "Selected benchmark config is for TransformerEngine but could not import the TransformerEngine library!"
+                )
             check_fp8_compute_capability()
 
         if "thunder" in self.compile and is_transformer_engine(self.low_precision_mode):
--- a/thunder/benchmarks/benchmark_litgpt.py
+++ b/thunder/benchmarks/benchmark_litgpt.py
@@ -20,11 +20,16 @@ from torch.distributed.algorithms._checkpoint.checkpoint_wrapper import (
 
 import thunder
 from thunder.tests.litgpt_model import Config, GPT, Block
-import transformer_engine.pytorch as te
-
-from lightning.fabric.plugins.precision.transformer_engine import TransformerEnginePrecision
 from lightning.fabric.utilities.throughput import measure_flops
 from lightning.fabric.utilities import Throughput
+
+
+try:
+    import transformer_engine.pytorch as te
+    from lightning.fabric.plugins.precision.transformer_engine import TransformerEnginePrecision
+    transformer_engine_available = True
+except ImportError:
+    transformer_engine_available = False
 
 
 world_size = int(os.environ.get("WORLD_SIZE", 1))
@@ -184,6 +189,9 @@ class Benchmark_litGPT:
                 print(
                     f"[WARNING] Bucketing mode is set to {self.bucketing_mode}. --fsdp_bucket_params will be ignored."
                 )
+                
+        if not transformer_engine_available and is_transformer_engine(low_precision_mode):
+            raise ImportError("Selected benchmark config is for TransformerEngine but could not import the TransformerEngine library!")
 
         if "thunder" in self.compile and is_transformer_engine(self.low_precision_mode):
             self.compile += "_transformerengine"
--- a/thunder/benchmarks/benchmark_litgpt.py
+++ b/thunder/benchmarks/benchmark_litgpt.py
@@ -46,10 +46,12 @@ def is_transformer_engine(low_precision_mode: str) -> bool:
 
 def get_context_manager(low_precision_mode: str, compile_mode: str) -> AbstractContextManager:
     if is_transformer_engine(low_precision_mode) and "thunder" not in compile_mode:
+
         @contextlib.contextmanager
         def fp8_autocast_context():
             with te.fp8_autocast(enabled=True):
                 yield
+
         return fp8_autocast_context
     else:
         return nullcontext
--- a/thunder/benchmarks/benchmark_litgpt.py
+++ b/thunder/benchmarks/benchmark_litgpt.py
@@ -2,8 +2,8 @@ import os
 import time
 import warnings
 from typing import Any
-from contextlib import nullcontext, AbstractContextManager
 import contextlib
+from contextlib import nullcontext, AbstractContextManager
 
 import torch
 import functools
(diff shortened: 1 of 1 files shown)


thunder/benchmarks/benchmark_litgpt.py
```diff
@@ -407,7 +425,8 @@ def train(self):
                     input_ids, targets = next(self.train_data_iter)
                     input_ids = input_ids.to(self.device)
                     targets = targets.to(self.device)
-                    logits = self.model(input_ids)
+                    with te.fp8_autocast(enabled=is_transformer_engine(self.low_precision_mode)):
```
t-vi, line 428: Should this also depend on whether we're running a thunder benchmark (similar to the `__init__`)? (also for the context manager below)

lantiga: I agree, let's create separate codepaths for with vs without TE.

wprazuch: Hello @t-vi, could you explain a little bit more? You mean that for thunder case, we do not need to provide a context manager?

t-vi: Yes, I have been wondering about the use of the context manager for thunder, and in particular the setup being conditioned to
```
# fp8-delayed-te precision for thunder is handled through compile
if is_transformer_engine(low_precision_mode) and "thunder" not in self.compile:
   ...
```

and then the context manager uses
```
with te.fp8_autocast(enabled=is_transformer_engine(self.low_precision_mode)):
```

We could have
- add `and "thunder" not in self.compile` to the enabled argument,
- a path without the context manager (or a `contextlib.nullcontext` manager) when not using `te.fp8_autocast`.

What happens when we hit the current context manager and we did not convert the module? To my mind, we need to be sure to have something that works in all four settings (thunder + no fp8, thunder + fp8, no thunder + no fp8, no thunder + fp8), and I'm not entirely sure that the current PR covers them.

wprazuch: I see. I thought it is also necessary to provide a context manager for thunder - I will make the required changes to straighten it out.

For the H100/GH200 remark, that is true. FP8-enabled script will throw error if we run for example for A100. I am not sure how to handle this - revert to pure BF16 run, or maybe throw an explicit error why it fails?

t-vi: > maybe throw an explicit error why it fails?

Yes, please! In the benchmarking context it is best to get error feedback when a requested config does not work.

kshitij12345: When using fp8 with `thunder`, `thunder.jit` takes care of wrapping the jitted function with `te.fp8_autocast`. Even, if the user runs this jitted function under their `te.fp8_autocast`, computation would still adhere to `thunder`'s context as it is the inner context. So, there is no need of `te.fp8_autocast` when using `thunder + transformer_engine_ex`.

See this [PR](https://github.com/Lightning-AI/lightning-thunder/pull/48) which added the support.

-----

For other cases, `te.fp8_autocast(enabled=True)` should have no effect when there are no TE layers in the model but it would be confusing to see this for a non-TE model. So, we should only enable it when TE mode is active.

One way to achieve this would be to optional switch context between `contextlib.nullcontext` or `te.fp8_autocast` depending on `is_transformer_engine` as suggested by @t-vi 

Other approach could be to replace `self.model` with a function which will call the model under `fp8_autocast` context (see example below). I like as there would be fewer places where you will have to condition on `is_transformer_engine` and fewer places to forget adding this context.

```python
if is_transformer_engine(low_precision_mode) and "thunder" not in self.compile:
    te_precision = TransformerEnginePrecision(weights_dtype=torch.bfloat16, replace_layers=True)
    converted_model = te_precision.convert_module(self.model)
    def call_model_with_fp8autocast(*args, **kwargs):
         with te.fp8_autocast(enabled=True):
             return converted_model(*args, **kwargs)
    self.model = call_model_with_fp8autocast
```
-------

Also, another thing to note regarding benchmark is that `TransformerEnginePrecision` will replace `Linear` and `LayerNorm` with their TE equivalent while `thunder.jit` with `transformer_engine_ex` currently only replaces the `Linear`. So, the models won't be similar in terms of the layers which do their computation in FP8.


https://lightning.ai/docs/fabric/stable/api/generated/lightning.fabric.plugins.precision.TransformerEnginePrecision.html

wprazuch: @kshitij12345 @t-vi @lantiga I applied changes according to your comments - hopefully I did not misunderstand your comments. Let me know if something else is needed or requires modification :)

thunder/benchmarks/benchmark_litgpt.py
```diff
@@ -188,6 +199,17 @@ def __init__(
...
+        # fp8-delayed-te precision for thunder is handled through compile
+        if is_transformer_engine(low_precision_mode) and "thunder" not in self.compile:
+            te_precision = TransformerEnginePrecision(weights_dtype=torch.bfloat16, replace_layers=True)
+            converted_model = te_precision.convert_module(self.model)
+
+            def call_model_with_fp8autocast(*args, **kwargs):
+                with te.fp8_autocast(enabled=True):
+                    return converted_model(*args, **kwargs)
+
+            self.model.__call__ = call_model_with_fp8autocast
```
t-vi, line 211: Is this safe?

wprazuch: The safer, or "less hacky" way would be to create an appropriate context manager depending on the condition:
```
if is_transformer_engine(low_precision_mode) and "thunder" not in self.compile
```
so it would return either null context, or `fp8_autocast`. So we could have a function returning a proper context manager object, and then use it in each model call. 
@kshitij12345 Or maybe you see a better way? I very liked your suggestion, but I don't know how to make it work without this `self.model.__call__ = call_model_with_fp8autocast` line 🙈

t-vi: I think I would feel much more at ease with the less hacky way of using a context manager, but clever seems OK, just overriding `self.model.__call__` looks risky (after staring down way to many Python corner cases).

kshitij12345: Ouch, apologies for proposing a risky solution. Let's stick to using appropriate ctx managers.

thunder/benchmarks/benchmark_litgpt.py
lantiga: I'm in favor of a completely separate path when TE is not used (`nullcontext` is useful to avoid code duplication but in this case the impact is minimal at the cost of clarity).

Also let's make importing TE optional at the top of the script, so we don't strictly require it when we run benchmarks. You can try-catch the import and set a boolean variable accordingly.

wprazuch: @lantiga What do you mean by "completely separate path"? Do you mean creating dedicated `def train(self)` method, sth like `def train_fp8(self):`? Or creating a separate module for fp8 training (`benchmark_litgpt_fp8.py`)?

thunder/benchmarks/benchmark_litgpt.py
wprazuch: @t-vi @kshitij12345 I added the changes for context managers, but I think the code will change according to the [recent comment](https://github.com/Lightning-AI/lightning-thunder/pull/689#discussion_r1665365705)

thunder/benchmarks/benchmark_litgpt.py
```diff
@@ -188,6 +224,8 @@ def __init__(
         self.model = self.init_model()
         print(f"Time to instantiate model: {time.perf_counter() - t0:.02f} seconds.")
 
+        self.model = get_converted_model(low_precision_mode, self.compile, self.model)
```
lantiga, line 254: here too we could be more explicit: just condition of `if use_transformer_engine` (as defined in the other comment) and convert the model, and leave it alone in the else

wprazuch: Done

thunder/benchmarks/benchmark_litgpt.py
```diff
@@ -407,7 +445,8 @@ def train(self):
                     input_ids, targets = next(self.train_data_iter)
                     input_ids = input_ids.to(self.device)
                     targets = targets.to(self.device)
-                    logits = self.model(input_ids)
+                    with self.context_manager():
```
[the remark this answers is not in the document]
wprazuch: Done

lantiga (approved):
Looks good!

PR #692, opened 2024-07-01
Fix proxy repr for non-tensor proxy

kshitij12345:
Repro
```python
import thunder

t = thunder.core.trace.TraceCtx()
with thunder.core.trace.tracectx(t):
    p = thunder.core.proxies.NumberProxy("foo", 1, python_type=int)
    t = thunder.core.proxies.TensorProxy("t", shape=(1,), dtype=thunder.core.dtypes.float16, device=thunder.core.devices.Device("cpu"), requires_grad=True)
    print(p)
    print(t)

```

Output
```python
  File "/home/kkalambarkar/lightning-thunder/scratchpad/test.py", line 40, in <module>
    print(p)
  File "/home/kkalambarkar/lightning-thunder/thunder/core/proxies.py", line 137, in __repr__
    return f'<{type(self).__name__}(name="{self.name}", dtype={self.dtype}, shape={self.shape}>'
AttributeError: 'NumberProxy' object has no attribute 'dtype'
```

### DIFF
--- a/thunder/core/proxies.py
+++ b/thunder/core/proxies.py
@@ -134,7 +134,9 @@ class Proxy(VariableInterface, ProxyInterface):
         return self.__class__(name=name)
 
     def __repr__(self) -> str:
-        return f'<{type(self).__name__}(name="{self.name}", dtype={self.dtype}, shape={self.shape}>'
+        # All subclasses of Proxy will have `self.name`, so this generic implementation relies on that.
+        # To have a specific repr for a subclass, override the implementation for that subclass.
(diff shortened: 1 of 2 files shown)


kshitij12345:
@k223kim - unable to request review from the dropdown, so requesting via comment

t-vi (approved):
Seems good, thank you @kshitij12345 for adding a test.

k223kim:
Hi @kshitij12345, thanks for catching and updating this. I should have considered the fact that not all Proxies have shapes.. Thank you!

PR #704, opened 2024-07-03
Update dtype repr in trace

kshitij12345:
Currently, dtype argument in a trace is printed as  `dtypes.thunder.dtypes.float16` which looks a bit odd (and potentially lead to errors). This PR updates it back to `dtypes.float16`

See example below for full repro.

```python
import thunder
import torch

def foo(x):
    return x.to(torch.float16)

x = torch.randn(3, 3)
jfoo = thunder.jit(foo)
jfoo(x)
```

Output
```python
from torch import Tensor
import torch

@torch.no_grad()
@no_autocast
def computation(x):
  # x: "cpu f32[3, 3]"
  t0 = Tensor.to(x, copy=False, dtype=torch.float16)  # t0: "cpu f16[3, 3]"
    # t0 = ltorch.to(x, None, None, device=None, dtype=torch.float16, copy=False, memory_format=None)  # t0: "cpu f16[3, 3]"
      # t0 = prims.convert_element_type(x, dtypes.thunder.dtypes.float16)  # t0: "cpu f16[3, 3]"
  del x
  return t0
```

### DIFF
--- a/thunder/core/codeutils.py
+++ b/thunder/core/codeutils.py
@@ -227,7 +227,10 @@ def prettyprint(
         unflattened_str = unflattened_str.replace(f"'{_quote_marker}", "")
         return unflattened_str
     if isinstance(x, dtypes.dtype):
-        return m(f"{str(x)}")
+        # str(x) -> thunder.dtypes.foo
+        # For consistency with previous repr,
+        # remove `thunder.` from the representation.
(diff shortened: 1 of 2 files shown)


t-vi:
previously, this was `dtypes.float16`.

lantiga (approved):
LGTM

PR #705, opened 2024-07-03
autocast: support mixed dtypes

kshitij12345:
Fixes: https://github.com/Lightning-AI/lightning-thunder/issues/678

In PyTorch eager, enabling autocast allows mixed inputs to operator like `linear` which expect dtypes to be same. This works in PyTorch eager, as dispatcher applies the conversion first and then passes the converted input to the operator.

To mimick similar behavior, here we update the mapping of `torch` -> `thunder.torch` for all operators which have autocast rule, to apply the conversion rule and call the mapped function if autocast was enabled otherwise to directly map the function.

### DIFF
--- a/thunder/__init__.py
+++ b/thunder/__init__.py
@@ -391,7 +391,7 @@ def jit(
                 autocast_cpu_dtype=str(autocast_cpu_dtype),
             )
             autocast_thunder_dtype = autocast_cpu_dtype if pytorch.is_autocast_cpu_enabled() else autocast_gpu_dtype
-            cd.autocast_thunder_dtype = autocast_thunder_dtype
+            cache_info.update(autocast_thunder_dtype=str(autocast_thunder_dtype))
 
         cache_info["is_autocast_enabled"] = is_autocast_enabled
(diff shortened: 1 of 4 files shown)


t-vi (approved):
Thank you @kshitij12345 !

As discussed with @kshitij12345 , while it would be nice to "record without autocast" and then do autocast as a transform", this comes with a bunch of caveats, so this seems to be a pragmatic approach.
Note that you can still do an autocast transform, it just isn't the mechanism used when interacting with the torch autograd context manager.

thunder/__init__.py
```diff
@@ -571,14 +572,6 @@ def get_computation_and_inputs(*args, **kwargs):
             computation_trc = dce(computation_trc)
             computation_traces.append(computation_trc)
 
-            if is_autocast_enabled:
```
IvanYashchuk, line 574: Before this PR it was possible to apply autocast with a context manager to any symbol with a registered autocast rule.
Now it's only for functions decorated with "torchsymbol" *and* with a registered autocast rule.

Issue #694, opened 2024-07-02
Fail to interpret `megatron.core.tensor_parallel.layers.LinearWithGradAccumulationAndAsyncCommunication`


crcrpar:
*Note*: If you have a model or program that is not supported yet but should be, please use the program coverage template.

#### Code sample

```python
import torch
from nemo.collections.nlp.modules.common.megatron.mlp import ParallelMLP
from megatron.core.tensor_parallel.layers import LinearWithGradAccumulationAndAsyncCommunication

import thunder


def f(x: torch.Tensor, weight: torch.Tensor, bias: torch.Tensor) -> tuple[torch.Tensor]:
    return LinearWithGradAccumulationAndAsyncCommunication.apply(
        x,
        weight,
        bias,
        False,  # gradient_accumulation_fusion
        False,  # allreduce_dgrad
        False,  # sequence_parallel
        None,  # grad_output_buffer
    )


def main():
    device = torch.device("cuda")
    s, b, hp = 1025, 1, 256
    with device:
        l = torch.nn.Linear(256, 1024).to(device)
        weight = l.weight
        bias = l.bias
        x = torch.randn((s, b, hp))

    jitted = thunder.jit(f)
    computation_traces, *_ = thunder.compile_data(jitted).get_computation_and_inputs(x, weight, bias)
    print(computation_traces)
    y = jitted(x, weight, bias)


if __name__ == "__main__":
    main()
```

error message:
```
$ python megatron_core_col_linear_fwd.py
[NeMo W 2024-07-02 05:02:32 nemo_logging:349] /opt/pytorch/Megatron-LM/megatron/core/tensor_parallel/layers.py:279: FutureWarning: `torch.cuda.amp.custom_fwd(args...)` is deprecated. Please use `torch.amp.custom_fwd(args..., device_type='cuda')` instead.
      def forward(

[NeMo W 2024-07-02 05:02:32 nemo_logging:349] /opt/pytorch/Megatron-LM/megatron/core/tensor_parallel/layers.py:291: FutureWarning: `torch.cuda.amp.custom_bwd(args...)` is deprecated. Please use `torch.amp.custom_bwd(args..., device_type='cuda')` instead.
      def backward(ctx, grad_output):

[...]

PR #707, opened 2024-07-04
Add lookaside for `torch.autograd.function.Function.apply`

crcrpar:
Fixes https://github.com/Lightning-AI/lightning-thunder/issues/694#issuecomment-2201959134

### DIFF
--- a/thunder/core/jit_ext.py
+++ b/thunder/core/jit_ext.py
@@ -891,6 +891,20 @@ def _general_jit_named_buffers_lookaside(obj: Any, *args, **kwargs):
     )
 
 
+@general_jit_lookaside(torch.autograd.function.Function.apply.__func__)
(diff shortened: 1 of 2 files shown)


t-vi (approved):
Thank you @crcrpar . At some point we might move the warning into the lookaside, but let's have this now.

lantiga:
Great @crcrpar. Why do we need to `skip_inplace_functionalization`?

crcrpar:
> Great @crcrpar. Why do we need to `skip_inplace_functionalization`?

thank you.
We don't need it. I vaguely remember that I turned it off because the test cases don't use any in-place thus functionalization isn't needed

lantiga:
I would avoid turning defaults off ad-hoc, they will be less likely to break for unforseen interactions.

crcrpar:
100% agreed. opened 712 :)

Issue #460, opened 2024-05-25
Support for slices in advanced indexing (was 'NoneType' object error using thunder.jit with NeMo Stable Diffusion)


athitten:
### 🐛 Bug

Applying `thunder.jit` to the UNet stage of NeMo's Stable Diffusion raises an error: `'NoneType' object is not callable`
This happens for the `gt` method(greater than).

### To Reproduce

1. Apply the diff attached to NeMo
2. Use the command below to run Stable Diffusion in NeMo

`python examples/multimodal/text_to_image/stable_diffusion/sd_train.py trainer.precision=16 trainer.num_nodes=1 trainer.devices=1 ++exp_manager.max_time_per_run=00:00:03:00 trainer.max_steps=20 model.micro_batch_size=1 model.global_batch_size=1 model.data.synthetic_data=True exp_manager.exp_dir=/workspace/TestData/multimodal/stable_diffusion_train model.inductor=False model.cond_stage_config._target_=nemo.collections.multimodal.modules.stable_diffusion.encoders.modules.FrozenCLIPEmbedder ++model.cond_stage_config.version=openai/clip-vit-large-patch14 ++model.cond_stage_config.max_length=77 ~model.cond_stage_config.restore_from_path ~model.cond_stage_config.freeze ~model.cond_stage_config.layer model.unet_config.from_pretrained=null model.first_stage_config.from_pretrained=null model.unet_config.use_flash_attention=False model.unet_config.attention_resolutions=\[1\] model.unet_config.channel_mult=\[1\]`

Please let me know if any more details are needed. Thanks!
CC: @tfogal 
[unet.patch](https://github.com/Lightning-AI/lightning-thunder/files/15440910/unet.patch)

cc @apaz-cli @tfogal

crcrpar:
`gt` itself seems to be available as we have it in https://github.com/Lightning-AI/lightning-thunder/blob/7d6e540a6ec0bebcb6bedc5ceadbc82d1b982b65/thunder/torch/__init__.py#L1543

tfogal:
With relatively more recent NeMo versions, we appear to hit issues with different modes of indexing:
```
NotImplementedError: key=(t4, slice(None, None, None)) mixes basic and advanced indexing that is not currently supported
```
Could this be related to #187 ?

Full traceback:
```
[rank0]: Traceback (most recent call last):
[...]

Issue #187, opened 2024-04-15
Advanced idexing NoneType support


riccardofelluga:
## 🐛 Bug

In some edge cases it is possible to encounter a None value as index along with tensor indices. In such cases out  `_advanced_indexing` clang op fails.

### To Reproduce

Repro script:

```python
import torch; import thunder

a = torch.randn(15, 15, 15)
indices  = torch.tensor([1, 2, 3])

def apply_indices(a, b):
    return a[b, b, None]

jfn= thunder.jit(apply_indices)

jfn(a, indices)
```

This fails with the following:

```shell
File /lightning-thunder/thunder/clang/__init__.py:699, in _advanced_indexing(a, key)
    697 if x is Ellipsis:
    698     num_ellipses += 1
--> 699 utils.check(
    700     isinstance(x, (EllipsisType, TensorLike)),
    701     lambda: f"Advanced indexing currently only supports tensors as sequence elements (possibly with a starting ellipsis), but found an object of type {type(x)}",
    702 )
    703 if isinstance(x, TensorLike):
    704     utils.check(
    705         dtypes.is_nonboolean_integer_dtype(x.dtype) and (x.ndim == 1 or x.ndim == 0),
    706         lambda: f"Advanced indexing currently only supports zero or one-dimensional integer tensors, but found a tensor with dtype {x.dtype} and {x.ndim} dimensions",
    707     )

File /lightning-thunder/thunder/core/baseutils.py:103, in check(cond, s, exception_type)
     98 """Helper function for raising an error_type (default: RuntimeError) if a boolean condition fails.
     99 
    100 s is a callable producing a string to avoid string construction if the error check is passed.
    101 """
    102 if not cond:
--> 103     raise exception_type(s())

RuntimeError: Advanced indexing currently only supports tensors as sequence elements (possibly with a starting ellipsis), but found an object of type <class 'NoneType'>
```

### Expected behavior

That's actually a good question, in theory it should create a new axis at the position but sometimes it doesn't, like in this case where the output shape is `(3, 1, 15)`. Corrections about this statement are very welcome.

### Additional context

Such code can be found as part of Mixtral 8x7B.

cc. @mruberry

t-vi:
@k223kim wants to work on this. Thank you!

k223kim:
Hey Team ⚡️, as Tom mentioned, I will be working on this issue. Thanks!

riccardofelluga:
[...]

PR #710, opened 2024-07-04
some more indexing options (mix of basic, advanced, lists in place of tensors for advanced)

### LINKED ISSUE #643: Indexing with slice and list
Repro:

```
def fn(input_ids):
   return input_ids[:, [-1, 0]]

a = torch.randn(5, 5)
jfn = thunder.jit(fn)
jfn(a)
```

Needed for transformers BERT:

```
import thunder, torch, transformers

m = transformers.BertForSequenceClassification(transformers.BertConfig())
x = torch.randint(1, 10, (1, 32))
jm = thunder.jit(m)
jm(x)
```

Might ideally be fixed along with #460 and #187 

cc @apaz-cli @tfogal

k223kim:
<details>

</details>

## What does this PR do?

Fixes #643, #460, #187.

## PR review

## Did you have fun?

⚡️

### DIFF
--- a/thunder/clang/__init__.py
+++ b/thunder/clang/__init__.py
@@ -866,8 +866,10 @@ def _advanced_indexing(a: TensorLike, /, key) -> TensorLike:
     else:
         # NOTE No ellipsis case
         permute_shape = permute_shape + list(range(len(key), len(a.shape)))
-        a = transpose(a, tuple(permute_shape))
-        key = tuple(permute_key)
+        # check if we need to permute
+        if permute_shape != list(range(a.ndim)):
+            a = transpose(a, tuple(permute_shape))
+            key = tuple(permute_key)
 
         subtensor_shape = a.shape[: len(key)]
         remaining_shape = list(a.shape[len(key) :])
--- a/thunder/clang/__init__.py
+++ b/thunder/clang/__init__.py
@@ -727,14 +727,14 @@ def _advanced_indexing(a: TensorLike, /, key) -> TensorLike:
     # this handles both cases when key is list or a tuple that contains lists
     if isinstance(key, Sequence):
         if isinstance(key, list):
-            key = tensor_from_sequence(key, dtype=dtypes.int32, device=a.device)
+            key = tensor_from_sequence(key, dtype=dtypes.int8, device=a.device)
         else:
             key_ = []
             for x in key:
                 if isinstance(x, list):
-                    x = tensor_from_sequence(x, dtype=dtypes.int32, device=a.device)
+                    x = tensor_from_sequence(x, dtype=dtypes.int8, device=a.device)
                     key_.append(x)
-                if isinstance(x, (EllipsisType, TensorLike)):
+                elif isinstance(x, (EllipsisType, TensorLike)):
                     key_.append(x)
                 # intentionally skip indices with basic indexing
             key = tuple(key_)
@@ -889,11 +889,15 @@ def getitem(a: TensorLike, /, key) -> TensorLike:
     if isinstance(key, TensorLike) or (isinstance(key, Sequence) and not isinstance(key, tuple)):
         return _advanced_indexing(a, key)
 
-    a = _basic_indexing(a, key)
+    # a = _basic_indexing(a, key)
 
-    if sig.advanced:
-        a = _advanced_indexing(a, key)
+    if not sig.advanced:
+        return _basic_indexing(a, key)
 
+    if isinstance(key, tuple):
+        for x in key:
+            if isinstance(x, TensorLike) or isinstance(x, Sequence):
+                return _advanced_indexing(a, key)
     return a
 
 
--- a/thunder/clang/__init__.py
+++ b/thunder/clang/__init__.py
@@ -514,7 +514,7 @@ def _get_indexing_signature(key: Any) -> IndexingSignature:
         return sig
 
     # TensorLike triggers advanced indexing.
-    if isinstance(key, TensorLike):
+    if isinstance(key, (TensorLike)):
         sig.advanced.append((None, None))
         return sig
 
@@ -598,7 +598,14 @@ def _basic_indexing(a: TensorLike, /, key) -> TensorLike:
     if key is None or isinstance(key, (Number, NumberProxy, slice, EllipsisType)):
         key = (key,)
 
+    _key = []
     for idx, x in enumerate(key):
+        # convert the advanced indexing key to None
+        if isinstance(x, (list, TensorLike)):
+            x = None
+            _key.append(x)
+        else:
+            _key.append(x)
         if x is Ellipsis:
             utils.check(ellipsis_idx is None, lambda: f"Found two (or more) ellipses in key={key}")
             ellipsis_idx = idx
@@ -611,6 +618,7 @@ def _basic_indexing(a: TensorLike, /, key) -> TensorLike:
                 unsqueeze_dims_post_ellipsis.append(idx)
         else:
             raise ValueError(f"Found unexpected value {x} in key={key}")
+    key = tuple(_key)
 
     utils.check(
         specified_slices <= len(a.shape),
@@ -714,6 +722,22 @@ def _advanced_indexing(a: TensorLike, /, key) -> TensorLike:
         isinstance(key, (TensorLike, Sequence)),
         lambda: f"Advanced indexing currently only supports keys that are ellipses, integer tensors or sequences, but got {key=}",
     )
+
+    # convert list to 1D tensor
+    # this handles both cases when key is list or a tuple that contains lists
+    if isinstance(key, Sequence):
+        if isinstance(key, list):
+            key = tensor_from_sequence(key, dtype=dtypes.int8, device=a.device)
+        else:
+            key_ = []
+            for key_idx, x in enumerate(key):
+                if isinstance(x, list):
+                    x = tensor_from_sequence(x, dtype=dtypes.int8, device=a.device)
+                if not isinstance(x, (EllipsisType, TensorLike)):
+                    # convert this into a tensor with all the indices from 0 to len(a.shape[key_idx])
+                    x = tensor_from_sequence(list(range(a.shape[key_idx])), dtype=dtypes.int8, device=a.device)
+                key_.append(x)
+            key = tuple(key_)
 
     if isinstance(key, (TensorLike)):
         key = utils.sequencify(key)
@@ -862,21 +886,26 @@ def getitem(a: TensorLike, /, key) -> TensorLike:
                 key = tuple(key[:key_idx]) + (slice(start, end),) + tuple(key[key_idx + 1 :])
                 return _basic_indexing(a, key)
 
-    utils.check(
-        not (len(sig.basic) > 0 and len(sig.advanced) > 0),
-        lambda: f"{key=} mixes basic and advanced indexing that is not currently supported",
-        NotImplementedError,
-    )
+    # utils.check(
+    #     not (len(sig.basic) > 0 and len(sig.advanced) > 0),
+    #     lambda: f"{key=} mixes basic and advanced indexing that is not currently supported",
+    #     NotImplementedError,
+    # )
 
     if isinstance(key, TensorLike) or (isinstance(key, Sequence) and not isinstance(key, tuple)):
         return _advanced_indexing(a, key)
 
-    if isinstance(key, tuple):
-        for x in key:
-            if isinstance(x, TensorLike) or isinstance(x, Sequence):
-                return _advanced_indexing(a, key)
+    # figure out a way to do advanced indexing first and then basic indexing
+    if sig.advanced:
+        a = _advanced_indexing(a, key)
+    # if isinstance(key, tuple):
+    #     for x in key:
+    #         if isinstance(x, TensorLike) or isinstance(x, Sequence):
+    #             a = _advanced_indexing(a, key)
 
-    return _basic_indexing(a, key)
+    if sig.basic:
+        a = _basic_indexing(a, key)
+    return a
 
 
 # Based on NumPy's https://numpy.org/doc/stable/reference/generated/numpy.moveaxis.html
--- a/thunder/clang/__init__.py
+++ b/thunder/clang/__init__.py
@@ -705,6 +705,10 @@ def _basic_indexing(a: TensorLike, /, key) -> TensorLike:
             # NOTE: this is redundant with the ValueError exception above
             raise ValueError(f"Found unexpected value {x} in key={key}")
 
+    # check if we can skip slice_prim
+    if all([x == 0 for x in start_indices]) and all([x == l for x, l in zip(end_indices, a.shape)]):
+        return a, tuple(new_key)
+
     result = prims.slice_prim(a, start_indices, end_indices, strides)
 
     if len(squeeze_dims) > 0:
@@ -850,8 +854,10 @@ def _advanced_indexing(a: TensorLike, /, key) -> TensorLike:
         del modified_key[0]
         permute_shape = [x + len(list(a.shape[: a.ndim - (seq_len - 1)])) - 1 for x in permute_shape]
         permute_shape = list(range(a.ndim - (seq_len - 1))) + permute_shape
-        a = transpose(a, tuple(permute_shape))
-        key = tuple(permute_key)
+        # check if we need to permute
+        if permute_shape != list(range(a.ndim)):
+            a = transpose(a, tuple(permute_shape))
+            key = tuple(permute_key)
 
         subtensor_shape = a.shape[a.ndim - (seq_len - 1) :]
         remaining_shape = list(a.shape[: a.ndim - (seq_len - 1)])
@@ -925,8 +931,10 @@ def _advanced_indexing(a: TensorLike, /, key) -> TensorLike:
     if all(k.ndim == 0 for k in modified_key if isinstance(k, TensorLike)):
         res = squeeze(res, (dim,))
     res = reshape(res, tuple(new_shape))
-    # permute back to original shape
-    res = transpose(res, tuple(permute_shape[: len(new_shape)]))
+    # check if we need to permute
+    if permute_shape != list(range(a.ndim)):
+        # permute back to original shape
+        res = transpose(res, tuple(permute_shape[: len(new_shape)]))
 
     return res
 
--- a/thunder/clang/__init__.py
+++ b/thunder/clang/__init__.py
@@ -727,16 +727,16 @@ def _advanced_indexing(a: TensorLike, /, key) -> TensorLike:
     # this handles both cases when key is list or a tuple that contains lists
     if isinstance(key, Sequence):
         if isinstance(key, list):
-            key = tensor_from_sequence(key, dtype=dtypes.int8, device=a.device)
+            key = tensor_from_sequence(key, dtype=dtypes.int32, device=a.device)
         else:
             key_ = []
-            for key_idx, x in enumerate(key):
+            for x in key:
                 if isinstance(x, list):
-                    x = tensor_from_sequence(x, dtype=dtypes.int8, device=a.device)
-                if not isinstance(x, (EllipsisType, TensorLike)):
-                    # convert this into a tensor with all the indices from 0 to len(a.shape[key_idx])
-                    x = tensor_from_sequence(list(range(a.shape[key_idx])), dtype=dtypes.int8, device=a.device)
-                key_.append(x)
+                    x = tensor_from_sequence(x, dtype=dtypes.int32, device=a.device)
+                    key_.append(x)
+                if isinstance(x, (EllipsisType, TensorLike)):
+                    key_.append(x)
+                # intentionally skip indices with basic indexing
             key = tuple(key_)
 
     if isinstance(key, (TensorLike)):
@@ -886,25 +886,14 @@ def getitem(a: TensorLike, /, key) -> TensorLike:
                 key = tuple(key[:key_idx]) + (slice(start, end),) + tuple(key[key_idx + 1 :])
                 return _basic_indexing(a, key)
 
-    # utils.check(
-    #     not (len(sig.basic) > 0 and len(sig.advanced) > 0),
-    #     lambda: f"{key=} mixes basic and advanced indexing that is not currently supported",
-    #     NotImplementedError,
-    # )
-
     if isinstance(key, TensorLike) or (isinstance(key, Sequence) and not isinstance(key, tuple)):
         return _advanced_indexing(a, key)
 
-    # figure out a way to do advanced indexing first and then basic indexing
+    a = _basic_indexing(a, key)
+
     if sig.advanced:
         a = _advanced_indexing(a, key)
-    # if isinstance(key, tuple):
-    #     for x in key:
-    #         if isinstance(x, TensorLike) or isinstance(x, Sequence):
-    #             a = _advanced_indexing(a, key)
 
-    if sig.basic:
-        a = _basic_indexing(a, key)
     return a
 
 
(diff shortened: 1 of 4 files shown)


thunder/clang/__init__.py
```diff
@@ -715,8 +727,49 @@ def _advanced_indexing(a: TensorLike, /, key) -> TensorLike:
         lambda: f"Advanced indexing currently only supports keys that are ellipses, integer tensors or sequences, but got {key=}",
     )
 
+    def _to_tensorproxies(x: list, device: devices.DeviceType):
```
mruberry, line 730: Why do you want to convert a list (or other sequence?) to a tensor?

mruberry:
Hey @k223kim!

What's your thinking on the overall strategy to support more (all?) advanced integer indexing?

My understanding may be outdated, but currently I think thunder supports all basic indexing cases, some advanced integer indexing cases, and no advanced boolean indexing cases. Supporting advanced boolean indexing is a whole issue because it involves data-dependent control-flow.

The reason we only support some advanced integer indexing cases is because we couldn't think of an elegant way to map the additional cases to existing thunder operations. One option for supporting more of these cases would be to write a meta function for them, but then just call torch's advanced indexing directly (adding a new primitive for it). The backward for that primitive would likely need to be another primitive that called torch's advanced indexing backward operation directly, too. I'm not sure how easy those operations are to access.

I wonder if we should generally try falling back to the torch operation, and look to devise more special cases where we can express advanced indexing logic by decomposing it to existing operations? Or do you think that we can decompose all advanced indexing cases to existing operations in a reasonable way?

cc @nikitaved

thunder/clang/__init__.py
k223kim: The reason behind this was because, before this PR, advanced indexing supported Tensor indexing. I thought it would be better to use the existing functionality when receiving a list.

k223kim:
Hi @mruberry ! It's been a while since I got a review from you :) great to see you! Thank you for the detailed feedback. 
Before answering your question, I am trying to understand where your comments are coming from. Isn't this PR specifically for adding list indexing to advanced indexing? I am not sure why you are talking about "advanced boolean indexing". Are you suggesting the implementation of list indexing in advanced indexing is not needed? Would you able to explain this for me? I would like to better understand your thoughts.

thunder/clang/__init__.py
t-vi: In order to do the linear indices / reduce to take we would want tensors for tensor arithmetic.
The list or sequence is a bit complicated due to tuple being special in getitem (so `a[[1, 2, 3]]` is the same as `a[((1, 2, 3),)]` but not as `a[(1, 2, 3)]` which is `a[1, 2, 3]`). For writing things, list clearly using a list is *the* correct thing.

thunder/clang/__init__.py
```diff
@@ -860,23 +965,18 @@ def getitem(a: TensorLike, /, key) -> TensorLike:
...
 
     if isinstance(key, TensorLike) or (isinstance(key, Sequence) and not isinstance(key, tuple)):
         return _advanced_indexing(a, key)
 
-    if isinstance(key, tuple):
-        for x in key:
-            if isinstance(x, TensorLike) or isinstance(x, Sequence):
-                return _advanced_indexing(a, key)
+    # perform basic indexing first
+    a, key = _basic_indexing(a, key)
```
t-vi, line 975: The question I had yesterday could more succinctly be asked: Should here (or likely more reasonably  `result = prims.slice_prim(a, start_indices, end_indices, strides)`) be an `if` to not perform `slice_prim` when we don't have anything to slice. (But it turns out that having this is not what prevents NVFuser to fail.)

thunder/clang/__init__.py
```diff
@@ -776,17 +829,42 @@ def _advanced_indexing(a: TensorLike, /, key) -> TensorLike:
...
+    for key_idx, key_val in advanced_keys:
+        permute_shape.append(key_idx)
+        permute_key.append(key_val)
+
+    # transpose so that we have basic keys first
     if has_ellipsis:
         del modified_key[0]
+        permute_shape = [x + len(list(a.shape[: a.ndim - (seq_len - 1)])) - 1 for x in permute_shape]
+        permute_shape = list(range(a.ndim - (seq_len - 1))) + permute_shape
+        a = transpose(a, tuple(permute_shape))
```
t-vi, line 853: Should this be with an if in case we already have the right permutation?

thunder/clang/__init__.py
```diff
@@ -776,17 +829,42 @@ def _advanced_indexing(a: TensorLike, /, key) -> TensorLike:
...
+        key = tuple(permute_key)
+
         subtensor_shape = a.shape[a.ndim - (seq_len - 1) :]
+        remaining_shape = list(a.shape[: a.ndim - (seq_len - 1)])
         dim = -1
         flattened = flatten(a, a.ndim - (seq_len - 1), -1)
     else:
         # NOTE No ellipsis case
+        permute_shape = permute_shape + list(range(len(key), len(a.shape)))
+        a = transpose(a, tuple(permute_shape))
```
t-vi, line 863: Same as above (Should this be with an if in case we already have the right permutation?).

t-vi (approved):
Hi @k223kim ! Great work!
I just checked that with NVFuser built after https://github.com/NVIDIA/Fuser/pull/2547 , we pass the CI, so I think we wait for the builds to pick that up an then merge.
If you wanted, you might look at guarding the `slice_prim` and transpose with `if`s to only do them if needed (but we can also do this in a follow-up).

thunder/clang/__init__.py
k223kim: Thank you for the explanation. I got confused last night. You are right. Let me update it so that it can check if we need indexing at all.

thunder/clang/__init__.py
k223kim: Yes, thank you! I have updated accordingly.

thunder/clang/__init__.py
mruberry: The thing I worry about with converting a list into a tensor is that the list could be constexpr (known at compile-time) while we don't really have a notion of tensors with constexpr data (although we could do that, too, I suppose).

k223kim: Do you think it'll be helpful to submit an issue regarding that? Would you be able to provide me an example of the scenario that you are particularly concerned about?

mruberry: No, that's OK -- it's a pretty niche concern we probably won't need to bother with for a long time

k223kim: Ok, cool. Let me resolve this!

mruberry:
> Hi @mruberry ! It's been a while since I got a review from you :) great to see you! 

Nice to work with you, too!

> Thank you for the detailed feedback. Before answering your question, I am trying to understand where your comments are coming from. Isn't this PR specifically for adding list indexing to advanced indexing? I am not sure why you are talking about "advanced boolean indexing". Are you suggesting the implementation of list indexing in advanced indexing is not needed? Would you able to explain this for me? I would like to better understand your thoughts.

Absolutely! Let's talking about indexing! fyi @nikitaved who may add additional info or correct me.

The best official documentation on indexing tensors is probably https://numpy.org/doc/stable/user/basics.indexing.html, and now there's the newer https://data-apis.org/array-api/2022.12/API_specification/indexing.html, but that document doesn't describe indexing with integer tensors. Let's talk about each separately.

The NumPy documentation has the following categories for indexing:

- basic indexing
- advanced integer indexing
- advanced boolean indexing
- combined advanced and basic indexing

Advanced indexing is triggered when one the key is any of:

- a non-tuple sequence object (like a list)
- an integer or boolean array
- a tuple with one or more sequences, integer arrays, or boolean arrays

Advanced boolean indexing (AKA "boolean array indexing") occurs when the key is a boolean array. This is an operation where the output metadata is dependent on the input's tensor data, and since the tensor data is not known at compile-time we cannot know the output metadata at compile-time. Thunder currently doesn't handle these operations (although there are some ideas about how to do so). The Array API documentation even has a note about this (https://data-apis.org/array-api/2022.12/API_specification/indexing.html#boolean-array-indexing). 

However, any key that doesn't include a boolean tensor does have output metadata that can be determined at compile-time. We could support a lot of advanced indexing by writing the meta function for these cases, then calling a special indexing primitive (which the torch executor would execute by calling PyTorch's getitem). One issue with this approach, however, is that PyTorch does not correctly implement NumPy's indexing, and so we'd have to write a meta function that describes PyTorch's indexing behavior, which could be tricky because I'm not sure PyTorch's indexing behavior is documented.

I don't think the NumPy documentation describing indexing is that great, but it does an OK job explaining the core concepts. Personally, I would have created different indexing categories.

Today Thunder supports:

- basic indexing
- advanced indexing with a sequence of 0D or 1D integer tensors (possibly starting with an ellipsis)
- indexing with a sequence that consists of a single number

This PR can focus narrowly on changing the second bullet from 

- advanced indexing with a sequence of 0D or 1D integer tensors (possibly starting with an ellipsis)

to

- advanced indexing with a sequence of 0D or 1D integer tensors or sequences (possibly starting with an ellipsis)

But it looks like this PR is also trying to improve support for combined indexing?

If the PR just wants to extend the current support for lists (which seems totally reasonable), then what about converting sequences to tensors here?

https://github.com/Lightning-AI/lightning-thunder/blob/6dfe7e939a19d1ef5ab259de8709a79f0104fa42/thunder/clang/__init__.py#L725

```
def _seq_to_tensor(x):
  if isinstance(x, Sequence):
    return tensor_from_sequence(x)
  return x

key = (_seq_to_tensor(k) for k in key)
```

Wouldn't that enable support for empty lists and lists that define a tensor with a single dimension?

k223kim:
Wow, thanks so much for the detailed explanation @mruberry ! I'm always learning from your comments.

Before anything, I would like to clarify one thing; although the PR is named to support 'advanced list indexing', it ended up also touching on combinations of basic and advanced due to the repro examples provided in each issue. Here is an example:
```
def fn(input_ids):
   return input_ids[:, [-1, 0]] 

a = torch.randn(5, 5)
jfn = thunder.jit(fn)
jfn(a)
```
Maybe it would have been less confusing if I had separated the PR into two pieces; one for adding support for list indexing and one for combining advanced and basic indexing. What do you think? (Also curious what @t-vi suggests)

Also, I think your suggestion is valid, however, that means that we wouldn't be able to support something like: `a[[1,2,3]]` (cause in this case, `key` would be the list itself), right? Is that also what you are intending?

mruberry:
> Wow, thanks so much for the detailed explanation @mruberry ! I'm always learning from your comments.
> 
> Before anything, I would like to clarify one thing; although the PR is named to support 'advanced list indexing', it ended up also touching on combinations of basic and advanced due to the repro examples provided in each issue. Here is an example:
> 
> ```
> def fn(input_ids):
>    return input_ids[:, [-1, 0]] 
> 
> a = torch.randn(5, 5)
> jfn = thunder.jit(fn)
> jfn(a)
> ```
> 
> Maybe it would have been less confusing if I had separated the PR into two pieces; one for adding support for list indexing and one for combining advanced and basic indexing. What do you think? (Also curious what @t-vi suggests)
> 

I think separating the work into a few PRs would be helpful because indexing is so tricky it can be hard to review multiple changes at once.

> Also, I think your suggestion is valid, however, that means that we wouldn't be able to support something like: `a[[1,2,3]]` (cause in this case, `key` would be the list itself), right? Is that also what you are intending?

I think that could work because the key is a list, which would trigger advanced indexing (the key is a non-tuple sequence), but the logic here would have to change:

https://github.com/Lightning-AI/lightning-thunder/blob/6dfe7e939a19d1ef5ab259de8709a79f0104fa42/thunder/clang/__init__.py#L718

I think we'd have to do something like create a sequence (lst,), just like we create sequences for tensors like (t,)?

k223kim:
Yeah that could definitely work. I wanted to clarify if that was also part of the "advanced indexing". Thanks for the explanation. 

However, I do want to understand why we would want to separate this PR if both list indexing and mixed indexing (combination of basic and advanced) is working..? Is there a particular concern regarding this implementation? Or is it for the sake of clarity?

thunder/clang/__init__.py
```diff
@@ -715,8 +731,49 @@ def _advanced_indexing(a: TensorLike, /, key) -> TensorLike:
...
+            x = tensor_from_sequence(x, dtype=dtypes.int64, device=device)
+        return x
+
+    basic_keys = []  # (key index, key)
+    advanced_keys = []  # (key index, key)
+
+    # convert list to 1D tensor
+    # this handles both cases when key is list or a tuple that contains lists
+    if isinstance(key, Sequence):
+        if isinstance(key, list):
```
t-vi, line 750: If we wanted to improve this, we could swap the if and else to condition on key being a TensorLike.

mruberry:
> Yeah that could definitely work. I wanted to clarify if that was also part of the "advanced indexing". Thanks for the explanation.
> 
> However, I do want to understand why we would want to separate this PR if both list indexing and mixed indexing (combination of basic and advanced) is working..? Is there a particular concern regarding this implementation? Or is it for the sake of clarity?

It would just make it easier to review. I think the PR adding the conversion of lists into tensors is straightforward, can be merged quickly, and has its own clear tests. All of that makes it a good candidate to separate from the more complicated combined indexing.

t-vi:
FWIW, the thing that is blocking this PR is that the new tests seems to trigger https://github.com/NVIDIA/Fuser/pull/2547 and the fix for that is in git but the packages we pull for PyTorch 2.3 have stopped updating 2 months ago.
Happily @xwang233 is looking into that, and hopefully we can resolve it soon, but the holdup is very orthogonal to the PR itself and at 130 lines plus tests, the PR is largish but does not seem to be excessively complicated.

The scoping of this PR has been discussed and is informed by the user needs, which is a mode that has served us well to make progress in the past few weeks.

> I think we'd have to do something like create a sequence (lst,), just like we create sequences for tensors like (t,)?

As far as I can tell we do handle that case, from the code, the tests, and this:

```
import torch, thunder

def fn(x):
    return x[[1, 2]]

jfn = thunder.jit(fn)
torch.testing.assert_(jfn(torch.arange(5)), fn(torch.arange(5)))
```

I mean, if there is something wrong with the patch, we can either fix it here, or preferably in a follow-up. To my mind it is OK, and people can review or not review based on their appetite.

thunder/clang/__init__.py
```diff
@@ -598,7 +598,18 @@ def _basic_indexing(a: TensorLike, /, key) -> TensorLike:
     if key is None or isinstance(key, (Number, NumberProxy, slice, EllipsisType)):
         key = (key,)
 
+    _key = []
```
mruberry, line 601: Part of the reason I'd like to separate this PR more into updates to advanced indexing and then support for combined indexing is that I'd like to better understand updating basic indexing to support the combined indexing. I think a discussion of approaches for combined indexing support would be really interesting.

thunder/clang/__init__.py
```diff
@@ -598,7 +598,18 @@ def _basic_indexing(a: TensorLike, /, key) -> TensorLike:
...
+    # keeps track of keys that are not squeeze related
+    new_key = []
     for idx, x in enumerate(key):
+        if not isinstance(x, (NumberProxy, Number)):
+            new_key.append(x)
+        if isinstance(x, (list, TensorLike)):
```
mruberry, line 607: My big questions here are:

- does PyTorch's getitem handle those cases correctly? If not, should happen? Should we try to mimic PyTorch's behavior, or mimic NumPy's behavior?
- what combined indexing case or cases does this PR want to support?

See https://github.com/Lightning-AI/lightning-thunder/blob/6dfe7e939a19d1ef5ab259de8709a79f0104fa42/thunder/tests/opinfos.py#L3502 for some existing discrepancies between PyTorch and NumPy.

I think we previously decided (maybe @nikitaved remembers better) that we should implement NumPy's indexing behavior, and we can leave being consistent with PyTorch (or not) for future work.

The NumPy documentation describes two cases of combined indexing here https://numpy.org/doc/stable/user/basics.indexing.html#combining-advanced-and-basic-indexing:

- the advanced indices are separated by a slice, ellipsis, or newaxis
- the advanced indices are all next to each other

Is this PR addressing one or both of those cases?

With basic indexing, should we require that the advanced indexing elements of a key be replaced by ":" before calling the basic indexing operation? That might be more explicit if someone is trying to call it directly, and would avoid this check in basic indexing.

thunder/clang/__init__.py
```diff
@@ -693,12 +705,16 @@ def _convert_none(x):
             # NOTE: this is redundant with the ValueError exception above
             raise ValueError(f"Found unexpected value {x} in key={key}")
 
+    # check if we can skip slice_prim
+    if all([x == 0 for x in start_indices]) and all([x == l for x, l in zip(end_indices, a.shape)]):
```
mruberry, line 709: This is a performance optimization, right? If so, let's make the comment state that a little more clearly.

What if the strides are not equal to 1? Don't we still need take the slice?

thunder/clang/__init__.py
mruberry: I'm a little confused by the cases here -- maybe we could go through them in a comment? It seems like there should be some way to simplify the case logic.

Also, for this check (as currently written) isn't it any non-tuple sequence that we'd like to identify?

thunder/clang/__init__.py
```diff
@@ -715,8 +731,49 @@ def _advanced_indexing(a: TensorLike, /, key) -> TensorLike:
...
+        if isinstance(key, list):
+            key = _to_tensorproxies(key, a.device)
+            if isinstance(key, list):
+                # handle list of tensors
+                for ii, i in enumerate(key):
+                    advanced_keys.append((ii, i))
+        else:
+            key_ = []
+            for key_idx, x in enumerate(key):
+                if isinstance(x, list):
```
mruberry, line 759: Wouldn't this check be any for any Sequence, not just list?

thunder/tests/opinfos.py
```diff
@@ -3595,6 +3595,51 @@ def make_idx(dim_length: int, indices: int):
     idx0 = make_idx(7, 9)
     yield SampleInput(a, (Ellipsis, idx0))
 
+    # list indexing
```
mruberry, line 3598: let's also add a couple cases to the basic indexing cases where:

- all slices are from 0 to the length of their dimension (and so trigger the performance optimization) and have stride of 1
- all slices are from 0 to the length of their dimension (and so trigger the performance optimization) and have strides != 1

thunder/clang/__init__.py
k223kim: To answer  to best of my knowledge, here are my takes:
-  Thunder currently handles both slice and newaxis as basic indices. (see below)
https://github.com/Lightning-AI/lightning-thunder/blob/7d62ae1c7da8fb57e09da3e95ced734e34480400/thunder/clang/__init__.py#L598
https://github.com/Lightning-AI/lightning-thunder/blob/7d62ae1c7da8fb57e09da3e95ced734e34480400/thunder/clang/__init__.py#L649
- Ellipsis are handled in both advanced and basic indexing
- This PR does not necessarily assume that the advanced indices are all next to each other. That is why I had the `transpose` operation. (this allows us to rearrange so that basic indices come first and advanced indices come after; which leads to 'advanced indices are all next to each other')

thunder/tests/opinfos.py
k223kim: Let me add this :) thank you!

thunder/clang/__init__.py
k223kim: What other sequences do we want to cover in advanced indexing? (Slices are covered in basic indexing)

thunder/clang/__init__.py
k223kim: You are right. I missed that. Thank you for pointing it out. Let me update this!

thunder/clang/__init__.py
k223kim: My intention here was to convert any "advanced indices" with slice(None,None,None) so that basic indexing could "skip" it (i.e. no effect to basic indexing). To update that advanced key, I needed a variable to keep those basic keys + converted advanced keys.

thunder/clang/__init__.py
k223kim: I would like to understand what other Sequence we want to cover in advanced indexing. Would you be able to explain that? Or provide me an example?

t-vi:
The failing test `test_cse_rematerialization` seems tricky because we optimize out one operation and apparently the names are not that stable.
I would suggest updating as follows:

```diff
diff --git a/thunder/tests/test_nvfuser.py b/thunder/tests/test_nvfuser.py
index 7fcf960d..f2437ad9 100644
--- a/thunder/tests/test_nvfuser.py
+++ b/thunder/tests/test_nvfuser.py
@@ -365,10 +365,25 @@ def test_cse_rematerialization(executor, device, _):
     # fusion groups 1 and 6 correspond with the apply_rotary_emb function
     # Nvfuser with recomputation should use precomputed cos and sin values.
     assert len(fusion_bsyms[1].args) == len(fusion_bsyms[6].args)
-    assert fusion_bsyms[1].subsymbols[0].output.name == "freqs_cos"
-    assert fusion_bsyms[1].subsymbols[1].output.name == "freqs_sin"
-    assert fusion_bsyms[6].subsymbols[0].output.name == "freqs_cos"
-    assert fusion_bsyms[6].subsymbols[1].output.name == "freqs_sin"
+
+    # Below, we check that freqs_sin and tos1 (the latter being freqs_cos) are used
+    # in the same operation in both fusions.
+
+    (fusion1_freqs_sin_arg,) = [a for a in fusion_bsyms[1].args if a.name == "freqs_sin"]
+    (fusion1_tos1_arg,) = [a for a in fusion_bsyms[1].args if a.name == "tos1"]
+    (fusion6_freqs_sin_arg,) = [a for a in fusion_bsyms[6].args if a.name == "freqs_sin"]
+    (fusion6_tos1_arg,) = [a for a in fusion_bsyms[6].args if a.name == "tos1"]
+
+    (fusion1_freqs_sin_user,) = [s for s in fusion_bsyms[1].subsymbols if s.args[0] is fusion1_freqs_sin_arg]
+    (fusion6_freqs_sin_user,) = [s for s in fusion_bsyms[6].subsymbols if s.args[0] is fusion6_freqs_sin_arg]
+
+    assert fusion1_freqs_sin_user.sym is fusion6_freqs_sin_user.sym
+    assert fusion1_freqs_sin_user.args[1:] == fusion6_freqs_sin_user.args[1:]
+    (fusion1_tos1_user,) = [s for s in fusion_bsyms[1].subsymbols if s.args[0] is fusion1_tos1_arg]
+    (fusion6_tos1_user,) = [s for s in fusion_bsyms[6].subsymbols if s.args[0] is fusion1_tos1_arg]
+
+    assert fusion1_tos1_user.sym is fusion6_tos1_user.sym
+    assert fusion1_tos1_user.args[1:] == fusion6_tos1_user.args[1:]
 
 
 # Tests that two separated nvFuser regions can be merged when they don't depend
```

thunder/clang/__init__.py
nikitaved: NumPy is the definition of (advanced)indexing (https://numpy.org/doc/stable/user/basics.indexing.html) without any surprises. PyTorch tries to comply with NumPy. If not today, it will at some point. Sticking to NumPy is robust and future-proof.

thunder/clang/__init__.py
k223kim: Let me update this in a follow up PR that can handle more advanced indexing!

t-vi:
Great work @k223kim . Thank you for your insights @mruberry @lantiga . At a week of continuous improvements, let's merge this checkpoint and build from there in a follow-up PR.
If we want, we can also take a few minutes in the technical meeting to day to chat about the next steps for the advanced indexing.

thunder/clang/__init__.py
mruberry: Following the NumPy documentation, I think any Sequence is something that we should attempt to convert to a tensor?

Issue #701, opened 2024-07-03
[TransformerEngine] Support `backward(retain_graph=True)`


kshitij12345:
```python
from thunder.benchmarks.targets import LitGPTConfig, LitGPTBenchmark, backward_only
import torch
import thunder
from thunder.executors.transformer_engineex import transformer_engine_ex

cfg: LitGPTConfig = LitGPTConfig.from_name("Llama-2-7b-hf")
cfg.n_layer = 3
b = LitGPTBenchmark(cfg, batchdims=(2,), device="cuda:0", dtype=torch.bfloat16, requires_grad=True)

args, kwargs = b.make_batch()
fn = thunder.jit(b.fn(), executors=[transformer_engine_ex,])

# backward_only creates a graph and calls `torch.autograd.backward` with `retain_graph=True`.
backward_fn, backward_setup = backward_only(fn, *args, **kwargs)
backward_args = backward_setup()
backward_fn(*backward_args)

# Second usage fails with 
# File "/home/kkalambarkar/lightning-thunder/thunder/executors/transformer_engineex.py", line 352, in _te_functional_linear_backward_impl
#     with enable_grad(ctx.saved_tensors[2]):
# IndexError: tuple index out of range
backward_fn(*backward_args)

```

cc: @IvanYashchuk

kshitij12345:
One of the reason for the issue could that we have our own version of Context which may probably behaves differently when backward is called with `retain_grad`.

t-vi:
I think this may be the famous clearing of collections...

IvanYashchuk:
PyTorch requires all tensor data to be saved with `ctx.saved_for_backward(*tensors)` when it's not done objects can be deleted preemptively.

kshitij12345:
Even if we fix this, TransformerEngine itself has problem with running the backward multiple times with `retain_graph=True`.

```python
import torch
from transformer_engine.pytorch import Linear as TELinear, fp8_autocast

m = TELinear(16, 16)
x = torch.randn(16, 16, device='cuda')

with fp8_autocast(True):
    o = m(x).sum()

o.backward(retain_graph=True)

# this fails with
# AssertionError: FP8 execution requires 2D input matrices with height divisible by 8 and width divisible by 16, but got tensor with dims=[0]
# looks like TELinear.backward mutates the context object such that it is not reusable.
o.backward()

```

IvanYashchuk:
[...]

PR #713, opened 2024-07-04
Group saved for backward tensors produced by TransfomerEngine together with all other saved for backward tensors

IvanYashchuk:
TLDR: `Tensor.backward(retain_graph=True)` is a public PyTorch feature that may be used in a lot of code, we should support it. Thunder without TransformerEngine works, but with TransformerEngine there was a bug in Thunder code that is fixed in this PR. There's still a bug on the TransformerEngine's side (https://github.com/NVIDIA/TransformerEngine/issues/990).

This PR also bumps the required minimal version of TransformerEngine to 1.8+.

Previously we were hiding "saved_tensors" inside an opaque context object that TransformerEngine populates inside `_Linear.forward` and we were relying on passing a dict representation of this object to be passed from `ThunderFunction.forward` to `ThunderFunction.backward` instead of dealing with it the same way as we deal with all other tensors to be saved for backward here
https://github.com/Lightning-AI/lightning-thunder/blob/e75ddc1353178d5d74fcc5d29154efd8f77605a5/thunder/executors/torch_autograd.py#L73
A dict representation was needed to let Thunder clear it during backward and release the memory occupied by saved by TE tensors. However, we already have a mechanism to clear saved tensors and also that works correctly with PyTorch's `retain_graph=True`, so we should be using it.
For Thunder's general "saved tensor" list to include also TransformerEngine's saved tensors we need to modify the meta functions to model it explicitly like we do in all other forward meta functions.

The code snippet from https://github.com/Lightning-AI/lightning-thunder/issues/701 still doesn't work, but the problem is not in the Thunder code.

Here's how the forward and backward traces changed:
```py
import torch
import thunder
from thunder.executors.transformer_engineex import transformer_engine_ex

def foo(x, w):
    return torch.nn.functional.linear(x, w)
x = torch.randn(16, 16, requires_grad=True, device='cuda')
w = torch.randn(16, 16, requires_grad=True, device='cuda')

fn = thunder.jit(foo, executors=[transformer_engine_ex,])
fn(x, w)

print(thunder.last_traces(fn)[-1])
print(thunder.last_backward_traces(fn)[-1])
```
```diff
def augmented_forward_fn(x, w):
  # x: "cuda:0 f32[16, 16]"
  # w: "cuda:0 f32[16, 16]"
- (t8, ctx_te_1) = te_linear_0(x, w, None)
+  (t8, (t3, t4, t5, _, t6, t7, _), ctx_te_1) = te_linear_0(x, w, None)
- return {'output': t8, 'flat_args': [x, w], 'flat_output': (t8,)}, ((), (ctx_te_1,))
+  return {'output': t8, 'flat_args': [x, w], 'flat_output': (t8,)}, ((t3, t4, t5, t6, t7), (ctx_te_1,))
```
```diff
def backward_fn(saved_for_backward, cotangents):
  # saved_for_backward: "Collection"
  # cotangents: "Collection"
-  _, C1, = saved_for_backward
+  C0, C1, = saved_for_backward
  clear_mutable_collection(saved_for_backward)
  del saved_for_backward
  t9, = cotangents
  clear_mutable_collection(cotangents)
  del cotangents
-
+  t3, t4, t5, t6, t7, = C0
+  clear_mutable_collection(C0)
+  del C0
  ctx_te_3, = C1
  clear_mutable_collection(C1)
  del C1
- (t12, t13, _) = te_functional_linear_backward(t9, (16, 16), (16, 16), None, ctx_te_3)
+  (t12, t13, _) = te_functional_linear_backward((16, 16), (16, 16), None, ctx_te_3, (t3, t4, t5, None, t6, t7, None), t9)
- clear_mutable_collection(ctx_te_1)
+  del ctx_te_3, t3, t4, t5, t6, t7, t9
  te_sync_fp8_meta_bwd()
  return (t12, t13)
```


kshitij12345:
On TE v1.8 ('1.8.0+37280ec') - I am seeing failure when running - `pytest thunder/tests/test_transformer_engine_executor.py`. Looks like the number of saved tensors has changed. Since, `_Linear` is a non-public API, it's inputs and outputs change frequently. The [PR](https://github.com/NVIDIA/TransformerEngine/pull/707) adding stable functional interface should be merged soon (it is currently in review). Till then we can either add switches based on version and return output accordingly or maybe wait till the mentioned PR is merged and then use the stable functional API.

```python
    def augmented_forward_fn(x, w):
      # x: "cuda:0 f32[16, 16]"
      # w: "cuda:0 f32[16, 16]"
>     (t8, (t3, t4, t5, _, t6, t7, _), ctx_te_19) = te_linear_18(x, w, None)
E     ValueError: not enough values to unpack (expected 7, got 6)
```

IvanYashchuk:
> On TE v1.8 ('1.8.0+37280ec') - I am seeing failure when running - `pytest thunder/tests/test_transformer_engine_executor.py`. Looks like the number of saved tensors has changed. Since, `_Linear` is a non-public API, it's inputs, outputs (and other details) change frequently. The [PR](https://github.com/NVIDIA/TransformerEngine/pull/707) adding stable functional interface should be merged soon (it is currently in review). Till then we can either add switches based on version and return output accordingly or maybe wait till the mentioned PR is merged and then use the stable functional API.
> 
> Failure -
> 
> ```python
>     def augmented_forward_fn(x, w):
>       # x: "cuda:0 f32[16, 16]"
>       # w: "cuda:0 f32[16, 16]"
> >     (t8, (t3, t4, t5, _, t6, t7, _), ctx_te_19) = te_linear_18(x, w, None)
> E     ValueError: not enough values to unpack (expected 7, got 6)
> ```

Thank you, we already have a number of these switches, and I'll add another one.

IvanYashchuk:
Instead of adding a switch I have increased the minimal required version of TransformerEngine to be 1.8 and now the saved tensor in the meta functions models the saving of 6 objects as in TE 1.8 and the current TE main.

kshitij12345 (approved):
LGTM, have verified that the tests pass on v1.8 (stable), v1.9 and v1.10 (latest main). Thanks @IvanYashchuk 

cc: @t-vi

IvanYashchuk:
Awesome! Thank you for checking!

t-vi (approved):
Thank you @IvanYashchuk  @kshitij12345

PR #718, opened 2024-07-05
`iter(TensorProxy)` implementation

nikitaved:
Could that fix https://github.com/Lightning-AI/lightning-thunder/issues/717 based on https://github.com/Lightning-AI/lightning-thunder/issues/717#issuecomment-2210682835?

Not yet sure whether the whole approach is actually legit...

### DIFF
--- a/thunder/core/proxies.py
+++ b/thunder/core/proxies.py
@@ -1316,6 +1316,15 @@ class TensorProxy(Proxy, TensorProxyInterface):
 
         return method_or_value
 
+    def __iter__(self):
+        # NOTE: this implementation is equivalent to torch.Tensor.__iter__
+
+        if self.ndim == 0:
+            raise TypeError("iteration over a 0-dim tensor")
+
+        unbound_tuple = self.unbind(0)
+        return iter(unbound_tuple)
+
     #
     # Default attribute
     #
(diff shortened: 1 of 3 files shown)


t-vi:
I know I suggested lookaside, but as @kshitij12345 observed, it might be even easier to just add `TensorProxy.__iter__` directly (it will be run in the interpreter, so does the same as lookaside + impl that does it).
Maybe add a comment that this is typically run inside the Thunder interpreter.

t-vi (approved):
Thank you @nikitaved (and apologies for suggesting the more complicated initially).
Also thank you @kshitij12345 for the analysis and repro.

thunder/core/proxies.py
```diff
@@ -1316,6 +1316,17 @@ def __getattr__(self, attr: str, /):
...
+    def __iter__(self):
+        # NOTE: this implementation is equivalent to torch.Tensor.__iter__
+
+        if self.ndim == 0:
+            raise TypeError("iteration over a 0-dim tensor")
+
+        import thunder.torch
+
+        unbound_tuple = thunder.torch.unbind(self)
+        return iter(unbound_tuple)
```
kshitij12345, line 1328: I think we can also do `iter(self.unbind()` as `unbind` is also a registered method.

https://github.com/Lightning-AI/lightning-thunder/blob/0d804443b2d499d897d0e5cb3cc96d508ec2454a/thunder/torch/__init__.py#L1256-L1262

nikitaved: Hmm, did not know that things changed with the API :)

nikitaved: Applied! Thank you!

PR #721, opened 2024-07-05
Exception types for interpolate()

tfogal:
Follow-ups from #679:

* Use NotImplementedError for the exception type.
* Add some negative tests for this to ensure something runs in CI.


t-vi (approved):
Thank you @tfogal

PR #722, opened 2024-07-06
Reduce cyclomatic complexity by 1

crcrpar:


thunder/core/transform_common.py
```diff
@@ -467,16 +476,15 @@ def functionalize_inplace_ops(
     For example, :func:`thunder.torch.add_` is represented as a :class:`thunder.core.symbol.BoundSymbol`
     whose `subsymbols` are :func:`thunder.torch.add` and :func:`thunder.core.prims.copy_`. This function
     replaces it with a :class:`~thunder.core.symbol.BoundSymbol` of :func:`~thunder.torch.add`.
+
+    .. note::
+
+        Currently if an arg of ``computation_trace`` is passed to multiple in-place ops,
```
t-vi, line 482: I wonder how hard it would be to make this an error instead of a note...

t-vi (approved):
Seems good to have, thank you @crcrpar .
For a follow-up: How hard/bad would it be to make the wrong update of an arg an error?

thunder/core/transform_common.py
crcrpar: I bet you've noticed this one but #723 is trying to support such cases :)

t-vi: Right. :)

Issue #720, opened 2024-07-05
wrong result with in-place operations.


jjsjann123:
### To Reproduce

The program below fails the assert on updating `a`.

```
import thunder
 
def foo(a):
    a.relu_()
    a.add_(1.0)
    c = a + 1
    return c
    
import torch
    
dtype = torch.float32
    
size0 = 4
size1 = 8
size2 = 6
a = torch.randn(size0, size1, size2, device="cuda").to(dtype=dtype)
    
a_ref = a.clone().detach()
out_ref = foo(a_ref)
    
jfoo = thunder.jit(foo)
 
out = jfoo(a)
assert out.allclose(out_ref)
assert a.allclose(a_ref)
```

Gives a computational trace as:
```
import torch
from thunder.executors.torchex import no_autocast

@torch.no_grad()
@no_autocast
def computation(a):
  # a: "cuda:0 f32[4, 8, 6]"
  [result] = nvFusion0(a)
    # t0 = prims.gt(a, 0.0)  # t0: "cuda:0 b8[4, 8, 6]"
    # t1 = prims.where(t0, a, 0.0)  # t1: "cuda:0 f32[4, 8, 6]"
    # t3 = prims.add(t1, 1.0)  # t3: "cuda:0 f32[4, 8, 6]"
    # result = prims.add(t3, 1.0)  # result: "cuda:0 f32[4, 8, 6]"
    # prims.copy_(t1, a)
  del a
  return result
```

`a` is copied from `t1` instead of `t3`. Looks like we just missed to update the source buffer from inplace update. cc'ing @crcrpar

jjsjann123:
closing since this is a duplicate for @crcrpar 's comment here: https://github.com/Lightning-AI/lightning-thunder/issues/657#issuecomment-2208655632

Issue #657, opened 2024-06-26
Desired inplace idioms (add yours here)


t-vi:
We do support some inplace thanks to @crcrpar 's great patches. This has enable running additional models already, but is limited to a few cases of particular interest.

However, we would want to see and prioritize other important inplace ideoms, so we would appreciate if you chimed in with needs you or your favourite model has.

Please do try to post
- a minimal code snippet,
- if you are aware of a particular model needing this, please also add this.

We don't want to support all corner cases of inplace (eg "does reshape produce a view or not") but we do care a lot about enabling users and models. Thank you!

crcrpar:
`thunder.jit` the following function with nvfuerex fails with the message below.
By moving the copy for `a += b` to the end of a trace and replacing `a += b` with `t0 = a + b`, I expect it to work.
```python
def f(a: torch.Tensor, b: torch.Tensor) -> tuple[torch.Tensor, torch.Tensor]:
    a += b
    c = torch.exp(a)
    d = torch.tanh(b)

    e = c.view(-1)
    e.add_(d.flatten())

    d.div_(a)
    return c, d, e / 2.0
```

```
Traceback (most recent call last):
  File "/home/mkozuki/ghq/github.com/Lightning-AI/lightning-thunder/snippet.py", line 55, in <module>
    main()
  File "/home/mkozuki/ghq/github.com/Lightning-AI/lightning-thunder/snippet.py", line 24, in main
    c, d, e = jit_f(a, b)
  File "/home/mkozuki/ghq/github.com/Lightning-AI/lightning-thunder/thunder/__init__.py", line 676, in fn_
    cache_entry, inps, pro_to_epi = get_computation_and_inputs(*args, **kwargs)
  File "/home/mkozuki/ghq/github.com/Lightning-AI/lightning-thunder/thunder/__init__.py", line 223, in cache_info_wrapper
    res = fn(*args, **kwargs)
  File "/home/mkozuki/ghq/github.com/Lightning-AI/lightning-thunder/thunder/__init__.py", line 615, in get_computation_and_inputs
    thunder.core.transform_common._inplace_copy_sanity_check(computation_trc)
  File "/home/mkozuki/ghq/github.com/Lightning-AI/lightning-thunder/thunder/core/transform_common.py", line 86, in _inplace_copy_sanity_check
    check(copy_to_out, "output")
[...]

PR #723, opened 2024-07-06
Support func args passed to multiple in-place ops

crcrpar:
rel:
- https://github.com/Lightning-AI/lightning-thunder/issues/657#issuecomment-2208655632
- #720 

cc @jjsjann123

### DIFF
--- a/thunder/core/transform_common.py
+++ b/thunder/core/transform_common.py
@@ -463,6 +463,85 @@ def is_functionalizable(bsym: BoundSymbol) -> bool:
     )
 
 
+def collect_required_copy_bsyms_for_args(
+    trace: Trace,
+    removed_copy_bsyms: list[BoundSymbol],
+) -> tuple[list[BoundSymbol], dict[VariableInterface, TensorProxy]]:
+    from thunder.core.utils import find_producer_symbols
+    from thunder.torch import _syms_returning_views
+
+    flat_args = tuple(
+        filter(
+            lambda p: isinstance(p, TensorProxy),
+            tree_flatten((trace.args, trace.kwargs))[0],
+        )
+    )
+    trace_args_set = ProxyDict()
+    for a in flat_args:
+        trace_args_set[a] = a
+
+    copy_dst_to_copy_bsym: dict[VariableInterface, BoundSymbol] = {}
+    required_copy_bsyms: list[BoundSymbol] = []
+    unused_copy_bsyms: list[BoundSymbol] = []
+    inplace_args: list[TensorProxy] = []
+    copy_destinations: list[TensorProxy] = []
+    for idx, bsym in enumerate(removed_copy_bsyms):
+        if (copy_to := bsym.flat_proxy_args[1]) in trace_args_set:
+            required_copy_bsyms.append(bsym)
+            copy_dst_to_copy_bsym[variableify(copy_to)] = bsym
+            inplace_args.append(copy_to)
+        else:
+            unused_copy_bsyms.append(bsym)
+
+        copy_destinations.append(copy_to)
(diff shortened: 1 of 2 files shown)


thunder/core/transform_common.py
```diff
@@ -463,6 +463,119 @@ def is_functionalizable(bsym: BoundSymbol) -> bool:
...
+
+        first_bsym: BoundSymbol = s_bsyms[0]
+        if first_bsym.sym in _syms_returning_views:
+            if all(
+                (bsym.sym.tags and prims.OpTags.IN_PLACE in bsym.sym.tags) or bsym.sym in _syms_returning_views
+                for bsym in s_bsyms[1:]
+            ):
+                new_copy_dst: TensorProxy = list(filter(lambda p: isinstance(p, TensorProxy), first_bsym.flat_args))[0]
+                check(new_copy_dst in trace_args_set, lambda: f"{new_copy_dst=} not in {trace_args_set}")
+                copy_src: TensorProxy = bsym.flat_proxy_args[0]
```
crcrpar, line 524: this does read maladroit to me because I do expect this check to does the equivalent or even the same as `elif` condition, except `new_copy_dst` check.
I however somehow fail to make it happen.

Also, the check should use `view_args` with its variables are `variableify`'ed.
```python
        if first_bsym.sym in _syms_returning_views:
            if all(
                (bsym.sym.tags and prims.OpTags.IN_PLACE in bsym.sym.tags) or bsym.sym in _syms_returning_views
                for bsym in s_bsyms[1:]
            ):
                new_copy_dst: TensorProxy = list(filter(lambda p: isinstance(p, TensorProxy), first_bsym.flat_args))[0]
                check(new_copy_dst in view_args, lambda: f"{new_copy_dst=} not in {view_args}")
                copy_src: TensorProxy = bsym.flat_proxy_args[0]
```

t-vi:
Here it is. :)

thunder/core/transform_common.py
```diff
@@ -467,16 +589,15 @@ def functionalize_inplace_ops(
     For example, :func:`thunder.torch.add_` is represented as a :class:`thunder.core.symbol.BoundSymbol`
     whose `subsymbols` are :func:`thunder.torch.add` and :func:`thunder.core.prims.copy_`. This function
     replaces it with a :class:`~thunder.core.symbol.BoundSymbol` of :func:`~thunder.torch.add`.
+
+    .. note::
+
+        Currently if an arg of ``computation_trace`` is passed to multiple in-place ops,
+        the return values would be correct but the arg itself would not be correctly updated.
+        See https://github.com/Lightning-AI/lightning-thunder/issues/657#issuecomment-2208655632.
```
crcrpar, line 597:
```suggestion
```

crcrpar:
dunno why this gets conflicts but, anyway

thunder/core/transform_common.py
t-vi: Maybe we could add a comment so people looking won't be confused (I'm always confused, but more so when the branches of an if look too similar.)

t-vi (approved):
Looks good to me, I wonder if we want a note about aliases as args?

```
 def f(a, b):
        return a.exp_().sin_() + b.exp_().sin_()

a = torch.randn(5, 5)
f(a, a)
```

crcrpar:
> I wonder if we want a note about aliases as args?
> 
> ```
>  def f(a, b):
>         return a.exp_().sin_() + b.exp_().sin_()
> 
> a = torch.randn(5, 5)
> f(a, a)
> ```

What's the expected behavior of both pytorch eager and thunder for this `f`?

PR #724, opened 2024-07-08
sort: a prim with grad support

nikitaved:
This enables sort/stable sort in Thunder.

This is also a required piece to be able to remove data-dependent `where` in NeMo (cc @tfogal) .


t-vi (approved):
Thank you @nikitaved

PR #730, opened 2024-07-08
Fix a ddp test for #191.

wujingyue:
See https://github.com/Lightning-AI/lightning-thunder/issues/191#issuecomment-2163742283 for the context. 

This is separated out from #731 in order to keep PRs small.


t-vi (approved):
Thank you @wujingyue .

Seems good. In general, I don't think we need to split <100 line PRs more unless there is a particular reason...

PR #744, opened 2024-07-10
Remove some nanoGPT cases from benchmarks/targets.py, they are replaced by LitGPT

IvanYashchuk:
At the beginning of the project, we focused on measuring improvements for the nanoGPT model implementation. We switched to LitGPT long ago and it's time to remove some of the benchmark cases.

The batch size for `test_llama2_mlp_7b` and `test_llama2_mlp_7b` was decreased to 2 because it's a common max batch size that can be used in real training run.

`test_open_llama_7b` was removed and it was never working because the batch size is too large. There should be a parametrized benchmark for all litgpt configs instead of separate benchmark functions per config.

nanoGPT Gelu benchmark was upgraded to use LitGPT configurations instead.

cc @crcrpar

### DIFF
--- a/thunder/benchmarks/targets.py
+++ b/thunder/benchmarks/targets.py
@@ -306,7 +306,7 @@ sdpa_executors_ids = (
 
 
 # Sample command to run this benchmark:
-# pytest thunder/benchmarks/targets.py -k "test_litgpt_sdpa_grad" --benchmark-group-by='param:config,param:bs' --benchmark-columns='min,max,mean,stddev,median'
+# pytest thunder/benchmarks/targets.py -k "test_litgpt_sdpa" --benchmark-group-by='param:config,param:bs,param:compute_type'
 @pytest.mark.parametrize(
     "executor,",
     sdpa_executors,
--- a/thunder/benchmarks/targets.py
+++ b/thunder/benchmarks/targets.py
@@ -271,6 +271,8 @@ def test_nanogpt_cross_entropy(benchmark, executor: None | Callable, compute_typ
     benchmark_for_compute_type(compute_type, benchmark, fn, args, kwargs)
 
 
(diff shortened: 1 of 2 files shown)


crcrpar (approved):

thunder/benchmarks/targets.py
```diff
@@ -168,20 +165,63 @@ def interpreter_fwd(module: Callable):
...
-        config="gpt2-xl", device="cuda:0", dtype=thunder.bfloat16, requires_grad=is_requires_grad(compute_type)
+@pytest.mark.parametrize(
+    "config,",
+    get_configs_for_gelu(),
+)
+def test_litgpt_gelu(benchmark, executor: Callable, bs: int, compute_type: ComputeType, config: str):
+    gelu_bench: Benchmark = LitGPTGeluBenchmark(
+        config=config,
+        batchdims=(bs,),
+        device="cuda:0",
```
riccardofelluga, line 222: LGTM! I just have a question here, is there a particular reason why we specify the device index here instead of leaving it to the default device?

riccardofelluga (approved):

t-vi (approved):
Thank you @IvanYashchuk @crcrpar @riccardofelluga 

@IvanYashchuk do you want to address @riccardofelluga comment before merging or leave as cuda:0 for now?

thunder/benchmarks/targets.py
IvanYashchuk: No, there's no special reason. It was like that before this PR, see removed line 184 https://github.com/Lightning-AI/lightning-thunder/pull/744/files#diff-118f5dfdb54f6ed69a3249fda1ba903ec309955fdea5186d1de7bc0169f929c2L184

IvanYashchuk:
Leave as cuda:0.

PR #746, opened 2024-07-10
Replace `time.time_ns()` with `time.perf_counter_ns()` where possible

### LINKED ISSUE #745: Use `time.perf_counter_ns()` instead of `time.time_ns()`
Thank you for this great project!

I feel that there's some problem with time measurement.  `time.time_ns()` is used to measure processing time such as [thunder/executors/passes.py#L222](https://github.com/Lightning-AI/lightning-thunder/blob/57d48524f5aafc56066f5fe64af9aa4ae02236c3/thunder/executors/passes.py#L222). However, `time.time_ns()` isn't monotonic, meaning that it's possible that a later reading returns a lower number. The resolution isn't always the most accurate.

I would suggest switching `time.time_ns()` over to [`time.perf_counter_ns()`](https://docs.python.org/3/library/time.html#time.perf_counter) as it's usually used for performance measurement.

mshr-h:
Fix #745

### DIFF
--- a/thunder/__init__.py
+++ b/thunder/__init__.py
@@ -408,7 +408,7 @@ def jit(
         # TODO RC1 Add module and function checks to prologue (make it a compile option)
 
         # Checks cache
-        cs.last_trace_cache_start = time.time_ns()
+        cs.last_trace_cache_start = time.perf_counter_ns()
         if (cd.cache_option is CACHE_OPTIONS.CONSTANT_VALUES) or (cd.cache_option is CACHE_OPTIONS.SYMBOLIC_VALUES):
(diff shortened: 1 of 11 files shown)


t-vi:
@mshr-h thank you for changing this. The PR is currently labeled draft/PR, but from my pov we might merge quickly as it spans a number of files. WDYT?

mshr-h:
@t-vi Thank you for your comment. I made it as wip because I was waiting for the CI pipeline to pass. Marked as ready for review.

t-vi (approved):
Thank you @mshr-h

PR #749, opened 2024-07-10
[in-place] Improve how to collect necessary `prims.copy_` to function args/kwargs

crcrpar:
## What does this PR do?

Improve how to collect bsyms of `copy_src` to the real `copy_dst`.

It turned out to be quite tricky to correctly set `stop_proxies` to
`thunder.core.utils.find_producer_symbols` especially when `copy_src` is
the result of some operation whose operands include at least two of
func's args.

This PR recursively traverse a trace / bound symbols from `copy_src` until (a) op returns a new tensor or (b) operand is function's args/kwargs.
If (a), the `copy_src` would not share the storage of function's args/kwargs. If (b) the `copy_src` is highly likely to share the storage with one of function's args/kwargs.

Also, this PR adds bsym update to avoid unwanted return values swap such as function's args is returned but it's replaced due to the functionalization's swap_map.

### DIFF
--- a/thunder/core/transform_common.py
+++ b/thunder/core/transform_common.py
@@ -487,16 +487,14 @@ def _get_prod_bsym_with_arg(
     first_tensor = _get_first_tensor_arg(prod_bsym)
     if first_tensor is None:
         return None
-    if first_tensor in trace_args:
-        if inplace_or_view(prod_bsym):
+
+    if inplace_or_view(prod_bsym):
+        if first_tensor in trace_args:
             return prod_bsym
         else:
-            return None
-    else:
-        if inplace_or_view(prod_bsym):
             return _get_prod_bsym_with_arg(first_tensor, producer_map, trace_args)
-        else:
-            return None
+    else:
+        return None
 
 
 def collect_required_copy_bsyms_for_args(
--- a/thunder/core/transform_common.py
+++ b/thunder/core/transform_common.py
@@ -584,12 +584,82 @@ def functionalize_inplace_ops(
 
     In thunder, an in-place is an out-of-place or functional op followed by :func:`~thunder.core.prims.copy_`.
     This function replaces such in-place ops with out-of-place ops.
-    If any of an in-place op's arguments are ``computation_trace.args`` or ``computation_trace.kwargs``,
-    the copy is moved to the end of the trace.
 
-    For example, :func:`thunder.torch.add_` is represented as a :class:`thunder.core.symbol.BoundSymbol`
-    whose `subsymbols` are :func:`thunder.torch.add` and :func:`thunder.core.prims.copy_`. This function
-    replaces it with a :class:`~thunder.core.symbol.BoundSymbol` of :func:`~thunder.torch.add`.
+    This function returns an empty list if no in-place ops are found in ``computation_trace``.
+    If any are found, functionalization is done in two steps. The first step is to canonicalize the trace
+    by making sure that any operands of in-place ops have multiple consumers. The second step is to either
+    remove `prims.copy_` from the trace before collecting required ones. The required `prims.copy_`s are ones
+    whose destination is ``computation_trace``'s args/kwargs.
+
+    For the following ``f``, this function creates two traces below.
+    Let's take a look at what this functionalization generates for the following ``f``.
+
+    .. code-block:: python
+        :name: input-func
+
+        def f(x):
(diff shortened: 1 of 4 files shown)


thunder/core/transform_common.py
```diff
@@ -583,12 +584,82 @@ def functionalize_inplace_ops(
 
     In thunder, an in-place is an out-of-place or functional op followed by :func:`~thunder.core.prims.copy_`.
     This function replaces such in-place ops with out-of-place ops.
-    If any of an in-place op's arguments are ``computation_trace.args`` or ``computation_trace.kwargs``,
-    the copy is moved to the end of the trace.
 
-    For example, :func:`thunder.torch.add_` is represented as a :class:`thunder.core.symbol.BoundSymbol`
-    whose `subsymbols` are :func:`thunder.torch.add` and :func:`thunder.core.prims.copy_`. This function
-    replaces it with a :class:`~thunder.core.symbol.BoundSymbol` of :func:`~thunder.torch.add`.
+    This function returns an empty list if no in-place ops are found in ``computation_trace``.
```
crcrpar, line 588: @IvanYashchuk could you take a look at this change?

t-vi:
Hi, just a quick note that I'm still wraping my head around this patch, it's not forgotten.

thunder/core/transform_common.py
```diff
@@ -463,133 +463,289 @@ def is_functionalizable(bsym: BoundSymbol) -> bool:
...
-    view_args = []
-    for a in filter(lambda a: variableify(a) not in copy_dst_to_copy_bsym, flat_args):
-        first_consumer = consumers[a][0]
-        if first_consumer.sym in _syms_returning_views:
-            view_args.append(a)
-    stop_proxies = tuple(inplace_args + view_args)
+        new_bsyms: list[BoundSymbol] = []
+        reshaped_copy_src = copy_src
+        with tracectx(trace):
+            if copy_src.shape != new_copy_dst.shape:
```
t-vi, line 535: So here, is my understanding correct that we adapt the shape to cope with .view in between but we don't need broadcasting because we don't do that yet (and because our `copy_` is not `torch.Tensor.copy_`)?

thunder/core/transform_common.py
```diff
@@ -463,133 +463,289 @@ def is_functionalizable(bsym: BoundSymbol) -> bool:
...
+    for var_t in new_dst_to_copy_bsym:
+        if var_t not in dst_to_copy_bsym:
+            _, bsyms = new_dst_to_copy_bsym[var_t]
+            required_copy_bsyms[trace_args_set[unvariableify(var_t)]] = bsyms
+
+    sorted_copy_bsyms: list[BoundSymbol] = []
+    for bsym_or_list in required_copy_bsyms:
+        if bsym_or_list is None:
+            continue
+        if isinstance(bsym_or_list, list):
```
t-vi, line 573: So this is/can be a list for the case when a reshape is necessary? Or are there cases when there are multiple `copy_` (I would expect not to). Could it make sense to make it always a list (and sometimes a one-element one)?

t-vi (approved):
Thank you @crcrpar . I've added a couple of comments to test my understanding, but LGTM.

thunder/core/transform_common.py
crcrpar: I think there're two reasons

- `torch.Tensor.copy_` is not covered in `thunder/torch/__init__.py`
- thunder.jit errors out if a view tensor of different numel from the original is the operand of an in-place op here: https://github.com/Lightning-AI/lightning-thunder/blob/e646e4fce78542410a77c8be7de38a659860b536/thunder/core/transform_common.py#L444-L451

thunder/core/transform_common.py
crcrpar: > So this is/can be a list for the case when a reshape is necessary?

Correct.

> Could it make sense to make it always a list (and sometimes a one-element one)?

I don't have a strong preference

crcrpar: 3197cf9e37203e257b91cad892605bc7c19e6c94 switched to list

thunder/core/transform_common.py
t-vi: Thank you!

thunder/core/transform_common.py
t-vi: Thank you!

PR #754, opened 2024-07-11
ggml quantization in thunder v0

t-vi:
Notebook demoing a ggml transform (without specialized kernels yet, but e.g. reducing memory consumption).

The top part is just the litgpt generation demo litgpt itself.

Note that I needed to add an eval hack for "meta" device in check same tensors in order to be able to trace through the model with meta weights.


lantiga:
@t-vi tests pass now, should we? : )

lantiga (approved):

PR #759, opened 2024-07-11
feat: convert to torch dtype for finfo

### LINKED ISSUE #457: TypeError with torch.finfo()
### 🐛 Bug

NeMo's Stable Diffusion uses HuggingFace Transformers that uses `torch.finfo` in one of its operations. When `thunder.jit` is used with Stable Diffusion, it gives an error:
```
TypeError (note: full exception trace is shown but execution is paused at: _run_module_as_main)
finfo(): argument 'type' (position 1) must be torch.dtype, not floating
```
although torch supports the `dtype` in `torch.finfo(dtype)` to be either `torch.float32` or just `float32`.

CC: @tfogal 

cc @tfogal

k223kim:
<details>

</details>

## What does this PR do?

Fixes #457. The following code works as expected:
```
import torch, thunder

def fn(x):
    return torch.finfo(x.dtype).min

a = torch.randn(5, 5)

torch_output = fn(a)
print(torch_output) # -3.4028234663852886e+38

jfn = thunder.jit(fn)

thunder_output = jfn(a)
print(thunder_output) # -3.4028234663852886e+38

torch.testing.assert_close(torch_output, thunder_output)
```

Although, I am not sure how I can add a test case within `tests` for this. Any advice would be appreciated!

## PR review

## Did you have fun?

### DIFF
--- a/thunder/core/interpreter.py
+++ b/thunder/core/interpreter.py
@@ -1613,12 +1613,6 @@ def check_self(obj, potential_method):
 
 
 def plausibly_wrapper_of(wrapper, value):
-    # convert dtype if needed
-    if isinstance(value, thunder.dtypes.dtype):
-        if thunder.dtypes.to_torch_dtype(value) == wrapper.value:
-            return True
-        else:
-            False
     if wrapper.value is value or wrapper.original_value is value:
         return True
     if callable(value) or True:
--- a/thunder/core/interpreter.py
+++ b/thunder/core/interpreter.py
@@ -3570,6 +3570,8 @@ def _call_handler(
     func_or_null = stack.pop_wrapped()
     if frame.call_shape_kwnames is not None:
         kwnames = frame.call_shape_kwnames
+        if len(kwnames) == 3:
+            print("")
         assert len(args) >= len(kwnames)
         kwargs = dict(zip(kwnames, args[-len(kwnames) :]))
         args = args[: -len(kwnames)]
@@ -6247,8 +6249,12 @@ def _setup_frame_and_run_python_function(
             p = p.replace(default=wrap_const(p.default))
         params.append(p)
     sig = sig.replace(parameters=params)
-    bound = sig.bind(*args, **kwargs)
-    bound.apply_defaults()
+    try:
+        bound = sig.bind(*args, **kwargs)
+        bound.apply_defaults()
+    except:
+        print("")
+        bound = None
 
     locals_dict: dict[str, Any] = dict(bound.arguments)
 
--- a/thunder/core/interpreter.py
+++ b/thunder/core/interpreter.py
@@ -6249,12 +6249,8 @@ def _setup_frame_and_run_python_function(
             p = p.replace(default=wrap_const(p.default))
         params.append(p)
     sig = sig.replace(parameters=params)
-    try:
-        bound = sig.bind(*args, **kwargs)
-        bound.apply_defaults()
-    except:
-        print("")
-        bound = None
+    bound = sig.bind(*args, **kwargs)
+    bound.apply_defaults()
 
     locals_dict: dict[str, Any] = dict(bound.arguments)
 
--- a/thunder/core/interpreter.py
+++ b/thunder/core/interpreter.py
@@ -16,7 +16,6 @@ import sys
 import traceback
 import weakref
 import torch
-import thunder
 from typing import Any, Literal, TypedDict
 from collections.abc import Callable, Iterable, Iterator, Mapping, MutableMapping, Sequence, Set, Sized
 import collections
--- a/thunder/core/interpreter.py
+++ b/thunder/core/interpreter.py
@@ -3570,8 +3570,6 @@ def _call_handler(
     func_or_null = stack.pop_wrapped()
     if frame.call_shape_kwnames is not None:
         kwnames = frame.call_shape_kwnames
-        if len(kwnames) == 3:
-            print("")
         assert len(args) >= len(kwnames)
         kwargs = dict(zip(kwnames, args[-len(kwnames) :]))
         args = args[: -len(kwnames)]
--- a/thunder/core/interpreter.py
+++ b/thunder/core/interpreter.py
@@ -16,6 +16,7 @@ import sys
 import traceback
 import weakref
 import torch
+import thunder
 from typing import Any, Literal, TypedDict
 from collections.abc import Callable, Iterable, Iterator, Mapping, MutableMapping, Sequence, Set, Sized
 import collections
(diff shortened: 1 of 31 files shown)


thunder/core/jit_ext.py
```diff
@@ -1054,6 +1054,8 @@ def is_from_torch(fn):
 
         has_tensor_arg = False
         for a in args:
+            if is_opaque(fn) and isinstance(a.value, thunder.dtypes.dtype):
+                a.value = thunder.dtypes.to_torch_dtype(a.value)
```
t-vi, line 1058: Unfortunately, I think this was a red herring, sorry about that! We need to have something after unwrapping, which is not here.
With that, I think it might be better to have a `finfo` function in thunder/torch/__init__.py that just does the thunder dtype -> torch.dtype conversion, calls `torch.finfo` and returns the result (not sure if we need to handle the result type but I think we might try as is).

thunder/core/interpreter.py
```diff
@@ -16,6 +16,7 @@
 import traceback
 import weakref
 import torch
+import thunder
```
t-vi, line 19: I don't think this needs thunder
```suggestion
```

thunder/core/jit_ext.py
```diff
@@ -915,6 +915,23 @@ def _general_jit_torch_autograd_function_apply_lookaside(obj: Any, *args, **kwar
...
 
 
+@run_once
+def _warn_custom_torch_info_function():
+    warnings.warn("`torch.finfo` is not implemented in Thunder. Calling `torch.finfo` directly.")
+
+
+@general_jit_lookaside(torch.finfo)
+def _general_jit_torch_finfo_lookaside(obj: Any):
+    _warn_custom_torch_info_function()
```
t-vi, line 925: I don't think we need this warning. This is only for autograd.Function not implementing the backward. So we also don't need the `def _warn_custom_torch_info_function():`.

thunder/core/jit_ext.py
```diff
@@ -915,6 +915,23 @@ def _general_jit_torch_autograd_function_apply_lookaside(obj: Any, *args, **kwar
...
+    thunder_dtype = unwrap(obj)
+    torch_dtype = thunder.dtypes.to_torch_dtype(thunder_dtype)
+    res = torch.finfo(torch_dtype)
+
+    pr = ProvenanceRecord(PseudoInst.OPAQUE, inputs=[wrap_const(res).provenance])
```
t-vi, line 930: If we want to wrap this manually, I think the inputs should be `[wrap_const(torch.finfo).provenance, obj.provenance]` (so the things you need to know from the call).

thunder/core/jit_ext.py
```diff
@@ -915,6 +915,23 @@ def _general_jit_torch_autograd_function_apply_lookaside(obj: Any, *args, **kwar
...
+def _warn_custom_torch_info_function():
+    warnings.warn("`torch.finfo` is not implemented in Thunder. Calling `torch.finfo` directly.")
+
+
+@general_jit_lookaside(torch.finfo)
+def _general_jit_torch_finfo_lookaside(obj: Any):
+    _warn_custom_torch_info_function()
+    thunder_dtype = unwrap(obj)
+    torch_dtype = thunder.dtypes.to_torch_dtype(thunder_dtype)
+    res = torch.finfo(torch_dtype)
```
t-vi, line 928: If we want to do this manually, the we probably want something like
```
try:
  torch_dtype = thunder.dtypes.to_torch_dtype(thunder_dtype)
  res = torch.finfo(torch_dtype)
except BaseException as e:
  return do_raise(e) 
```
because the lookasides are part of the interpreter, so they need to do `do_raise` instead of throwing exceptions.

t-vi (commented):
Hi @k223kim , great work! I think with `@interpreter_needs_wrapping` it might be easier (because it would do the `unwrap` and `wrap` and the `try: except:`  for us but this approach also works.

thunder/core/jit_ext.py
k223kim: Oh, I did not fully understand how that worked. Thanks for the explanation!

thunder/core/jit_ext.py
k223kim: Thanks for the explanation Tom!

k223kim:
Hey Tom! Thanks for the detailed comments and feedback. I got to learn how lookaside works! 🚀

thunder/core/jit_ext.py
```diff
@@ -915,6 +915,20 @@ def _general_jit_torch_autograd_function_apply_lookaside(obj: Any, *args, **kwar
     return _interpret_call(custom_forward, wrapped_ctx, *args_, **kwargs_)
 
 
+@general_jit_lookaside(torch.finfo)
+def _general_jit_torch_finfo_lookaside(obj: Any):
+    try:
+        thunder_dtype = unwrap(obj)
```
t-vi, line 921: It's nitpicky, but  would move the unwrap outside the try...except as an error there is an error in the interpreter itself.

thunder/core/jit_ext.py
```diff
@@ -915,6 +915,20 @@ def _general_jit_torch_autograd_function_apply_lookaside(obj: Any, *args, **kwar
...
+        torch_dtype = thunder.dtypes.to_torch_dtype(thunder_dtype)
+        res = torch.finfo(torch_dtype)
+    except BaseException as e:
+        return do_raise(e)
+
+    pr = ProvenanceRecord(PseudoInst.OPAQUE, inputs=[wrap_const(torch.finfo).provenance])
```
t-vi, line 927: We should add obj.provenance here, too.
```suggestion
    pr = ProvenanceRecord(PseudoInst.OPAQUE, inputs=[wrap_const(torch.finfo).provenance, obj.provenance])
```

thunder/core/jit_ext.py
k223kim: Oh sorry, that makes more sense. let me update that!

thunder/core/jit_ext.py
k223kim: Oh, I thought I had it.. will update this.

thunder/tests/test_jit_general.py
```diff
@@ -435,6 +435,19 @@ def foo(a, b):
...
+
+    jfoo = thunder.jit(foo)
+    a = torch.randn((2, 2), device="cpu")
+
+    actual = jfoo(a)
+    expected = foo(a)
+
+    assert actual == expected
+
+
```
t-vi, line 450: Maybe it would be neat to have both this and the previous check, like

```suggestion
   def bar(a):
        return torch.finfo(a.dtype).min

    jbar = thunder.jit(bar)
    a = torch.randn((2, 2), device="cpu")

    actual = jbar(a)
    expected = bar(a)

    assert_close(actual, expected)


```

k223kim: Yes, me add this right now!

t-vi (approved):
Thank you @k223kim

PR #769, opened 2024-07-15
benchmark - ignore activation checkpoint with thunder

kshitij12345:
Fixes - https://github.com/Lightning-AI/lightning-thunder/issues/582

Activation checkpointing is not supported with `thunder` and the benchmarking script already throws a warning. In this PR, we set the `self.checkpoint_activations=False` when compiling with thunder (to avoid going through the activation checkpointing path which leads to errors due to the same - see #582).


IvanYashchuk (approved):

t-vi (approved):

PR #772, opened 2024-07-16
Split `test_ddp` into `test_ddp`, `test_fsdp` and `test_ops`

crcrpar:
## What does this PR do?

As the file keeps getting bigger, I find it reasonable to split it into three.
I checked the number of test cases in `thunder/tests/distributed` is the same as the main.


t-vi (approved):
Makes total sense
Thank you @crcrpar

Issue #750, opened 2024-07-10
type inference: mismatched dtype in cat operator


tfogal:
## 🚀 Model / language coverage

The following code results in a

```
RuntimeError: Expected dtype thunder.dtypes.float32 but found thunder.dtypes.int64_!
```

error in `cat`s implementation. It seems we end up confused about the proper dtype of the `second` tensor.

```python
#!python3
import torch
import thunder
import einops

def foo(input_ids, inputs_embeds):
  batch_size, sequence_length, hidden_size = inputs_embeds.shape

  media_features = torch.randn((2,1,1,256,5120), dtype=torch.float16)
  num_images_per_sample = media_features.size(1)
  num_patches = media_features.size(3) * media_features.size(2)

  media_end_id = 32005
  sorted_media_end_positions_mask, media_end_positions_mask_sort_idx = (
      # NOTE: to(torch.long) is needed because PyTorch does not have sort for boolean tensors on CUDA
      (input_ids == media_end_id).to(torch.long).sort(dim=-1, descending=True, stable=True)
  )

  padded_media_indices = torch.where(
    sorted_media_end_positions_mask.to(torch.bool),
    media_end_positions_mask_sort_idx - num_patches + 1,
    sequence_length
  )
  padded_media_indices = padded_media_indices.unsqueeze(-1) + torch.arange(
    num_patches, device=padded_media_indices.device
  ).repeat(*padded_media_indices.shape, 1)
  padded_media_indices = padded_media_indices.reshape(batch_size, -1)
  padded_media_indices = einops.repeat(padded_media_indices, 'b s -> b s h', h=hidden_size)

  second = torch.zeros((batch_size, num_patches, hidden_size), device=inputs_embeds.device)
  # Note: thunder can be made to work by explicitly setting the dtype:
  #   second = torch.zeros((batch_size, num_patches, hidden_size), dtype=torch.float32, device=inputs_embeds.device)
  #print(f"ii dt:shape={inputs_embeds.dtype}:{inputs_embeds.shape}")
  #print(f"2nd dt:shape={second.dtype}:{second.shape}")
  updated_input_embeds = torch.cat(
    (inputs_embeds, second), dim=1
  )
  return updated_input_embeds

at = torch.zeros((2,384), dtype=torch.int64)
bt = torch.randn((2,384, 5120), dtype=torch.float32)

foo(at, bt)

thfoo = thunder.jit(foo)
thfoo(at, bt)
```

[...]

PR #775, opened 2024-07-16
[WIP] support inferring dtype with torch.get_default_dtype for factory functions

kshitij12345:
Fixes: #750

Changes - 
1. Stash `torch.get_default_dtype` in `cache_info`. Also, this adds a check to the prologue trace to verify that jitted fn is called with same default dtype - see example prologue below.
2. Factory functions infer the dtype based on `torch.get_default_dtype` from cache_info (if dtype is not passed explicitly)
3. We don't support changing the default dtype in the `jitted fn` as reordering and fusion can lead to issue - it is a loud error for now (we can revisit in follow-up if required).

Repro:
```python
import torch
import thunder

def foo(x: torch.Tensor) -> torch.Tensor:
    o = torch.zeros(x.shape, device=x.device)
    return o

jfoo = thunder.jit(foo)
o = jfoo(torch.randn(3, 3))
print(o.dtype)
print(thunder.last_prologue_traces(jfoo)[0])
print()
print(thunder.last_traces(jfoo)[0])
```

Prologue Trace
```python
import thunder
import torch

@torch.no_grad()
@no_autocast
  # args: "Any"
  prims.check_len(args, 1)
  # kwargs: "Any"
  prims.check_len(kwargs, 0)
  t_0: "cpu f32[3, 3]" = args[0]
  prims.check_tensor_metadata(t_0, (3, 3), 'cpu', torch.float32, False)
  cache_info_default_dtype: "<class 'torch.dtype'>" = cache_info['default_dtype']
  # NOTE - We bake the torch.dtype in trace (check_tensor_metadata also does the same).
  prims.check_literal_like(cache_info_default_dtype, torch.float32)
  prims.check_number_type_and_value(cache_info_is_autocast_enabled, False)
  prims.check_number_type_and_value(cache_info_no_grad_sync, False)
  return ((), ())
```

Computation Trace
```python
import thunder
import torch

@torch.no_grad()
@no_autocast
def computation():
  # /home/kkalambarkar/lightning-thunder/scratchpad/test.py:64:             o = torch.zeros(x.shape, device=x.device)
  o = ltorch.zeros((3, 3), device=devices.Device("cpu"), dtype=None)  # o: "cpu f32[3, 3]"
    # o = ltorch.full((3, 3), 0, device=devices.Device("cpu"), dtype=None)  # o: "cpu f32[3, 3]"
      # o = prims.full((3, 3), 0, device=devices.Device("cpu"), dtype=dtypes.float32)  # o: "cpu f32[3, 3]"
  return o

```

### DIFF
--- a/thunder/torch/__init__.py
+++ b/thunder/torch/__init__.py
@@ -85,6 +85,24 @@ _torch_to_thunder_function_map: dict[Callable, Callable] = {}
 _inplace_to_out_of_place: dict[Callable, tuple[Callable, int]] = {}
 
 
+# Helpers for factory functions to get default dtypes.
+def get_default_dtype():
+    # `thunder.jit` will create cache info and stash the default dtype
+    # observed at the beginning of jitting.
+    cache_info = thunder._get_cache_info()
+
+    # Currently, changing dtype during the jitted function is unsupported.
+    utils.check(
+        cache_info["default_dtype"] == torch.get_default_dtype(),
+        lambda: "Default dtype is changed during the execution of jitted function. This is currently unsupported.",
+    )
+    return torch.get_default_dtype()
+
+
+def maybe_get_default_dtype(dtype):
+    return dtype or get_default_dtype()
+
+
 # A wrapper that executes the operations within the torch language context
 # NOTE because this module defines the torch language context, a reference to itself
 #   is acquired by inspecting the __module__ attribute of the is_available function defined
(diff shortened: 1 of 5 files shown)


thunder/torch/__init__.py
```diff
@@ -85,6 +85,24 @@
 _inplace_to_out_of_place: dict[Callable, tuple[Callable, int]] = {}
 
 
+# Helpers for factory functions to get default dtypes.
+def get_default_dtype():
+    # `thunder.jit` will create cache info and stash the default dtype
+    # observed at the beginning of jitting.
+    cache_info = thunder._get_cache_info()
```
kshitij12345, line 92: This won't work with the deprecated `thunder.compile` (which is why I had to update `executor.make_callable_with_info(fn)` to `executor.make_callable(fn)`). But I assume that is ok.

t-vi: if we could remove make_callable_with_info, it'd be perfect. :)

kshitij12345: Updating to `make_callable` worked for the failing test. Resolving the comment :)

t-vi (approved):
Supergood, thank you @kshitij12345

PR #776, opened 2024-07-16
apply `sort_waits` if `dist_prims.wait` is in a trace

### LINKED ISSUE #765: Apply `thunder.distributed.utils.sort_waits` appropriately when possible
https://github.com/Lightning-AI/lightning-thunder/pull/624 adds the support of `torch.distributed` ops notwithstanding the value of `async_op`. 
Even when `async_op=True`, thunder calls `wait` when the op is called, not when the `handle.wait` is called (Even, `handle.wait` is currently ignored).
So, it'd be better/required to apply `sort_waits` so that we have (at least) decent overlap of computation and communication.

Currently `sort_waits` and https://github.com/Lightning-AI/lightning-thunder/blob/f07027eb1f8ab97b39c5faa5a5b4a4beb1b1c91d/thunder/distributed/utils.py#L54 are only applied when `use_fsdp=True`.

cc @carmocca @crcrpar

crcrpar:
## What does this PR do?

Fixes #765

### DIFF
--- a/thunder/__init__.py
+++ b/thunder/__init__.py
@@ -569,14 +569,6 @@ def jit(
             cs.last_interpreter_log = last_interpreter_log
             cs.last_interpreted_instructions = (i for i in last_interpreter_log if isinstance(i, dis.Instruction))
 
-            # note(crcrpar): deliberately apply sort_waits optionally as per
-            # https://github.com/Lightning-AI/lightning-thunder/blob/7a804e56d9cd695eb03a1fc8d178253d06e7548a/thunder/distributed/utils.py#L155-L159
-            if cd.disable_torch_autograd_support:
-                from thunder.distributed.utils import maybe_sort_waits
-
-                computation_trc = maybe_sort_waits(computation_trc)
-                computation_traces.append(computation_trc)
-
             computation_trc = dce(computation_trc)
             computation_traces.append(computation_trc)
 
@@ -605,11 +597,22 @@ def jit(
                     computation_traces.append(computation_trc)
                 cs.last_computation_transformation_stop = time.perf_counter_ns()
 
+                from thunder.executors.passes import transform_for_execution as transform_for_execution_pass
+                from thunder.executors.passes import _transform_for_operator_executor_execution
+                from thunder.distributed.utils import maybe_sort_waits
+
+                computation_trc = _transform_for_operator_executor_execution(computation_trc, cd.executors_list)
+                is_transformed, computation_trc = maybe_sort_waits(computation_trc)
+                if is_transformed:
+                    computation_traces.append(computation_trc)
+
                 with langctxs.langctx(cd.langctx):
                     extraces = transform_for_execution(
                         computation_trc,
                         executors_list=cd.executors_list,
+                        use_del_last_used=True,
                     )
+                computation_traces.append(computation_trc)
                 computation_trc = extraces[-1]
 
             if not compile_options.get("disable_inplace_copy_check", False):
--- a/thunder/__init__.py
+++ b/thunder/__init__.py
@@ -601,7 +601,7 @@ def jit(
                 from thunder.executors.passes import _transform_for_operator_executor_execution
                 from thunder.distributed.utils import maybe_sort_waits
 
-                with langctx.langctx(cd.langctx):
+                with langctxs.langctx(cd.langctx):
                     tmp_comp_trc = _transform_for_operator_executor_execution(computation_trc, cd.executors_list)
                 is_transformed, tmp_comp_trc = maybe_sort_waits(tmp_comp_trc)
                 if is_transformed:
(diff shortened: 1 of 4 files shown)


thunder/tests/distributed/test_ops.py
```diff
@@ -272,6 +272,60 @@ def lc_foo(
...
+            if t._provenance is not None and t._provenance.pss == "Sort Waits":
+                trace_with_waits_sorted = t
+                break
+
+        comm_idx = len(t.bound_symbols)
+        for idx, bsym in enumerate(trace_with_waits_sorted.bound_symbols):
+            if bsym.sym.id in {all_gather_prim_impl.id, reduce_scatter_prim_impl.id}:
+                comm_idx = idx
+            if bsym.sym.id == wait_prim_impl.id:
+                self.assertGreater(idx, comm_idx + 2)
```
crcrpar, line 327: The goal of this check is to make sure that wait gets distant from the comm by `sort_waits`. In-place comms are expressed as a pair of thunder's functional comm and copy. So initially they are next to each other.

thunder/__init__.py
```diff
@@ -595,11 +595,24 @@ def get_computation_and_inputs(*args, **kwargs):
                     computation_traces.append(computation_trc)
                 cs.last_computation_transformation_stop = time.perf_counter_ns()
 
+                from thunder.executors.passes import transform_for_execution as transform_for_execution_pass
+                from thunder.executors.passes import _transform_for_operator_executor_execution
+                from thunder.distributed.utils import maybe_sort_waits
+
+                with langctxs.langctx(cd.langctx):
+                    tmp_comp_trc = _transform_for_operator_executor_execution(computation_trc, cd.executors_list)
```
kiya00, line 603: we have this `_transform_for_operator_executor_execution` also in the `transform_for_execution`, does it affect anything? I noticed the bucketing was applied before `_transform_for_operator_executor_execution`, now it's after the `_transform_for_operator_executor_execution`, don't know if it matters

thunder/executors/torch_autograd.py
```diff
@@ -249,8 +250,14 @@ def split_forward_backward(computation_trc: TraceCtx, compile_data, compile_stat
                 compile_data.fn.bucketing_strategy != FSDPBucketingStrategy.NONE,
             )
             bw_extrace = sort_waits(bw_extrace)
-    if getattr(compile_data.fn, "use_ddp", False):
+    use_ddp: bool = getattr(compile_data.fn, "use_ddp", False)
+    if use_ddp:
         bw_extrace = sort_waits(bw_extrace)
+    if (not use_ddp) and (not use_fsdp):
```
kiya00, line 256: Is this sorting also needed for inplace comms when using ddp/fsdp?

thunder/__init__.py
crcrpar: To effectively apply `sort_waits` to a trace, `wait`s need to be a BoundSymbol. If a trace has a BoundSymbol representing an in-place distributed op, its subsymbols are a pair of out-of-place dist op, wait, and copy and there does not exist a boundsymbol whose sym is wait.
The call of `_transform_for_operator_executor_execution` here flattens such bsyms of in-place dist ops and `tmp_comp_trc` would have bsyms of waits, if the computation_trc has bsyms of in-place dist ops.

thunder/executors/torch_autograd.py
crcrpar: inplace comms are not used in thunder's ddp & fsdp, currently.
This check is rather to avoid redundant application for ddp or unwanted application for fsdp

kiya00: >inplace comms are not used in thunder's ddp & fsdp, currently. This check is rather to avoid redundant application for ddp or unwanted application for fsdp.

I get it now, thanks for the above explanation and trace example

kiya00 (approved):

thunder/executors/torch_autograd.py
kiya00: it seems for fsdp, the maybe_sort_wait in `__init__` will be applied after the limit_in_flight_allgathers on fwd trace, will it prefer the allgather again and break the limit_in_flight_allgathers ?

t-vi:
Thank you @crcrpar @kiya00

thunder/__init__.py
```diff
@@ -597,11 +597,24 @@ def get_computation_and_inputs(*args, **kwargs):
                     computation_traces.append(computation_trc)
                 cs.last_computation_transformation_stop = time.perf_counter_ns()
 
+                from thunder.executors.passes import transform_for_execution as transform_for_execution_pass
+                from thunder.executors.passes import _transform_for_operator_executor_execution
+                from thunder.distributed.utils import maybe_sort_waits
+
+                with langctxs.langctx(cd.langctx):
```
IvanYashchuk, line 604: I think `langctxs.langctx(cd.langctx)` should be applied as a decorator on the `get_computation_and_inputs` function as a whole. I will create a PR for this.

PR #777, opened 2024-07-16
rename (registering) general_jit_lookaside to register_general_jit_lookaside

kshitij12345:
Currently there are two - `general_jit_lookaside` in `jit_ext.py`

This registers a lookaside - (we should call it `register_general_jit_lookaside` and this is what the PR does)
https://github.com/Lightning-AI/lightning-thunder/blob/7a804e56d9cd695eb03a1fc8d178253d06e7548a/thunder/core/jit_ext.py#L717-L722

This actually gets the lookaside for a given function - (this PR leaves it untouched)
https://github.com/Lightning-AI/lightning-thunder/blob/7a804e56d9cd695eb03a1fc8d178253d06e7548a/thunder/core/jit_ext.py#L1038

### DIFF
--- a/thunder/core/jit_ext.py
+++ b/thunder/core/jit_ext.py
@@ -714,7 +714,7 @@ _general_jit_lookaside_map.update(
 )
 
 
-def general_jit_lookaside(diverted_fn):
+def register_general_jit_lookaside(diverted_fn):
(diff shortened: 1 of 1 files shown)


t-vi (approved):
Thank you @kshitij12345

PR #779, opened 2024-07-16
remove add_executors_list definition as it is already imported from extend.add_executors_list

kshitij12345:
Minor clean-up

It is already imported here - https://github.com/Lightning-AI/lightning-thunder/blob/323d7df47c47b09db168e9ac3c8e1df5a4cd6c58/thunder/__init__.py#L168

Also, `itertools` is not imported in `thunder/__init__.py` so this (removed) definition will fail.

### DIFF
--- a/thunder/__init__.py
+++ b/thunder/__init__.py
@@ -232,19 +232,6 @@ def _get_cache_info():
     return _cache_info_ctx.get()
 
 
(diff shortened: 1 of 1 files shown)


t-vi (approved):
Thank you @kshitij12345

Issue #614, opened 2024-06-18
[Feature request] Optional debugging option to get trace with information on tensor strides along with tensor shapes


parthmannan:
## 🚀 Feature Request

Currently we have computation traces with the generated tensor shapes as part of comments next to the computation like
```
t908 = torch.nn.functional.linear(t907, t19, t17)  # t908: "cuda:0 bf16[1, 2048, 4096]"
    # t908 = ltorch.linear(t907, t19, t17)  # t908: "cuda:0 bf16[1, 2048, 4096]"
      # t908 = prims.linear(t907, t19, t17)  # t908: "cuda:0 bf16[1, 2048, 4096]"
```

However, there are some situations where stride information becomes necessary for debugging like #583 where a stride difference was creating an illegal memory access in one of the executors.
While my understanding is that Thunder consciously has made a decision to not include stride information in order to let backends manage strides on their own. 
This feature request does not aim to change that. Given a set of fixed input tensors, can there be a way of generating computation traces with the tensor shapes and strides? There can be a mandatory requirement that such a trace can only be generated after one full iteration has executed so that strides can be recorded.

t-vi:
Hi @parthmannan , thank you for filing this!

Let me pick your brain a bit: Maybe we could have an advanced debugging tutorial where we define a symbol with an impl printing tensor information (such as stride) and a transformation that inserts calling that symbol for every TensorProxy result. Would that help you?

parthmannan:
That sounds pretty useful and should suffice the requirement. Would calling this transformation generate a full computation trace for every TensorProxy result with the required tensor information?

And I am guessing this is far more useful than just having a trace with strides as one can define a symbol that prints any attribute of a tensor like `requires_grad` etc.?

(Side comment) Re: tutorial - While a tutorial would be awesome to start but it does require users to modify model execution code. The easiest debugging method is just enabling debug logs using env variables like `TORCH_COMPILE_DEBUG` or `CUBLASLT_LOG_LEVEL` etc.
This doesn't need to be an option for something niche like strides but in general, our debugging is a little more complex with users required to call functions to grab traces, print them out etc.

mruberry:
triage review:

- let's provide a notebook showing how to debug issues like this
[...]

PR #783, opened 2024-07-16
Add Debug Transform to support collecting runtime details eg. input strides, memory allocation, etc

kshitij12345:
Ref: #614 

This PR adds a `DebugTransform` which adds debug symbols before all the computation symbols.

`DebugTransform` in the constructor takes a callback with signature `callback(computation_bsym, *args, **kwargs) -> str` which is invoked by the debug symbol. The callback gets the computation_bsym it is tied to and the runtime args and kwargs for the same. The callback should return a string capturing the details it is interested in and this is put in the header of the corresponding debug bsym.

Eg.

```python
import torch
import thunder
import itertools

dim = 4096


class Model(torch.nn.Module):
    def __init__(self) -> None:
        super().__init__()
        self.fc1 = torch.nn.Linear(dim, dim)
        self.fc2 = torch.nn.Linear(dim, dim)

    def forward(self, x):
        x = self.fc1(x)
        x = torch.nn.functional.relu(x)
        x = self.fc2(x)
        return x


model = Model().to(device="cuda")
x = torch.randn(4, dim, dim, device="cuda")

# Transform
from thunder.dev_utils.debug_transform import DebugTransform

def callback(bsym, *args, **kwargs):
    # String returned from the callback will be used in the header for debug_symbol.
    input_strides = []
    for arg in itertools.chain(args, kwargs.values()):
        if isinstance(arg, torch.Tensor):
            input_strides.append(arg.stride())

    output_str = (
        f"{bsym.sym.name}\n"
        f"Input Strides: {[stride for stride in input_strides]}\n"
        f"Memory Allocated: {torch.cuda.memory_allocated()}\n"
    )
    return output_str


stride_and_mem_profile_transform = DebugTransform(callback=callback)

jmodel = thunder.jit(
    model,
    post_optimization_transforms=[
        stride_and_mem_profile_transform ,
    ],
)
o = jmodel(x)
o.sum().backward()

trace = thunder.last_traces(jmodel)[-1]
bwd_trace = thunder.last_backward_traces(jmodel)[-1]

print(trace)
# print(bwd_trace)

```

Output
```python
# Constructed by Debug Transform
import torch
import torch.nn.functional

@torch.no_grad()
@no_autocast
def augmented_forward_fn(x, t_fc1_bias, t_fc1_weight, t_fc2_bias, t_fc2_weight):
  # x: "cuda:0 f32[4, 4096, 4096]"
  # t_fc1_bias: "cuda:0 f32[4096]"
  # t_fc1_weight: "cuda:0 f32[4096, 4096]"
  # t_fc2_bias: "cuda:0 f32[4096]"
  # t_fc2_weight: "cuda:0 f32[4096, 4096]"
  # linear
  # Input Strides: [(16777216, 4096, 1), (4096, 1), (1,)]
  # Memory Allocated: 402685952
  debug_linear_1(x, t_fc1_weight, t_fc1_bias)
  t0 = torch.nn.functional.linear(x, t_fc1_weight, t_fc1_bias)  # t0: "cuda:0 f32[4, 4096, 4096]"
    # t0 = ltorch.linear(x, t_fc1_weight, t_fc1_bias)  # t0: "cuda:0 f32[4, 4096, 4096]"
      # t0 = prims.linear(x, t_fc1_weight, t_fc1_bias)  # t0: "cuda:0 f32[4, 4096, 4096]"
  # nvFusion0
  # Input Strides: [(16777216, 4096, 1)]
  # Memory Allocated: 671121408
  debug_nvFusion0_2(t0)
  [t1, t2] = nvFusion0(t0)
    # t1 = prims.gt(t0, 0.0)  # t1: "cuda:0 b8[4, 4096, 4096]"
    # t2 = prims.where(t1, t0, 0.0)  # t2: "cuda:0 f32[4, 4096, 4096]"
  del t0
  # linear
  # Input Strides: [(16777216, 4096, 1), (4096, 1), (1,)]
  # Memory Allocated: 738230272
  debug_linear_3(t2, t_fc2_weight, t_fc2_bias)
  t3 = torch.nn.functional.linear(t2, t_fc2_weight, t_fc2_bias)  # t3: "cuda:0 f32[4, 4096, 4096]"
    # t3 = ltorch.linear(t2, t_fc2_weight, t_fc2_bias)  # t3: "cuda:0 f32[4, 4096, 4096]"
      # t3 = prims.linear(t2, t_fc2_weight, t_fc2_bias)  # t3: "cuda:0 f32[4, 4096, 4096]"
  return {'output': t3, 'flat_args': [x, t_fc1_bias, t_fc1_weight, t_fc2_bias, t_fc2_weight], 'flat_output': (t3,)}, ((t1, t2, t_fc2_weight, x), ())
```


t-vi:
Neat! :rocket:

t-vi (approved):
Thank you @kshitij12345

parthmannan:
This is awesome! Would be a very useful tool in debugging.
One question - Maybe this just needs some simple modification to the callback. Is there a way to get the information on symbols inside the larger symbol (not sure about the terminology). For example, we can see the information on t0 which is the input to the nvFusion block but not the tensors generated within the block like t1 and t2.

riccardofelluga:
> For example, we can see the information on t0 which is the input to the nvFusion block but not the tensors generated within the block like t1 and t2.

It's possible to get that info by querying the fusion. I was going to add a helper function to do that and get the nvfuser repro out of the trace #387. Would it be more useful to get a repro or just the tensor information?

parthmannan:
>would it be more useful to get a repro to execute the fusion or just the tensor information?
Both have different use cases I believe. In brief, there is a need for repro to be able to quickly test nvFuser execution and there is a need to understand tensor information inside the fusion to understand if an unexpected shape/stride combo made nvFuser behave in an unexpected (or non-performant way).

In terms of usefulness, this would be my priority - 
1. Understand tensor information for the fusion block inputs
2. Repro execute nvFuser fusion block (possibly with the option to change inputs)
3. Find tensor information inside a fusion block

t-vi:
> 2. Repro execute nvFuser fusion block (possibly with the option to change inputs)

If you have the trace from last_traces, you can access the fusion definition and such through the Python environment of the trace (probably not exactly what you are looking for, but maybe where such information could be added):
```
lt = thunder.last_traces(jf)[-1]
fdw = lt.python_ctx()['nvFusion0']  # we might eventually do this more nicely
fdw.last_used  # this is the fusion definition
```

I don't know if it keeps the metadata of the inputs, printing the repro seems inlined into .execute in the except branch but might be cool to pull out into its own helper.
(the above is a snippet from the code for [our podcast](https://www.youtube.com/watch?v=HtL-T1nw0Rw))

t-vi:
One thing I've been thinking of is whether it might be neat to be able to use the "repro on error" function outside the error path (with a debug flag or so) to get the info3... @jjsjann123 wdyt?

kshitij12345:
> Repro execute nvFuser fusion block (possibly with the option to change inputs)

I think one problem with current approach of adding symbol before the actual computation symbol is - You maybe able to get the bound symbol for (in this case) fusion block in the callback but it won't have any information as the block hasn't executed yet.

Eg. Callback
```python

def callback(bsym, *args, **kwargs):
    output_str = (
        f"{bsym.sym.name}\n"
    )
    if bsym.sym.is_fusion:
        sym_name = bsym.sym.name
        _, call_ctx, _ = bsym.gather_ctxs()
        print(call_ctx[sym_name])  # FusionDefinitionWrapper(nvFusion0: (gt, where))
        print(call_ctx[sym_name].last_used)  # Will print None (as the FusionDefinition hasn't been called yet.)

    return output_str

```

PR #786, opened 2024-07-16
scatter is both a function and a method

tfogal:
## What does this PR do?

Allows `scatter` and `scatter_` to be methods in addition to raw functions.

This is needed for NeVA: https://github.com/NVIDIA/NeMo/blob/ca94e032cf9f2c1920f7d5e55dcaa0de9c2abd70/nemo/collections/multimodal/models/multimodal_llm/neva/neva_model.py#L256.

## PR review

### DIFF
--- a/thunder/torch/__init__.py
+++ b/thunder/torch/__init__.py
@@ -2776,7 +2776,7 @@ def gather(a: TensorLike, /, dim: int, index: TensorLike) -> TensorLike:
 
 # NOTE: PyTorch uses `src` for torch.Tensor arguments and `value` for scalars
 # when referencing the source of the values
-@torchsymbol(torch.scatter)
+@torchsymbol(torch.scatter, is_method=True)
 def scatter(
     a: TensorLike,
     /,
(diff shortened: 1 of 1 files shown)


thunder/torch/__init__.py
```diff
@@ -2802,7 +2802,7 @@ def scatter(
         return clang.scatter(a, index, value, dim)
 
 
-@torchsymbol(torch.Tensor.scatter_, tags=(prims.OpTags.IN_PLACE,))
+@torchsymbol(torch.Tensor.scatter_, is_method=True, tags=(prims.OpTags.IN_PLACE,))
```
t-vi, line 2805: Do we need the `is_method` here? My impression was that it's implicit when the symbol is on torch.Tensor...

tfogal: Yes, we do. Without it the linked-to line in NeVA will die in thunder with `AttributeError: Unknown attribute scatter_. Did you mean: '_return_value'?`

tfogal:
Tests that are failing do not fail locally and don't seem related; marking ready for review.

thunder/torch/__init__.py
t-vi: Right, sorry. :)

t-vi (approved):
Thank you @tfogal

thunder/torch/__init__.py
nikitaved: Oops, missed that. Thank you, @tfogal !

PR #788, opened 2024-07-17
remove OverridenKVCache and fix some peculiar  cases of `prims.copy_` + NVFuser

nikitaved:
As per title, we can remove `OverridenKVCache` in our `litgpt` tests since the missing op, `index_copy_` is in the system.
Additionally, this PR
* Fixes https://github.com/Lightning-AI/lightning-thunder/issues/789
* Fixes https://github.com/Lightning-AI/lightning-thunder/issues/791

### DIFF
--- a/thunder/executors/nvfuserex_impl.py
+++ b/thunder/executors/nvfuserex_impl.py
@@ -786,6 +786,22 @@ class nvFuserExecutor(FusionExecutor):
         #   (Used to name fusions like nvFusion0, nvFusion1, ...)
         fusion_counter: int = 0
         for bsyms in bound_symbol_groups:
+            # Related to in-place ops.
+            # prims.copy_ is a no-op for NVFuser in a sense that it does not
+            # generate nor runs any kernels.
+            # For that reason we should avoid fusing prims.copy_ unless
+            # it comes after other non-copy symbols in a fusion.
+            # See the following relevant issues:
+            # https://github.com/Lightning-AI/lightning-thunder/issues/789
+            # https://github.com/Lightning-AI/lightning-thunder/issues/791
+            # NOTE: filter all first "dangling" no-op copies
+            while len(bsyms) > 0 and bsyms[0].sym.id is prims.PrimIDs.COPY_:
+                fused_bsyms.append(bsyms[0])
+                bsyms = bsyms[1:]
+
+            if len(bsyms) == 0:
+                continue
+
             # TODO The following allows generating single node fusions, which
             #   may be suboptimal for real-world performance.
             #   Provide a mechanism to switch between "test" and "perf" modes
--- a/thunder/tests/litgpt_model.py
+++ b/thunder/tests/litgpt_model.py
@@ -118,47 +118,8 @@ configs = [
 name_to_config = {config["name"]: config for config in configs}
 
 
-class OverridenKVCache(nn.Module):
-    def __init__(
-        self,
-        k_shape: tuple[int, int, int, int],
-        v_shape: tuple[int, int, int, int],
-        device: torch.device | None = None,
-        dtype: torch.dtype | None = None,
-    ) -> None:
-        super().__init__()
-        self.register_buffer("k", torch.zeros(k_shape, device=device, dtype=dtype), persistent=False)
-        self.register_buffer("v", torch.zeros(v_shape, device=device, dtype=dtype), persistent=False)
-
-    def forward(self, input_pos: torch.Tensor, k: torch.Tensor, v: torch.Tensor) -> tuple[torch.Tensor, torch.Tensor]:
-        # move the buffer to the activation dtype for when AMP is used
-        self.k = self.k.to(k.dtype)
-        self.v = self.v.to(v.dtype)
-        # update the cache
-        # NOTE: `torch._dynamo.is_compiling` is being deprecated, we should update this once all versions have `torch.compiler.is_compiling`.
-        is_compiling = (
-            torch.compiler.is_compiling if hasattr(torch.compiler, "is_compiling") else torch._dynamo.is_compiling
-        )
-        if is_compiling():
-            # inductor doesn't support `index_add` with bfloat16
-            k = self.k.index_copy_(2, input_pos, k)
-            v = self.v.index_copy_(2, input_pos, v)
-            return k, v
-        # See issue: "Support more indexing operators (index_copy and index_add)"
-        k = self.k = torch.index_add(self.k, 2, input_pos, k)
-        v = self.v = torch.index_add(self.v, 2, input_pos, v)
-        # THUNDER bug: cannot return self.k, self.v here (may be cuda graphs related - no minimum repro)
-        return k, v
-
-    def reset_parameters(self) -> None:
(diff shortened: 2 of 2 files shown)


thunder/tests/litgpt_model.py
```diff
@@ -134,20 +134,10 @@ def forward(self, input_pos: torch.Tensor, k: torch.Tensor, v: torch.Tensor) ->
...
-            # inductor doesn't support `index_add` with bfloat16
-            k = self.k.index_copy_(2, input_pos, k)
-            v = self.v.index_copy_(2, input_pos, v)
-            return k, v
-        # See issue: "Support more indexing operators (index_copy and index_add)"
-        k = self.k = torch.index_add(self.k, 2, input_pos, k)
-        v = self.v = torch.index_add(self.v, 2, input_pos, v)
-        # THUNDER bug: cannot return self.k, self.v here (may be cuda graphs related - no minimum repro)
+        k = self.k.index_copy_(2, input_pos, k)
+        v = self.v.index_copy_(2, input_pos, v)
```
kshitij12345, line 140: If this patch works then I don't think we need `OverridenKVCache` at all and can directly rely on `KVCache` from litgpt. Wdyt?

https://github.com/Lightning-AI/litgpt/blob/97ce737b27eababfabd7e94286373df9901df94f/litgpt/model.py#L382-L406

nikitaved: Yes, this is the goal. I will follow your suggestion upon green CIs.

t-vi: While it is not apparent from the tests, if you run llama 3 with this, it will not work (out of place `index_copy` will produce sensical results, in-place `index_copy_` not). I have not yet found out why, but as we intend to do benchmarking, this would be interesting to know.

t-vi: (this is from the branch with tom/ggmlquant from PR #754 )

nikitaved: Hmm, it breaks the test, simply removing `OverridenKVCache`, that is...

nikitaved: The tests are still failing, on the GPU...

t-vi: I would recommend to do it in two/tree steps:
```
k = self.k = torch.index_copy(self.k, 2, input_pos, k)
v = self.v = torch.index_copy(self.v, 2, input_pos, v)
```
should work
```
k = self.k = torch.index_copy_(self.k, 2, input_pos, k)
v = self.v = torch.index_copy_(self.v, 2, input_pos, v)
```
breaks / needs to be fixed
Ultimate goal:
```
k = torch.index_copy_(self.k, 2, input_pos, k)
v = torch.index_copy_(self.v, 2, input_pos, v)
```

nikitaved:
Looks like the issue could be a bit deeper. Related:
https://github.com/Lightning-AI/lightning-thunder/issues/789
https://github.com/Lightning-AI/lightning-thunder/issues/791

nikitaved:
The fix is in! @jjsjann123 , could you please also have a look and tell me whether you approve the fix in the NVFuser fusion logic.

thunder/tests/litgpt_model.py
nikitaved: I have update the PR to include a potential fix...

nikitaved: OK, some GPU CIs are green now! :)

thunder/tests/litgpt_model.py
```diff
@@ -118,47 +118,8 @@
 name_to_config = {config["name"]: config for config in configs}
 
 
-class OverridenKVCache(nn.Module):
```
tfogal, line 121: It's not clear why we're removing a test here... is there some reason this code is fundamentally invalid? Why wouldn't we want to test it?

thunder/executors/nvfuserex_impl.py
```diff
@@ -786,6 +786,22 @@ def _can_fuse_node(n: Node):
...
+            # generate nor runs any kernels.
+            # For that reason we should avoid fusing prims.copy_ unless
+            # it comes after other non-copy symbols in a fusion.
+            # See the following relevant issues:
+            # https://github.com/Lightning-AI/lightning-thunder/issues/789
+            # https://github.com/Lightning-AI/lightning-thunder/issues/791
+            # NOTE: filter all first "dangling" no-op copies
+            while len(bsyms) > 0 and bsyms[0].sym.id is prims.PrimIDs.COPY_:
+                fused_bsyms.append(bsyms[0])
+                bsyms = bsyms[1:]
```
tfogal, line 800: * It seems we're looking at the absolute order of operations, rather than a dataflow order. But maybe we don't have infra to do so via dataflow?
* Should logic like this go into `_should_fuse`, above, instead of here? @jjsjann123

tfogal (commented):
Thanks. I may have missed something in the motivation for dropping the test option. Could you educate me on where you're coming from here?

For the nvfuser aspect, in general I think our fusion logic should be revisited. This seems like an extension in the same vein as what's there, and as such seems reasonable. I also think that the higher-level rework is a big can of worms, as we're presently experiencing with the small-scale change of bookends. So I'm inclined to say we should just go with what's here and accrue a little bit more debt pending a larger rewrite.
But it's more reasonable for @jjsjann123 to make that call than me.

thunder/tests/litgpt_model.py
nikitaved: The test is there, this KVCache was used to avoid missing ops in Thunder. See `test_litgpt_variants_kvcache`

thunder/executors/nvfuserex_impl.py
tfogal: Oh, side note: could I ask you to have the comment define what "dangling" means here?

nikitaved: I had a similar idea. Tried the easiest solution first

thunder/tests/litgpt_model.py
tfogal: Ahh, I understand now. Thank you!

tfogal:
+1 from me but I'd prefer we wait for Jie's thoughts before merging.

tfogal (approved):

t-vi (approved):
There goes my chance to merge without review.
Thank you @nikitaved @tfogal

thunder/executors/nvfuserex_impl.py
tfogal: Yeah, that's exactly the kind of thing I was worried about :-)

As you're highlighting, we have a lot of implicit assumptions here on the sorting of ops because we are not relying on dataflow. Such a requirement is implicit on earlier code's behavior, and there's really not a strong reason for that earlier behavior to exist (at least before this).

This is a good definition of fragility--when disconnected, distant pieces of code impose specific requirements on one another.

One of the reasons to ping Jie to get his thoughts is that maybe this fragility is fine here: if our thought is that we're going to scrap much of the fusion logic entirely in a rewrite, a little bit of extra debt here probably is not worth worrying about.

PR #790, opened 2024-07-17
Always raise error from the benchmarking script

mpatel31415:
<details>

</details>

## What does this PR do?
Adds to the benchmark_litgpt.py:
* Always raising and error from every worker to later catch and manage it. 
* Setting flag TORCH_NCCL_ASYNC_ERROR_HANDLING = 1 to prevent hanging of the benchmarks - it should not impact the performance: [info](https://gitlab-master.nvidia.com/dl/JoC/mxlg/mixology-logs/-/merge_requests/35#8a052024ee1084ceed6825e357472f1582141fbb:~:text=also%20this%20information%3A-,on%20github,-so%20I%20hope) We also tested it for a few cases. 
* Parametrization warmup and max iters
* 
## PR review

## Did you have fun?

Yes

### DIFF
--- a/thunder/benchmarks/benchmark_litgpt.py
+++ b/thunder/benchmarks/benchmark_litgpt.py
@@ -510,7 +510,7 @@ class Benchmark_litGPT:
                     if self.throughput:
                         self.throughput.update(
                             time=(t1 - t0),
-                            flops=self.perf_metrics.get("model_flops"),
+                            flops=self.perf_metrics["model_flops"],
                             batches=i,
                             samples=(i * self.micro_batch_size * self.gradient_accumulation_steps),
                             lengths=(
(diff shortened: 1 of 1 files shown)


parthmannan (approved):

thunder/benchmarks/benchmark_litgpt.py
```diff
@@ -611,6 +615,7 @@ def benchmark_main(return_metrics_as_json=False, json_path="", **kwargs) -> None
         # Helps catch OutOfMemory Errors and post processing of errors
         if global_rank in [0, None]:
             print("An error occurred:", type(error).__name__, "–", error)
+        raise
```
crcrpar, line 618: if we do raise, what would be the benefit(s) of using `try-except`?

crcrpar (approved):

t-vi (approved):
Thank you @mpatel31415 @parthmannan @crcrpar

thunder/benchmarks/benchmark_litgpt.py
mpatel31415: I guess I tried to be compatible with previous version if Parth Mannan would like to parse the messages in the same way as previously, but actually it won't be compatible anyway, because the error will be raised and json results won't be saved. 
I've removed the try except block.

PR #792, opened 2024-07-17
convert thunder types to torch types

### LINKED ISSUE #573: Returning `dtype` or `device` from jitted function returns thunder's dtype or device (not torch.{dtype/device}).
I think this can be a problem where part of the program is called with `thunder.jit` and it's output is used for non-compatible parts (eg. https://github.com/Lightning-AI/lightning-thunder/issues/461). These non-compatible parts may expect these objects to be `torch` types.

```python
import thunder
import torch

def foo(x):
    return x.device, x.dtype

jfoo = thunder.jit(foo)
x = torch.randn(4,)
o = jfoo(x)
print(o)  # (cpu, float32)
print(type(o[0]), type(o[1]))  # <class 'thunder.core.devices.Device'> <class 'thunder.core.dtypes.floating'>

# # Constructed by Delete Last Used (took 0 milliseconds)
# import thunder.core.devices as devices
# import thunder.core.dtypes as dtypes
# import torch
# from thunder.executors.torchex import no_autocast

# @torch.no_grad()
# @no_autocast
# def computation():
#   return (devices.Device("cpu"), dtypes.float32)
```

k223kim:
<details>

</details>

## What does this PR do?

Fixes #573.

## PR review

- I am not sure if it is the best practice to have `transform_to_torch_types` independently from other functions
- Or if `transform_to_torch_types` should be in `common.py`
- I think other tests already tests this implementation. Should I still add another test for this PR?

## Did you have fun?

⚡️


thunder/common.py
```diff
@@ -858,3 +859,20 @@ def _fn(*args, **kwargs) -> tuple[Any, list[TraceCtx]]:
...
+    def map_to_torch(x: Any) -> Any:
+        if isinstance(x, thunder.dtypes.dtype):
+            return thunder.dtypes.to_torch_dtype(x)
+        elif isinstance(x, thunder.devices.Device):
+            return thunder.devices.to_torch_device(x)
+        return x
+
+    last = trace.bound_symbols[-1]
+    new_args = tree_map(map_to_torch, last.args)
+    if last.sym.id == prims.PrimIDs.RETURN:
```
t-vi, line 875: Could this be an assert instead? (and probably before we assign new_args...)

thunder/tests/test_core.py
```diff
@@ -2969,12 +2969,12 @@ def fn(x):
     jfn = thunder.jit(fn)
     actual_dtype = jfn(x)
 
-    assert actual_dtype == thunder.dtypes.float32
+    assert actual_dtype == torch.float32
```
t-vi, line 2972: In this and the tests below (the above is special and excepmpt), could you adopt the pattern of
`assert fn(x) == jfn(x)` please?

t-vi (commented):
Looks great, I'm still not enchanted about having this in the main jit codeflow, but let's be pragmatic about this until we find a better place.
Two minor comments and we should be good to go.

t-vi (approved):
Thank you @k223kim

Issue #796, opened 2024-07-17
Support for torch autocasting


tfogal:
## 🚀 Model / language coverage

Having an issue with a customer model because we do not emulate or respect autocast. This in turn means the model does not fit.

### Pitch

This is an NVIDIA customer model.

### Alternatives / Potential work-arounds

I tried just casting the model to fp16 explicitly via `to()`, but there are type issues with this approach (probably another bug, still simplifying it for a second issue).

### Minimal Repro

```python
import torch
import torch.nn as nn
import thunder

class ProxyModel(torch.nn.Module):
    def __init__(self):
        super().__init__()

        self.mp2d = nn.MaxPool2d(kernel_size=(3, 3), stride=(2, 2), padding=(1, 1), dilation=1, ceil_mode=False)
        self.c0 = nn.Conv2d(96, 96, kernel_size=(1, 1), stride=(1, 1))
    
    def forward(self):
        with torch.amp.autocast('cuda', enabled=True):
            i0 = torch.randn((22, 96, 120, 160), device="cuda")
            t0 = self.mp2d(i0)
            t1 = self.c0(t0)
            print(f"t1 shape:dtype={t1.shape}:{t1.dtype}")
            return t1

m = ProxyModel().cuda()
print("eager:")
with torch.amp.autocast('cuda', enabled=True):
  eager = m()
  mdl = thunder.jit(m)
  print("thunder:")
  thund = mdl()
  print(f"eager shape:dtype={eager.shape}:{eager.dtype}")
  print(f"thunder shape:dtype={thund.shape}:{thund.dtype}")
```

Note that `autocast`ing twice is irrelevant; torch will make `t1` be `torch.float16` with either of the two. I only put them in twice to make it clear that thunder is ignoring the `with` context both when it's part of the `jit`ted function and when it's coming from the outer context.

t-vi:
As far as I can tell, we need autocast for conv layers in thunder.core.transforms. I'll try.

PR #797, opened 2024-07-18
add autocast for conv1/2/3d

t-vi:
Fixes: #796

### DIFF
--- a/thunder/tests/test_autocast.py
+++ b/thunder/tests/test_autocast.py
@@ -169,3 +169,29 @@ def test_autocast_mixed_dtype_inputs():
         jit_out = jfoo(x, w)
 
     torch.testing.assert_close(eager_out, jit_out)
+
+
+@pytest.mark.parametrize("dim", [1, 2, 3])
+def test_autocast_convolution(dim):
+    conv_fn = getattr(torch.nn.functional, f"conv{dim}d")
+
+    def foo(x, w, b=None):
+        return conv_fn(x, w, b)
+
+    x = torch.randn(1, 2, *(dim * (8,)))
+    w = torch.randn(3, 2, *(dim * (4,)))
+    b = torch.randn(3)
+
+    jfoo = thunder.jit(foo)
+
+    with torch.autocast("cpu", torch.bfloat16):
+        eager_out = foo(x, w, b)
+        jit_out = jfoo(x, w, b)
+
+    torch.testing.assert_close(eager_out, jit_out)
+
+    with torch.autocast("cpu", torch.bfloat16):
+        eager_out = foo(x, w)
+        jit_out = jfoo(x, w)
+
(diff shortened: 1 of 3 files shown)


thunder/torch/__init__.py
```diff
@@ -3875,8 +3878,17 @@ def pad_lo_hi_dilation_seq():
     padding, a = process_padding_str(int_to_seq(padding), stride, dilation, a)
     # }
 
-    res = clang.convolution(
-        a, weight, bias, stride, padding, dilation, False, (0,) * dim, groups  # transposed  # output_padding
+    res = conv_function(
+        a,
```
nikitaved, line 3882: nit: looks like `conv_function == _convolution_autocast_impl` accepts kwargs here, but `_convolution_autocast_impl` misses kwargs in its definition.

thunder/tests/test_autocast.py
```diff
@@ -169,3 +169,29 @@ def foo(x, w):
...
+
+    with torch.autocast("cpu", torch.bfloat16):
+        eager_out = foo(x, w, b)
+        jit_out = jfoo(x, w, b)
+
+    torch.testing.assert_close(eager_out, jit_out)
+
+    with torch.autocast("cpu", torch.bfloat16):
+        eager_out = foo(x, w)
+        jit_out = jfoo(x, w)
```
nikitaved, line 195: I wonder, will the autograd be affected?

t-vi: you mean the backward?

nikitaved: Yeah :)

thunder/torch/__init__.py
t-vi: Yes, but this is to be able to route dtype through without burdening the conv helper to know about dtypes.

thunder/tests/test_autocast.py
t-vi: Added it with terrible bounds...

nikitaved: Hmmm, maybe `rand` vs `randn` could do better?

t-vi: Well, so I would suspect that there still is some dragon lingering somewhere, because I cannot really think of  much reason to have thunder + autocast return wildly different results than torch + autocast even for wild inputs, but TBH, I would postpone investigating that.

t-vi: but you are right, switching to rand buys me an order of magnitude, cool, thank you!

t-vi: Turns out that Thunder does funny, slow things for the conv backward.

t-vi: https://github.com/Lightning-AI/lightning-thunder/issues/799 <-- filed an issue

nikitaved (approved):
LGTM! Thank you, @t-vi !

PR #798, opened 2024-07-18
Error out if alias tensor args are consumed by in-place ops

crcrpar:
<details>

</details>

## What does this PR do?

Fixes # (issue).

## PR review

## Did you have fun?

### DIFF
--- a/thunder/__init__.py
+++ b/thunder/__init__.py
@@ -251,6 +251,7 @@ CacheEntry = namedtuple(
         "backward_fn",
         "backward_traces",
         "return_none_instead_of_grads",
+        "vanilla_tensor_args",
     ],
 )
 
@@ -413,6 +414,7 @@ def jit(
                         backward_fn,
                         backward_traces,
                         _return_none_instead_of_grads,
+                        _vnilla_args,
                     ) = cache_entry
                     try:
                         cs.last_prologue_execution_start = time.perf_counter_ns()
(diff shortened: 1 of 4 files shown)


thunder/core/transform_common.py
```diff
@@ -750,11 +752,40 @@ def computation(x):
...
+        if is_functionalizable(bsym):
+            any_functionalizable = True
+        bsym_to_idx[bsym] = idx
+        idx_to_bsym[idx] = bsym
+
+    if not any_functionalizable:
         return []
 
+    # Step 0: Detect and error out aliased func args/kwargs. sparse tensors are ignored.
+    flat_args, _ = tree_flatten((computation_trace.args, computation_trace.kwargs))
```
t-vi, line 772: I don't think we should need the tree flatten here (at least for the general jit)?

t-vi (changes requested):
Hi @crcrpar , thank you for working on this!

I am not sure that we're checking at the right place - and that we should be putting this on the TensorProxies:
- to my mind, we need to do the aliasing check on every invocation (including cached), with all the speed requirement implications, so maybe only for tensors where it is relevant...
- if we wanted to support this down the road (shared weights might be an important case depending on whether they come out as different or same args), we would need to check for this in the prologue (but this is future probably).

thunder/core/transform_common.py
IvanYashchuk: It's still allowed in Thunder to construct a trace with non-empty dictionary for keyword arguments. All transforms should be generic when possible and error out when it's designed to work with just one specially constructed trace.

thunder/__init__.py
```diff
@@ -396,6 +417,12 @@ def get_computation_and_inputs(*args, **kwargs):
         cache_info["no_grad_sync"] = no_grad_sync
         return_none_instead_of_grads = is_fsdp_enabled and no_grad_sync
 
+        # NOTE(crcrpar): If a callable is free from in-place ops whose operand is args and/or their views
+        # alaises wouldn't matter, thus it'd be better to nullify this entry in such cases.
+        # It however would require the functionalized computation trace to interact with `cache_info`,
+        # which seems to break the consistency of cache_info, leading to a failure in cache_info check.
+        cache_info["alias_tensor_indices"] = _alias_tensor_of_args_kwargs(*args, **kwargs)
```
t-vi, line 424: While you would not want to change the cache info, you can grab the prologue check for it and remove it. It would be better to actually not collect the information in this case, but I guess that's how it is for now.

crcrpar: If we were to support the combo of in-place ops and alias tensor args, I do find it better to have separate traces whether or not the args include alias tensors

PR #800, opened 2024-07-18
partial > extra kwargs

t-vi:
The extra_kwargs I added to the autograd helper raised an eyebrow with @nikitaved so we remove that here in favour of partial which keeps the logic where it belongs.


nikitaved (approved):
👍

PR #804, opened 2024-07-18
don't use inspect.signature, prefer code object

### LINKED ISSUE #803: replace use of inspect.signature in interpreter._setup_frame_and_run_python_function
Perhaps ironically, the signature of `inspect.Parameter.__init__` is rather wrong because it does not reflect the defaults for the two keyword-only arguments (`default` and `annotation`) even though the `__kwdefaults__` are giving them:
```
In: inspect.signature(inspect.Parameter.__init__)
Out: <Signature (self, name, kind, *, default, annotation)>

In : inspect.Parameter.__init__.__kwdefaults__
Out: {'default': inspect._empty, 'annotation': inspect._empty}
```

Jitting HuggingFace transformers BERT does hit this in `transformers.pytorch_utils.apply_chunking_to_forward`.

t-vi:
Fixes #803 

The reason this is needed is that inspect uses the (global) sentinel `inspect._empty` to signify absence of (e.g. default) values. But this means that we cannot properly inspect things in the inspect module itself, and libraries do this (e.g. transformers when running BERT).
  
This patch avoiding signature makes the problem of signatures inside inspect being potentially misleading go away.

### DIFF
--- a/thunder/core/interpreter.py
+++ b/thunder/core/interpreter.py
@@ -6238,19 +6238,41 @@ def _setup_frame_and_run_python_function(
 ):
     fn = unwrap(wrapped_fn)
 
+    # We bind the provided args + kwargs to the function parameters.
+    # In the local var names (co_varnames), we will first have the
+    # positional (-only and positional-or-kw) args, then the keyword-only
+    # args, then optionally *args and optionally then **kwargs
+    # (of the signature). Here we populate locals_dict with those values.
+    # TODO: could populate localsplus directly instead of going through
+    #       locals_dict
+
     code: CodeType = extract_code(fn)
 
-    # adjustments for "hidden" instructions (EXTENDED_ARGS, CACHE, ...)
     locals_dict: dict[str, Any] = {}
     defaults = (*(wrap_const(v) for v in fn.__defaults__),) if fn.__defaults__ is not None else ()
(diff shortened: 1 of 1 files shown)


thunder/core/interpreter.py
```diff
@@ -6238,41 +6238,114 @@ def _setup_frame_and_run_python_function(
...
-        if name.startswith("."):
-            locals_dict[name] = locals_dict.pop(f"implicit{name[1:]}")
+    locals_dict: dict[str, Any] = {}
+    defaults = (*(wrap_const(v) for v in fn.__defaults__),) if fn.__defaults__ is not None else ()
+    kwdefaults = {k: wrap_const(v) for k, v in fn.__kwdefaults__.items()} if fn.__kwdefaults__ is not None else {}
+    has_varargs = (fn.__code__.co_flags & inspect.CO_VARARGS) > 0
+    has_varkwargs = (fn.__code__.co_flags & inspect.CO_VARKEYWORDS) > 0
+
+    # for the function signature varargs, we just need keep track of the
+    # number of provided positional args in args, but to to pupulate the
```
lantiga, line 6258:
```suggestion
    # number of provided positional args in args, but to to populate the
```

lantiga (approved):
Looks great

PR #810, opened 2024-07-19
fix torch.autocast re-enable on prims

### LINKED ISSUE #725: torch.autocast context manager got broken for prims.linear
## 🐛 Bug

Before https://github.com/Lightning-AI/lightning-thunder/pull/705 Thunder applied autocast transform to all functions invoked within a torch.autocast region. In https://github.com/Lightning-AI/lightning-thunder/pull/705 the logic has changed to apply autocast dispatch only for "torchsymbols" introducing a regression for all other symbols.

Here's a reproducer that is run on before the above PR is merged and after:

```py
import thunder
import torch

def foo(x, w):
    return thunder.prims.linear(x, w, None)

device = torch.device("cuda")
with device:
    # fp32 inputs
    x, w = torch.randn(16, 16), torch.randn(16, 16)
    print(x.dtype, w.dtype)
    
    jfoo = thunder.jit(foo)
    jit_out = jfoo(x, w)

```
trace before c816506d4dc61dfb6da8bc6fe4c34de8a6706399:
```py
def computation(x, w):
  # x: "cuda:0 f32[16, 16]"
  # w: "cuda:0 f32[16, 16]"
  [t0, t1] = nvFusion0(w, x)
    # t0 = prims.convert_element_type(x, dtypes.bfloat16)  # t0: "cuda:0 bf16[16, 16]"
    # t1 = prims.convert_element_type(w, dtypes.bfloat16)  # t1: "cuda:0 bf16[16, 16]"
  del w, x
  t2 = torch.nn.functional.linear(t0, t1, None)  # t2: "cuda:0 bf16[16, 16]"
    # t2 = ltorch.linear(t0, t1, None)  # t2: "cuda:0 bf16[16, 16]"
      # t2 = prims.linear(t0, t1, None)  # t2: "cuda:0 bf16[16, 16]"
  del t0, t1
  return t2
```
trace after c816506d4dc61dfb6da8bc6fe4c34de8a6706399:
```py
def computation(x, w):
  # x: "cuda:0 f32[16, 16]"
  # w: "cuda:0 f32[16, 16]"
  t0 = torch.nn.functional.linear(x, w, None)  # t0: "cuda:0 f32[16, 16]"
    # t0 = ltorch.linear(x, w, None)  # t0: "cuda:0 f32[16, 16]"
      # t0 = prims.linear(x, w, None)  # t0: "cuda:0 f32[16, 16]"
  del x, w
  return to
```

cc @crcrpar @apaz-cli

kshitij12345:
Fixes #725 

Changes
* Have mainly moved the logic of applying autocast rule from `torchsymbol` to `Symbol.call`. Though the utility to apply this rule lives in `transforms.py`.
* Added a `disable_autocast` ctx manager to disable re-applying autocast rule on the symbol.

Maybe moving forward if we see similar cases where we want to preprocess a symbol while generating the first trace, we can have a more general dispatch mechanism where we apply multiple of such transformations (though this would require us to be careful about the order of application). As per offline discussion with Ivan, this could be useful extension point for inserting the dist_prims.synchronize in distributed setting.

cc: @IvanYashchuk

### DIFF
--- a/thunder/core/transforms.py
+++ b/thunder/core/transforms.py
@@ -4011,11 +4011,11 @@ def autocast(func: Callable, dtype: dtypes.dtype):
 
 
 # State to enable and disable autocast (while interpreter is generating the computation trace).
-autocast_enabled = True
+_autocast_enabled = True
 
 
 def is_autocast_enabled():
-    return autocast_enabled
+    return _autocast_enabled
 
 
 # NOTE: Disable Autocast
@@ -4024,13 +4024,13 @@ def is_autocast_enabled():
 # Without disabling autocast we will have infinite recursion.
 @contextmanager
 def disable_autocast():
-    global autocast_enabled
-    default = autocast_enabled
-    autocast_enabled = False
+    global _autocast_enabled
+    default = _autocast_enabled
+    _autocast_enabled = False
     try:
         yield
     finally:
-        autocast_enabled = default
+        _autocast_enabled = default
 
 
 def maybe_apply_autocast(sym):
--- a/thunder/core/transforms.py
+++ b/thunder/core/transforms.py
@@ -1,5 +1,5 @@
 from collections import namedtuple
-from contextlib import nullcontext
+from contextlib import nullcontext, contextmanager
 from dataclasses import dataclass, replace
 from enum import auto, Enum
 from itertools import chain, compress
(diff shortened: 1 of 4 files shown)


thunder/core/transforms.py
```diff
@@ -18,6 +18,7 @@
 import os
 import dataclasses
 
+import thunder
```
IvanYashchuk, line 21: I'm surprised that there are no import errors doing this!

IvanYashchuk (approved):
Thank you!

t-vi (approved):

PR #814, opened 2024-07-20
skip `test_torch_compile_cat_nvfuser_phi2_tanh` if bf16 is not available

crcrpar:


thunder/tests/test_torch_compile_executor.py
```diff
@@ -34,6 +34,9 @@ def test_torch_compile_litgpt():
 @pytest.mark.skipif(not is_inductor_supported(), reason="inductor unsupported")
 @requiresCUDA
 def test_torch_compile_cat_nvfuser_phi2_tanh():
+    if not torch.cuda.is_bf16_supported():
+        pytest.skip("Your CUDA device does not support bfloat16")
```
t-vi, line 38: Nit1: I think I'd prefer a skipif decorator (but probably below the requiresCUDA).
Nit2: Is this `torch.cuda.is_bf16_supported()` we want or `thunder.tests.bf16.device_supports_bf16("cuda")`?

crcrpar: nit1: what would be the difference you'd see by making this a decorator? consistency with the existing skipif?
nit2: wouldn't the latter thunder's be just a wrapper of `is_bf16_supported`?

t-vi: 1. To my mind, we have mostly standardized to skip_if decorators except for things that are exceedingly complicated.
2. @tfogal added the thunder.tests.bf16 because the torch check was too generous (reported support even if it's just emulated or so).

crcrpar:
i'm checking this on a machine with volta gpus

thunder/tests/test_torch_compile_executor.py
tfogal: Yeah, as TomV mentions, torch's test is just if bf16 can be "stored", not that bf16 can actually be used. I agree that it is confusing, sorry :-(. As Tom mentions, Torch will happily tell us that bf16 is usable on Volta.

tfogal (approved):

t-vi:
Thank you @crcrpar @tfogal

PR #817, opened 2024-07-22
Explicitly define the RNG state tensor CPU tensor to avoid failure in CUDA device context

kiya00:
## What does this PR do?

I didn't explicitly specify the state to be a cpu tensor when using torch.tensor in the rng state, it leads to failure when using the cuda device context manager. It can be fixed by setting the tensor to be a cpu tensor

## PR review

## Did you have fun?


t-vi (approved):
Thank you @kiya00

Issue #812, opened 2024-07-20
Dtype mismatch in cat: bfloat16 and float16


tfogal:
## 🚀 Model / language coverage

First, I applied this diff to thunder:
```
diff --git a/thunder/core/utils.py b/thunder/core/utils.py
index 271dcdf3..346f264e 100644
--- a/thunder/core/utils.py
+++ b/thunder/core/utils.py
@@ -237,6 +237,10 @@ def check_same_dtype(*args):
             if dtype is None:
                 dtype = typ

+            if not are_same_dtypes(dtype, typ):
+                import traceback
+                print(f"mismatched types: {dtype}, {typ}")
+                traceback.print_stack()
             check(
                 are_same_dtypes(dtype, typ),
                 lambda: f"Expected dtype {dtype} but found {typ}!",
```

The diff was necessary to get the beginning of the output below, which conveys that a `cat` operator is what is at fault:

```
  File "/home/tfogal/dev/thunder/thunder/core/langctxs.py", line 132, in _fn
    result = fn(*args, **kwargs)
  File "/home/tfogal/dev/thunder/thunder/torch/__init__.py", line 812, in cat
    return clang.cat(tensors, dim)
  File "/home/tfogal/dev/thunder/thunder/core/langctxs.py", line 132, in _fn
    result = fn(*args, **kwargs)
  File "/home/tfogal/dev/thunder/thunder/clang/__init__.py", line 1289, in cat
    return prims.cat(tensors, dim)
  File "/home/tfogal/dev/thunder/thunder/core/symbol.py", line 272, in __call__
    result = self.meta(*args, **kwargs)
  File "/home/tfogal/dev/thunder/thunder/core/langctxs.py", line 132, in _fn
    result = fn(*args, **kwargs)
  File "/home/tfogal/dev/thunder/thunder/core/prims.py", line 2983, in cat_meta
    utils.check_same_dtype(*tensors)
  File "/home/tfogal/dev/thunder/thunder/core/utils.py", line 243, in check_same_dtype
    traceback.print_stack()
mismatched types: thunder.dtypes.bfloat16, thunder.dtypes.float16
[...]

PR #819, opened 2024-07-22
cat: support inputs with mixed dtypes

kshitij12345:
Fixes : https://github.com/Lightning-AI/lightning-thunder/issues/812

TODO - Look into `test_vjp_correctness` failure

### DIFF
--- a/thunder/clang/__init__.py
+++ b/thunder/clang/__init__.py
@@ -1286,7 +1286,20 @@ def unfold(a: TensorProxy, /, dim: int, size: int, step: int) -> TensorProxy:
 @clangop()
 def cat(tensors: list[TensorProxy], dim: int):
     """Concatenates the given sequence of tensors in the given dimension."""
-    return prims.cat(tensors, dim)
+    # Upcast to tensors only if we have more than 1 tensor.
+    # NumPy and PyTorch support upcasting with mixed dtypes.
+    if len(tensors) > 1:
+        _, output_dtype = utils.elementwise_type_promotion(
+            *tensors, type_promotion_kind=utils.ELEMENTWISE_TYPE_PROMOTION_KIND.PRESERVE
+        )
+        promoted_tensors = []
+        for t in tensors:
+            if t.dtype != output_dtype:
+                t = prims.convert_element_type(t, output_dtype)
+            promoted_tensors.append(t)
+    else:
+        promoted_tensors = tensors
+    return prims.cat(promoted_tensors, dim)
 
 
 @clangop()
@@ -1299,7 +1312,7 @@ def stack(tensors: list[TensorProxy], dim: int):
             s == shapes[0], lambda: f"tensors must be of the same shape, tensor at {i} is {s} instead of {shapes[0]}"
         )
     tensors_ = [unsqueeze(t, dim) for t in tensors]
-    return prims.cat(tensors_, dim)
+    return cat(tensors_, dim)
 
 
 @clangop()
--- a/thunder/clang/__init__.py
+++ b/thunder/clang/__init__.py
@@ -1286,7 +1286,7 @@ def unfold(a: TensorProxy, /, dim: int, size: int, step: int) -> TensorProxy:
 @clangop()
 def cat(tensors: list[TensorProxy], dim: int):
     """Concatenates the given sequence of tensors in the given dimension."""
-    # Upcast to tensors only if we have more than 1 tensor.
+    # Upcast tensors only if we have more than 1 tensor.
     # NumPy and PyTorch support upcasting with mixed dtypes.
     if len(tensors) > 1:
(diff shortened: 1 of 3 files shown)


kshitij12345:
test_vjp_correctness failure : `check_vjp` verifies the output against numerically computed Jacobian. However, with mixed inputs dtypes test (float and double) input, the numerical differentiation output is slightly different (for the float part of inputs) leading to mismatch. I would think that this is expected and we should probably just increase the tolerance for the test (or maybe if there is a way to increase tolerance only for this sample).

cc: @IvanYashchuk (original author of test)

Smaller Repro
```python
from functools import partial
import torch
import thunder
from thunder.tests.make_tensor import make_tensor, make_tensor_like
from thunder.core.pytree import tree_map, tree_flatten
from thunder.tests.test_grad import numerical_jvp, _dot, vjp as thunder_vjp, Sequence, flatten_func, _make_differentiable_wrapper


# Copied from `test_grad` with minor changes to support torch eager
# and to create `u` and `v` as ones tensor (for easier reasoning).
def check_vjp(f, *primals, comp, eager=False):
    """Check that the vector-Jacobian product of a function is correct.

    Args:
        f (callable): The function to differentiate.
        *primals (torch.Tensor): The input tensors.
        executor (str): The executor to use. Defaults to "torch".
        atol (float): Absolute tolerance. Defaults to None.
        rtol (float): Relative tolerance. Defaults to None.

    Raises:
        AssertionError: If the vector-Jacobian product is not correct.
    """
    # Let f be a function from vectors of size n to vectors of size m.
    # Its Jacobian is a matrix J of size m x n.
    # The adjoint property is J^* J = I, where J^* is the conjugate transpose (adjoint) of J.
    # J^* is a matrix of size n x m.
    # For any vector v of size m, J^* v is a vector of size n.
    # For any vector u of size n, J u is a vector of size m.
    # The dot product of J^* v and u is the same as the dot product of v and J u.
    # This function checks that the dot product of J^* v and u is the same as the dot product of v and J u.
    # 〈J u, v〉 == 〈u, J* v〉
    # Since u and v can be arbitrary, we take u = rand_like(primals), and v = rand_like(f(primals)).
    # We compute J u using numerical_jvp, and J* v using Thunder's vjp. That way we check correctness of Thunder's vjp.
    # Using finite differences we can compute J u, but we can't compute J* v, without computing full J, which is expensive.

    # make = partial(make_tensor_like, low=0, high=1)
    make = partial(torch.ones_like)

    u = tree_map(make, primals)
    if eager:
      outs_p, J_u = numerical_jvp(f)(primals, u)
    else:
      outs_p, J_u = numerical_jvp(thunder.compile(f, disable_torch_autograd_support=True, disable_preprocessing=True))(primals, u)

    multiple_results = isinstance(outs_p, Sequence)

    v = tree_map(make, outs_p)
    if eager:
      _, J_star_v = torch.autograd.functional.vjp(f, primals, v)
    else:
      _, J_star_v = thunder.compile(thunder_vjp(f), disable_torch_autograd_support=True, disable_preprocessing=True)(primals, v)

    if not multiple_results:
        v = (v,)
        J_u = (J_u,)

    J_u_v = _dot(J_u, v)
    u_J_star_v = _dot(u, J_star_v)
    if J_u_v.isnan().any():
        # TODO: find a better way to handle NaNs in finite differences
        return  # skip this sample
    comp(J_u_v, u_J_star_v)


# Check thunder - torch.cat
f = thunder.torch.cat
primals = ((torch.ones(1, requires_grad=True), torch.ones(1, dtype=torch.double, requires_grad=True)),)
kwargs = {"dim": 0}
flat_op, flat_args, spec = flatten_func(f, primals, kwargs)

filtered_op, filtered_args = _make_differentiable_wrapper(flat_op, flat_args)
# AssertionError: Scalars are not close!

# Expected 2.0 but got 1.999905726350911.
# Absolute difference: 9.427364908898284e-05 (up to 1e-07 allowed)
# Relative difference: 4.713682454449142e-05 (up to 1e-07 allowed)
check_vjp(filtered_op, *filtered_args, comp=torch.testing.assert_close)

# Check eager - torch.cat
f = torch.cat
flat_op, flat_args, spec = flatten_func(f, primals, kwargs)
filtered_op, filtered_args = _make_differentiable_wrapper(flat_op, flat_args)

# AssertionError: Scalars are not close!

# Expected 2.0 but got 1.999905726350911.
# Absolute difference: 9.427364908898284e-05 (up to 1e-07 allowed)
# Relative difference: 4.713682454449142e-05 (up to 1e-07 allowed)
check_vjp(filtered_op, *filtered_args, comp=torch.testing.assert_close, eager=True)

```

t-vi (approved):
Looks great.

t-vi:
Thank you @kshitij12345

Issue #818, opened 2024-07-22
`prims.div` inconsistencies between executors for integer tensors


nikitaved:
Likely the cause of https://github.com/Lightning-AI/lightning-thunder/issues/808

```python
In [1]: import torch, thunder

In [2]: def f(x, y):
   ...:     return thunder.prims.div(x, y)
   ...: 

In [3]: x = torch.randint(1, 10, (3, 3), device="cuda")

In [4]: y = torch.randint(1, 10, (3, 3), device="cuda")

In [5]: jf = thunder.jit(f)

In [6]: jf(x.cpu(), y.cpu())
Out[6]: 
tensor([[0, 1, 1],
        [0, 0, 2],
        [0, 0, 0]])

In [7]: jf(x, y)
Out[7]: 
tensor([[0.2500, 1.3333, 1.0000],
        [0.3333, 0.1250, 2.2500],
        [0.1429, 0.3750, 0.5000]], device='cuda:0')

In [8]: trace = thunder.last_traces(jf)[-1]

In [9]: print(trace)
# Constructed by Delete Last Used (took 0 milliseconds)
import torch
from thunder.executors.torchex import no_autocast

@torch.no_grad()
@no_autocast
def computation(x, y):
  # x: "cuda:0 i64[3, 3]"
  # y: "cuda:0 i64[3, 3]"
  [t0] = nvFusion0(x, y)
    # t0 = prims.div(x, y)  # t0: "cuda:0 i64[3, 3]"
  del x, y
  return t0

```

cc @apaz-cli

Issue #808, opened 2024-07-19
Integer divison `a // b` is broken with NVFuser. Broken decomposition and/or meta?


nikitaved:
Take a look at the following repro:
```python
In [1]: import torch, thunder

In [2]: def f(x, y):
   ...:     return x // y
   ...: 

In [3]: x = torch.randint(1, 10, (3, 3), device="cuda")

In [4]: y = torch.randint(1, 10, (3, 3), device="cuda")

In [5]: f(x, y)
Out[5]: 
tensor([[0, 0, 0],
        [0, 1, 0],
        [0, 1, 4]], device='cuda:0')

In [6]: jf = thunder.jit(f)

In [7]: jf(x, y)
Out[7]: 
tensor([[0.8889, 0.7500, 0.5000],
        [0.4444, 1.0000, 0.4286],
        [0.3333, 1.5000, 4.0000]], device='cuda:0')

In [8]: thunder.last_traces(jf)[-1]
Out[8]: 
# Constructed by Delete Last Used (took 0 milliseconds)
import torch
from thunder.executors.torchex import no_autocast

@torch.no_grad()
@no_autocast
def computation(x, y):
  # x: "cuda:0 i64[3, 3]"
  # y: "cuda:0 i64[3, 3]"
  [result] = nvFusion0(x, y)
    # t0 = prims.signbit(x)  # t0: "cuda:0 b8[3, 3]"
    # t1 = prims.signbit(y)  # t1: "cuda:0 b8[3, 3]"
    # t2 = prims.ne(t0, t1)  # t2: "cuda:0 b8[3, 3]"
    # t3 = prims.fmod(x, y)  # t3: "cuda:0 i64[3, 3]"
    # t4 = prims.ne(t3, 0)  # t4: "cuda:0 b8[3, 3]"
    # t5 = prims.bitwise_and(t2, t4)  # t5: "cuda:0 b8[3, 3]"
    # t6 = prims.div(x, y)  # t6: "cuda:0 i64[3, 3]"
    # t7 = prims.convert_element_type(t5, dtypes.int64)  # t7: "cuda:0 i64[3, 3]"
    # result = prims.sub(t6, t7)  # result: "cuda:0 i64[3, 3]"
  del x, y
  return result
```
[...]

PR #821, opened 2024-07-22
`prims.div`: equalize NVFuser prims with PyTorch

nikitaved:
NVFuser implementation of `prims.div` used `reciprocal`. That caused upcasting for integer inputs and, ultimately, rendered meta checks incorrect.

Fixes https://github.com/Lightning-AI/lightning-thunder/issues/808
Fixes https://github.com/Lightning-AI/lightning-thunder/issues/818

### DIFF
--- a/thunder/tests/test_nvfuser.py
+++ b/thunder/tests/test_nvfuser.py
@@ -961,3 +961,33 @@ def test_rm_unused_inputs_of_nvfusion(executor, device, _):
     out_ref = foo(t, ab)
 
     assert out.equal(out_ref)
+
+
+# TODO: we should improve our consistency testing
+# to also include checks for the result of meta functions.
+@instantiate(
+    dtypes=(thunder.int64, thunder.int32),
+    executors=(nvFuserExecutor,),
+)
+def test_div_truediv_integer_tensors_consistency_nvfuser(executor, device, thunder_dtype):
+    dtype = ltorch.to_torch_dtype(thunder_dtype)
+
+    def div(a, b):
+        return thunder.prims.div(a, b)
+
+    def truediv(a, b):
+        return a // b
+
+    def make_integer_tensor():
+        half_len = 5
+        t = torch.tensor([*range(-half_len, 0), *range(1, half_len + 1)], device=device, dtype=dtype)
+        perm = torch.randperm(2 * half_len)
+        return t[perm]
+
+    x = make_integer_tensor()
+    y = make_integer_tensor()
+
+    for f in (thunder.jit(div), thunder.jit(truediv)):
+        rout = f(x.cpu(), y.cpu()).to(device)
+        jout = f(x, y)
+        assert rout.equal(jout)
--- a/thunder/executors/nvfuserex_impl.py
+++ b/thunder/executors/nvfuserex_impl.py
@@ -1740,14 +1740,17 @@ def bitwise_xor(a: TensorProxy | Number, b: TensorProxy | Number, *, fd: FusionD
 register_supported(PrimIDs.BITWISE_XOR, bitwise_xor, _elementwise_binary_check)
 
 
-# TODO nvFuser's div operation is not equivalent to the div primitive
-#   (mruberry) I need to investigate if nvFuser exposes a truncation division operation
 def div(a: TensorProxy | Number, b: TensorProxy | Number, *, fd: FusionDefinition, lc_to_nv_map: dict) -> Any:
     nva = getnv(a, fd, lc_to_nv_map)
     nvb = getnv(b, fd, lc_to_nv_map)
 
+    a_dtype = dtypes.to_dtype(a)
+    b_dtype = dtypes.to_dtype(b)
+
+    if dtypes.is_integer_dtype(a_dtype) and dtypes.is_integer_dtype(b_dtype):
+        return fd.ops.div(nva, nvb)
+
     # NOTE It's currently significantly faster for nvFuser to multiply the reciprocal than divide
-    # return fd.ops.div(nva, nvb)
     return fd.ops.mul(nva, fd.ops.reciprocal(nvb))
 
 


thunder/tests/test_nvfuser.py
```diff
@@ -961,3 +961,33 @@ def foo(t, ab):
...
+        return thunder.prims.div(a, b)
+
+    def truediv(a, b):
+        return a // b
+
+    def make_integer_tensor():
+        half_len = 5
+        t = torch.tensor([*range(-half_len, 0), *range(1, half_len + 1)], device=device, dtype=dtype)
+        perm = torch.randperm(2 * half_len)
+        return t[perm]
```
nikitaved, line 985: I will move these into OpInfo tests in a follow-up PR. We also have some inconsistencies with floating types. Wanna knock them one by one

t-vi (approved):
Thank you @nikitaved, awesome stuff!

thunder/tests/test_nvfuser.py
```diff
@@ -961,3 +961,33 @@ def foo(t, ab):
...
+    executors=(nvFuserExecutor,),
+)
+def test_div_truediv_integer_tensors_consistency_nvfuser(executor, device, thunder_dtype):
+    dtype = ltorch.to_torch_dtype(thunder_dtype)
+
+    def div(a, b):
+        return thunder.prims.div(a, b)
+
+    def truediv(a, b):
+        return a // b
```
mruberry, line 979: This is floor division, not true division

thunder/tests/test_nvfuser.py
```diff
@@ -961,3 +961,33 @@ def foo(t, ab):
...
+        perm = torch.randperm(2 * half_len)
+        return t[perm]
+
+    x = make_integer_tensor()
+    y = make_integer_tensor()
+
+    for f in (thunder.jit(div), thunder.jit(truediv)):
+        rout = f(x.cpu(), y.cpu()).to(device)
+        jout = f(x, y)
+        assert rout.equal(jout)
```
mruberry, line 993: I don't think these results should be equal (I think we're exposing a bug in the nvFuser implementation of `prims.div`). `prims.div` should implement truncation division, which is not the same as Python's floor division. See, for example, the PyTorch executor's implementation of prims.div:

https://github.com/Lightning-AI/lightning-thunder/blob/44d2f3c9f02166945b2d737b42d86f28e4133b65/thunder/executors/torchex.py#L907

mruberry: fyi @kevinstephano @jjsjann123

nikitaved: `div` calls to `prims.div`, so they should be equivalent.

thunder/tests/test_nvfuser.py
nikitaved: Good catch! Made a type - will fix that. Thank you!

thunder/tests/test_nvfuser.py
mruberry: Yes, but doesn't `truediv` call floor division?

nikitaved: Floor division has a decomp that is more that just `prims.div`

nikitaved: The goal for now is to be able to handle integer tensors. I plan a follow-up to address all dtypes and consistency in general. Currently, our OpInfos skip NVFuser tests plus vjp.

mruberry: I understand and I'm not arguing for a high priority change, I'd just like to get us on the same page about this detail.

When `prims.div(a, b)` is called then it should perform a truncation division.

When `a // b` is written then it should perform a floor division.

And truncation division and floor division are different operations.

That's why I'm suggesting that this test is exposing an issue. Does that make sense? Maybe a quick VC would be simpler.

thunder/tests/test_nvfuser.py
nikitaved: C division has always been the true division for me :)

thunder/tests/test_nvfuser.py
nikitaved: They are equivalent for integer tensors, no?

nikitaved: I am confused -- I am not comparing these two operations head to head, I just make sure that the cpu version is consistent with the nvfuser version for both `prims.div` and `* // *`.

mruberry: No, but it's a common misconception that they are. If the result is negative then they round differently. Floor division rounds door, truncation division rounds towards zero. 

In Python

```
-9 // 5
-2
```

In C++

```
#include <iostream>

int main() {
    int a = (-9 / 5);
    std::cout << a; 
}
```

prints -1. This is because -9 / 5 is -1.8, and floor division takes the floor of -1.8, which is -2, while truncation division truncates the fractional part of -1.8, giving -1.

mruberry: > fyi @kevinstephano @jjsjann123

Aha! @IvanYashchuk told me that nvFuser's div is performing truncation division for integers now. I mistakenly thought it wasn't. That's why this test passes

t-vi: while we are looking at torchex, I think the div prim impl has a typo (twice a instead of a and b) in the if condition:
https://github.com/Lightning-AI/lightning-thunder/blob/44d2f3c9f02166945b2d737b42d86f28e4133b65/thunder/executors/torchex.py#L908

mruberry: > while we are looking at torchex, I think the div prim impl has a typo (twice a instead of a and b) in the if condition:
> 
> https://github.com/Lightning-AI/lightning-thunder/blob/44d2f3c9f02166945b2d737b42d86f28e4133b65/thunder/executors/torchex.py#L908

I think you're right!

PR #823, opened 2024-07-22
test grad for normalize: increase eps to make test less flaky

t-vi:
I can run that test 100 times locally and not hit it, but let's not have the CI failure rate go up too much.


t-vi:
Expedited merge to make CI more reliable.

PR #828, opened 2024-07-23
avoid hard coding of output shape in distributed/test_ops

crcrpar:
## What does this PR do?

As per title, removed the hard coding of output shape from the recently added test using `torch.distributed` ops


t-vi (approved):
Thank you @crcrpar

PR #830, opened 2024-07-23
[inplace] move copy to after its dst's last use

### LINKED ISSUE #693: Move `copy` for arg's in-place update right after its last consumer
Would it make sense to try to move the copy as early as possible by analyzing usage, instead of defaulting to the end, since this will have memory implications?

Maybe we can open an issue for this. Also let's add an issue with the proposal for batchnorm.

_Originally posted by @lantiga in https://github.com/Lightning-AI/lightning-thunder/pull/675#pullrequestreview-2147432669_

crcrpar:
closes #693

### DIFF
--- a/thunder/core/transform_common.py
+++ b/thunder/core/transform_common.py
@@ -490,7 +490,7 @@ def _get_prod_bsym_with_arg(
 def collect_required_copy_bsyms_for_args(
     trace: Trace,
     removed_copy_bsyms: list[tuple[int, BoundSymbol]],
-) -> tuple[list[BoundSymbol], dict[VariableInterface, TensorProxy]]:
+) -> tuple[list[list[BoundSymbol]], dict[VariableInterface, TensorProxy]]:
     """Identify required pairs of copy src and copy dst."""
 
     flat_args: tuple[TensorProxy, ...] = tuple(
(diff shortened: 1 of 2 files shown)


thunder/tests/test_inplace_functionalization.py
```diff
@@ -355,6 +355,33 @@ def f_with_view(a):
     torch.testing.assert_close(x, x_ref)
 
 
+@instantiate(
+    dtypes=NOTHING,
+)
+def test_multiple_views_before_inplace_to_base(executor, device, _):
+
+    # ref: https://github.com/pytorch/pytorch/blob/29e2e2a/test/test_functionalization.py#L159-L169
+    def f(x):
```
crcrpar, line 364: I tried one test case following https://github.com/Lightning-AI/lightning-thunder/issues/657#issuecomment-2242979740, the main branch failed to create a correct trace but this pr seems to succeed

crcrpar: The trace however doesn't seem to copy the value of `x` after `add_` to the views explicitly, so it seems like that we implicitly depend on (a) thunder preserves `prims.copy_` if an operand of in-place ops is arg.

thunder/core/transform_common.py
```diff
@@ -854,14 +853,33 @@ def computation(x):
...
-        removed_copy_bsyms,
-    )
     functionalized_computation_trace = from_trace(computation_trace)
-    return_bsym = new_bsyms[-1].from_bsym_swap_proxies(swap_map_for_return)
-    functionalized_computation_trace.bound_symbols = new_bsyms[:-1] + required_copy_bsyms + [return_bsym]
     functionalized_computation_trace.set_provenance(TraceProvenance("Functionalize in-place ops"))
+    required_copy_bsyms = collect_required_copy_bsyms_for_args(intermediate_trace, removed_copy_bsyms)
+
+    producer_map = producers(new_bsyms[:-1])
+    bsym_to_copy_bsyms: dict[BoundSymbol, list[BoundSymbol]] = {}
```
t-vi, line 861: I know I'm bad at this myself, but could add a comment to this saying (if I understand it correctly) that these are the bsyms not needing processing to follow the processed bsym that is the key?

I must admit I wonder a bit if this could be clearer if the two loops below could be just a loop over the bsyms doing the processing or copying.

thunder/tests/test_inplace_functionalization.py
t-vi: Hm. I'm not sure I understand this and if it is safe...

t-vi (commented):
Thank you @crcrpar . Sorry about the delay, it always takes me a while to understand what's going on (entirely because of me). I have a few smallish comments and then it seems good to merge.

t-vi (approved):
lgtm. Thank you @crcrpar

Issue #826, opened 2024-07-22
Expected the batch dimensions of a (((384,),)) and the batch dimensions of b (()) to be the same


tfogal:
## 🚀 Model / language coverage

Examine fails on one of the NeVA subgraphs (as divided by torch.compile):

```
Expected the batch dimensions of a (((384,),)) and the batch dimensions of b (()) to be the same
```

### Pitch

This is for #343.

### Alternatives / Potential work-arounds

Assigning to me until I can get this into a reasonable state.

### Minimal Repro

PR #838, opened 2024-07-23
torch.matmul - fix autocast rule (and update the error message)

kshitij12345:
Fixes https://github.com/Lightning-AI/lightning-thunder/issues/826

Minimal Repro:
```python
import torch
import thunder

def foo(a, b):    
  return torch.matmul(a, b)

a = torch.rand([3, 1, 2], dtype=torch.float, device='cuda')
b = torch.rand([2, 3], dtype=torch.bfloat16, device='cuda')

  foo(a, b)
  jfoo = thunder.jit(foo)
  jfoo(a, b)

```

Reason for failure - autocast rule for matmul before this PR called `prims.matmul`. However, this is incorrect as `ltorch.matmul` takes care of reshaping inputs to be compliant with the correct shapes expected by prims.matmul (and this step was skipped when autocast rule was applied).

Fix - Update autocast ltorch.matmul rule to call `ltorch.matmul` after casting the dtypes.

**NOTE** - Have also tweaked the error message from prim.matmul to remove extra parentheses. The new error message now reads - `Expected the batch dimensions of a (3,) and the batch dimensions of b () to be the same`.

Generated Trace for above snippet (with patch from this PR) - 
```python
import torch

@torch.no_grad()
@no_autocast
def computation(a, b):
  # a: "cuda:0 f32[3, 1, 2]"
  # b: "cuda:0 bf16[2, 3]"
  [t0] = nvFusion0(a)
    # t0 = prims.convert_element_type(a, dtypes.bfloat16)  # t0: "cuda:0 bf16[3, 1, 2]"
  del a
  t2 = torch.matmul(t0, b)  # t2: "cuda:0 bf16[3, 1, 3]"
    # t2 = ltorch.matmul(t0, b)  # t2: "cuda:0 bf16[3, 1, 3]"
      # t5 = prims.broadcast_in_dim(b, (3, 2, 3), (1, 2))  # t5: "cuda:0 bf16[3, 2, 3]"
      # t2 = prims.matmul(t0, t5)  # t2: "cuda:0 bf16[3, 1, 3]"
  del t0, b
  return t2
```

### DIFF
--- a/thunder/tests/test_autocast.py
+++ b/thunder/tests/test_autocast.py
@@ -232,3 +232,27 @@ def test_autocast_convolution(dim, requires_grad):
 
         for eg, jg in zip(eager_grads, jit_grads):
             torch.testing.assert_close(eg, jg, rtol=5e-2, atol=5e-2)
+
+
+@pytest.mark.parametrize("requires_grad", [False, True])
+def test_autocast_torch_matmul(requires_grad):
+    def foo(a, b):
+        output = torch.matmul(a, b)
+        return output
+
+    a = torch.rand([3, 1, 2], dtype=torch.float, device="cpu", requires_grad=requires_grad)
+    b = torch.rand([2, 3], dtype=torch.bfloat16, device="cpu", requires_grad=requires_grad)
+
+    with torch.autocast("cpu", torch.bfloat16):
+        expected = foo(a, b)
+        actual = thunder.jit(foo)(a, b)
+
+    torch.testing.assert_close(actual, expected)
+
+    if requires_grad:
+        go = torch.randn_like(expected)
+        eager_grads = torch.autograd.grad(expected, [a, b], go)
+        jit_grads = torch.autograd.grad(actual, [a, b], go)
+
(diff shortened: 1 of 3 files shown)


thunder/tests/test_autocast.py
```diff
@@ -232,3 +232,27 @@ def foo(x, w, b=None):
...
             torch.testing.assert_close(eg, jg, rtol=5e-2, atol=5e-2)
+
+
+@pytest.mark.parametrize("requires_grad", [False, True])
+def test_autocast_torch_matmul(requires_grad):
+    def foo(a, b):
+        output = torch.matmul(a, b)
+        return output
+
+    a = torch.rand([3, 1, 2], dtype=torch.float, device="cpu", requires_grad=requires_grad)
```
mruberry, line 243: It might be interesting to test both a float x bfloat16 input, and a float x float input, on both cpu and cuda devices

mruberry (approved):

t-vi:
Thank you @kshitij12345 @mruberry

PR #847, opened 2024-07-23
bump to pytorch 2.3.1

t-vi:
our docker build fails for stable with missing symbol from nvfuser vs. PyTorch maybe this helps


Borda (approved):

t-vi:
Per the stable docker build failing, this is not quite what we are looking for yet.

@xwang233 I could probably use a hint what PyTorch I need to play well with the nvfuser from pip.

PR #848, opened 2024-07-24
grad transform to properly identify proxy on saved_for_backward

### LINKED ISSUE #541: grad transform `forward_and_backward_from_trace` is not handling NumberProxy properly in saved_for_backward
## 🐛 Bug

construct_trace used in grad transform mistakenly converts all saved_for_backward into proxies.

https://github.com/Lightning-AI/lightning-thunder/blob/0342223ac7851beb07d7df731389db777b58f1ac/thunder/core/transforms.py#L74-L77
https://github.com/Lightning-AI/lightning-thunder/blob/0342223ac7851beb07d7df731389db777b58f1ac/thunder/core/transforms.py#L3619-L3620

For a simple program like this:
```py
import torch
import thunder

def foo(t, ab):
    return t * ab

jfoo = thunder.jit(foo)

dtype = torch.float32
t = torch.randn(5, 3, device="cuda").to(dtype=dtype)
t.requires_grad_()

ab = 0.5

out = jfoo(t, ab)
```

The transform gives us:
```py
===fwd trc===
import thunder
import torch

@torch.no_grad()
@no_autocast
def augmented_forward_fn(t):
  # t: "cuda:0 f32[5, 3]"
  t0 = ltorch.mul(t, 0.5)  # t0: "cuda:0 f32[5, 3]"
    # t0 = prims.mul(t, 0.5)  # t0: "cuda:0 f32[5, 3]"
  return ({'output': t0, 'flat_args': [t], 'flat_output': (t0,)}, (((t, 0.5), None, ([0.5],)),))

=== saved_for_backward=(((t, 0.5), None, ([0.5],)),)

===bwd trc===
import thunder
import torch

@torch.no_grad()
@no_autocast
  # cotangents: "Collection"
  C2, = saved_for_backward
  t1, = cotangents
  _, _, C4, = C2
  C5, = C4
  f1, = C5
  t7 = ltorch.mul(f1, t1)  # t7: "cuda:0 f32[5, 3]"
    # t7 = prims.mul(f1, t1)  # t7: "cuda:0 f32[5, 3]"
  return [t7]

--- bw_flat_saved_for_backward=[t, [FloatProxy name=f0, value=0.5], None, [FloatProxy name=f1, value=0.5]]
```

We have `saved_for_backward=(((t, 0.5), None, ([0.5],)),)`, which was later translated to `bw_flat_saved_for_backward=[t, [FloatProxy name=f0, value=0.5], None, [FloatProxy name=f1, value=0.5]]` in backward trace and that is not right.

The backward trace is treating that as a proxy `f1`, instead of baking static numbers in.

### Context

We need to support NumberProxy passed from fwd to bwd. Currently all NumberProxies are baked in as constant here: https://github.com/Lightning-AI/lightning-thunder/blob/0342223ac7851beb07d7df731389db777b58f1ac/thunder/core/transforms.py#L3449

When we remove that line, it breaks rematerialization pass, since it assumed `saved_for_backward` is consistent between forward and backward.
https://github.com/Lightning-AI/lightning-thunder/blob/0342223ac7851beb07d7df731389db777b58f1ac/thunder/core/rematerialization.py#L635
https://github.com/Lightning-AI/lightning-thunder/blob/0342223ac7851beb07d7df731389db777b58f1ac/thunder/core/rematerialization.py#L643

Note that it's using the same `new_required_for_backward` for both forward and backward. In the example above, we'll have fwd trace transformed like
```py
fw_trace=# Constructed by Transform for execution (took 1 milliseconds)
import torch

@torch.no_grad()
@no_autocast  
def augmented_forward_fn(t):
  # t: "cuda:0 f32[5, 3]"
  [t0] = nvFusion0(t)
    # t0 = prims.mul(t, 0.5)  # t0: "cuda:0 f32[5, 3]"
  return {'output': t0, 'flat_args': [t], 'flat_output': (t0,)}, ((), (0.5,))

===
new_fw_trace=# Constructed by Rematerialization
import torch

@torch.no_grad()
@no_autocast
def augmented_forward_fn(t):
  # t: "cuda:0 f32[5, 3]"
  [t0] = nvFusion0(t)
    # t0 = prims.mul(t, 0.5)  # t0: "cuda:0 f32[5, 3]"
  return {'output': t0, 'flat_args': [t], 'flat_output': (t0,)}, ((), (f1,))
```

Where the new_fw_trace is trying to save a `f1` NumberProxy, which is never a number proxy on fwd graph.

jjsjann123:
Fixes #541

backward trace needs to correctly use number proxy and scalar number in `save_for_backward`. Right now it's not the case where `thunder.trace` used to construct backward trace converts all inputs to proxies.

This PR restores proper proxy-ness by examine the `save_for_backward` returned by forward trace.

### DIFF
--- a/thunder/core/transforms.py
+++ b/thunder/core/transforms.py
@@ -3737,11 +3737,14 @@ def forward_and_backward_from_trace(trace: Trace, torch_autograd=False) -> Forwa
     flat_saves, _ = tree_flatten(saved_for_backward)
 
     def backward_fn(saved_for_backward, cotangents):
-        flat_saves_proxified, saves_spec  = tree_flatten(saved_for_backward)
-        flat_filtered = [proxified if isinstance(entry, Proxy) else entry for proxified, entry in zip(flat_saves_proxified, flat_saves)]
+        flat_saves_proxified, saves_spec = tree_flatten(saved_for_backward)
+        flat_filtered = [
+            proxified if isinstance(entry, Proxy) else entry
+            for proxified, entry in zip(flat_saves_proxified, flat_saves)
+        ]
         saved_for_backward = tree_unflatten(flat_filtered, saves_spec)
         env = reconstruct_forward_env_for_backward(trace, saved_for_backward)
-        
+
         if torch_autograd:
             cotangents = tree_unflatten(cotangents, output_spec)
         out = backward_pass(env, trace, cotangents)
--- a/thunder/core/transforms.py
+++ b/thunder/core/transforms.py
@@ -3737,6 +3737,7 @@ def forward_and_backward_from_trace(trace: Trace, torch_autograd=False) -> Forwa
     flat_saves, _ = tree_flatten(saved_for_backward)
 
     def backward_fn(saved_for_backward, cotangents):
+        # trace converts all saved_for_backward into proxy, we want to restore number scalars afterwards.
         flat_saves_proxified, saves_spec = tree_flatten(saved_for_backward)
         flat_filtered = [
             proxified if isinstance(entry, Proxy) else entry
@@ -3756,7 +3757,6 @@ def forward_and_backward_from_trace(trace: Trace, torch_autograd=False) -> Forwa
             out = tree_flatten(out)[0]
         return out
 
-    print("-- saved for backward ", saved_for_backward)
     backward_trace = construct_trace(rename_proxies=False)(backward_fn, saved_for_backward, cotangents)
 
     # We are done with constructing the forward and backward passes at this
--- a/thunder/core/transforms.py
+++ b/thunder/core/transforms.py
@@ -3733,8 +3733,15 @@ def forward_and_backward_from_trace(trace: Trace, torch_autograd=False) -> Forwa
     finally:
         reset_tracectx(tracectx_token)
 
+    saved_for_backward = forward_trace.output[1]
+    flat_saves, _ = tree_flatten(saved_for_backward)
+
     def backward_fn(saved_for_backward, cotangents):
+        flat_saves_proxified, saves_spec  = tree_flatten(saved_for_backward)
+        flat_filtered = [proxified if isinstance(entry, Proxy) else entry for proxified, entry in zip(flat_saves_proxified, flat_saves)]
+        saved_for_backward = tree_unflatten(flat_filtered, saves_spec)
         env = reconstruct_forward_env_for_backward(trace, saved_for_backward)
+        
         if torch_autograd:
             cotangents = tree_unflatten(cotangents, output_spec)
         out = backward_pass(env, trace, cotangents)
@@ -3746,7 +3753,7 @@ def forward_and_backward_from_trace(trace: Trace, torch_autograd=False) -> Forwa
             out = tree_flatten(out)[0]
         return out
 
-    saved_for_backward = forward_trace.output[1]
+    print("-- saved for backward ", saved_for_backward)
     backward_trace = construct_trace(rename_proxies=False)(backward_fn, saved_for_backward, cotangents)
 
     # We are done with constructing the forward and backward passes at this
--- a/thunder/tests/test_examine_memory.py
+++ b/thunder/tests/test_examine_memory.py
@@ -161,7 +161,7 @@ def test_nanogpt_block(executor, device, dtype):
         expected_return_calculated_mem = get_return_memory(fw_extrace.bound_symbols[-1]) - 4 * 2 * 1024 * 768
         assert expected_return_calculated_mem == sum(fw_alloc_mem[1].values())
 
-        assert bw_alloc_mem[0] == 361881600
+        assert bw_alloc_mem[0] == 361783296
         assert sum(bw_alloc_mem[1].values()) == get_return_memory(bw_extrace.bound_symbols[-1])
     if isinstance(executor, TorchTestExecutor):
         assert fw_alloc_mem[0] == 362863616
--- a/thunder/tests/test_grad.py
+++ b/thunder/tests/test_grad.py
@@ -1618,22 +1618,25 @@ def test_make_forward_backward_symbol_caching_with_executor():
 
 
 def test_grad_transform_saved_for_backward_proxy():
+    from thunder.core.proxies import Proxy
 
     def foo(a, c):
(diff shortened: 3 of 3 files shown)


thunder/tests/test_examine_memory.py
```diff
@@ -161,7 +161,7 @@ def test_nanogpt_block(executor, device, dtype):
         expected_return_calculated_mem = get_return_memory(fw_extrace.bound_symbols[-1]) - 4 * 2 * 1024 * 768
         assert expected_return_calculated_mem == sum(fw_alloc_mem[1].values())
 
-        assert bw_alloc_mem[0] == 361881600
+        assert bw_alloc_mem[0] == 361783296
```
jjsjann123, line 164: didn't investigate. But I assumed using less memory is a good thing?!

tfogal (approved):
i'd recommend adding some type annotations, hard to understand the core/transforms.py code just from a diff.
but 🤷, nbd.

t-vi (approved):
Thank you @jjsjann123 @tfogal

thunder/core/transforms.py
```diff
@@ -3733,8 +3733,19 @@ def ones_like(x):
...
+    flat_saves, _ = tree_flatten(saved_for_backward)
+
     def backward_fn(saved_for_backward, cotangents):
+        # trace converts all saved_for_backward into proxy, we want to restore number scalars afterwards.
+        flat_saves_proxified, saves_spec = tree_flatten(saved_for_backward)
+        flat_filtered = [
+            proxified if isinstance(entry, Proxy) else entry
+            for proxified, entry in zip(flat_saves_proxified, flat_saves)
+        ]
+        saved_for_backward = tree_unflatten(flat_filtered, saves_spec)
```
IvanYashchuk, line 3746: These lines do not change the function signature of the backward function and the saved numbers are still passed to the backward function from forward but silently ignored.
Luckily, there is a postprocessing step that removes saved tensors/numbers that are unused in the backward trace:
https://github.com/Lightning-AI/lightning-thunder/blob/f08088e65361484e34ec4ba105ea1e7b91cf91e9/thunder/core/transforms.py#L3791-L3796

thunder/core/transforms.py
```diff
@@ -3746,7 +3757,6 @@ def backward_fn(saved_for_backward, cotangents):
             out = tree_flatten(out)[0]
         return out
 
-    saved_for_backward = forward_trace.output[1]
     backward_trace = construct_trace(rename_proxies=False)(backward_fn, saved_for_backward, cotangents)
```
IvanYashchuk, line 3760: We're lucky that rename_proxies=False is used here. Otherwise, all of the "saved_for_backward" proxies would be new proxies with new names and the approach wouldn't work.

thunder/tests/test_grad.py
```diff
@@ -1615,3 +1615,28 @@ def foo(a):
...
+        return a * c
+
+    a = make_tensor((2, 2), device="cpu", dtype=torch.float32, requires_grad=True)
+    c = 2.0
+
+    dynamic_jit = thunder.jit(foo, cache="symbolic values")
+    static_jit = thunder.jit(foo)
+
+    out = dynamic_jit(a, c)
+    torch.autograd.backward(out, torch.rand_like(out), retain_graph=True)
```
IvanYashchuk, line 1633: It's unnecessary to run backward. The backward trace is always generated for each forward run.

thunder/tests/test_grad.py
```diff
@@ -1615,3 +1615,28 @@ def foo(a):
...
+    dynamic_trace = thunder.last_backward_traces(dynamic_jit)[-1]
+    # dynamic trace should save `c` as proxy for backward
+    assert any(map(lambda x: isinstance(x, Proxy), tree_flatten(dynamic_trace.args[0])[0]))
+
+    out = static_jit(a, c)
+    torch.autograd.backward(out, torch.rand_like(out), retain_graph=True)
```
IvanYashchuk, line 1639: It's unnecessary to run it for the test to pass as intended.

thunder/core/transforms.py
IvanYashchuk, line 3746: tree_flatten -> map -> tree_unflatten can be handled with a single tree_map call

IvanYashchuk (commented):
That's a clever approach!

There are a few possible improvements but the ship has sailed.

tfogal:
> There are a few possible improvements but the ship has sailed.

Could always improve these in a follow-up PR :-)
@jjsjann123 could you take a look?

IvanYashchuk:
> i'd recommend adding some type annotations, hard to understand the core/transforms.py code just from a diff. but 🤷, nbd.

Do you want to see more type annotations for intermediate variables? This function and helper functions that it uses include type annotations for inputs and outputs and this function has a docstring as well.

tfogal:
> > i'd recommend adding some type annotations, hard to understand the core/transforms.py code just from a diff. but 🤷, nbd.
> 
> Do you want to see more type annotations for intermediate variables? This function and helper functions that it uses include type annotations for inputs and outputs and this function has a docstring as well.

Oh, great that we have annotations and docstrings here!

But yeah, in general I'm in favor of localizing context more, and types are one way to help with that. Jumping into a function's header works, but scales poorly and doesn't translate to the web all that well.

thunder/core/transforms.py
jjsjann123: > and the saved numbers are still passed to the backward function from forward but silently ignored.

yeah that's intended~

thunder/core/transforms.py
jjsjann123: Just for my education. Would you mind elaborate on how new names would affect this PR? I though even with renaming, the sequence/order of saved_for_backward would still remain the same and what I did here is just replacing according to their index. i.e. whether a proxy has been renamed or not, wouldn't really affect the code change.

thunder/core/transforms.py
jjsjann123: Let me try if tree_flatten would work with zipped trees~~

thunder/core/transforms.py
IvanYashchuk: You're right, I read the code incorrectly.

PR #852, opened 2024-07-24
test_change_default_device_with_ctx seems to flakily pass

kshitij12345:
Ref https://github.com/Lightning-AI/lightning-thunder/issues/853

`test_change_default_device_with_ctx` seems to flakily pass and fails otherwise - removing `strict=True`. Will investigate more into this.


t-vi:
I think it is not the test, but the Inés before...

kshitij12345:
#849 should mostly fix this.

PR #855, opened 2024-07-24
CUDAGraphs: fix issues with buffer initializations

nikitaved:
We forget to properly init buffers prior to warm-up runs.
That was not caught by our tests because the GPT model, apparently, does not run kernels conditioned on inputs' values.

Thank you, @apaz-cli , @lantiga , for your help in finding this issue!


t-vi (approved):
Thank you @nikitaved

PR #856, opened 2024-07-24
require nvfuser version 0.2.8

jjsjann123:
What's in this PR:
* requiring nvfuser version 0.2.8 for inplace patch;
* removing code for old nvfuser API change.

### DIFF
--- a/thunder/executors/nvfuserex.py
+++ b/thunder/executors/nvfuserex.py
@@ -30,7 +30,7 @@ def nvfuser_version() -> LooseVersion | None:
 
 
 def required_nvfuser_version() -> LooseVersion:
-    return LooseVersion("0.0.1")
+    return LooseVersion("0.2.8")
 
 
 def nvfuser_available() -> bool:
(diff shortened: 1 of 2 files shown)


jjsjann123:
marking this as draft to ensure that we have proper nvfuser packages published and available in CI.

t-vi:
Marking as ready as the conditions seem to apply.

t-vi (approved):
Thank you @jjsjann123

thunder/executors/nvfuserex.py
```diff
@@ -30,7 +30,7 @@ def nvfuser_version() -> LooseVersion | None:
 
 
 def required_nvfuser_version() -> LooseVersion:
```
IvanYashchuk, line 32: If nvFuser is installed but the version is not new enough should Thunder issue a warning about this?

t-vi: I think an error would be also good.

Issue #640, opened 2024-06-24
Transforms as EarlyTransforms


t-vi:
Currently we have [grad](https://github.com/Lightning-AI/lightning-thunder/blob/fa55b090d88d769b926f67ea71087c5660f730bc/thunder/core/transforms.py#L1430) and [noop](https://github.com/Lightning-AI/lightning-thunder/blob/fa55b090d88d769b926f67ea71087c5660f730bc/thunder/core/transforms.py#L510) as AdditionalTransforms, however we might want to move them to early transforms to be able to work with epilogue-carrying functions.

Also, we have vjp, vmap and others that currently do now work well with jitted models and for which we should also do something (i.e. figure out what it means to change the input signature), what do we want to do when a module is involved? One option could be to just work on the arguments and ignore the model in the background. (If we want to support that at all.)

My idea would be to adapt these to work on the traces directly rather than the current indirection.

I'm filing this issue to discuss this, and then we see how to move things forward.

mruberry:
triage review: the design tag marks issues where we want people to review and propose a design (possibly with a linked PR to see a sample implementation), so look for those and let's try to call them out

t-vi:
I know I was the one to have everything called transform_trace, but it turns out that the scheme of transforms and subclasses creates a need to differentiate that would not be needed if we had just the Transform class with methods for the various stages, so: 
```
class Transform:
    def early_transform_traces(self, prologue_trace: Trace, computation_trace: Trace, epilogue_trace: Trace | None, **kwargs):
    def transform_module(self, model: thunder.ThunderModule):
    def transform_state_dict_for_submodule(...):
    def additional_transform_trace(self, computation_trace: Trace, **kwargs):  # but to be phased out
    def post_optimization_transform_trace(self, computation_trace: Trace, **kwargs):
```
The methods would default to returning unmodified args.

This allows users to pass in transforms regardless of their implementation (we currently have additional_transforms / early_transforms as separate arguments in places).

PR #861, opened 2024-07-25
Unify transforms

t-vi:
cf #640 and the discussion on transforms in the tech sessions (need to add the summary to the 640)


thunder/core/transform_common.py
```diff
@@ -331,20 +331,16 @@ def thunder_140410131706304(x, y):
...
-    """
-    EarlyTransform enables transforming prologue, computation and epilogue trace.
-    Note that the computation trace here is before the autograd transform, so any update to
-    the computation trace will also update backward trace.
-    """
-
-    def transform_traces(self, prologue_trace: Trace, computation_trace: Trace, epilogue_trace: Trace | None, **kwargs):
+# The transform class, at different points in the processing, the methods are executed. Typically, a subset of them is implemented.
+class Transform(ABC):
+    def transform_traces_early(
```
lantiga, line 336: We could adopt a name that is more explicit. Like `pre-<something>` or `post-<something>`.
Like `post-optimization` for the other method is a good one IMO. It will probably make names more future-proof.

Lightning uses `on_before_<event>` and `on_after_<event>` like `on_before_zero_grad` or `on_after_backward`, and it served us well.

thunder/core/transforms.py
```diff
@@ -518,33 +469,34 @@ def noop(cfn: Callable) -> Callable:
...
-    commented_trace.set_provenance(TraceProvenance(f"Comment Fusions (took {elapsed_time_millis} milliseconds)"))
+        commented_trace.set_provenance(TraceProvenance(f"Comment Fusions (took {elapsed_time_millis} milliseconds)"))
 
-    return commented_trace
+        return commented_trace
 
 
 def comment_fusions(cfn: Callable) -> Callable:
-    return add_post_optimization_transform(cfn, _comment_fusions_transform)
+    return add_ransform(cfn, _CommentFusionsTransform)
```
lantiga, line 499: `add_ransform` typo here

lantiga (commented):
Looks great, this will allow us to expand on more transform lifecycle hooks.

Also it will probably help with composition, as we may need to add tags or traits to transforms to express more granular ordering in the future.

thunder/core/transform_common.py
t-vi: Fair point for early: pre_prologue ?

for additionally I would not sweat it because we want to get rid of it.

lantiga (approved):
Looks great!

PR #865, opened 2024-07-25
nvFuser fusion repro helper function 

riccardofelluga:
<details>

- [x ] Was this discussed/approved via a Github issue? (no need for typos and docs improvements)
- [x ] Did you read the [contributor guideline](https://github.com/Lightning-AI/pytorch-lightning/blob/main/.github/CONTRIBUTING.md), Pull Request section?
- [x ] Did you make sure to update the docs?

</details>

## What does this PR do?

As per discussed in #387, #783, this PR adds an helper function to print a repro script for a specific nvFuser segment. I also added a small section in the documentation to explain the usage of the method as discussed in the issue. 

cc @borda @apaz-cli @tfogal

### DIFF
--- a/thunder/examine/__init__.py
+++ b/thunder/examine/__init__.py
@@ -9,7 +9,6 @@ from thunder.core.proxies import TensorProxy
 from thunder.core.symbol import BoundSymbol
 from thunder.torch import _torch_to_thunder_function_map
 from thunder.core.langctxs import resolve_language, LanguageContext, Languages
-from thunder.executors.nvfuserex_impl import FusionDefinitionWrapper
 import torch
 from warnings import warn
 
--- a/thunder/examine/__init__.py
+++ b/thunder/examine/__init__.py
@@ -219,7 +219,14 @@ def get_fusion_symbols(trace: TraceCtx, warn_if_fusions_unavailable: bool = True
     return fusions
 
 
-def get_nvFuser_repro(fusion: FusionDefinitionWrapper, /) -> str:
+def get_nvFuser_repro(trace: TraceCtx, fusion_name: str, /) -> str:
+    """
+    Helper function to get the repro of a specific nvFusion segment.
+    """
+    if (fusion := trace.python_ctx().get(fusion_name, None)) is None:
+        available_fusions = [name for name, _ in get_fusions(trace)]
+        assert False, f"Unable to find fusion '{fusion_name}' in trace. Available fusions are: {available_fusions}."
+
     msg = ""
     msg += (
         "import torch\nfrom nvfuser import FusionDefinition, DataType\n"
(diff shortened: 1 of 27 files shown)


thunder/executors/nvfuserex_impl.py
```diff
@@ -383,10 +383,12 @@ class FusionDefinitionWrapper:
     cache_info: None | Callable = None
     cache_clear: None | Callable = None
     last_used: None | FusionDefinition = None
+    last_inputs_meta: None | Sequence[tuple] = None
 
     def __call__(self, *args):
         fd = self.get_fd(to_descriptors(args))
         self.last_used = fd
+        self.last_inputs_meta = [(i.size(), i.stride(), i.dtype, i.device) for i in args if isinstance(i, torch.Tensor)]
```
t-vi, line 391: The question @kevinstephano and I had was whether this is perf sensitive.

thunder/examine/__init__.py
```diff
@@ -216,3 +216,48 @@ def get_fusion_symbols(trace: TraceCtx, warn_if_fusions_unavailable: bool = True
             fusions.append(bsym)
 
     return fusions
+
+
+def get_nvFuser_repro(trace: TraceCtx, fusion_name: str, /) -> str:
```
t-vi, line 221: I think we should get the fusions out of obscurity (I think even `trace.get_trace_object(name)` to get `trace.python_ctx()[name]` would do the trick) and then make this a method of the trace.
This would enhance discoverability. WDYT? (also cc @lantiga who mentioned that he wanted an easier way on our podcast)

Also Nit: let's keep lowercase function names, please. I know every exception is worthy in itself, but as a user of APIs, having exceptions at all is terrible.

thunder/examine/__init__.py
```diff
@@ -216,3 +216,48 @@ def get_fusion_symbols(trace: TraceCtx, warn_if_fusions_unavailable: bool = True
...
+    msg = ""
+    msg += (
+        "import torch\nfrom nvfuser import FusionDefinition, DataType\n"
+        f"{fusion.last_used}"
+        "with FusionDefinition() as fd:\n"
+        f"    nvfuser_fusion_id{fusion.last_used.id()}(fd)\n\n"
+    )
+
+    msg += "inputs = [\n"
+    for size, stride, dtype, device in fusion.last_inputs_meta:
```
t-vi, line 243: I'm fine with copying these bits from nvFuser, but I wonder if we should use it from nvFuser eventually.

t-vi (commented):
Great to see this @riccardofelluga , thank you!
I have a few comments, but I look forward to get this into Thunder.

thunder/executors/nvfuserex_impl.py
riccardofelluga: That is also what I was thinking and the main reason why I was trying to save as little info as possible. However, maybe an idea would be to enable this behavior only when the practitioner is using a debug transform. Let me see what it's possible

thunder/examine/__init__.py
riccardofelluga: Yeah I agree it's better to have a clear path to get to the fusion information from the trace without jumping hoops

Issue #862, opened 2024-07-25
allow tracing models on meta device


t-vi:
We want to facilitate running models that can only fit into memory after transforms. The current main thing in PyTorch to instantiate a model without memory is through `meta` devices.

So the plan is to enable:
- Trace through model on meta device with real input (say on GPU).
   - This runs into trouble when checking for "same device". We currently have a few hacks around it here and there, but it would be good to have a more structured way.
   - My idea would be to add `__thunder_device` to the meta tensors (and/or maybe as a default option for compilation) and then use that device for proxying meta tensors.
- Have transforms handle things on meta, ideally staying on meta. (I'll work this out for the bnb quant, and then we can see if/how to adapt the distributed transforms.)  
- Have a materialization transform that materializes models, optionally with initialization from the original code (using the transform_state_dict_for_submodule).

cc @apaz-cli

t-vi:
I see this as the materialization transform:
```
@requiresCUDA
def test_materialization():
    from thunder.transforms import MaterializationTransform

    config = litgpt_model.Config.from_name("llama2-like")
    with torch.device("cuda"):
        ref_m = litgpt_model.GPT(config).to(torch.bfloat16)
    with torch.device("meta"):
        m = litgpt_model.GPT(config).to(torch.bfloat16)
    for p in m.parameters():
        p.__thunder_device = torch.device("cuda")
    for b in m.buffers():
        p.__thunder_device = torch.device("cuda")

    init_from_sd = MaterializationTransform.from_original_state_dict(ref_m)
    jm = thunder.jit(
        m,
        transforms=MaterializationTransform("cuda", init=init_from_sd)
    )
    x = torch.randint(1, 255, (1, 10), device="cuda")
    input_pos = torch.arange(10, device="cuda")

    expected = ref_m(x, input_pos)
    actual = jm(x, input_pos)
    
    assert_close(actual, expected)
```
wdyt?

lantiga:
Looks super neat, to simplify we could also do
```python
    jm = thunder.jit(
        m,
        transforms=MaterializationTransform("cuda", init=ref_m.state_dict())
    )
```

t-vi:
I thought that would be neat, but this is what kept me from offering:
 we could have
- no init (but garbage on first run, so only for people who do evil stuff between creating traces / prologue and running),
- orig module init (through sd transforms),
- orig state dict init (through sd transforms),
- thunder module state dict.

[...]

PR #490, opened 2024-05-30
fsdp_v2 : fix weight sharing

kshitij12345:
Related: https://github.com/Lightning-AI/lightning-thunder/issues/257#issuecomment-2133070264

PR #873, opened 2024-07-28
DDP as a transform

### LINKED ISSUE #795: DDP as an Early Transform
Move DDP to the same transform logic as FSDP (in #359) and remove the distributed from jit_ext.py

davidgonmar:
## What does this PR do?
Adds DDP as a Transform (previously Early Transform), similar to what is done in https://github.com/Lightning-AI/lightning-thunder/pull/359.
Can be used by passing to thunder.ddp a ThunderModule (if passed a torch.nn.Module, previous behaviour is kept). If a ThunderModule is passed, a transform is added to the module.
The transform modifies the traces to insert the necessary syncs.

For testing, I modified two existing tests to test this new way of DDPing a module, and added a new one that tests weight sharing works efficiently (similar to the one in test_fsdp).

Fixes #795

## PR review


davidgonmar:
Marking it ready to review, any feedback is welcome.
For the tests, currently the basic cases + weight sharing case is tested. Should I try to change every test to use the 'new' way? Or is it better to change gradually?

t-vi (commented):
Hi @davidgonmar ! Looks like an awesome start.
So between back when we did FSDP and now, our view on transformations has evolved:
- we want to move `ddp_transform_module` to `DDPTransform.transform_module`,
- it might be interesting to move the materialization to the `MaterializationTransform` instead of having it here (#862 has a bit more).

I think we should do the first in this PR, for the second, we should probably do this separately, but could do this before or after merging this one.
WDYT?

thunder/distributed/__init__.py
```diff
@@ -177,6 +177,102 @@ def prep_shard(
...
+# - Marks the original function with appropiate attributes (use_ddp...)
+# - Broadcasts parameters if necessary
+# - It then registers a transform (callback that runs before prologue is executed) that transforms the
+#   prologue and compute trace, that insert syncs (and grad syncs for the backward, handled by thunder automatically.)
+
+
+def ddp_transform_module(
+    thunder_model: ThunderModule,
+    *,
+    broadcast_from=None,
```
crcrpar, line 190: could you give type hint to this keyword only argument? it feels like `int | None` though

thunder/tests/distributed/test_ddp.py
```diff
@@ -199,6 +205,78 @@ def fwd_loss(m, x):
         with thunder.ThunderModule.no_sync(model):
             fwd_loss(model, x)
 
+    @pytest.mark.skipif(torch.cuda.device_count() < 2, reason="Requires 2 devices")
+    def test_fsdp_weight_sharing(self):
```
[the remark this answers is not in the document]
davidgonmar: fixed:)

davidgonmar:
> Hi @davidgonmar ! Looks like an awesome start. So between back when we did FSDP and now, our view on transformations has evolved:
> 
> * we want to move `ddp_transform_module` to `DDPTransform.transform_module`,
> * it might be interesting to move the materialization to the `MaterializationTransform` instead of having it here ([allow tracing models on meta device #862](https://github.com/Lightning-AI/lightning-thunder/issues/862) has a bit more).
> 
> I think we should do the first in this PR, for the second, we should probably do this separately, but could do this before or after merging this one. WDYT?

I agree with the first one, working on it. As for the second one, I'm currently not even materializing the params on this transform (which I am now realizing seems wrong?). But yeah, I agree, moving common logic to other transforms is nice, but better to do on other PR. Should I still materialize them in this transform for this PR?

thunder/distributed/__init__.py
```diff
@@ -177,6 +177,102 @@ def prep_shard(
...
+        # Since we are doing no sharding, we do not need to materialize the params
+
+        # Broadcast parameters if requested
+        if broadcast_from is not None:
+            for pn, _ in submodule.named_parameters(recurse=False, prefix=module_name):
+                tdist.broadcast(
+                    thunder_model.get_parameter(pn), src=broadcast_from, group=process_group, async_op=False
+                )
+            for pn, _ in submodule.named_buffers(recurse=False, prefix=module_name):
+                tdist.broadcast(thunder_model.get_buffer(pn), src=broadcast_from, group=process_group, async_op=False)
```
crcrpar, line 240: PyTorch's `DistributedDataParallel` seems to broadcast params always as per https://github.com/pytorch/pytorch/blob/784a6ec5a30bd2d1831cb4f78183ad51696794e5/torch/nn/parallel/distributed.py#L823-L831

thunder/distributed/transforms/ddp_v2.py
```diff
@@ -0,0 +1,98 @@
...
+if TYPE_CHECKING:
+    from typing import Any
+    from torch.distributed import ProcessGroup
+    from thunder.core.trace import TraceCtx
+
+
+@dataclass(frozen=True)
+class DDPTraceTransform(Transform):
+    process_group: ProcessGroup
+    replicated_params: dict[str, Any]
```
crcrpar, line 23: IMHO this would be basically `dict[str, torch.nn.Parameter]` or `dict[str, torch.Tensor]`. would you know when/if values are other than those types?

thunder/tests/distributed/test_ddp.py
```diff
@@ -199,6 +205,78 @@ def fwd_loss(m, x):
...
+
+        with device:
+            jit_ddp_model = Model()
+            ddp_jit_model = Model()
+            x = torch.ones(4, 16)
+
+        # Check `jit(ddp(model))` works
+        jit_ddp_model.fc1.weight = jit_ddp_model.fc2.weight
+
+        jit_ddp_model = thunder.jit(thunder.distributed.ddp(jit_ddp_model), executors=["torch"])
```
crcrpar, line 269: out of curiosity, why is `executors=["torch"]`, not the default values which should include nvfuser executor?

crcrpar:
> * it might be interesting to move the materialization to the `MaterializationTransform` instead of having it here ([allow tracing models on meta device #862](https://github.com/Lightning-AI/lightning-thunder/issues/862) has a bit more).

Since DDP doesn't do parameter sharding, what would be the benefits of meta initialization for ddp?

thunder/tests/distributed/test_ddp.py
davidgonmar: I based the test on the one from test_fsdp.py. Not sure about why it is that way there, but I went with it. Is it better to leave the default ones?

thunder/distributed/transforms/ddp_v2.py
davidgonmar: Yes, should be torch.nn.Parameter afaik

davidgonmar:
> * we want to move `ddp_transform_module` to `DDPTransform.transform_module`,


is it safe to assume that transform_module will always be called before transform_traces_pre_prologue? I moved the logic that tracks some things (needed for the trace transform) to transform_module.

thunder/tests/distributed/test_ddp.py
crcrpar: > I based the test on the one from test_fsdp.py. Not sure about why it is that way there, but I went with it. Is it better to leave the default ones?

okay, I wasn't aware of the similar test case for FSDP. It's fine to leave this one as is.

@kshitij12345 would you happen to remember why? It seems that you wrote the reference test case in https://github.com/Lightning-AI/lightning-thunder/pull/490/

thunder/distributed/__init__.py
davidgonmar: > PyTorch's `DistributedDataParallel` seems to broadcast params always as per https://github.com/pytorch/pytorch/blob/784a6ec5a30bd2d1831cb4f78183ad51696794e5/torch/nn/parallel/distributed.py#L823-L831

The thunder.distributed.ddp(...) function (the user API) takes a broadcast_from that defaults to 0, but can be set to None to avoid broadcasting. In the PyTorch API, it always seems to broadcast parameters to 0. Isn't it better to let the user specify if they don't want to broadcast (for any weird reason)?.

thunder/distributed/transforms/ddp_v2.py
davidgonmar: should be always parameters, yes

thunder/distributed/__init__.py
crcrpar: uh... I seem to have been confused.
We allow `broadcast_from=None` to skip broadcast of ddp, but the default value is `0` as per https://github.com/Lightning-AI/lightning-thunder/blob/67ec2c6f4e72f6215f12941bf161915f8f458181/thunder/distributed/__init__.py#L188.
okay, no changes needed here :)

thunder/tests/distributed/test_ddp.py
kshitij12345: I think I used the code from issue description for writing the test - https://github.com/Lightning-AI/lightning-thunder/issues/257#issue-2258890286 and forgot to change the executors 😅

t-vi (approved):
@davidgonmar great PR, thank you!
Nit: Could you update the PR to not have these spurious whitespace changes, please?
Other than that, we're ready to go.

t-vi:
> Since DDP doesn't do parameter sharding, what would be the benefits of meta initialization for ddp?

Right. But would we know that nothing follows DDP that might benefit?

PR #870, opened 2024-07-26
cse fix

t-vi:
the cse mangled all our args when they were used in the same symbol.

:heart:  Debugging credit to @nikitaved :heart:

PR #875, opened 2024-07-29
`unpack_trivial` name check

riccardofelluga:
## What does this PR do?

Fixes the fix for CSE #870.

## Note

TLDR: always use  `unpack_trivial` with the kwarg `name` and/or always use `unpack_trivial.bind( ..., name="something")`.

CSE uses the Right Hand Side(RHS) of the symbols to perform its duty. When using `unpack_trivial` without specifying the variable name, it results in a RHS with `name=None`. Suffice to say that all the symbols that have this RHS are hashed to the same hash, therefore the CSE logic will treat them as duplicate and delete them.

With this PR I cleaned up a couple of prims that were reusing the same fn as `unpack_trivial` for bind post-processing and fixed the missing name in the quantization transform.


t-vi (approved):
Thank you @riccardofelluga

PR #876, opened 2024-07-29
quantization transform: properly update proxy metadata when changing the tensor

t-vi:


thunder/core/proxies.py
```diff
@@ -129,9 +129,20 @@ def __init__(self, name: str | None = None, *, prefix: None | str = None, histor
     def name(self) -> str:
         return self._name
 
+    def replace(self, **changes):
+        r"""returns a copy replacing \**changes. Note that the copy will use the current tracectx"""
```
crcrpar, line 133: I'm afraid I'm not quite following what this docstring would mean, could you update this?

Specifically "`\**changes" looks confusing to me

t-vi: how would you format this? (I found this as a way to write kwargs in docstrings and keep it proper rst for the doc generator.)

thunder/core/proxies.py
```diff
@@ -129,9 +129,22 @@ def __init__(self, name: str | None = None, *, prefix: None | str = None, histor
     def name(self) -> str:
         return self._name
 
+    def replace(self, **changes):
+        r"""Return a copy of the Proxy object with new values for the specified fields as given to the constructor as arguments.
+        Valid kewword arguments are `name`, ``history`.
```
crcrpar, line 134:
```suggestion
        Valid keyword arguments are ``"name"``, ``"history"``.
```

t-vi: I think it's with double backtics per https://bashtage.github.io/sphinx-material/rst-cheatsheet/rst-cheatsheet.html

lantiga (approved):
Looks great.

Super nitpick: the name `replace` suggests an in-place method, while this has copy semantics.

t-vi:
I stole the name `replace` from Python doing this (copies with replacement) on their immutable objects (string, code, function, ...).

PR #878, opened 2024-07-29
interpreter - Fix error when function is called with invalid arg.

kshitij12345:
```python
def fn(x):
        return x

jfn = thunder.jit(fn)
jfn(3, incorrect_arg=3)
```

Errors with 
```python
raise InterpreterError(msg) from e
E               thunder.core.interpreter.InterpreterError: Encountered exception NameError: name 'join' is not defined while tracing <function test_incorrect_args.<locals>.fn at 0x7c50c5ff4820>
```


t-vi (approved):
Thank you @kshitij12345

Issue #858, opened 2024-07-25
`TypeError: expected 2 arguments, got 1` in megatron transformer


tfogal:
## 🚀 Model / language coverage

```
[rank0]: Traceback (most recent call last):
[rank0]:   File "/home/tfogal/dev/nemo/./examples/multimodal/multimodal_llm/neva/neva_pretrain.py", line 118, in <module>
[rank0]:     main()
[rank0]:   File "/home/tfogal/dev/nemo/nemo/core/config/hydra_runner.py", line 129, in wrapper
[rank0]:     _run_hydra(
[rank0]:   File "/home/tfogal/env/lib/python3.10/site-packages/hydra/_internal/utils.py", line 394, in _run_hydra
[rank0]:     _run_app(
[rank0]:   File "/home/tfogal/env/lib/python3.10/site-packages/hydra/_internal/utils.py", line 457, in _run_app
[rank0]:     run_and_report(
[rank0]:   File "/home/tfogal/env/lib/python3.10/site-packages/hydra/_internal/utils.py", line 223, in run_and_report
[rank0]:     raise ex
[rank0]:   File "/home/tfogal/env/lib/python3.10/site-packages/hydra/_internal/utils.py", line 220, in run_and_report
[rank0]:     return func()
[rank0]:   File "/home/tfogal/env/lib/python3.10/site-packages/hydra/_internal/utils.py", line 458, in <lambda>
[rank0]:     lambda: hydra.run(
[rank0]:   File "/home/tfogal/env/lib/python3.10/site-packages/hydra/_internal/hydra.py", line 132, in run
[rank0]:     _ = ret.return_value
[rank0]:   File "/home/tfogal/env/lib/python3.10/site-packages/hydra/core/utils.py", line 260, in return_value
[rank0]:     raise self._return_value
[rank0]:   File "/home/tfogal/env/lib/python3.10/site-packages/hydra/core/utils.py", line 186, in run_job
[rank0]:     ret.return_value = task_function(task_cfg)
[rank0]:   File "/home/tfogal/dev/nemo/./examples/multimodal/multimodal_llm/neva/neva_pretrain.py", line 109, in main
[rank0]:     trainer.fit(model)
[rank0]:   File "/usr/local/lib/python3.10/dist-packages/pytorch_lightning/trainer/trainer.py", line 543, in fit
[rank0]:     call._call_and_handle_interrupt(
[...]

PR #880, opened 2024-07-29
fix class setattr

t-vi:
Fixes: #858 

Thank you @kshitij12345 for the repro, I added a test without the context manager bit, but all this would not have worked as easily without your repro.

### DIFF
--- a/thunder/tests/test_interpreter.py
+++ b/thunder/tests/test_interpreter.py
@@ -3305,3 +3305,23 @@ def test_incorrect_args():
     jfn = thunder.jit(fn)
     with pytest.raises(TypeError, match="got unexpected keyword arguments: incorrect_arg"):
         jfn(3, incorrect_arg=3)
+
+
+def test_class_setattr():
+    import torch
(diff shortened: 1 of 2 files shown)


thunder/tests/test_interpreter.py
```diff
@@ -3305,3 +3305,23 @@ def fn(x):
     jfn = thunder.jit(fn)
     with pytest.raises(TypeError, match="got unexpected keyword arguments: incorrect_arg"):
         jfn(3, incorrect_arg=3)
+
+
+def test_class_setattr():
+    import torch
+    import thunder
+    import contextlib
```
kshitij12345, line 3313: minor nit - move the imports to file level

kshitij12345 (approved):
LGTM, thanks for the quick fix @t-vi

lantiga (approved):

thunder/tests/test_interpreter.py
t-vi: Thank you!

PR #882, opened 2024-07-29
Add executor for flash attention 3

Fuzzkatt:
<details>

</details>

## What does this PR do?

Adds new thunder executor flash attention 3: https://github.com/Dao-AILab/flash-attention?tab=readme-ov-file#flashattention-3-beta-release. Currently relies on the user to install / build fa3 on their own; this executor then finds it and runs it over sdpaex if available. 

Do not merge yet; currently timing tests are showing fa3 running slower than sdpaex and need to debug why. 

## PR review

## Did you have fun?


thunder/executors/fa3ex.py
```diff
@@ -0,0 +1,153 @@
...
+fa3_fwd = fa3_ex.register_operator("fa3_fwd", meta=fa3_fwd_meta, fn=fa3_fwd_impl)
+fa3_bwd = fa3_ex.register_operator("fa3_bwd", meta=fa3_bwd_meta, fn=fa3_bwd_impl)
+
+
+def fa3_checker(query, key, value, attn_mask=None, dropout_p=0.0, is_causal=False, scale=None):
+    if not HAS_FA3:
+        return False
+    if attn_mask is not None or dropout_p != 0.0:
+        return False
+    return query.device.type == "cuda" and key.device == query.device and value.device == query.device
```
crcrpar, line 104: according to https://github.com/Dao-AILab/flash-attention/blob/main/README.md#flashattention-3-beta-release, shouldn't dtype be float16?

thunder/tests/test_fa3_executor.py
```diff
@@ -0,0 +1,47 @@
+import pytest
+import torch
+
+import thunder
+import thunder.core
+from thunder.executors.fa3ex import fa3_ex
```
crcrpar, line 6:
```
from thunder.executors.fa3ex import fa3ex or fas3_executor?
```

crcrpar (commented):
I took a brief look knowing it's being verified and left some nit-picky comments

lantiga:
Thank you @Fuzzkatt, fixed the title of the PR

thunder/tests/test_fa3_executor.py
Fuzzkatt: I copied this off test_sdpaex_executor.py following the import pattern at https://github.com/Lightning-AI/lightning-thunder/blob/d202ba3d868642e0b4d61cd1bf93794350d5a663/thunder/tests/test_sdpaex_executor.py#L6 and subsequent usage pattern at https://github.com/Lightning-AI/lightning-thunder/blob/d202ba3d868642e0b4d61cd1bf93794350d5a663/thunder/tests/test_sdpaex_executor.py#L87, this should be correct

thunder/executors/fa3ex.py
Fuzzkatt: Good catch, thanks!

mruberry:
This expect test needs to be updated to include the executor as a native executor:

https://github.com/Lightning-AI/lightning-thunder/blob/d202ba3d868642e0b4d61cd1bf93794350d5a663/thunder/tests/test_extend.py#L126

That's why at least some CI jobs are failing

thunder/__init__.py
```diff
@@ -171,12 +172,13 @@ def __version__():
...
 get_always_executors = extend.get_always_executors
 
+fa3_executor: None | extend.Executor = extend.get_executor("fa3")
 cudnn_executor: None | extend.Executor = extend.get_executor("cudnn")
 sdpa_executor: None | extend.Executor = extend.get_executor("sdpa")
 nvfuser_executor: None | extend.Executor = extend.get_executor("nvfuser")
 pytorch_executor: None | extend.Executor = extend.get_executor("torch")
 
-# Default executor list is [cudnn -> sdpa -> nvfuser -> torch -> python]
+# Default executor list is [fa3 -> cudnn -> sdpa -> nvfuser -> torch -> python]
```
mruberry, line 181: I like it! First in the default order.

However, maybe we should leave this change for a follow-up PR? Could we get mixology to run with fa3 enabled and first and compare the performance before making it the default? I'd like to develop some system for deciding what gets to be in the default executor list.

fyi @tfogal @IvanYashchuk @t-vi

thunder/executors/fa3ex.py
```diff
@@ -0,0 +1,153 @@
...
+    causal: bool = False,
+    softmax_scale: float | None = None,
+):
+    return thunder.TensorProxy(like=q), thunder.TensorProxy(like=q, shape=(*q.shape[:1], 1))
+
+
+def fa3_fwd_impl(
+    q: torch.Tensor, k: torch.Tensor, v: torch.Tensor, causal: bool = False, softmax_scale: float | None = None
+):
+    if not HAS_FA3:
```
mruberry, line 32: Why return `None, None` vs throwing an exception here?

thunder/executors/fa3ex.py
```diff
@@ -0,0 +1,153 @@
...
+        return None, None
+
+    if softmax_scale is None:
```
mruberry, line 35: A comment about this behavior would be nice

thunder/executors/fa3ex.py
```diff
@@ -0,0 +1,153 @@
...
+        v,
+        out,
+        softmax_lse,
+        dq,
+        dk,
+        dv,
+        softmax_scale,
+        causal,
+    )
+    dq = dq[..., : dout.shape[-1]]  # We could have padded the head dimension
```
mruberry, line 88: Would you elaborate on this comment?

thunder/executors/fa3ex.py
```diff
@@ -0,0 +1,153 @@
...
+
+def fa3_checker(query, key, value, attn_mask=None, dropout_p=0.0, is_causal=False, scale=None):
+    if not HAS_FA3:
+        return False
+    if attn_mask is not None or dropout_p != 0.0:
+        return False
+    return query.device.type == "cuda" and key.device == query.device and value.device == query.device
+
+
+def fa3_transform(
```
mruberry, line 107: How about `fa3_transform_fwd`? To be symmetric with `fa3_transform_bwd` and clarify which transform is referred to

mruberry: Oh wait, this is the execution transform! My mistake. What about `fa3_execution_transform`?

thunder/executors/fa3ex.py
```diff
@@ -0,0 +1,153 @@
...
+    q: thunder.TensorProxy,
+    k: thunder.TensorProxy,
+    v: thunder.TensorProxy,
+    attn_mask: None | thunder.TensorProxy,
+    dropout_p: float = 0.0,
+    is_causal: bool = False,
+    *,
+    scale: None | float = None,
+):
+    out, softmax_lse = fa3_fwd(q, k, v, is_causal, softmax_scale=scale)
```
mruberry, line 130: @Fuzzkatt Was it ultimately clear how to implement the grad transform?

thunder/executors/fa3ex.py
```diff
@@ -0,0 +1,153 @@
...
+    attn_mask: None | thunder.TensorProxy,
+    dropout_p: float = 0.0,
+    is_causal: bool = False,
+    scale: None | float = None,
+):
+    out, softmax_lse = fa3_fwd(q, k, v, is_causal, softmax_scale=scale)
+    return out
+
+
+def fa3_transform_bwd(
```
mruberry, line 120: How about `fa3_fwd_bwd_transform`? Or `fa3_transform_fwd_bwd`? Or `fa3_transform_grad`? Because it doesn't just do the backward, but also the forward

thunder/tests/test_fa3_executor.py
```diff
@@ -0,0 +1,47 @@
...
+    from flash_attn_interface import _flash_attn_forward, _flash_attn_backward, flash_attn_func
+
+    HAS_FA3 = True
+except:
+    HAS_FA3 = False
+
+
+@pytest.mark.parametrize("device,", ["cuda"])
+@requiresCUDA
+def test_sdpa(device: str, dtype: torch.dtype):
```
mruberry, line 20: This test isn't being executed because the test framework doesn't know how to instantiate the `dtype` parameter:

```
______________________ ERROR at setup of test_sdpa[cuda] _______________________
[gw4] linux -- Python 3.10.12 /usr/bin/python3.10
file /__w/3/s/thunder/tests/test_fa3_executor.py, line 18
  @pytest.mark.parametrize("device,", ["cuda"])
  @requiresCUDA
  def test_sdpa(device: str, dtype: torch.dtype):
E       fixture 'dtype' not found
>       available fixtures: benchmark, benchmark_weave, cache, capfd, capfdbinary, caplog, capsys, capsysbinary, cov, doctest_namespace, monkeypatch, no_cover, pytestconfig, record_property, record_testsuite_property, record_xml_attribute, recwarn, testrun_uid, tmp_path, tmp_path_factory, tmpdir, tmpdir_factory, worker_id
>       use 'pytest --fixtures [testpath]' for help on them.

/__w/3/s/thunder/tests/test_fa3_executor.py:18
```

Per @crcrpar 's point above, I think this needs to instantiate for the `float16` dtype

thunder/tests/test_fa3_executor.py
```diff
@@ -0,0 +1,47 @@
...
+# flash attn 3
+try:
+    from flash_attn_interface import _flash_attn_forward, _flash_attn_backward, flash_attn_func
+
+    HAS_FA3 = True
+except:
+    HAS_FA3 = False
+
+
+@pytest.mark.parametrize("device,", ["cuda"])
```
mruberry, line 18: Would you also add a test that the fa3 executor is not used when the inputs are unsupported? From the checker function and @crcrpar 's comment, that could inputs where the `attn_mask` is not None, the `dropout` is nonzero, the tensors are on the `cpu` device, and the dtype is torch.bfloat16 or torch.float32

mruberry:
Cool stuff, @Fuzzkatt! Things are looking good. I made some comments for your review.

thunder/__init__.py
IvanYashchuk: Previously before declaring the cuDNN executor to be part of the default list we verified that it gives speedup for the important models we test and we assume this speedup would hold for other shapes.
In the absence of a better smart system that can make the priority choices for us, I think we should follow the same process for this new executor:
1. Don't add it to the defaults
2. Add benchmark options to enable using this executor at the candidate priority placement
3. Run all benchmarks we have and analyze the results
4. I'd say if it provides speedup for 80% of the cases _and_ Llama 3 then it's good to be included in the defaults.

However, this executor is special that it doesn't support bf16 at the moment and all workloads we benchmark use bf16, so the process would yield useful results.

thunder/executors/fa3ex.py
Fuzzkatt: Fixed

thunder/executors/fa3ex.py
Fuzzkatt: Good idea, changed to raise exception

thunder/executors/fa3ex.py
Fuzzkatt: I copied this off of https://github.com/Dao-AILab/flash-attention/blob/main/hopper/flash_attn_interface.py#L179 (the original fa3 bwd)

thunder/executors/fa3ex.py
Fuzzkatt: I used the existing executors/sdpaex.py as a reference, which mostly matched up. I also did remember you mentioning using put_grads() to have fwd and bwd in the same function from our meeting with Tom a couple of weeks ago, that was helpful :)

thunder/tests/test_fa3_executor.py
Fuzzkatt: Good idea, added

thunder/executors/fa3ex.py
Fuzzkatt: Are you referring to the raising exception if not HAS_FA3 or the if softmax_scale is None part? The if softmax_scale is None part comes from https://github.com/Dao-AILab/flash-attention/blob/main/hopper/flash_attn_interface.py#L199

thunder/executors/fa3ex.py
Fuzzkatt: Changed to fa3_execution_transform to stay consistent with the arguments that fa3ex.register_implementation is asking for

thunder/executors/fa3ex.py
Fuzzkatt: Changed to fa3_grad_transform to stay consistent with the arguments that fa3ex.register_implementation is asking for

thunder/tests/test_fa3_executor.py
Fuzzkatt: My bad, I copied off https://github.com/Lightning-AI/lightning-thunder/blob/d202ba3d868642e0b4d61cd1bf93794350d5a663/thunder/tests/test_sdpaex_executor.py#L39 but removed all of https://github.com/Lightning-AI/lightning-thunder/blob/d202ba3d868642e0b4d61cd1bf93794350d5a663/thunder/tests/test_sdpaex_executor.py#L34-L36 instead of just the bfloat16 and float32 cases, fixed

Fuzzkatt:
As an update on the investigation into whether fa3 is actually performant over sdpa, fa2, and cudnn, I ran the benchmarking script provided in the fa3 repo: https://github.com/Dao-AILab/flash-attention/blob/main/hopper/benchmark_attn.py and got the attached results: [benchmark_attn_output.log](https://github.com/user-attachments/files/16433818/benchmark_attn_output.log). In summary, fa3 was better / comparable to cudnn on head_dim = 64, worse on head_dim=128, and better than fa2 on head_dim=256 (for some reason the benchmarking script stopped running cudnn kernels at head_dim=256). 

I also tried running the llama2.c example here https://github.com/Lightning-AI/lightning-thunder/tree/67ec2e9ec88721d21d5d6609227f12e2635365da/examples/llama2.c with the fa3_executor. It was unable to run, as fa3 fwd currently requires input to be contiguous: https://github.com/Dao-AILab/flash-attention/issues/1087. So we don't really have any full end to end model perf examples.

Given these findings, I would agree with Ivan and keep fa3 out of default executors for now while we test fa3 further / wait for it to be developed more to cover these missing functionality cases like bf16, attn_mask, dropout, non-contiguous input, etc.

thunder/__init__.py
Fuzzkatt: Moved it out of default executors, more details in my latest comment about fa3 perf in general: https://github.com/Lightning-AI/lightning-thunder/pull/882#issuecomment-2259275497

thunder/executors/fa3ex.py
mruberry: It would be nice to have a note somewhere, either as a comment or as a doc string on the function, explaining what the default value is (and ideally why),

thunder/executors/fa3ex.py
mruberry: If the code is copied from somewhere else, and we're not exactly sure why it's there, then a comment explaining its provenance would be helpful

thunder/executors/fa3ex.py
```diff
@@ -0,0 +1,156 @@
...
+
+def fa3_fwd_impl(
+    q: torch.Tensor, k: torch.Tensor, v: torch.Tensor, causal: bool = False, softmax_scale: float | None = None
+):
+    if not HAS_FA3:
+        raise Exception("fa3 not built, cannot use fa3 executor")  # checker should fail before getting here
+
+    if softmax_scale is None:
+        softmax_scale = q.shape[-1] ** (-0.5)
+    out, q, k, v, out_padded, softmax_lse, S_dmask = _flash_attn_forward(q, k, v, softmax_scale, causal)
```
mruberry, line 37: You mentioned elsewhere that the inputs to the operation must be contiguous. Would you add a comment about that requirement, and then call `torch.contiguous` on the inputs that must be contiguous here?

Would you also add a test with noncontiguous inputs, and verify the executor works as expected?

fyi @IvanYashchuk @jjsjann123 -- particular stride requirements are likely to come up more in the future. Instead of putting the contiguous calls into the implementation, it'd be nice if we could put them into the transformation for the function. But if we do that today then it will always trigger a copy, even if the input tensors are always contiguous

jjsjann123: > particular stride requirements are likely to come up more in the future. Instead of putting the contiguous calls into the implementation, it'd be nice if we could put them into the transformation for the function.

Does having it in the transformation mean we'll have a `stride_order` node in the trace?
Admittedly, having a `contiguous` call could be tricky, since it can be executed as a view or a copy. The ambiguity is tricky to handle if we want to have consistent behavior with eager.

> But if we do that today then it will always trigger a copy, even if the input tensors are always contiguous

If torch executor takes the contiguous call and figured it's a no-op, this will only produce a view to the buffer. So I think this statement is on nvfuser's handling of stride_order? But if that's fused, we aren't really making any real copy...

mruberry: I think that if it's in the transformation then yes, `stride_order` will be in the trace. 

Yes you're also right that we could check that our `stride_order` execution can no-op, but currently it doesn't:

https://github.com/Lightning-AI/lightning-thunder/blob/67ec2c6f4e72f6215f12941bf161915f8f458181/thunder/executors/torchex.py#L621

mruberry: Awesome, thanks for working that through with me!

So, @Fuzzkatt, what about putting `contiguous` calls in the transform, to ensure that the inputs are the expected format, and adding a test that noncontiguous inputs work as expected?

thunder/tests/test_fa3_executor.py
```diff
@@ -0,0 +1,93 @@
...
+@pytest.mark.parametrize("dtype", [torch.float16], ids=("float16"))
+@requiresCUDA
+def test_fa3(device: str, dtype: torch.dtype):
+
```
rdspring1, line 22: Would you add a device check for `fa3`?
For example:
```python
    # Flash Attention 3 is only supported on Hopper devices.
    # Skip this test on Ampere and prior devices.
    torch_device = torch.device("cuda")
    if not device_version_support(torch_device, CudaVersion(9, 0)):
        pytest.skip(f"flash attention 3 is not supported on {torch.cuda.get_device_name()}")
```

thunder/executors/fa3ex.py
```diff
@@ -0,0 +1,156 @@
...
+    grads = (dq, dk, dv)
+    return grads
+
+
+fa3_fwd = fa3_ex.register_operator("fa3_fwd", meta=fa3_fwd_meta, fn=fa3_fwd_impl)
+fa3_bwd = fa3_ex.register_operator("fa3_bwd", meta=fa3_bwd_meta, fn=fa3_bwd_impl)
+
+
+def fa3_checker(query, key, value, attn_mask=None, dropout_p=0.0, is_causal=False, scale=None):
+    if not HAS_FA3:
```
rdspring1, line 97: Does `fa3` support a specific cuda version?

`device_version_support` is a helper function in `thunder/tests/test_sdpaex_executor.py`. You should move it to a more general place.

```python
    # Flash Attention 3 is only supported on Hopper devices.
    if not device_version_support(query.device.type, CudaVersion(9, 0)):
        return False
```

mruberry, line 97: Following on @rdspring1's thoughts about the test, would you add a comment about which devices support fa3. If a device doesn't support fa3, will `HAS_FA3` be `False`, or is a separate check needed?

Fuzzkatt: I can add a comment. In general, fa3 is currently built for Hopper only. The assumption with this executor is that it will try to look for fa3 through import, which should only work if the user / container builds fa3 which should only work for Hopper. Thus if HAS_FA3 will be true if and only if the device is Hopper and the user / container has built fa3, so the current HAS_FA3 check should be sufficient.

Fuzzkatt: Under the requirements in this section https://github.com/Dao-AILab/flash-attention?tab=readme-ov-file#flashattention-3-beta-release, it mentions requires cuda >= 12.3. Do we want to add an additional check for sm >= 90? I think the current HAS_FA3 check is sufficient (see the below response to @mruberry's comment) and I was planning to just add some clarifying comments in the code explaining it, but I can also add another check if it would make things clearer.

thunder/tests/test_fa3_executor.py
Fuzzkatt: For test_fa3 specifically, I already have an if not HAS_FA3 pytest.skip(), which should already cover the case of non-Hopper devices, since HAS_FA3 will be true if and only if the device is Hopper and the user / container has built fa3.

thunder/executors/fa3ex.py
Fuzzkatt: Sure I can add the contiguous calls in the transform and add a test. One caveat: are we concerned about the additional overhead for H2D memcpy that a .contiguous() convert might add that was brought up in https://github.com/Dao-AILab/flash-attention/issues/1087? I can also try to run some timing tests if it would help

thunder/executors/fa3ex.py
mruberry: Great! I think a comment will make that very clear to future readers

thunder/executors/fa3ex.py
mruberry: You're correct that would be unfortunate, and in the future we'll want to do a better job selecting which executor to use (possibly with some benchmarking tools), but this PR doesn't need to worry about that. Practitioners can always disable the fa3 executor globally as a performance workaround. 

This is a good issue to leave for a follow-up.

Fuzzkatt: Added the contiguous() calls in both the fwd and bwd, as they ended up also being necessary in the bwd. The weird thing is that the actual _flash_attn_forward() and _flash_attn_backward() calls from fa3 itself that I'm calling into in the fwd_impl() and bwd_impl() of the thunder fa3 executor should already be doing these contiguous() calls: https://github.com/Dao-AILab/flash-attention/blob/main/hopper/flash_attn_interface.py#L18 and https://github.com/Dao-AILab/flash-attention/blob/main/hopper/flash_attn_interface.py#L44. But without adding these explicit calls in the transforms, I'm still hitting the 'Inputs must be contiguous' RuntimeError. I'll keep looking into it tomorrow.

thunder/executors/fa3ex.py
Fuzzkatt: Added clarifying comment

Fuzzkatt: Added clarifying comment

Fuzzkatt:
Documenting my benchmarking experiments here as they seem relevant. While testing the contiguous() and headdim changes today using llama2.c as the benchmark, I was able to finally run llama2.c with the fa3 executor after a couple of changes (adding contiguous() to the fwd and bwd impl of the fa3 executor, setting dim / nheads to 512 / 4 to fit the headdim reqs, and changing dtype to fp16). I got the following numbers:

dim / nheads: 512 / 4:
* thunder sdpa: 188 ms
* thunder cudnn: 185 ms
* thunder fa3: 200 ms
* torch.compile: 136 ms
* eager: 243 ms

Seems like fa3 ex is slower than both cudnn ex and sdpa ex in this case. I wonder how much of it has to do with the added .contiguous() calls, as the input tensors were definitely non-contiguous to begin with in this use case.

Fuzzkatt:
Also found out that fa3 fwd supports headdims (64, 128, 256,) while fa3 bwd supports only headdims (64, 128,). Currently the fa3_checker function seems to cover both the execution_transform (fwd only) and the grad_transform (fwd+bwd). Is there a way to differentiate between the two cases in the checker function? Ideally checker function would pass for fwd only on headdim 256 but fail for fwd+bwd on headdim 256, but I don't know how to do this yet.

thunder/executors/fa3ex.py
Fuzzkatt: Added the original source. From what I understand, it looks like the actual flash_attn_hopper_cuda.bwd() function may possibly pad the head dim (final dim) of the grads, so the next three lines simply remove the padding and set the final dim back to the original head dim.

thunder/executors/fa3ex.py
Fuzzkatt: Added clarifying comment

thunder/executors/fa3ex.py
```diff
@@ -37,6 +37,8 @@ def fa3_fwd_impl(
     if not HAS_FA3:
         raise Exception("fa3 not built, cannot use fa3 executor")  # checker should fail before getting here
 
+    # According to https://github.com/Dao-AILab/flash-attention?tab=readme-ov-file#how-to-use-flashattention, softmax_scale is
```
crcrpar, line 40: nit-picking but it'd be safer to use a permalink

```suggestion
    # According to https://github.com/Dao-AILab/flash-attention/blob/5018ac6/README.md?plain=1#L112, softmax_scale is
```

thunder/executors/fa3ex.py
```diff
@@ -90,7 +92,7 @@ def fa3_bwd_impl(
         softmax_scale,
         causal,
     )
-    dq = dq[..., : dout.shape[-1]]  # We could have padded the head dimension
+    dq = dq[..., : dout.shape[-1]]  # We could have padded the head dimension (from https://github.com/Dao-AILab/flash-attention/blob/main/hopper/flash_attn_interface.py#L179)
```
crcrpar: ditto

```suggestion
    dq = dq[..., : dout.shape[-1]]  # We could have padded the head dimension (from https://github.com/Dao-AILab/flash-attention/blob/5018ac6/hopper/flash_attn_interface.py#L179)
```

thunder/executors/fa3ex.py
mruberry: That is curious! 

I think we find that determining exactly when a kernel (or set of kernels) can be used is often the hardest part of integrating them with Thunder. fyi @tfogal and @IvanYashchuk. 

There are two thoughts here I think the project needs to be aware of:

First, incorporating this logic about when a particular kernel can and should be used is part of the Thunder project, and it's good we're very thorough, and in the future we'll have benchmarking facilities to help us make decisions about which executor to use.

Second, even though Thunder is intended to select the best kernels given the current program, we probably need to think more about letting practitioners know where they are losing performance. We have seen this issue in the past, where program was calling flash attention in a way that didn't use our fastest kernels, and this caused a huge performance impact and a lot of end user confusion.

I expect statistics on how operations are called, analysis of performance gaps between executors, automated benchmarking, and descriptions of how practitioners could utilize faster kernels will be part of thunder's Q4 planning.

mruberry:
> Documenting my benchmarking experiments here as they seem relevant. While testing the contiguous() and headdim changes today using llama2.c ([`67ec2e9`/examples/llama2.c](https://github.com/Lightning-AI/lightning-thunder/tree/67ec2e9ec88721d21d5d6609227f12e2635365da/examples/llama2.c)) as the benchmark, I was able to finally run llama2.c with the fa3 executor after a couple of changes (adding contiguous() to the fwd and bwd impl of the fa3 executor, setting dim / nheads to 512 / 4 to fit the headdim reqs, and changing dtype to fp16). I got the following numbers:
> 
> dim / nheads: 512 / 4:
> 
> * thunder sdpa: 188 ms
> * thunder cudnn: 185 ms
> * thunder fa3: 200 ms
> * torch.compile: 136 ms
> * eager: 243 ms
> 
> Seems like fa3 ex is slower than both cudnn ex and sdpa ex in this case. I wonder how much of it has to do with the added .contiguous() calls, as the input tensors were definitely non-contiguous to begin with in this use case.

Interesting datapoint. I'm surprised torch.compile is so much faster than thunder's sdpa. 

fyi @tfogal 

@jjsjann123 @IvanYashchuk Maybe in the future we can provide clear data of how much time each operator takes to execute without having to look at a profile?

Follow-up thought for @jjsjann123 -- maybe we should consider benchmarking operator executors, like the flash attention executors, on the actual inputs seen at runtime, and then specializing on tensor shape and strides? One interesting question that additional reporting would help us to understand is how often do tensor strides vary if their shapes are the same? I would be really surprised if the stride order of input tensors varied, even if their shapes varied. If that's so, can we do something with that knowledge?

mruberry:
Looks like there are still some test failures, @Fuzzkatt! Let us know if you have any more questions, and be sure to let us know when the CI is working and the PR is ready for another review. Seems like things are going well!

mruberry:
> Also found out that fa3 fwd supports headdims (64, 128, 256,) while fa3 bwd supports only headdims (64, 128,). Currently the fa3_checker function seems to cover both the execution_transform (fwd only) and the grad_transform (fwd+bwd). Is there a way to differentiate between the two cases in the checker function? Ideally checker function would pass for fwd only on headdim 256 but fail for fwd+bwd on headdim 256, but I don't know how to do this yet.

Interesting! This is a great detail.

First, I think it'd be OK for this PR to only handle (64, 128) headdims and make a note that forward-only also supports 256. Then file an issue requesting support for that.

Second, I don't think we support a difference in what an executor can run fwd-only vs fwd-bwd today. @IvanYashchuk may correct me. This sounds like a good issue for follow-up.

jjsjann123:
> > Documenting my benchmarking experiments here as they seem relevant. While testing the contiguous() and headdim changes today using llama2.c ([`67ec2e9`/examples/llama2.c](https://github.com/Lightning-AI/lightning-thunder/tree/67ec2e9ec88721d21d5d6609227f12e2635365da/examples/llama2.c)) as the benchmark, I was able to finally run llama2.c with the fa3 executor after a couple of changes (adding contiguous() to the fwd and bwd impl of the fa3 executor, setting dim / nheads to 512 / 4 to fit the headdim reqs, and changing dtype to fp16). I got the following numbers:
> > dim / nheads: 512 / 4:
> > 
> > * thunder sdpa: 188 ms
> > * thunder cudnn: 185 ms
> > * thunder fa3: 200 ms
> > * torch.compile: 136 ms
> > * eager: 243 ms
> > 
> > Seems like fa3 ex is slower than both cudnn ex and sdpa ex in this case. I wonder how much of it has to do with the added .contiguous() calls, as the input tensors were definitely non-contiguous to begin with in this use case.
> 
> Interesting datapoint. I'm surprised torch.compile is so much faster than thunder's sdpa.
> 
> fyi @tfogal
> 
> @jjsjann123 @IvanYashchuk Maybe in the future we can provide clear data of how much time each operator takes to execute without having to look at a profile?
> 
> Follow-up thought for @jjsjann123 -- maybe we should consider benchmarking operator executors, like the flash attention executors, on the actual inputs seen at runtime, and then specializing on tensor shape and strides? One interesting question that additional reporting would help us to understand is how often do tensor strides vary if their shapes are the same? I would be really surprised if the stride order of input tensors varied, even if their shapes varied. If that's so, can we do something with that knowledge?

Re: `maybe we should consider benchmarking operator executors, like the flash attention executors, on the actual inputs seen at runtime, and then specializing on tensor shape and strides?`
By `specializing` do you mean we want to generate a kernel with baked in shape/stride, or just that we are seeing varying shape/stride that triggers an nvfuser recompilation?
On the recompilation part, my intuition is that it's going to be rare. There could be some an artifacts coming from shape changes across size-1 and non-size-1... Would be interesting to see some wild workflow actually change tensor layout between iterations.

mruberry:
> > > Documenting my benchmarking experiments here as they seem relevant. While testing the contiguous() and headdim changes today using llama2.c ([`67ec2e9`/examples/llama2.c](https://github.com/Lightning-AI/lightning-thunder/tree/67ec2e9ec88721d21d5d6609227f12e2635365da/examples/llama2.c)) as the benchmark, I was able to finally run llama2.c with the fa3 executor after a couple of changes (adding contiguous() to the fwd and bwd impl of the fa3 executor, setting dim / nheads to 512 / 4 to fit the headdim reqs, and changing dtype to fp16). I got the following numbers:
> > > dim / nheads: 512 / 4:
> > > 
> > > * thunder sdpa: 188 ms
> > > * thunder cudnn: 185 ms
> > > * thunder fa3: 200 ms
> > > * torch.compile: 136 ms
> > > * eager: 243 ms
> > > 
> > > Seems like fa3 ex is slower than both cudnn ex and sdpa ex in this case. I wonder how much of it has to do with the added .contiguous() calls, as the input tensors were definitely non-contiguous to begin with in this use case.
> > 
> > 
> > Interesting datapoint. I'm surprised torch.compile is so much faster than thunder's sdpa.
> > fyi @tfogal
> > @jjsjann123 @IvanYashchuk Maybe in the future we can provide clear data of how much time each operator takes to execute without having to look at a profile?
> > Follow-up thought for @jjsjann123 -- maybe we should consider benchmarking operator executors, like the flash attention executors, on the actual inputs seen at runtime, and then specializing on tensor shape and strides? One interesting question that additional reporting would help us to understand is how often do tensor strides vary if their shapes are the same? I would be really surprised if the stride order of input tensors varied, even if their shapes varied. If that's so, can we do something with that knowledge?
> 
> Re: `maybe we should consider benchmarking operator executors, like the flash attention executors, on the actual inputs seen at runtime, and then specializing on tensor shape and strides?` 

> By `specializing` do you mean we want to generate a kernel with baked in shape/stride, or just that we are seeing varying shape/stride that triggers an nvfuser recompilation? On the recompilation part, my intuition is that it's going to be rare. There could be some an artifacts coming from shape changes across size-1 and non-size-1... Would be interesting to see some wild workflow actually change tensor layout between iterations.

I think we could consider building a "mini-dispatcher" for operators with multiple executors that caches on metadata and actual runtime strides. That would be cases where nvFuser probably wouldn't be involved?

But also interesting to consider the effects on nvFuser fusions, too

thunder/executors/fa3ex.py
Fuzzkatt: Fixed, thanks

thunder/executors/fa3ex.py
Fuzzkatt: Fixed, thanks

Fuzzkatt:
Did a round of CI cleanup / responding to suggestions. Should be ready for review again. 

Some thoughts on CI testing / results:
* Currently the tests are skipped if HAS_FA3 is False, which it definitely will be in CI. So the CI being green for test_fa3_executor.py just means it is no longer crashing (I think earlier I had the wrong number of parameters in parametrize or something), but gives no indication to whether the tests actually function or not. For that, I'm manually running them on an H100 node with fa3 built
* Comparing test_fa3_executor.py to its directly counterpart in test_sdpa_executor.py: https://github.com/Lightning-AI/lightning-thunder/blob/67ec2e9ec88721d21d5d6609227f12e2635365da/thunder/tests/test_sdpaex_executor.py#L39, I've elected to separate accuracy and whether it's used or not into different tests
* The accuracy check is surprisingly involved. Somewhat unintuitively, I can't just compare a thunder.jit(fn) for fa3 vs torch.nn.functional.scaled_dot_product_attention() as the numerical mismatch is gigantic (I'm actually not clear on why this doesn't work, if anyone knows that would be helpful). Instead, I looked to the fa3 sanity testing here: https://github.com/Dao-AILab/flash-attention/blob/main/hopper/test_flash_attn.py#L65. They have a handwritten attention_ref function, and their accuracy check involves running the handwritten thing twice, then checking if the fa3 result vs handwritten result is within 2x magnitude of the handwritten result vs handwritten result attempt 2 (plus some margin of error). I've added this to the accuracy check here, but I'm not sure if it's worth including as a thunder test; more specifically, I'm not sure where to draw the line on if we're testing thunder or fa3? Would like to hear people's thoughts
* I followed the sdpaex example for checking if fa3 is called into correctly here: https://github.com/Lightning-AI/lightning-thunder/blob/67ec2e9ec88721d21d5d6609227f12e2635365da/thunder/tests/test_sdpaex_executor.py#L62-L64, but this only checks if the forward is called. Is there an equivalent of this for checking if the backward is called?

thunder/tests/test_fa3_executor.py
```diff
@@ -0,0 +1,237 @@
...
+    seq_len = 128
+    num_heads = 4
+    dim_per_head = 32
+
+    device = "cuda"
+    dtype = torch.float16
+    attn_mask = None
+    dropout_p = 0.0
+
+    def check(device, dtype, attn_mask, dropout_p):
```
mruberry, line 211: nit: a comment describing this function would be helpful

thunder/tests/test_fa3_executor.py
```diff
@@ -0,0 +1,237 @@
...
+        query = torch.randn([batch, seq_len, num_heads, dim_per_head], device=device, dtype=dtype)
```
mruberry, line 212: Can all these tensors really have values between -1 and 1?

thunder/tests/test_fa3_executor.py
```diff
@@ -0,0 +1,237 @@
...
+
+        cfn = thunder.jit(fn, executors=[fa3_ex])
+
+        thunder_result = cfn(query, key, value, attn_mask=attn_mask, dropout_p=dropout_p)
+
+        # Verifies fa3 was not called
+        extrace = thunder.last_traces(cfn)[-1]
+        assert not any(bsym.sym.name == "fa3_fwd" for bsym in extrace.bound_symbols)
+
+    check("cpu", dtype, attn_mask, dropout_p)
```
mruberry, line 229: Comments here to describe the cases would be nice

thunder/tests/test_fa3_executor.py
```diff
@@ -0,0 +1,237 @@
...
+        attention_drop = attention
+    output = torch.einsum("bhts,bshd->bthd", attention_drop, v * dropout_scaling)
+    if query_padding_mask is not None:
+        output.masked_fill_(rearrange(~query_padding_mask, "b s -> b s 1 1"), 0.0)
+    return output.to(dtype=dtype_og), attention.to(dtype=dtype_og)
+
+
+# verify that fa3 kernel is accurate
+@requiresCUDA
+def test_fa3_accuracy():
```
mruberry, line 114: The fa3 version of the operation being so numerically distinct from the actual PyTorch operation is a concern.

Would you file an issue for follow-up, and add a comment here that links the issue?

I don't think we should try to address the numerics in this PR

mruberry (approved):

mruberry:
@Fuzzkatt Cool! I made a few comments requesting more comments and an issue to follow-up on the numerics.

thunder/tests/test_fa3_executor.py
Fuzzkatt: Opened an issue with fa3 directly, as this is happening independently of Thunder: https://github.com/Dao-AILab/flash-attention/issues/1128.

thunder/tests/test_fa3_executor.py
Fuzzkatt: Added

thunder/tests/test_fa3_executor.py
Fuzzkatt: Added

thunder/tests/test_fa3_executor.py
Fuzzkatt: Confirmed that sdpa and fa3 expect inputs in a different dimensional ordering. Added a new fa3_accuracy_vs_sdpa test with the correct input dimensional ordering

Fuzzkatt:
Some various updates: 
* As of the latest, fa3 no longer requires contiguous inputs: https://github.com/Dao-AILab/flash-attention/issues/1109#issuecomment-2269964760, so I removed the explicit .contiguous() checks
* It looks like bf16 is now supported by fa3 (this is not yet updated with in their readme but I tested manually)
* I added a separate fa3 vs torch sdpa accuracy check based on the discussion in https://github.com/Dao-AILab/flash-attention/issues/1128

mruberry:
> Some various updates:
> 
> * As of the latest, fa3 no longer requires contiguous inputs: [FA3 No perf gain noticed. Dao-AILab/flash-attention#1109 (comment)](https://github.com/Dao-AILab/flash-attention/issues/1109#issuecomment-2269964760), so I removed the explicit .contiguous() checks
> * It looks like bf16 is now supported by fa3 (this is not yet updated with in their readme but I tested manually)
> * I added a separate fa3 vs torch sdpa accuracy check based on the discussion in [Flash attn 3 has large numerical mismatches with torch spda Dao-AILab/flash-attention#1128](https://github.com/Dao-AILab/flash-attention/issues/1128)

Awesome updates!

mruberry:
I think this is ready to merge; any more changes you'd like to make, @Fuzzkatt ?

Fuzzkatt:
Just made a final small change to check that fa3_bwd is being used in test_fa3_used thanks to @IvanYashchuk's hint. @mruberry, this now looks ready to merge for me, feel free to resolve conversations and press the button whenever CI finishes turning green. Thanks everyone for reviewing and commenting!

mruberry:
This is OK to merge. @t-vi, the CI failures are a flaky

```
FAILED thunder/tests/test_grad.py::test_vjp_correctness_polygamma_torch_cpu_thunder.dtypes.float64 - AssertionError: Scalars are not close!

Expected -42.53858118981584 but got -42.53857107089372.
Absolute difference: 1.0118922119772833e-05 (up to 1e-07 allowed)
Relative difference: 2.3787634276329376e-07 (up to 1e-07 allowed)
```

and are not related to this PR. But it may need to be merged manually to override them.

@IvanYashchuk, let's widen the test tolerance for polygamma, or change its inputs?

PR #885, opened 2024-07-30
CI: cope with pt nightly SDPAParams change

t-vi:


t-vi:
Expedited merging as CI fix.

Issue #728, opened 2024-07-08
CUDA Graphs region includes unnecessary "del" symbols resulting in an invalid Python program


IvanYashchuk:
## 🐛 Bug

First, apply the following patch:
```diff
diff --git a/thunder/executors/cudagraphex.py b/thunder/executors/cudagraphex.py
index 61e482e5..38bbf81f 100644
--- a/thunder/executors/cudagraphex.py
+++ b/thunder/executors/cudagraphex.py
@@ -155,6 +155,8 @@ class CUDAGraphExecutor(FusionExecutor):
         return fusion_bsym
 
     def can_fuse(self, bsym: BoundSymbol):
+        from thunder.executors.torchex import item
+
         curr_tracectx = get_tracectx()
         assert hasattr(curr_tracectx, "clear_collection_names")
 
@@ -180,6 +182,7 @@ class CUDAGraphExecutor(FusionExecutor):
             prims.PrimIDs.RETURN,
             # Data-dependent ops
             prims.PrimIDs.ITEM,
+            item.id,
         }
 
         if bsym.sym.id in do_not_fuse_sym_set:
```
Then run the following reproducer:
```py
import thunder
import torch

def func(x):
    return x.sum().item()

a = torch.randn(3, 3, device="cuda")

jfunc = thunder.jit(func, use_cudagraphs=True)
jfunc(a) # Fails with UnboundLocalError: local variable 't0' referenced before assignment
```
```py
File thunder.computation_4:10, in computation(x)
      6 @torch.no_grad()
      7 @no_autocast
      8 def computation(x):
      9   # x: "cuda:0 f32[3, 3]"
---> 10   [t0] = CUDAGraph0(x)
     11   f1 = Tensor.item(t0)  # f1: "float ?"
     12   return f1

File ~/dev/lightning-thunder/thunder/executors/cudagraphex.py:93, in CUDAGraphCallable.__call__(self, *args)
     89     static_inputs_mask = tuple(isinstance(arg, torch.nn.Parameter) for arg in args)
     91 args_descriptor = to_arg_descriptor(*args)
---> 93 graph, static_inputs, static_outputs = build_cuda_graph(self.fn, args_descriptor, static_inputs_mask)
     95 for static_input, arg in utils.safe_zip(static_inputs, args):
     96     if id(static_input) != id(arg) and isinstance(static_input, torch.Tensor) and isinstance(arg, torch.Tensor):

[...]

PR #888, opened 2024-07-30
cudagraphs before del

t-vi:
The current del handling in the cudagraphs transforms runs into trouble when not fusing everything because dels go to the wrong place.

This is needed in order to exclude (even experimentally) things from the cudagraph.

However, it misses
- handling autograd's transform for execution (@IvanYashchuk any advice if we can move the delete last used to after autograd, i.e. to `thunder.jit.fn_`?)
- a testing strategy. This only shows when not merging everything (I manually excluded `bnb_matmul_nf4` by hardcoding a change to the executor and hit this, maybe allowing operators to say if they're cudagraph-unfusable?)

Also cc @nikitaved if he has comments.

### DIFF
--- a/thunder/executors/cudagraphex.py
+++ b/thunder/executors/cudagraphex.py
@@ -169,6 +169,12 @@ class CUDAGraphExecutor(FusionExecutor):
             curr_tracectx.clear_collection_names.add(bsym.args[0].name)
             return False
 
+        # We let DEL to get fused (but should not see them much), unless the deleted proxy is a CollectionProxy
+        # consumed by the `clear_mutable_collection` symbol
+        if bsym.sym.id == prims.PrimIDs.DEL and bsym.args[0].name in curr_tracectx.clear_collection_names:
+            return False
+        # }
+
         do_not_fuse_sym_set = {
             # Skip the very beginning and the very end of the trace
             prims.PrimIDs.UNPACK_TRIVIAL,
--- a/thunder/executors/cudagraphex.py
+++ b/thunder/executors/cudagraphex.py
@@ -138,6 +138,10 @@ class CUDAGraphExecutor(FusionExecutor):
         inputs = [unvariableify(inp) for inp in sorted(region.inputs, key=lambda var: var.proxy.name)]
         outputs = [unvariableify(out) for out in sorted(region.outputs, key=lambda var: var.proxy.name)]
 
+        from thunder.executors.passes import _del_last_used
+
+        region.bound_symbols = _del_last_used(region.bound_symbols, outputs)
+
         fusion_name = f"CUDAGraph{fusion_counter}"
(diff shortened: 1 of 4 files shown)


IvanYashchuk:
I'm sorry it's unclear what problem you're seeing with `del`. Is it possible that you're hitting the same problem reported in https://github.com/Lightning-AI/lightning-thunder/issues/728?

thunder/executors/cudagraphex.py
```diff
@@ -138,6 +138,10 @@ def fuse(self, region: Region, fusion_counter: int, num_static_inputs: None | in
         inputs = [unvariableify(inp) for inp in sorted(region.inputs, key=lambda var: var.proxy.name)]
         outputs = [unvariableify(out) for out in sorted(region.outputs, key=lambda var: var.proxy.name)]
 
+        from thunder.executors.passes import _del_last_used
+
+        region.bound_symbols = _del_last_used(region.bound_symbols, outputs)
```
IvanYashchuk, line 143: Does Thunder need region transforms?

What's the benefit of having `del` calls at all inside the CUDA Graphed region? CUDA Graphs are used to escape Python and `del` is a Python-only concept.

t-vi: Would we not need them for memory reuse inside the cudagraph?

t-vi: OK, I'll remove these bits then. :)

t-vi: Thinking about it more, I am still confused here. For very large graphs we would be glad to re-use memory, I think, and I would hope to achieve this by `del`ing during capture...

t-vi:
> I'm sorry it's unclear what problem you're seeing with del. Is it possible that you're hitting the same problem reported in https://github.com/Lightning-AI/lightning-thunder/issues/728?

Yeah, I'm seeing fused dels of returns. I think it is better to do the del after use after (and within) the cudagraph because it is the final form. The trouble in the CI comes from me not touching autograd yet.

IvanYashchuk:
> handling autograd's transform for execution (@IvanYashchuk any advice if we can move the delete last used to after autograd, i.e. to thunder.jit.fn_?)

Yes, do it for the forward trace, I wouldn't touch the backward trace. The alternative is to move `cudagraphex.fusion_pass` inside the torch autograd path. Any backward trace manipulation should be handled inside this function https://github.com/Lightning-AI/lightning-thunder/blob/010c29889876fa761ec1b5feb145129d745e3a52/thunder/executors/torch_autograd.py#L109
The existence of a backward trace as a standalone object outside of this function should be hidden and it should always be attached to a forward trace which is done in https://github.com/Lightning-AI/lightning-thunder/pull/214. Before that PR is merged you are free to reorganize the compilation flow however you feel is convenient to follow the logic and however it brings runtime or trace time benefits.

PR #893, opened 2024-07-31
ci: split GPU test to two pools

t-vi:
Thank you @Borda for great advice and branch.


Borda (approved):
it is running the [expected machine](https://dev.azure.com/Lightning-AI/lightning/_build/results?buildId=210563&view=logs&j=47e66f3c-897a-5428-da11-bf5c7745762e&t=3a07c7c5-d402-5444-ef43-19ed5f298875)

t-vi:
But it seems that the distributed is very slow?

Borda:
> But it seems that the distributed is very slow?

true, it goes from 14min to 24min so may be worse to investigate if it is HW or CUDA issue

t-vi:
The A100 pool still is inexplicably slow, so I'm closing this for now, we might look into it again when we need to increase the parallel jobs.

PR #894, opened 2024-07-31
Add nvFuser fusion region repro util and update DebugTransform

riccardofelluga:
<details>

</details>

## What does this PR do?

From discussions in #387 and #783, this PR adds two helper functions:
- One to print a repro script for a specific nvFuser segment; and
- Another to get a fusion definition from the trace(avoids manually querying the context).

I also added a small section in the documentation to explain the usage of the helper function.

Also in this PR an update to the DebugTransform such that it is possible to debug symbols before and after any interesting section of the trace. I don't really like that transform since it's an hack but for now it does the job. I think in the future we should consider adding a debug prim to intercept inputs and add information to a trace.

## Usage:
As described in the documentation to use this feature you need two things:

1. Add the `nv_store_fusion_inputs=True` compile option; and
2. Call `get_nvfuser_repro` with the trace and the name of the fusion you want to see the repro of.

Note: ~This PR depends on nvFuser PR [#2732](https://github.com/NVIDIA/Fuser/pull/2732). Hence I will convert this PR to not a draft when that one passes.~ MERGED!

### DIFF
--- a/docs/source/basic/inspecting_traces.rst
+++ b/docs/source/basic/inspecting_traces.rst
@@ -189,7 +189,7 @@ which prints::
     }
   }
 
-Moreover, if you are just interested in running a specific nvFuser segment, Thunder offers an handy helper fuction. The ``get_nvFuser_repro()`` function takes a trace and a fusion name as input and returns it's repro script::
+Moreover, if you are just interested in running a specific nvFuser segment, Thunder offers an handy helper function. The ``get_nvFuser_repro()`` function takes a trace and a fusion name as input and returns it's repro script::
 
   from thunder.examine import get_nvFuser_repro
 
--- a/docs/source/basic/inspecting_traces.rst
+++ b/docs/source/basic/inspecting_traces.rst
@@ -189,7 +189,7 @@ which prints::
     }
   }
 
-Moreover, if you are just interested in running a specific nvFuser region without Thunder, you can use an handy helper function. The ``get_nvfuser_repro()`` function takes a trace and a fusion name as input and returns it's repro script::
+Moreover, if you are just interested in running a specific nvFuser region without Thunder, you can use a handy helper function. The ``get_nvfuser_repro()`` function takes a trace and a fusion name as input and returns it's repro script::
 
   from thunder.examine import get_nvfuser_repro
   ...
--- a/docs/source/basic/inspecting_traces.rst
+++ b/docs/source/basic/inspecting_traces.rst
@@ -189,17 +189,25 @@ which prints::
     }
   }
 
-Moreover, if you are just interested in running a specific nvFuser segment, Thunder offers an handy helper function. The ``get_nvFuser_repro()`` function takes a trace and a fusion name as input and returns it's repro script::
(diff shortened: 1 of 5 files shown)


docs/source/basic/inspecting_traces.rst
```diff
@@ -189,6 +189,46 @@ which prints::
     }
   }
 
+Moreover, if you are just interested in running a specific nvFuser region without Thunder, you can use an handy helper function. The ``get_nvfuser_repro()`` function takes a trace and a fusion name as input and returns it's repro script::
```
crcrpar, line 192:
```suggestion
Moreover, if you are just interested in running a specific nvFuser region without Thunder, you can use a handy helper function. The ``get_nvfuser_repro()`` function takes a trace and a fusion name as input and returns it's repro script::
```

thunder/executors/nvfuserex_impl.py
```diff
@@ -479,6 +484,9 @@ def create_fusion_definition_wrapper(
     #   objects is captured by names in the trace.
     # These properties are distinct from the inputs and outputs to the trace itself, which
     #   may contain duplicates and whose order must be preserved.
+    store_inputs: None | bool = get_compile_option(
+        "nv_store_fusion_inputs", "Enable nvFuser store fusion inputs for repro."
```
crcrpar, line 488: "Enable nvFuser to store fusion inputs for repro" or "Let nvFuser store fusion inputs for repro"?

thunder/examine/__init__.py
```diff
@@ -216,3 +216,41 @@ def get_fusion_symbols(trace: TraceCtx, warn_if_fusions_unavailable: bool = True
...
+
+    if fusion.last_used is None:
+        assert (
+            False
+        ), "Fusion definition needs to be executed to record the inputs. You must execute the fusion first before you can query the repro."
+
+    if fusion.last_inputs is None:
+        assert (
+            False
+        ), "Fusion definition inputs need to be recorded. Use compile option 'nv_store_fusion_inputs=True' while tracing."
```
crcrpar, line 253: why using `assert`, rather than raise?

thunder/executors/nvfuserex_impl.py
riccardofelluga: Yeah I red the descriptions of the other nv compile options and I went for something similar but it better to phrase it differently

thunder/examine/__init__.py
riccardofelluga: There is not particular reason, let me switch to raise

kshitij12345 (approved):
LGTM @riccardofelluga, thank you!

Only question I had was - do you think we should have a small test for `get_nvfuser_repro`?

riccardofelluga:
Thanks @kshitij12345 ! It is true that there is no thing such as excess of caution 😁 however since the method relies on a feature from nvFuser it might be risky to test it against an hardcoded result because it could change on their side.

t-vi (approved):
Thank you @riccardofelluga @kshitij12345 @crcrpar

PR #899, opened 2024-07-31
improve fusion stability

### LINKED ISSUE #898: Nondeterministic reordering of bsyms in `fuse_bound_symbols` (e.g. NVFuser)
When looking at CUDAGraphs, we found that there is a "nondeterministic" (i.e. depends on not controlled conditions) reordering of bound symbols through `fuse_bound_symbols`.
Additionally, the ordering of arguments to fusions depended on the lexicographic order of the input proxy names, which is not great for stability (i.e. identical blocks will get fusions with different orders of arguments) and thus caching.

For example, the following will fail very often on a NVFuser-Enabled thunder installation:

```
import torch
import thunder
class LlamaMLPLike(torch.nn.Module):
    def __init__(self) -> None:
        super().__init__()
        self.fc_1 = torch.nn.Linear(32, 32)
        self.fc_2 = torch.nn.Linear(32, 32)
        self.proj = torch.nn.Linear(32, 32)

    def forward(self, x: torch.Tensor) -> torch.Tensor:
        x_fc_1 = self.fc_1(x)
        x_fc_2 = self.fc_2(x)
        x = torch.nn.functional.silu(x_fc_1) * x_fc_2
       return self.proj(x)


with torch.device("cuda"):
   mlp = torch.nn.Sequential(*[LlamaMLPLike() for _ in range(16)]).requires_grad_(False)
j = thunder.jit(mlp)
j(torch.randn(32, 32, device="cuda"))
lt = thunder.last_traces(j)[-1]
assert all((i%2 +1 == i_2) for i, i_2 in enumerate([int(s.args[1].name.split('_')[-2]) for s in lt.bound_symbols if s.sym.name == "linear" and "fc" in s.args[1].name]))
```

I'll send a patch.

cc @apaz-cli

t-vi:
This stabilizes fusion things orderings (bsyms and args), which is useful for having similar blocks produce similar graph sections even after transform for execution.
Closes #898

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -3,6 +3,6 @@ looseversion ==1.3.0
 lightning-utilities >=0.7.0
 numpy >=1.23.0,<2  # not yet ready for numpy 2
 igraph >=0.10.4
-optree >=0.9.2
+optree >=0.11.0
 opt_einsum >= 3.3.0
 mpmath <1.4.0  # todo: teporarl pin for `NameError: name '_C' is not defined`
--- a/thunder/core/pytree.py
+++ b/thunder/core/pytree.py
@@ -57,6 +57,8 @@ def tree_flatten(args, namespace=""):
 # while generating the split functions.
 tree_map = partial(optree.tree_map, none_is_leaf=True, namespace=OPTREE_NAMESPACE)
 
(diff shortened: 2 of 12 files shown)


t-vi:
There still is a reordering at least in rematerialization. I'll update.

lantiga (approved):
🚀

Issue #864, opened 2024-07-25
dynamo : HF Bert- `NameError: name 't0' is not defined.`


kshitij12345:
See comment below for minimal repro -

Repro - (requires `transformers==4.42.4` as present in `requirements/test.txt`)
```python
from datasets import load_dataset
from transformers import AutoModelForSequenceClassification
from transformers import AutoTokenizer
import torch
from tqdm.auto import tqdm
from torch.optim import AdamW
from transformers import get_scheduler
import torch
from transformers import AutoModelForSequenceClassification
from torch.utils.data import DataLoader

dataset = load_dataset("yelp_review_full")

tokenizer = AutoTokenizer.from_pretrained("bert-base-cased")

def tokenize_function(examples):
    return tokenizer(examples["text"], padding="max_length", truncation=True)

tokenized_datasets = dataset.map(tokenize_function, batched=True)

tokenized_datasets = tokenized_datasets.remove_columns(["text"])
tokenized_datasets = tokenized_datasets.rename_column("label", "labels")
tokenized_datasets.set_format("torch")

small_train_dataset = tokenized_datasets["train"].shuffle(seed=42).select(range(100))
small_eval_dataset = tokenized_datasets["test"].shuffle(seed=42).select(range(100))

train_dataloader = DataLoader(small_train_dataset, shuffle=True, batch_size=8)
eval_dataloader = DataLoader(small_eval_dataset, batch_size=8)

model = AutoModelForSequenceClassification.from_pretrained("bert-base-cased", num_labels=5)
device = torch.device("cuda") if torch.cuda.is_available() else torch.device("cpu")
model.to(device)

optimizer = AdamW(model.parameters(), lr=5e-5)

num_epochs = 3
num_training_steps = num_epochs * len(train_dataloader)
lr_scheduler = get_scheduler(
    name="linear", optimizer=optimizer, num_warmup_steps=0, num_training_steps=num_training_steps
)

progress_bar = tqdm(range(num_training_steps))

model.train()

import thunder
class ThunderJitBackend:
    def __init__(self, **compile_options) -> None:        
        self.thunder_jit_fns = []
        self.dynamo_graphs = []
        self.cnt = 0
        self.compile_options = compile_options

    def compile(self, gm, sample_args):
        self.dynamo_graphs.append(gm)
        gm.real_recompile()
        thunder_jit_fn = thunder.jit(gm, **self.compile_options)
        self.thunder_jit_fns.append(thunder_jit_fn)
[...]

PR #901, opened 2024-07-31
fix - prims.reshape implementation with torchex

kshitij12345:
Fixes https://github.com/Lightning-AI/lightning-thunder/issues/864

The issue here was that there was a discrepancy between the `torch.reshape` (which uses `ltorch.reshape` for meta) implementation and the semantics of `prims.reshape`.

`ltorch.reshape` short-circuits if the `reshape` is a no-op when same shape as input is provided. (`ltorch.reshape` is implemented with `clang.reshape`)
https://github.com/Lightning-AI/lightning-thunder/blob/67ec2c6f4e72f6215f12941bf161915f8f458181/thunder/clang/__init__.py#L1043-L1051

However, `prims.reshape` always returns a new TensorProxy

https://github.com/Lightning-AI/lightning-thunder/blob/67ec2c6f4e72f6215f12941bf161915f8f458181/thunder/core/prims.py#L3098-L3112

This difference in behaviour leads to incorrect values being swapped as output in `transform_for_execution` leading to failures.

------
**NOTE** - I was planning to add a check to `__transform_for_operator_executor_execution` to verify that all inputs proxies to the trace have a consumer but I found that this is not necessarily true (we have a test case for a function where nothing is returned)

https://github.com/Lightning-AI/lightning-thunder/blob/fa307a5656d2b7000a07a489f3d434aa7ba8cfd6/thunder/tests/test_core.py#L303-L306

### DIFF
--- a/thunder/examine/memory_caculation.py
+++ b/thunder/examine/memory_caculation.py
@@ -17,6 +17,7 @@ thunder_alias_operator_list = (
     "transpose",
     "expand",
     "reshape",
+    "torch_prims_reshape_impl",  # torchex implementation of prims.reshape.
     "permute",
     "contiguous",
     "torch_wait_prim_impl",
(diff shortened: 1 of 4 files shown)


t-vi (approved):
Looks good, thank you @kshitij12345

PR #909, opened 2024-08-01
interpreter: avoid reference cycles with methods

### LINKED ISSUE #886: The input of thunder.jit function is not freed until gc.collect is used
### To Reproduce
```py
from torch import nn
import torch
import thunder

import gc

class MyModule(nn.Module):
    def __init__(
        self,
        out_features: int,
    ):
        super().__init__()
        self.activation = nn.ReLU()
        self.layer_norm = thunder.jit(nn.LayerNorm(out_features), executors=[thunder.pytorch_executor])
        # self.layer_norm = nn.LayerNorm(out_features)


    def forward(self, x):
        y = self.activation(x)
        p = self.layer_norm(y)
        return p

torch.manual_seed(20)
m = MyModule(4).to(device="cuda", dtype=torch.float32)
a = torch.randn(2, 4, device="cuda", dtype=torch.float32)

torch.cuda.reset_peak_memory_stats("cuda")
torch.cuda.memory._record_memory_history()
for i in range(4):
    out = m(a)
    # gc.collect()
    
print(torch.cuda.max_memory_allocated() / (1000**2))
torch.cuda.memory._dump_snapshot("other_reluln_thunderfwd_collect.pickle")
```
1.using `self.layer_norm = nn.LayerNorm(out_features)`: the memory of `y`(input of layer_norm) is freed normally after use
2.using `self.layer_norm = thunder.jit(nn.LayerNorm(out_features), executors=[thunder.pytorch_executor])`: the memory of `y` allocated **in the first iteration** is not freed until the program ends
3.using `self.layer_norm = thunder.jit(nn.LayerNorm(out_features), executors=[thunder.pytorch_executor])`+`gc.collect`: the memory of `y` in the first iter is freed until gc.collect is used

The expectation is that thunder will have the same behavior as torch

case 1 memory snapshot:
image
case 2 memory snapshot:
image
case 3 memory snapshot:
image

cc: @t-vi @IvanYashchuk 

cc @apaz-cli

t-vi:
Fixes #886 

Of course, all the credit is with @kiya00 and @kshitij12345 for finding the ref cycle.

### DIFF
--- a/thunder/core/interpreter.py
+++ b/thunder/core/interpreter.py
@@ -1644,7 +1644,14 @@ def wrap_attribute(plain_result, obj, name):
     pr = ProvenanceRecord(PseudoInst.LOAD_ATTR, inputs=[obj.provenance, name.provenance])
     result = wrap(plain_result, provenance=pr)
 
-    obj.attribute_wrappers[name.value] = result
+    # we want to avoid reference cycles (note that Python also does this)
+    # see https://github.com/Lightning-AI/lightning-thunder/issues/886 for
+    # discussion
+    if not (
+        isinstance(plain_result, (MethodType, BuiltinMethodType, MethodWrapperType))
+        and plain_result.__self__ is obj.value
+    ):
+        obj.attribute_wrappers[name.value] = result
 
     check_self(obj, result)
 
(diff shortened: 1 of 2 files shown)


lantiga (approved):
Great fix, now we know why Python does this too

PR #913, opened 2024-08-02
`state_dict` transform for `fsdp(jit(model))`

crcrpar:
## What does this PR do?

Transform for `fsdp(jit(model)).state_dict()` by all-gathering and unpadding params

### DIFF
--- a/thunder/core/module.py
+++ b/thunder/core/module.py
@@ -187,7 +187,10 @@ class ThunderModule(pytorch.nn.Module):
 
     def state_dict(self, *args: Any, **kwargs: Any) -> Any:
         # this is broken for transformed modules!!!
-        return self._model.state_dict(*args, **kwargs)
+        state_dict = self._model.state_dict(*args, **kwargs)
+        for transform in self._lc_transforms:
+            state_dict = transform.transform_state_dict(state_dict)
+        return state_dict
 
     def load_state_dict(self, *args: Any, **kwargs: Any) -> Any:
         # this is broken for transformed modules!!!
--- a/thunder/core/module.py
+++ b/thunder/core/module.py
@@ -189,7 +189,7 @@ class ThunderModule(pytorch.nn.Module):
         # this is broken for transformed modules!!!
         state_dict = self._model.state_dict(*args, **kwargs)
         for transform in self._lc_transforms:
-            state_dict = transform.transform_state_dict(state_dict)
+            state_dict = transform.transform_state_dict(self, state_dict)
         return state_dict
 
     def load_state_dict(self, *args: Any, **kwargs: Any) -> Any:
--- a/thunder/core/transform_common.py
+++ b/thunder/core/transform_common.py
@@ -367,6 +367,7 @@ class Transform(ABC):
 
     def transform_state_dict(
         self,
+        model: thunder.ThunderModule,
         state_dict: dict[str, Parameter | Tensor],
     ) -> dict[str, Parameter | Tensor]:
         return state_dict
--- a/thunder/core/transform_common.py
+++ b/thunder/core/transform_common.py
@@ -20,6 +20,8 @@ from thunder.core.utils import ProxyDict, producers, check, consumers
 if TYPE_CHECKING:
     from thunder.core.proxies import ProxyInterface
     from thunder.core.symbol import Symbol, VariableInterface
+    from torch import Tensor
+    from torch.nn import Parameter
 
 
 #
@@ -361,6 +363,12 @@ class Transform(ABC):
 
         Note that state dict keys do not include the submodule name as prefix.
         """
+        return state_dict
+
+    def transform_state_dict(
+        self,
+        state_dict: dict[str, Parameter | Tensor],
+    ) -> dict[str, Parameter | Tensor]:
         return state_dict
 
     def transform_trace_additionally(self, computation_trace: Trace, **kwargs):
--- a/thunder/distributed/transforms/fsdp_v2.py
+++ b/thunder/distributed/transforms/fsdp_v2.py
@@ -31,7 +31,7 @@ __all__ = [
 
 
 @dataclass
-class VisitorTransform:
+class FSDPParamUnpaddingVisitor:
     prod_bsym_to_unsharded_and_padding: dict[BoundSymbol, tuple[TensorProxy, int]]
     swap_map: dict[VariableInterface, TensorProxy] = field(init=False, default_factory=dict)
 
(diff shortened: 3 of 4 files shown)


thunder/core/transform_common.py
```diff
@@ -363,6 +364,9 @@ def transform_state_dict_for_submodule(
         """
         return state_dict
 
+    def transform_state_dict(self, model: thunder.ThunderModule, state_dict: dict[str, Any]) -> dict[str, Any]:
```
t-vi, line 367: I know it's a bit more tedious, but in order for this to be composable with things like materialization, quantization etc. without using too much memory, would it be possible not to introduce `transform_state_dict` but use the `transform_state_dict_for_submodule`? (also, it would be complicated to interleave the two)

crcrpar: How to distinguish load_state_dict from state_dict?

t-vi: Oh right, so to my mind, I think this is the reverse so how about we
- add a `reverse_transform_state_dict_for_submodule` that mimics the `transform_state_dict_for_submodule`  but for the opposite direction,
- add a `original_state_dict()` method calling this with the transforms in reverse order and for all submodules,
- also add a `save_original_state_dict(filename)` method that will actually do the same as `torch.save(self.original_state_dict(), filename)` but does so in an incremental way (i.e. does not materialize the full state dict in memory).

WDYT?

thunder/core/transform_common.py
```diff
@@ -363,6 +365,13 @@ def transform_state_dict_for_submodule(
         """
         return state_dict
 
+    def transform_state_dict(
```
t-vi, line 368: can we make a `reverse_transform_state_dict_for_submodule` instead, please?
(we also probably need the new parameters (`shared_names` and `processed_names` for shared params))

thunder/distributed/transforms/fsdp_v2.py
```diff
@@ -166,3 +170,41 @@ def transform_traces_pre_prologue(self, prologue_trace, computation_trace, epilo
                 provenance="fsdp pass with unpadding",
             )
         return prologue_trace, new_computation_trace, epilogue_trace
+
+    def transform_state_dict(
```
t-vi, line 174: I would like to make this per submodule (similar to `transform_state_dict_for_submodule` but the reverse for the option to deal with it incrementally.

thunder/core/module.py
```diff
@@ -187,7 +187,10 @@ def __getattr__(self, name: str) -> Any:
 
     def state_dict(self, *args: Any, **kwargs: Any) -> Any:
         # this is broken for transformed modules!!!
-        return self._model.state_dict(*args, **kwargs)
+        state_dict = self._model.state_dict(*args, **kwargs)
+        for transform in self._lc_transforms:
```
t-vi, line 191: I think the transforms should be iterated in reverse.
```suggestion
        for transform in reverse(self._lc_transforms):
```

thunder/core/module.py
```diff
@@ -187,7 +187,10 @@ def __getattr__(self, name: str) -> Any:
 
     def state_dict(self, *args: Any, **kwargs: Any) -> Any:
         # this is broken for transformed modules!!!
-        return self._model.state_dict(*args, **kwargs)
+        state_dict = self._model.state_dict(*args, **kwargs)
```
t-vi, line 190: This should probably stat with `self.state_dict()` and then reverse transform it.
I would love to have code like this, but in a method `original_state_dict`.

t-vi (commented):
I think this a mechanism like this would be great to have and I would be keen to implement it for quantization! (And I never thought of gathering to undo sharding.) However, I think that we would want to do it incrementally and with a bit of a different name.

So to my mind, I think this something like the following would be ideal
- add a `reverse_transform_state_dict_for_submodule` that mimics the `transform_state_dict_for_submodule`  but for the opposite direction,
- add a `original_state_dict()` method calling this with the transforms in reverse order and for all submodules,
- also add a `save_original_state_dict(filename)` method that will actually do the same as `torch.save(self.original_state_dict(), filename)` but does so in an incremental way (i.e. does not materialize the full state dict in memory). - as we want to deal with things that don't fit into memory.

WDYT?

thunder/core/module.py
crcrpar: why?

t-vi: So the transform_module / transform_state_dict_for_submodule is applied in the order of the transforms. This is a sort of inverse transformation.
Say (not working yet) you have a quantization and a fsdp transform.
Then in load_original_state_dict you have
```
sd for untransformed ("original" module), unquantized unsharded
--> quant_trafo.transform_state_dict_for_submodule
sd for quantized, unsharded module
--> fsdp_trafo.transform_state_dict_for_submodule
sd for quantized, sharded module
```
Now the `original_state_dict` would need to  go the other way:
```
sd for quantized, sharded module
--> fsdp_trafo.reverse_transform_state_dict_for_submodule
sd for quantized, unsharded module
--> quant_trafo.reverse_transform_state_dict_for_submodule
sd for untransformed ("original" module), unquantized unsharded
```
for this to work, we would need the reverse order of the transforms.

mpatel31415:
Hi! Can I please know what is the status of this MR? I wanted to look at to today, but I got closed in the meantime :)

thunder/distributed/transforms/fsdp_v2.py
crcrpar: I'm yet to realize why "state_dict_for_submodule" is important, not vanilla state dict

crcrpar: and what is the way to reconstruct `state_dict` from state_dict_per_module?

crcrpar:
@mpatel31415 I'm going to open a new pull request instead of updating this one. There've been many changes that have something to do with this PR or what this tries to achieve

thunder/distributed/transforms/fsdp_v2.py
t-vi: We don't want to materialize the entire state dict for models where the state dict does not fit into memory untransformed. Pretty much the same thing as why we do transform_state_dict_for_submodule or why the classic fsdp materialization did that submodule by submodule.
We can reconstruct `state_dict` as similar to what we do in `load_original_state_dict`, but here is the thing: in litgpt we have a thing for incrementally saving state dicts, so we can adapt that to get the state dict straight to disk: https://github.com/Lightning-AI/litgpt/blob/1bd860000a8daa3bbd263ba5eed9911189db0dbd/litgpt/utils.py#L238

PR #914, opened 2024-08-02
remove `transform_trace_additionally` 

### LINKED ISSUE #640: Transforms as EarlyTransforms

k223kim:
<details>

</details>

## What does this PR do?

partially fixes #640.

## PR review

## Did you have fun?

⚡️


thunder/__init__.py
```diff
@@ -589,39 +589,39 @@ def get_computation_and_inputs(*args, **kwargs):
                     # by split_forward_backward
                     extraces = cs.last_traces
 
-            if backward_trc is None:
```
t-vi, line 592: This deletes too much. Only the for loop L593-605 should go

k223kim: Ok! Sounds good. Let me update that!

thunder/__init__.py
```diff
@@ -637,16 +638,7 @@ def get_computation_and_inputs(*args, **kwargs):
 
             if backward_trc is None:
                 ## EPILOGUE and TRANSFORMS should not mix...
```
t-vi, line 640: can you remove this comment, too please?
also the cs.last_computation_transformation_start/stop could be removed both from the computation here and indeed from the compile stats class.
Finally, we want to drop the `transform_additionally` from the `Transform` class, as it will not be used after this PR.

thunder/core/transforms.py
```diff
@@ -1440,16 +1440,24 @@ def grad_func(*args, **kwargs):
...
     class _GradTransform(Transform):
-        def transform_trace_additionally(self, trc: Trace, *, executors_list: Sequence[Any]) -> Trace:
+        def transform_traces_pre_prologue(
+            self,
+            prologue_trc: Trace,
+            computation_trc: Trace,
+            epilogue_trc: Trace | None,
+            *,
+            executors_list: Sequence[Any],
+        ) -> Trace:
```
t-vi, line 1450:
```suggestion
        ) -> Trace:
            computation_trace = dce(computation_trace)
```
As discussed on slack, but here for transparency: we need a dce pass here...

thunder/core/transforms.py
```diff
@@ -1485,7 +1493,7 @@ def symbol_to_eval(bound_symbol):
 )
 
 
-def eval_trace(trace, *args, symbol_mapper=symbol_to_eval, with_env=False, **kwargs):
+def eval_trace(trace, *args, symbol_mapper=symbol_to_eval, with_env=False, allow_duplicates=False, **kwargs):
```
t-vi, line 1496:
```suggestion
def eval_trace(trace, *args, symbol_mapper=symbol_to_eval, with_env=False, **kwargs):
```
to not need `allow_duplicates` here. (And we don't need the changes below.)

thunder/__init__.py
```diff
@@ -569,6 +569,7 @@ def get_computation_and_inputs(*args, **kwargs):
             transform: Transform
             for transform in transforms:
                 thunder.core.utils.check_type(transform, Transform)
+
```
t-vi, line 572: Nitpicking: could you please remove the whitespace-only changes?

t-vi (commented):
Hi @k223kim . Great! The old transform hook is almost gone. :rocket: 
I left a few comments regarding and in addition to what we discussed on slack.

thunder/__init__.py
k223kim: yes absolutely! (wanted to remove it after I was sure that `transform_additionally` can be completely removed :)

thunder/__init__.py
k223kim: sorry, I think this was due to precommit..

thunder/core/transforms.py
k223kim: However, when we do want to allow duplicates (for other reasons) won't it be better to keep these changes?

t-vi: So we don't allow the duplicates because it would be problematic, so I would venture we don't add the override option until we really know we need it.

t-vi:
Could you rebase / resolve the confilicts, please?

thunder/core/transforms.py
k223kim: Sounds great! Let me update it :)

thunder/core/transforms.py
```diff
@@ -1440,16 +1439,24 @@ def grad_func(*args, **kwargs):
...
+            executors_list: Sequence[Any],
+        ) -> Trace:
             # Using trc.python_callable() makes it impossible to retrace the
             # function because the python_callable uses python_ctx which replaces
             # symbol occurrences with its symbol._call_ctx function
-            @wraps(trc.python_callable())
+            @wraps(computation_trc.python_callable())
             def python_callable(*args, **kwargs):
-                return eval_trace(trc, *args, **kwargs)
+                computation_trc = dce(computation_trc)
```
t-vi, line 1455: I would move this up to right at the beginning of `transform_traces_pre_prologue` then you don't need to modify the trace here.

t-vi (approved):
Looks great, thank you @k223kim !

PR #916, opened 2024-08-02
Trace cleanup, nvfuser version warning

apaz-cli:
1. Rename `compile_and_exec()` to something more sensible (since it does not in fact execute the trace in the way you would think it does by the name)
2. Rename confusing variable names
3. Add warning for if nvfuser is out of date.
4. Better error message for if a member of the executor list is `None`. It's been debated for a while if None is a valid argument or not, but at the moment if you pass None you get an `AttributeError` on `ex._provenance`.

### DIFF
--- a/thunder/core/baseutils.py
+++ b/thunder/core/baseutils.py
@@ -456,7 +456,7 @@ def build_callable(fn_name: str, python_str: str, file_name: str, ctx: dict) -> 
     try:
         code: CodeType = compile(python_str, program_name, mode="exec")
         exec(code, ctx)
-        _callable: Callable = ctx[fn_name] # Grab from globals()
+        _callable: Callable = ctx[fn_name]  # Grab from globals()
         return _callable
     except Exception as e:
         print("Encountered an exception while trying to compile the following program:")
--- a/thunder/core/baseutils.py
+++ b/thunder/core/baseutils.py
@@ -442,10 +442,10 @@ def print_base_printable(x: Any, /) -> str:
 _exec_ctr = 0
 
 
-def compile_and_exec(fn_name: str, python_str: str, program_name: str, ctx: dict) -> Callable:
+def build_callable(fn_name: str, python_str: str, file_name: str, ctx: dict) -> Callable:
     global _exec_ctr
 
-    program_name = f"{program_name}_{_exec_ctr}"
+    program_name = f"{file_name}_{_exec_ctr}"
 
     # simple cache hack
     mtime = None  # this signals that the cache should not be invalidated(!)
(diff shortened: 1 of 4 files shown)


thunder/extend/__init__.py
```diff
@@ -458,6 +458,8 @@ def resolve_executors(executors: None | Sequence[Executor | str]) -> tuple[Execu
                 continue
             else:
                 resolved_executors.append(ex)
+        elif e is None:
```
t-vi, line 461: No. If we want to do this, we need to check for "other than executor" and error with what we got.
`executors=(my_exec, thunder.get_default_executors())`

thunder/core/baseutils.py
```diff
@@ -442,10 +442,10 @@ def print_base_printable(x: Any, /) -> str:
 _exec_ctr = 0
 
 
-def compile_and_exec(fn_name: str, python_str: str, program_name: str, ctx: dict) -> Callable:
+def build_callable(fn_name: str, python_str: str, file_name: str, ctx: dict) -> Callable:
```
t-vi, line 445: I must admit I'm puzzled.

thunder/extend/__init__.py
apaz-cli: What do you mean by "other than executor"? As in, check if it isn't a string or an executor, not just None?

thunder/core/baseutils.py
apaz-cli: Yeah. Basically, `ctx` gets repurposed as `globals()` by `exec()`. Then we pull the function object from that globals dict. Don't know why we decided to do it this way, but it works.

thunder/extend/__init__.py
apaz-cli: Fixed. I think the `None` case is going to be really common though, so I retained it for a more informative error message.

t-vi: I would prefer the None to show the same error message, because it expresses that it should be an executor (with the string being the lazy option) in that list and it saves unneeded special casing.

thunder/core/baseutils.py
t-vi: OK.

t-vi (approved):
Seems good, thank you @apaz-cli but let's lose the None case.

thunder/extend/__init__.py
```diff
@@ -458,6 +458,10 @@ def resolve_executors(executors: None | Sequence[Executor | str]) -> tuple[Execu
...
```
t-vi, line 461:
```suggestion
```

thunder/extend/__init__.py
```diff
@@ -458,6 +458,10 @@ def resolve_executors(executors: None | Sequence[Executor | str]) -> tuple[Execu
...
+            raise ValueError(f"The executor list cannot contain None. Executor list: {executors}")
```
t-vi, line 462:
```suggestion
```

PR #918, opened 2024-08-03
Added `LangctxImplWrapper` in place of `@wraps`.

apaz-cli:
Moving towards a pickle-able `TraceCtx` object. They contain bound symbols, which contain symbols, which contain meta functions, which are wrapped in a language context with a function decorated with `@wraps`. But those wrappers are functions with nonlocal variables, so they keep a frame object alive, and that frame object contains a reference to `globals()`, which contains modules, which cannot be pickled. A problem.

So, to resolve this problem, I've created an object to hold the necessary information for the wrapper, without it keeping a frame object alive.

Next on the list is to tackle `sym.module`/`sym._module`. After that, `TraceCtx` objects should be pickleable.

### DIFF
--- a/thunder/core/langctxs.py
+++ b/thunder/core/langctxs.py
@@ -1,7 +1,7 @@
 from contextvars import ContextVar
 from typing import Optional, Any
 from collections.abc import Callable
-from functools import wraps
+from functools import wraps, update_wrapper
 from contextlib import contextmanager
 from enum import Enum, auto
 
@@ -107,6 +107,23 @@ class Languages(Enum):
     PRIMS = auto()
 
 
+
+# Can be pickled, unlike @wraps and partial(). Serves the exact same purpose as wraps().
+class LangctxImplWrapper:
+    def __init__(self, fn: Callable, _langctx: Any | LanguageContext):
+        self.fn = fn
+        self._langctx = _langctx
+        update_wrapper(self, fn)
+
+    def __call__(self, *args, **kwargs):
+        try:
+            tok = set_langctx(self._langctx)
+            result = self.fn(*args, **kwargs)
+            return result
+        finally:
+            reset_langctx(tok)
+
+
 # Decorator and context manager for setting the language context with a function
 #   or region
 #
@@ -125,16 +142,7 @@ class langctx:
         self.langctx: LanguageContext = _langctx
 
     def __call__(self, fn: Callable, /) -> Callable:
-        @wraps(fn)
-        def _fn(*args, **kwargs):
-            try:
-                tok = set_langctx(self.langctx)
-                result = fn(*args, **kwargs)
-                return result
-            finally:
-                reset_langctx(tok)
-
-        return _fn
+        return LangctxImplWrapper(fn, self.langctx)
 
     def __enter__(self):
         self.tok = set_langctx(self.langctx)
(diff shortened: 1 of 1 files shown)


apaz-cli:
It's funny that my commit message mentioned someone named `wraps` on github. I assume they get a lot of those.

t-vi:
Hi @apaz-cli . Thank you for working on pickling traces. It is very dear to us.

But let's not do this and keep symbols as they are.

We only need include pickle the .module and the .name of the symbol (with the expectation of finding it there when we unpickle, any symbol not obtainable either though that or one of the registered executors has a bug) while reducing a symbol. Please do *not* try to pickle or unpickle symbols themselves, we absolutely need thunder to use the ones that exist in the unpickled trace.

So again, the three things that need to be pickled and restored are
- the bound symbols
- the proxies,
- the trace objects.

If you need advice around how to pickle symbols or have any doubts around it, please chat me up on slack, I'm on the train tomorrow but should have reasonable internet with not too large gaps from ~my noon to ~my midnight.

apaz-cli:
@t-vi Unpickling a function looks inside the module and loads the correct function from that module. If it can't do that, the function object isn't pickleable. So, we should be good on that front.

Also, you can't pickle modules. Or at least `thunder.core.prims` cannot be pickled. So, what you're describing doesn't sound like it's possible through `pickle.dumps()`.

We could instead disconnect the bound symbols from the trace context, pickle the rest of the trace context, and use some other method for rebuilding the symbols, re-registering them with the appropriate executors, and then putting the bound symbols back together before finally putting them back on the unpickled trace object. But that didn't sound like a V0 implementation, hence why I wanted to pickle the trace directly as originally planned.

t-vi:
So standard Python objects have this: functions, types --> look up, objects --> create a new instance.

It is crucial that symbols need to be looked up, not recreated even though to Python they are objects.

What I had in mind is either doing that on the bound symbol itself or if you want on the symbol like in the following quick experiment:
```
import thunder, pickle, sys

def lookup_symbol(name, executor, module):
    if executor is None:
        if module not in sys.modules:
            raise RuntimeError("cannot find module {module} needed for symbol {name}")
        sym = getattr(sys.modules[module], name)
        assert isinstance(sym, thunder.core.symbol.Symbol)
        return sym
    else:
        # find the executor in all executors and then the symbol
        raise NotImplementedError()


def __reduce__(self):
    return (lookup_symbol, (self.name, self.executor if self.executor is None else self.executor.name, self.module if self.module is None else self.module.__name__))

thunder.core.symbol.Symbol.__reduce__ = __reduce__

# works
assert pickle.loads(pickle.dumps(thunder.core.prims.unpack_trivial)) is thunder.core.prims.unpack_trivial
````

Now the `__reduce__` for other things can return symbols and just work.

The trouble you'll run into is that the proxies (and possible also the bound symbols) want a tracectx to be created.
For the v0 we might either just use a throwaway tracectx for them or just solve this by not aiming to pickle these by themselves, but only writing a `__reduce__` and `unpack_trace` for the trace itself that roughly creates a trace similar to how e.g. this function

https://github.com/Lightning-AI/lightning-thunder/blob/510d8bf34975e46117ef42538a140a5ff2cca2ab/thunder/transforms/quantization.py#L49.

does, except that the bound symbols will need to be created from say, a tuple of (sym, bind_args, bind_kwargs) or so.

PR #921, opened 2024-08-05
[benchmark] migrate to fsdp/ddp after jit, from fsdp/ddp before jit

crcrpar:
- [ ] run benchmarks

cc @mpatel31415

### DIFF
--- a/thunder/benchmarks/benchmark_litgpt.py
+++ b/thunder/benchmarks/benchmark_litgpt.py
@@ -247,13 +247,16 @@ class Benchmark_litGPT:
         if self.checkpoint_activations:
             self.setup_activation_checkpointing()
 
+        # Compile the model
+        self.model = self.setup_compile(self.model)
+
+        # Add required transforms to convert model into thunder's distributed parallel
+        self.model = self.apply_thunder_transforms(self.model)
+
         # Initialize the optimizer after the model is sharded if using FSDP
         self.optimizer = configure_optimizers(
             self.model, weight_decay, learning_rate, (beta1, beta2), device_type="cuda"
         )
-
-        # Compile the model
-        self.model = self.setup_compile(self.model)
 
         # Setup the Dummy dataloader for training
         self.train_dataloader = self.setup_dummy_dataloader()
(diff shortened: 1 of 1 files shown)


thunder/benchmarks/benchmark_litgpt.py
```diff
@@ -275,80 +278,85 @@ def init_model(self):
...
+                sharding_strategy=sharding_strategy,
+                bucketing_strategy=bucketing_strategy,
+            )
+            return model
+        else:
+            raise ValueError(f"Invalid {self.distributed_mode=}")
+
     def setup_distributed(self, model):
-        if self.distributed_mode == "none":
+        if self.distributed_mode == "none" or "thunder" in self.compile:
```
parthmannan, line 313: Not needed but just a thought - Since we are applying Thunder distributed transforms in a separate function, can we move the check to the `setup_distributed` call on line 244 just so that it is clear that `setup_distributed` doesn't do any Thunder-related things from just reading the init function calls? wdyt?

thunder/benchmarks/benchmark_litgpt.py
```diff
@@ -393,7 +401,15 @@ def setup_compile(self, model):
 
                 executors.insert(0, transformer_engine_ex)
 
-            model = thunder.jit(model, executors=executors)
+            transforms = []
+            if self.distributed_mode == "fsdp":
+                from thunder.transforms import MaterializationTransform
```
parthmannan, line 406: I probably need to learn what this is doing but if there could be a ELI5 on the PR for this change, that would be great.

parthmannan:
There may already be a plan for this since this is still a draft and apologies if I am jumping ahead but given the big increase in memory consumption with this change and many workloads from @mpatel31415's benchmarks already failing with OOM error, do you think we can make this change optional?

thunder/benchmarks/benchmark_litgpt.py
crcrpar: Without this transform, the materialization of parameters is deferred to the first forward call, after the optimizer's construction while the fused optimizer requires all the params to be materialized on a certain device. This `MaterializationTransform` is helpful in this regard. Of course, I'll add some comment to the code but just so you know.

parthmannan:
Just so I understand the snapshot above, the blue markers are memory allocation during the training step right? Do we know the reason why `fsdp(jit(model))` has higher consumption? Is it because once the model is jit, it cannot shard all the parameters as some get hidden inside fusion blocks? Or perhaps, fsdp after jit means it has to AllGather larger buckets due to fusion blocks?

crcrpar:
> Just so I understand the snapshot above, the blue markers are memory allocation during the training step right? Do we know the reason why `fsdp(jit(model))` has higher consumption? Is it because once the model is jit, it cannot shard all the parameters as some get hidden inside fusion blocks? Or perhaps, fsdp after jit means it has to AllGather larger buckets due to fusion blocks?

The piled layers that's spanning horizontally in the bottom capture seem to indicate that materialize and/or sharding of `fsdp(jit(model))` leaves some parameters dangling somewhere while they are not used.

I do think it's because `fsdp(jit(model))` copies params before sharding, after materializing while `jit(fsdp(model))` directly applies sharding to params.
#932 would be a workaround when a model is initialized with `torch.device("meta")`.

t-vi (approved):
Great step!
The memory issues discussed are OK after the FSDP series?
I will set up things to merge but leave resolving the conversations to you, so you can decide when to merge.
Thank you @crcrpar @parthmannan .

PR #924, opened 2024-08-05
Set the default language context to be the torch language context

IvanYashchuk:
We already document the torch language context as the default one:
https://github.com/Lightning-AI/lightning-thunder/blob/e05785ed35633b61a8dbb276a4f78a8b514d8e5c/thunder/__init__.py#L289
https://github.com/Lightning-AI/lightning-thunder/blob/e05785ed35633b61a8dbb276a4f78a8b514d8e5c/thunder/common.py#L242-L243

but the language context is not set in all cases where method resolution is needed and that led to developers to add the `langctx("torch")` decorators/context managers in places where it's not needed but where it's needed it was confusing why do we require setting a PyTorch language when we're using and accelerating PyTorch. This PR modifies the main function developers are expected to use to get a current language context and when nothing is explicitly set the default option now is a PyTorch language context.
When people want to use another context they can still do that, but now the default mode eases the cognitive load of using and extending Thunder for PyTorch use cases.

Thanks to @ptrblck and the team for the feedback!

### DIFF
--- a/thunder/core/transforms.py
+++ b/thunder/core/transforms.py
@@ -586,16 +586,6 @@ def clear_grads(module: torch.nn.Module) -> None:
         b.grad = None
 
 
-from thunder.core.interpreter import make_opaque
-from thunder.core.langctxs import langctx, Languages
-
-
-# TODO RC1 Replace with langctx
-def torchctx(fn):
-    _fn = langctx(Languages.TORCH)(fn)
-    return make_opaque(_fn)
-
-
 _grad_fn_map: dict[Any, Callable] = {}
 
 
@@ -639,7 +629,6 @@ register_grad(pids.UNPACK_SEQUENCE, prims.unpack_sequence)
 #
 # Data movement and transformation operator grads
 #
-@torchctx
 def _convert_element_type_prim_grad(a: Number | TensorProxy, dtype: type | dtypes.dtype) -> Number | TensorProxy:
     fwd = prims.convert_element_type(a, dtype)
 
@@ -678,7 +667,6 @@ register_grad(pids.UNIFORM, _uniform_grad)
 #
 
 
-@torchctx
 def _broadcast_in_dim_prim_grad(
     a: TensorProxy, shape: Sequence[int], broadcast_dimensions: Sequence[int]
 ) -> TensorProxy:
@@ -707,7 +695,6 @@ def _broadcast_in_dim_prim_grad(
 register_grad(pids.BROADCAST_IN_DIM, _broadcast_in_dim_prim_grad)
 
 
-@torchctx
 def _cat_prim_grad(tensors: list[TensorProxy], /, dim: int) -> TensorProxy:
     fwd = prims.cat(tensors, dim)
 
@@ -742,7 +729,6 @@ def _reshape_prim_grad(a: TensorProxy, shape: tuple[int, ...]) -> TensorProxy:
 register_grad(pids.RESHAPE, _reshape_prim_grad)
 
 
-@torchctx
 def _slice_prim_grad(
     a: TensorProxy, start_indices: Sequence[int], end_indices: Sequence[int], strides: None | Sequence[int] = None
 ) -> TensorProxy:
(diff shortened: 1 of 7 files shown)


t-vi (approved):
Thank you @IvanYashchuk

thunder/core/transforms.py
```diff
@@ -639,7 +629,6 @@ def put_grads(a, g):
 #
 # Data movement and transformation operator grads
 #
-@torchctx
```
mruberry, line 642: The removal of `@torchctx` here is interesting.

If a practitioner sets a different language context, how would these operations run in the torch language context?

IvanYashchuk: I will go through this list and check what functions require torch language context. We should use it only when needed for method resolution that is specific to torch, like `.size`.

Should the user-supplied language context be used only for trace acquisition?

Also, NumPy language is broken (https://github.com/Lightning-AI/lightning-thunder/issues/927):
```py
In [1]: import thunder.numpy
---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
Cell In[1], line 1
----> 1 import thunder.numpy

File ~/dev/lightning-thunder/thunder/numpy/__init__.py:7
      3 from collections.abc import Callable
      5 import numpy as np
----> 7 from thunder.core.langctx import langctx, Languages
      8 from thunder.numpy.langctx import register_method
     10 from thunder.core.proxies import TensorProxy

ModuleNotFoundError: No module named 'thunder.core.langctx'
```

mruberry: > I will go through this list and check what functions require torch language context. We should use it only when needed for method resolution that is specific to torch, like `.size`.
> 
> Should the user-supplied language context be used only for trace acquisition?

Yes. Whenever running a transform then the transform should define its language context. If the transform doesn't need to define a language context (because it doesn't access any tensor attributes) then I suppose it can get away without doing so. Another option would be to automatically use the torch language context for any transform, unless it explicitly specifies another language context.

> 
> Also, NumPy language is broken (#927):
> 
> ```python
> In [1]: import thunder.numpy
> ---------------------------------------------------------------------------
> ModuleNotFoundError                       Traceback (most recent call last)
> Cell In[1], line 1
> ----> 1 import thunder.numpy
> 
> File ~/dev/lightning-thunder/thunder/numpy/__init__.py:7
>       3 from collections.abc import Callable
>       5 import numpy as np
> ----> 7 from thunder.core.langctx import langctx, Languages
>       8 from thunder.numpy.langctx import register_method
>      10 from thunder.core.proxies import TensorProxy
> 
> ModuleNotFoundError: No module named 'thunder.core.langctx'
> ```

It's not surprising to me that the NumPy language context is broken. We could argue that we should only support the torch language context, and other practitioner-facing language contexts shouldn't even exist so we can remove some of the language context complexity, but that could be a pain later. We'd also have to think about how the clang and prim language contexts would continue to work (or not).

IvanYashchuk: I have added wrapping all functions that are registered with `register_grad` with torch language context in https://github.com/Lightning-AI/lightning-thunder/pull/924/commits/d1f71d92077b36b128327c16543d6f30a7e11a68. If someone wants to write the derivative rule in a different context they can still do that because the inner context overwrites outer contexts.

thunder/core/transforms.py
```diff
@@ -587,24 +588,22 @@ def clear_grads(module: torch.nn.Module) -> None:
...
 _grad_fn_map: dict[Any, Callable] = {}
 
 
 def register_grad(sym_or_id: Symbol | Any, gradfn: Callable) -> None:
     id: Any = sym_or_id
     if isinstance(sym_or_id, Symbol):
         id = sym_or_id.id
-    _grad_fn_map[id] = gradfn
+
+    # The gradfn are expected to be written in terms of torch functions by
```
mruberry, line 599: Nice comment

mruberry (approved):
Cool!

We should fix the `numel` uses sometime (as the issue tracks)

Issue #871, opened 2024-07-26
Recomputation of saved for backward tensors


IvanYashchuk:
## 🚀 Feature

In Thunder, we have one memory-saving optimization called [`rematerialization`](https://github.com/Lightning-AI/lightning-thunder/blob/7e1f9bd95c29b9642173088e4d5c86c84db434cc/thunder/core/rematerialization.py#L505) but it's applicable only for fusion regions and does nothing for computation outside the fusion regions. Rematerialization is not only a memory-saving optimization but also performance-improving because the general expectation the less number of global memory reads and write a fused region needs to do the better performance is even though more computation is introduced into the fused region.

However, having just this memory-saving pass is not enough for large context window or large parameter count models. The structure of backward computation requires additional information about the intermediates of the forward computation, this is usually referred to as “activations” or in PyTorch code “saved for backward tensors”. The requirement to store these tensors in GPU memory contributes to the large memory consumption peaks. Here’s an example of how the total size of saved for backward tensors scales with the number of layers for the Thunder-compiled LitGPT model with “longchat-7b-16k” configuration:
![saved-tensor-usage-vs-n_layers-longchat-7b-16k](https://github.com/user-attachments/assets/61ae70b8-343f-4e73-ba68-3f582468e277)
67% of the total memory is occupied by saved-for-backward tensors.

PyTorch implements activation checkpointing via a special `torch.autograd.Function` that saves the required inputs in the detached form for recomputation in backward and performs the forward computation under `no_grad` context. Backward uses the stashed inputs and the forward function to restore missing tensors, which are used to invoke the original backward function.
Thunder doesn’t support tracing `torch.autograd.Function.backward` (https://github.com/Lightning-AI/lightning-thunder/issues/666), it also doesn’t support a `no_grad` context manager inside the traced function, it doesn’t recognize PyTorch’s Autograd graph manipulations like `detach()` and `requires_grad_(True/False)`.

PyTorch requires users to use special markers to apply saved tensor recomputation. In Thunder, we can do this automatically.

Thunder’s advantage is that we control both forward and backward computations and we can implement a trace transformation that would recompute “saved for backward tensors”.

There are two places where we could insert the logic for tensor recomputation:

1. Right after the forward and backward traces are created and before the “transform for execution” phase.
[...]

PR #926, opened 2024-08-05
Add get_saved_for_backward_tensors

IvanYashchuk:
It's not documented anywhere that the saved tensors in the traces we generate for PyTorch Autograd integration are stored in `trace.output[1][0]` and it can be inferred only by reading the code. This function will be helpful for https://github.com/Lightning-AI/lightning-thunder/issues/871.

cc @borda @apaz-cli


thunder/core/vjp_utils.py
```diff
@@ -178,3 +178,45 @@ def bw_fn(*args, **kwargs):
...
+        >>> print(saved_tensors)
+        (<TensorProxy(name="a", dtype=thunder.dtypes.float32, shape=())>, <TensorProxy(name="b", dtype=thunder.dtypes.float32, shape=())>)
+    """
+    # First let's check if the trace was generated by Thunder's automatic differentiation
+    # The name is defined at
+    # https://github.com/Lightning-AI/lightning-thunder/blob/681aef744786c85f925a590e6b5e6b5162932adf/thunder/core/transforms.py#L3643
+    utils.check(
+        lambda: trace.fn.__name__ == "augmented_forward_fn",
+        "The trace must be generated by Thunder's automatic differentiation",
+    )
```
riccardofelluga, line 211: Maybe this check is not enough, while testing I ran into the case where `trace.fn` is `None`. Such behavior skips this check and fails on the `trace.output` call with a `NoneType` error. (Likely because the tensors do not require grad in my test)

repro:
```python
import thunder
import torch
from thunder.core.vjp_utils import get_saved_for_backward_tensors
def forward(a, b): return a*b
a = torch.randn((1, 5), device="cuda")
b = torch.randn((1, 5), device="cuda")
jf = thunder.jit(forward)
jf(a, b)
trace = thunder.last_traces(jf)[-1]
get_saved_for_backward_tensors(trace)
```
Outputs:
```bash
  File "/opt/lightning-thunder/test.py", line 10, in <module>
    get_saved_for_backward_tensors(trace)
  File "/opt/lightning-thunder/thunder/core/vjp_utils.py", line 218, in get_saved_for_backward_tensors
    saved_tensors = trace.output[1][0]
  File "/opt/lightning-thunder/thunder/core/proxies.py", line 1513, in __getitem__
    return method(self, key)
TypeError: 'NoneType' object is not callable
```

thunder/tests/test_grad.py
```diff
@@ -1640,3 +1640,21 @@ def foo(a, c):
...
+
+
+def test_get_saved_for_backward_tensors():
+    from thunder.core.vjp_utils import get_saved_for_backward_tensors
+
+    def func(a, b):
+        return a * b
+
+    a = make_tensor((2, 2), device="cpu", dtype=torch.float32, requires_grad=True)
+    b = make_tensor((2, 2), device="cpu", dtype=torch.float32, requires_grad=True)
```
riccardofelluga, line 1652: Probably better to also test when the tensors do not require grad

riccardofelluga (changes requested):

thunder/core/vjp_utils.py
```diff
@@ -178,3 +178,45 @@ def bw_fn(*args, **kwargs):
...
+        >>> trace = thunder.last_traces(jitted_forward)[-1]
+        >>> saved_tensors = get_saved_for_backward_tensors(trace)
+        >>> print(saved_tensors)
+        (<TensorProxy(name="a", dtype=thunder.dtypes.float32, shape=())>, <TensorProxy(name="b", dtype=thunder.dtypes.float32, shape=())>)
+    """
+    # First let's check if the trace was generated by Thunder's automatic differentiation
+    # The name is defined at
+    # https://github.com/Lightning-AI/lightning-thunder/blob/681aef744786c85f925a590e6b5e6b5162932adf/thunder/core/transforms.py#L3643
+    utils.check(
+        lambda: trace.fn.__name__ == "augmented_forward_fn",
```
kshitij12345, line 209: This should be bool - `trace.fn.__name__ == "augmented_forward_fn"` and the next argument should be a lambda producing string.

IvanYashchuk: Thanks for catching this!

thunder/tests/test_grad.py
IvanYashchuk: Done.

thunder/core/vjp_utils.py
IvanYashchuk: Thanks, I added a test that checks that the error is thrown.

IvanYashchuk:
@t-vi, could you please review the code?

nikitaved (approved):
Great step in reducing the number of magic indices! We still have them lurking somewhere...

t-vi (approved):
Thank you @IvanYashchuk @riccardofelluga @nikitaved

PR #920, opened 2024-08-05
Do not reorder between copy_ and return statement

shino16:
## What does this PR do?

Fixes #912. Registers `COPY_` symbols before the return statement as its dependencies.

Issue #912, opened 2024-08-02
`fuse_bound_symbols` (toposort on bsyms) can put `return` before `copy_` on arguments


shino16:
## 🐛 Bug

[`fuse_bound_symbols`](https://github.com/Lightning-AI/lightning-thunder/blob/fa307a5656d2b7000a07a489f3d434aa7ba8cfd6/thunder/executors/data_dependent_partition.py#L300) (toposort on bsyms) puts `return` before `copy_` on arguments.

This causes AssertionError when `torchcompile` or `cudagraphex` executor is applied on in-place operations.

#### Code sample

```py
def f(a):
    a += 1

jitted = thunder.jit(f, executors=[thunder.get_executor("torchcompile")])
jitted(torch.tensor(0))
```

```
...
  File "/opt/pytorch/lightning-thunder/thunder/common.py", line 873, in transform_to_torch_types
    assert last.sym.id == prims.PrimIDs.RETURN
AssertionError
```

Before reaching this line, `TorchCompileExecution` generates
```py
# Constructed by Transform for execution (took 3 milliseconds)
import torch
from thunder.executors.torchex import no_autocast

@torch.no_grad()
@no_autocast
def computation(a):
  # a: "cpu i64[]"

  # /opt/pytorch/lightning-thunder/mshinokawa/sandbox/debug.py:31: 	    a += 1
  return None
  TorchCompile0(a)
    # t0 = ltorch.add(a, 1, alpha=None)  # t0: "cpu i64[]"
      # t0 = prims.add(a, 1)  # t0: "cpu i64[]"
    # prims.copy_(t0, a)
```

Returning `a` at the end of `f` does not fix this.

```py
# Constructed by Transform for execution (took 3 milliseconds)
import torch
from thunder.executors.torchex import no_autocast

@torch.no_grad()
@no_autocast
def computation(a):
  # a: "cpu i64[]"

  # /opt/pytorch/lightning-thunder/mshinokawa/sandbox/debug.py:5: 	    a += 1
  return a
  TorchCompile0(a)
    # t0 = ltorch.add(a, 1, alpha=None)  # t0: "cpu i64[]"
      # t0 = prims.add(a, 1)  # t0: "cpu i64[]"
    # prims.copy_(t0, a)
```

## Cause

[...]

PR #931, opened 2024-08-06
Handle torchex.copy_ eqally to prims.copy_

### LINKED ISSUE #930: fuse_bound_symbols skip the side effect of Tensor.copy_
## 🐛 Bug

#### Code sample

```py
def f(a):
    a += 1

jitted = thunder.jit(
)
jitted(torch.tensor(0))
```
```
  ...
AssertionError
```

## Possible fixes

shino16:
Fixes #930. This is a follow-up of #920, applying the second fix mentioned in #930.

### DIFF
--- a/thunder/executors/data_dependent_partition.py
+++ b/thunder/executors/data_dependent_partition.py
@@ -103,7 +103,8 @@ class Graph:
                 for copy_node in copy_nodes:
                     node.parents.add(copy_node)
                     copy_node.children.add(node)
-            elif bsym.sym.id is PrimIDs.COPY_:
+            # torch executor replaces prims.copy_ with torchex.copy_
+            elif bsym.sym.id in (PrimIDs.COPY_, "copy_"):
                 copy_nodes.append(node)
 
         for bsym_id, node in enumerate(bsym_id_to_node_map):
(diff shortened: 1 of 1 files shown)


thunder/executors/data_dependent_partition.py
```diff
@@ -103,7 +103,8 @@ def __init__(self, trace: TraceCtx):
                 for copy_node in copy_nodes:
                     node.parents.add(copy_node)
                     copy_node.children.add(node)
-            elif bsym.sym.id is PrimIDs.COPY_:
+            # torch executor replaces prims.copy_ with torchex.copy_
+            elif bsym.sym.id in (PrimIDs.COPY_, "copy_"):
```
crcrpar, line 107: could you check if `torchex.copy_.id` instead of "copy_" works?

shino16: Thanks, I changed it to comparison between syms.

IvanYashchuk:
Hey @shino16, I'm sorry I missed your original issue (https://github.com/Lightning-AI/lightning-thunder/issues/912). Let's take a step back and reevaluate possible solutions. I don't think data_dependent_partition.py should know anything about copies. One possible solution is to introduce fake dependencies to bound symbols (https://github.com/Lightning-AI/lightning-thunder/issues/912#issuecomment-2271733942).
Let's not merge this PR for now.

IvanYashchuk (approved):
While a long-term solution is being explored let's get this fix in as it's a small change to what's already in the codebase.

t-vi (approved):
Stamping, thank you @shino16 @IvanYashchuk @crcrpar

Issue #478, opened 2024-05-29
fsdp(jit(...)) transform can use more memory compared to jit(fsdp(...))


kshitij12345:
As `fsdp(jit(...))` holds on to the original parameters as well as the sharded parameters, it can lead to higher memory usage. I think a work-around can be to initialize the original model on `meta` device. But if using `meta` is the only correct way then we should add a warning if user does otherwise.

```python
import os
import torch
import torch.distributed as tdist
import thunder
import thunder.distributed

if __name__ == "__main__":
    tdist.init_process_group(backend="nccl")
    LOCAL_RANK = int(os.environ["LOCAL_RANK"])
    device = torch.device("cuda", LOCAL_RANK)
    torch.set_default_device(device)

    class Model(torch.nn.Module):
        def __init__(self) -> None:
            super().__init__()
            self.p1 = torch.nn.Parameter(torch.ones(1024, 1024))

        def forward(self, x):
            return self.p1 + x

    with device:
        model = Model()
        input_t = torch.randn(1)

    # jit(fsdp(...))
    # Memory Allocated - 6291968
    # model = thunder.distributed.fsdp(model)
    # model = thunder.jit(model, executors=["torch"])

    # fsdp(jit(...))
    # Memory Allocated - 10486272
    model = thunder.jit(model, executors=["torch"])
    model = thunder.distributed.fsdp(model)

    _ = model(input_t)

    if LOCAL_RANK == 0:
        print(torch.cuda.memory_allocated())

```

cc: @t-vi 

cc @carmocca @awaelchli @crcrpar

mruberry:
triage review:

- let's mutate the module we're given (so we don't preserve its tensors in memory and use more memory)
- practitioners can preserve the current behavior by copying the module before giving it to jit
- we should be careful that retracing has the information needed (possibly observing original values) to work as expected

crcrpar:
@mruberry 

> * we should be careful that retracing has the information needed (possibly observing original values) to work as expected

could you elaborate on what it means?

t-vi:
Two parts:
- The tracing (= new run through the Python interpreter, on cache miss) needs to happen on the original module (currently does, we won't be changing that).
- When we want to move tensors we override to "meta", we need to be sure that this does not screw up the tracing (likely by keeping the original device somewhere and using that when constructing the TensorProxy).

[...]

PR #932, opened 2024-08-06
[fsdp as a transform] Free memories occupied by materialized params that are initialized with `torch.device("meta")`

crcrpar:
## What does this PR do?

This is an effort to fix a part of #478 by passing materialized, original parameters to the fsdp trace transform in order for the transform to be able to release the memory consumed by the original parameters after (1) acquiring the original trace, (2) modifying prologue trace and computation trace accordingly to the shape information of sharded parameters, and (3) inserting the required `dist_prims.synchronize`.

### Caveats
This won't fix the increased memory usage if input models are initialized on a cuda device.

cc @IvanYashchuk @kshitij12345

### DIFF
--- a/thunder/distributed/__init__.py
+++ b/thunder/distributed/__init__.py
@@ -499,7 +499,9 @@ def fsdp_transform_module(
         # This is done before sharding in case the materialization logic depends on the tensor shape.
         # The tradeoff is that all of a module's direct parameters need to fit in device.
         # Each module only initializes its own parameters and not those of its children (recurse=False)
-        if any(t.is_meta for t in chain(module_copy.parameters(recurse=False), module_copy.buffers(recurse=False))):
+        if is_meta_initialized := any(
+            t.is_meta for t in chain(module_copy.parameters(recurse=False), module_copy.buffers(recurse=False))
+        ):
             # TODO: we could also support calling a "param_init_fn" argument like PyTorch
             _materialize(module_copy, device)
             for n, p in module_copy.named_parameters(recurse=False, prefix=module_name):
@@ -551,7 +553,12 @@ def fsdp_transform_module(
                 thunder_model._overrides_parameters[pn], global_rank, world_size, pn, allow_padding_for_fsdp=True
             )
             new_shape = thunder_model._overrides_parameters[pn].shape
-            sharded_params[pn] = (old_shape, new_shape, thunder_model._overrides_parameters[pn].device)
+            sharded_params[pn] = (
+                old_shape,
+                new_shape,
+                thunder_model._overrides_parameters[pn].device,
+                p if is_meta_initialized else None,
+            )
 
             # Track the original param and it's corresponding copied shard and metadata.
             shared_params[p] = {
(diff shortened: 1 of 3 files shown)


thunder/distributed/__init__.py
```diff
@@ -499,7 +499,9 @@ def fsdp_transform_module(
         # This is done before sharding in case the materialization logic depends on the tensor shape.
         # The tradeoff is that all of a module's direct parameters need to fit in device.
         # Each module only initializes its own parameters and not those of its children (recurse=False)
-        if any(t.is_meta for t in chain(module_copy.parameters(recurse=False), module_copy.buffers(recurse=False))):
+        if is_meta_initialized := any(
```
IvanYashchuk, line 502: I was confused by this name because "meta initialization" should not use any device memory.
```suggestion
        if is_meta_materialized := any(
```

thunder/distributed/transforms/fsdp_v2.py
```diff
@@ -116,6 +118,14 @@ def transform_traces_pre_prologue(self, prologue_trace, computation_trace, epilo
                             # TODO have a more principled way to update this?
                             a0, _, _, *a2pp = c.args
                             c.args = (a0, tuple(new_shape), thunder_device_str, *a2pp)
+                    # `orig_p` is not None means the input module is initialized with `torch.device("meta")`,
+                    # meaning we should better remove the materialized to free up the unexpected memory.
+                    if orig_p is not None:
+                        orig_p.untyped_storage().resize_(0)
```
IvanYashchuk, line 124: "the input module is initialized with meta and _then materialized_". I couldn't understand why do we need to free the memory of meta-initialized tensors because the "materialization" step is missing in the comment.

IvanYashchuk (commented):
@t-vi, is there a plan on fixing memory problems with "fsdp_v2"? This PR is good as a workaround, but it still feels like a hack. Why is a copy of parameters kept in memory instead of replacing them inplace?

t-vi:
Could we not keep it in meta here and use the materialization transform?

crcrpar:
> Could we not keep it in meta here and use the materialization transform?

it doesn't seem to work well in https://github.com/Lightning-AI/lightning-thunder/pull/921, if the diff there is correct.

t-vi:
> @t-vi, is there a plan on fixing memory problems with "fsdp_v2"? This PR is good as a workaround, but it still feels like a hack. Why is a copy of parameters kept in memory instead of replacing them inplace?

We need the original tensors to appear in the trace before transforms.
The plan is
- to initialize on the meta device (with `.__thunder_device = torch.device("cuda")` to allow CUDA inputs to be traced) - (I'll write a `PrepareMetaDevice` transform that does nothing but set the `.__thunder_device` on the original module's Parameters/Bufers),
- implement `.transform_module` in Transforms as leaving things on the meta device and
- implement `.transform_state_dict_for_submodule` in Transforms to be able to propagate parameters from initialization of the original module or a state dict of the original module, 
- have `MaterializationTransform` do the actual materialization + initialization through using the `.transform_state_dict_for_submodule`.

PR #934, opened 2024-08-06
autoregister method

### LINKED ISSUE #933: automatic registration - method variants
Repro
```python
import torch
import thunder

# Method variants fail with `AttributeError: Unknown attribute arccos`
def foo(x):
    # return torch.sinc(x)  # x.sinc()
    return torch.arccos(x)  # x.arccos()

x = torch.randn(3, 3, device='cuda')
jfoo = thunder.jit(foo)
y = jfoo(x)

```

Fails with
```python
  File "/home/kkalambarkar/lightning-thunder/thunder/core/proxies.py", line 1475, in __getattr__
    baseutils.check(method_or_value is not None, lambda: f"Unknown attribute {attr}", exception_type=AttributeError)
  File "/home/kkalambarkar/lightning-thunder/thunder/core/baseutils.py", line 103, in check
    raise exception_type(s())
AttributeError: Unknown attribute arccos
```

Potential Fix -
I think the `register_default_torch_op` is missing the registeration of `method` and `property` on Proxies.

This should probably handled similar to the code below - where `register_{method, property}` is called.
https://github.com/Lightning-AI/lightning-thunder/blob/36f46dde9b280784d8ff2a3a24e3a975fc67ffe1/thunder/torch/__init__.py#L186-L197

**EDIT** - the following patch seems to work
```patch
diff --git a/thunder/torch/__init__.py b/thunder/torch/__init__.py
index ab44018..343bc93 100644
--- a/thunder/torch/__init__.py
+++ b/thunder/torch/__init__.py
@@ -5392,6 +5392,9 @@ def register_default_torch_op(torchfn: Callable, m):
 
     from thunder.core.transforms import augmented_forward_impls, backward_impls
 
+    if m is torch.Tensor:
+        register_method(torchfn.__name__, torchfn)
+
     augmented_forward_impls[sym.id] = augmented_forward_adaptor(op)
 
     _vjp_impl_wrapper = partial(_vjp_impl, torchfn)
```

cc: @kiya00 

cc @apaz-cli

kshitij12345:
Fixes #933
 
Changes - 
1. We use `register_method` (for torch.Tensor.*) to register method on `TensorProxy`. 
2. We had conflicting registration for `torch.Tensor.size` and couple other methods. This was because they are directly registered as a method on TensorProxy (and not via torchsymbol. so they don't show up in `_torch_to_thunder_function_map`). Have updated to `check_overlap_ops` to also check for registered methods.
https://github.com/Lightning-AI/lightning-thunder/blob/36f46dde9b280784d8ff2a3a24e3a975fc67ffe1/thunder/torch/__init__.py#L245-L252


kiya00 (approved):
Thank you @kshitij12345 for the fix

t-vi (approved):
Thank you @kshitij12345

t-vi:
also thank you @kiya00

PR #937, opened 2024-08-07
fix - TransformerEngine with FSDP Bucketing

kshitij12345:
Running
```python
torchrun --local-ranks-filter 0 --standalone --max-restarts=0 --nproc-per-node=2  thunder/benchmarks/benchmark_litgpt.py --model_name Llama-3-8B --max_iters 10 --warmup_iters 2 --distributed_mode fsdp --shard_mode zero3 --bucketing_mode block --low_precision_mode fp8-delayed-te --n_layer 1 --compile thunder &> output_thunder_fp8.txt
```

Leads to
```
[rank0]:   File "/opt/pytorch/lightning-thunder/thunder/executors/torch_autograd.py", line 227, in split_forward_backward
[rank0]:     fw_extrace = sort_communication_ops(fw_extrace)
[rank0]:   File "/opt/pytorch/lightning-thunder/thunder/distributed/utils.py", line 112, in sort_communication_ops
[rank0]:     bsym_list_to_dag(execution_trace.bound_symbols)[1],
[rank0]:   File "/opt/pytorch/lightning-thunder/thunder/core/transforms.py", line 148, in bsym_list_to_dag
[rank0]:     producer = producers[inp]
[rank0]:   File "/opt/pytorch/lightning-thunder/thunder/core/utils.py", line 951, in __getitem__
[rank0]:     return self._dict[key_]
[rank0]: KeyError: 'ctx_te_1'
```

Problem - 
This issue happens because in `torch_autograd.py` certain passes (eg. bucketing) may update the returned value (including tensors saved for backward and backward metadata) in forward trace. We then update inputs passed to the backward trace. This is only done for saved tensors. However, for TE executor, we also have proxy to store the metadata which needs to be updated in backward trace. 

Solution - 
We update the part which updates backward trace inputs based on the updates in forward trace to also update the saved metadata as it may contain proxies.

Test - 
I didn't know of another operation which produces backward metadata as proxy so have written a test with TE (unfortunately this is not directly tested in CI)


t-vi:
@kshitij12345 could you take a look at the conflicts, please?
Who would be the best person with TE foo to take a look?

kshitij12345:
I think @IvanYashchuk @crcrpar would be the ones to review the PR.

thunder/executors/torch_autograd.py
IvanYashchuk, line 1: To other reviewers: I know that there is a lot of indexing going on here but all changes in this file make sense to me who created this file and are correct.

thunder/tests/distributed/test_fsdp.py
```diff
@@ -917,5 +976,32 @@ def test_fsdp_transformer_engine(executor, devices, dtype, thunder_fsdp_strategy
...
+        pytest.mark.skipif(not TE_AVAILABLE, reason="TransformerEngine is not installed."),
+        pytest.mark.skipif(not is_fp8_supported, reason=fp8_support_reason),
+        # See NOTE: Setting `NVTE_TORCH_COMPILE`
+        # NOTE: We don't pass `clear=True` to `unittest.mock.patch.dict` as that may clear paths
+        # from environment leading to picking up of incorrect dependencies in the spawned process.
+        unittest.mock.patch.dict(os.environ, {"NVTE_TORCH_COMPILE": "0"}),
+    ),
+)
+@distributed_wrapper("test_fsdp_transformer_engine_bucketing", _test_fsdp_transformer_engine_bucketing)
+def test_fsdp_transformer_engine_bucketing(executor, devices, dtype, thunder_fsdp_strategy_and_bucketing):
```
IvanYashchuk, line 1002: This test fails on the current main branch, right?

thunder/tests/distributed/test_fsdp.py
```diff
@@ -866,6 +866,65 @@ def is_same_across_ranks(t):
         return comparison_exceptions
 
 
+def _test_fsdp_transformer_engine_bucketing(input_data):
```
IvanYashchuk, line 869: I think it should be possible to construct a test with a mock test symbol like this test creates a `sincos` operation https://github.com/Lightning-AI/lightning-thunder/blob/86cd9c3b5824064f8faa024ae0427fb297c04add/thunder/tests/test_grad.py#L740
But I'd say we don't need to go this far and this test in combination with `test_fsdp_transformer_engine_bucketing` is good enough.

IvanYashchuk (approved):
Thank you very much for including "Problem - Solution - Test" in the pull request description!

thunder/executors/torch_autograd.py
```diff
@@ -177,11 +177,24 @@ def split_forward_backward(computation_trc: TraceCtx, compile_data, compile_stat
     # trace.
     original_bw_saved_tensors_for_backward = bw_trace.args[0][0]
     new_fw_saved_tensors_for_backward = get_saved_for_backward_tensors(fw_extrace)
-    swap_map = {
+
+    # saved meta data (this could also contain proxies)
+    original_bw_saved_meta_for_backward = bw_trace.args[0][1]
+    new_fw_saved_meta_for_backward = fw_extrace.output[1][1]
```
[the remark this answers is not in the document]
kshitij12345: Yes, I think it would be more clear to have 3 buckets so that passes are aware of places where they may have to make updates.

thunder/tests/distributed/test_fsdp.py
kshitij12345: Yes, that is correct. It fails with

```python
        # Raises the first assertion if any occurred
        root_results = results[0]
        if len(root_results) > 0:
>           raise (root_results[0])
E           KeyError: 'ctx_te_1'
```

t-vi (approved):

Issue #896, opened 2024-07-31
TypeError: rms_forward_affine_mixed_dtypes(): incompatible function arguments


tfogal:
## 🚀 Model / language coverage

```
[rank0]:   File "/home/tfogal/dev/nemo/nemo/collections/nlp/modules/common/megatron/language_model.py", line 824, in forward
[rank0]:     encoder_output = self.encoder(
[rank0]:   File "/home/tfogal/dev/thunder/thunder/core/interpreter.py", line 6372, in _impl
[rank0]:     return fn.__func__(fn.__self__, *args, **kwargs)
[rank0]:   File "/usr/local/lib/python3.10/dist-packages/torch/nn/modules/module.py", line 1716, in _wrapped_call_impl
[rank0]:     return self._call_impl(*args, **kwargs)
[rank0]:   File "/home/tfogal/dev/thunder/thunder/core/interpreter.py", line 6372, in _impl
[rank0]:     return fn.__func__(fn.__self__, *args, **kwargs)
[rank0]:   File "/usr/local/lib/python3.10/dist-packages/torch/nn/modules/module.py", line 1727, in _call_impl
[rank0]:     return forward_call(*args, **kwargs)
[rank0]:   File "/home/tfogal/dev/thunder/thunder/core/interpreter.py", line 6372, in _impl
[rank0]:     return fn.__func__(fn.__self__, *args, **kwargs)
[rank0]:   File "/home/tfogal/dev/nemo/nemo/collections/nlp/modules/common/megatron/transformer.py", line 1632, in forward
[rank0]:     hidden_states = layer(
[rank0]: TypeError: rms_forward_affine_mixed_dtypes(): incompatible function arguments. The following argument types are supported:
[rank0]:     1. (arg0: torch.Tensor, arg1: Tuple[int, ...], arg2: torch.Tensor, arg3: float) -> list[torch.Tensor]

[rank0]: Invoked with: <TensorProxy(name="t4161", dtype=thunder.dtypes.float16, shape=(384, 2, 5120))>, torch.Size([5120]), <TensorProxy(name="t4162", dtype=thunder.dtypes.float16, shape=(5120,))>, 1e-05
```

[full log of the run](https://github.com/user-attachments/files/16444645/neva.txt)

### Pitch

This comes up in the pure-thunder path to supporting NeVA

### Alternatives / Potential work-arounds

We could just use the dynamo frontend, for now.

### Minimal Repro

Still working on this ...

[...]

PR #938, opened 2024-08-07
support apex fused_rms_norm

kshitij12345:
Fixes : #896 

This PR adds a lookaside for `FusedRMSNormAffineMixedDtypesFunction.forward`  and corresponding symbol for the operation which is used by the lookaside. This  lookaside is enabled by default if the required import from `apex` (fused_layer_norm_cuda) is available. `apex_ex` (currently used for apex's cross-entropy) is re-used to register and implement this new symbol. So user has to also pass the `apex_ex` to the `executors` argument.

Also, we add compile option to disable `apex_ex` from claiming `cross_entropy` and newly added `FusedRMSNormAffineMixedDtypesFunction.forward`, so that using `apex_ex` doesn't make it mandatory to always use all the available apex operations.

NOTE - Lookaside is registered in `apex_fused_rms_norm.py` instead of usual `jit_ext.py` as this is operation may not be available always.


crcrpar:
IMHO it'd be reasonable to rename & update https://github.com/Lightning-AI/lightning-thunder/blob/7433b861b2fe0747462ff35818180a1a57a6ad6f/thunder/executors/apex_entropyex.py for this, what's your take?

kshitij12345:
> IMHO it'd be reasonable to rename & update https://github.com/Lightning-AI/lightning-thunder/blob/7433b861b2fe0747462ff35818180a1a57a6ad6f/thunder/executors/apex_entropyex.py for this, what's your take?

The reason I kept it seperate was that both these executors rely on different imports - entropyex relies on `xentropy_cuda` while rms_norm relies on `fused_layer_norm_cuda`. It is not necessary that user will always have both installed, so it would be odd (if we unify these under 1 executor) that this executor will only partially support certain operators depending on available packages. Because of this, I think we should keep them separate.

https://github.com/NVIDIA/apex/blob/59b80ee8df79cec125794949327f29913c328746/apex/normalization/fused_layer_norm.py#L69

riccardofelluga:
I do support @crcrpar comment about updating the apex executor and keeping one single executor for apex. To check if an op is available or not we could use the checks like it's done in nvfuser executor. My main pain point on having two different imports for apex is that it can become pretty bad for practitioners to have to list all the different executors and give them as input to the jit call. We could also gate the ops behind compiler flag in the case the practitioners want to disable certain apex modules. For ref apex has table of different modules and i fear the day we have an executor for each row of this table 😱 https://github.com/NVIDIA/apex/tree/7b2e71b0d4013f8e2f9f1c8dd21980ff1d76f1b6?tab=readme-ov-file#custom-ccuda-extensions-and-install-options

crcrpar (commented):
APEX has FusedRMSNorm `torch.autograd.Function`s and MixedFusedRMSNorm ones. IIRC/IIUC the latter ones were implemented for RMSNorm layers whose input and weight have different dtypes as in https://github.com/NVIDIA/apex/blob/59b80ee8df79cec125794949327f29913c328746/csrc/layer_norm_cuda.cpp#L358-L376.

Would you happen to know whether or not they are actually of different dtypes? If not, as mixed ones are used, we'd need to have some lookaside but it'd be possible to use non-mixed rms norm ones.

kshitij12345:
> Would you happen to know whether or not they are actually of different dtypes? If not, as mixed ones are used, we'd need to have some lookaside but it'd be possible to use non-mixed rms norm ones.

Nemo uses the `MixedFusedRMSNorm` in `ParallelMLP` (Ref 1). I am not sure but depending on the configuration (with autocast, etc), it may happen that weights and inputs have different dtype.  So, I think we should just take care of MixedFusedRMSNorm for now. Wdyt?

NOTE - The weights are always initialized to `float` - Ref 2

Ref 1- https://github.com/tfogal/NeMo/blob/030360457ed1b51e6e5f15c2109c78cf5b76700d/nemo/collections/nlp/modules/common/megatron/mlp.py#L214-L218

Ref 2 - https://github.com/NVIDIA/apex/blob/59b80ee8df79cec125794949327f29913c328746/apex/normalization/fused_layer_norm.py#L401-L405

thunder/tests/test_apex_fused_norms.py
```diff
@@ -0,0 +1,35 @@
...
+fused_layer_norm_cuda = pytest.importorskip("fused_layer_norm_cuda")
+from apex.normalization.fused_layer_norm import FusedRMSNormAffineMixedDtypesFunction
+from thunder.executors.apex_fused_rms_norm import register_apex_fused_rms_norm
+from thunder.executors.apex_entropyex import apex_ex
+import thunder
+
+
+@pytest.mark.parametrize("requires_grad", [True, False])
+@pytest.mark.parametrize("memory_efficient", [True, False])
+def test_apex_fused_norm(requires_grad, memory_efficient):
```
crcrpar, line 14: IMHO it'd be better if the name is e.g. `test_apex_fused_rms_norm` for clarity. what do you think?

kshitij12345: Agreed

docs/source/reference/executors/apex_entropyex.rst
```diff
@@ -10,4 +10,4 @@ Executor of `NVIDIA/apex <https://github.com/nvidia/apex>`_.
 .. autosummary::
     :toctree: generated/
 
-    apex_available
```
kshitij12345, line 13: TODO - Update the name of this file.

crcrpar (approved):

thunder/executors/apex_fused_rms_norm_impl.py
```diff
@@ -0,0 +1,123 @@
...
+)
+apex_ex.register_implementation(fused_rms_norm_backward, fused_rms_norm_backward)
+
+
+# Register the lookaside.
+def maybe_register_apex_fused_rms_norm_lookaside() -> None:
+    if APEX_FUSED_NORMS_AVAILABLE:
+
+        @thunder.core.jit_ext.register_general_jit_lookaside(FusedRMSNormAffineMixedDtypesFunction.forward)
+        @thunder.core.jit_ext.interpreter_needs_wrap
```
t-vi, line 117: Could this be handled by `replaces=FusedRMSNormAffineMixedDtypesFunction.forward` on the register_operator instead?

t-vi (commented):
Looks great, thank you @kshitij12345 . I think the lookaside can be gotten easier (see comment).
I wonder whether we could abstract this to make the next thing easier (I imagine that the fused_rms_norm is quite typical of apex if not PyTorch extension operators in general). WDYT? (Though not necessarily in this PR.)  

Thank you @crcrpar for reviewing.

kshitij12345:
> I wonder whether we could abstract this to make the next thing easier (I imagine that the fused_rms_norm is quite typical of apex if not PyTorch extension operators in general). WDYT? (Though not necessarily in this PR.)

I agree, making it easier would be good.

thunder/executors/apex_fused_rms_norm_impl.py
t-vi: Seems good to me.

t-vi (approved):
lgtm

PR #942, opened 2024-08-08
Make `TraceCtx` objects pickleable with `dill`

apaz-cli:
For some reason, the following works:
```py
import thunder, pickle, torch

def fn(a, b):
      return a + b

tm = thunder.jit(fn)
a, b = torch.randn(2, 5)
tm(a, b)
trace = thunder.last_traces(tm)[0]

assert str(pickle.loads(pickle.dumps(trace))) == str(trace)
```

But the moment you set `device='cuda'` it does not, and fails with:
```py
  File "/teamspace/studios/this_studio/repro.py", line 11, in <module>
    assert str(pickle.loads(pickle.dumps(trace))) == str(trace)
TypeError: cannot pickle 'mappingproxy' object
```

It turns out that python dataclasses do not play nicely with pickle. I talked to the cpython maintainers about it. The problem is that dataclasses create a map of all their fields behind the scenes, and the map that it creates is immutable and cannot be recreated by default by pickle since it is a builtin type and has no `__setattr__`. They've been meaning to support it eventually, but never got around to it. 

For some reason we run into this problem in the computation trace, but not the prologue trace. Genuinely no idea as to why. The read-only dict is there on every dataclass, no matter what. But why does it become a problem, for only the computation trace and not the prologue trace, and only when `device='cuda'`? Genuinely no clue, and too exhausted to care.

There is a module called `dill` that implements serialization of `mappingproxy` objects. It's basically a part of the stdlib, but hasn't been pulled into cpython yet largely because nobody has written a PEP. I tried it, and it works. So I propose we pull in the extra dependency.

There's also the issue that certain symbols are named different things than in their modules, and so they can't be looked up. To resolve this, I've added the `_python_name` field so that those symbols can be serialized.

### Next on the list for follow-up PRs:
1. Create a frontend function that serializes the result of `thunder.jit()`, to whatever extend we can implement
2. Save and load traces beyond just the first
3. Save and load the `nvFusion` cache
4. Investigate saving and loading kernels generated by `torch.compile`


requirements/base.txt
```diff
@@ -6,3 +6,4 @@ igraph >=0.10.4
 optree >=0.11.0
 opt_einsum >= 3.3.0
 mpmath <1.4.0  # todo: teporarl pin for `NameError: name '_C' is not defined`
+dill >=0.3.8 # Support for 3.12
```
t-vi, line 9: You continue to insist on dill, I seem to not need it. Why is that?

t-vi (commented):
Awesome!

requirements/base.txt
apaz-cli: For some reason, it happens when `device='cuda'`. But not without it. If you re-run with cuda as the device do you run into the issue?

thunder/core/prims.py
```diff
@@ -486,6 +488,7 @@ def _check_tensor_shape_and_metadata_meta(
     "check_tensor_metadata",
     meta=_check_tensor_shape_and_metadata_meta,
     tags=(OpTags.DONT_DCE,),
+    python_name="check_tensor_shape_and_metadata",
```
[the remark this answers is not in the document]
apaz-cli: Yeah, agreed.

requirements/base.txt
t-vi: I am using cuda. Can you try without dill and see if it works?
Thunder has a thing where it refuses to create cuda devices when there is no cuda (which is bad, arguably, but that's the status quo), but I can save and recover cuda as well.

apaz-cli: @t-vi I used your exact script that you posted on slack in a studio on an A10G. The one difference that gets it to fail is that I set `a, b = torch.randn(2, 5, device='cuda')`. If I `import dill as pickle` instead of importing normal pickle, everything works.

t-vi:
Fair enough, so it seems that it is easy to trigger different states of processing (e.g. has `bound_symbol.rhs` been set, ...).
From what I see, two things that keep us from using pickle are:
- modules in `trace._import_ctx` (cannot pickle module, would need manual intervention), 
- `thunder.utils.FrozenDict` instance (this would be easy to fix by adding a `__reduce__`).
I guess the dill use is not harmful (except for the versioning, likely).

t-vi (approved):
Great thank you, LGTM.

thunder/core/symbol.py
```diff
@@ -138,6 +138,7 @@ class Symbol:
     executor: None | Any = None
     python_impl: None | Callable = None
     _print_as_impl: bool = False  # If not None, w
+    _python_name: str | None = None
```
IvanYashchuk, line 141: Why is "name" or "python_impl.__name__" not sufficient?

t-vi: See the above examples?
The obvious alternative would be to have name always be the name of the member of the module, given that they typically need a bespoke python_printer anyways.

riccardofelluga:
@apaz-cli would it be possible for the follow-up PRs to open a tracking issue? So that it's easier to track next time, i've almost missed this ;)

PR #949, opened 2024-08-09
Use type inferred from `fill_value` when dtype is None for torch.full

### LINKED ISSUE #948: Thunder's handling of torch.full is inconsistent with torch's behavior

## 🐛 Bug

```
import torch
import thunder

def func():
    return torch.full((1,), False, device = "cuda")
o =func()
print(o)

jfunc = thunder.jit(func)
o1=jfunc()
print(o1)
```
outputs:
```
tensor([False], device='cuda:0')
tensor([0.], device='cuda:0')
```

Torch uses the type inferred from `fill_value`, Thunder uses the default dtype

kiya00:
It seems to be sufficient for `torch.full` without using the default dtype.

Fixes #948 .


mruberry:
+ @kshitij12345, who I think added the default dtype call here

thunder/tests/opinfos.py
```diff
@@ -5775,6 +5775,8 @@ def full_sample_generator(op, device, dtype, requires_grad, **kwargs):
 
     for shape, fill_value in cases:
         yield SampleInput(shape, fill_value, device=device, dtype=dtype)
+    yield SampleInput((2,), False, device=device)
```
mruberry, line 5778: What about:

```
for shape, fill_value in cases:
        yield SampleInput(shape, fill_value, device=device, dtype=dtype)

	# Tests dtype is inferred correctly
	yield SampleInput(shape, fill_value, device=device)
```

?

mruberry (commented):
Hey @kiya00! I added @kshitij12345 to help us sanity check since I think he added these lines. I made a suggestion for how to tweak the test coverage, too.

thunder/torch/__init__.py
```diff
@@ -566,7 +566,7 @@ def full(
     shape: Sequence[int], fill_value: NumberLike, *, device: None | DeviceLike = None, dtype: None | dtypeLike = None
 ) -> TensorLike:
     device = to_device(maybe_get_default_device(device))
-    dtype = to_dtype(maybe_get_default_dtype(dtype))
+    dtype = to_dtype(dtype)
```
kshitij12345, line 569: Looking at `torch.full` implementation 
https://github.com/pytorch/pytorch/blob/cd307fb0b1a833f9297d2233653b514ed4aa3163/aten/src/ATen/native/TensorFactories.cpp#L584-L604

When dtype is not passed explicitly, dtype is determined by
```python
if fill_value == bool:
   dtype = bool
elif fill_value.is_integer:
   dtype = LongInt
elif fill_value.is_complex:
   dtype = default_complex_dtype # complex type corresponding to default dtype
else: # floating
   dtype = default_floating_dtype
```

So, I think we need to get dtype from default torch dtype if fill_value is floating or complex.

kshitij12345 (commented):
Thanks for catching this @kiya00!

I have a comment and I think we need to update other ops which decompose to `full` eg. `ones`, `zeros`, etc to pass default dtype explicitly to `full` (as they return tensor with default dtype if dtype is not passed)

https://github.com/Lightning-AI/lightning-thunder/blob/019557e1cbd2944a4d8d719954ee8a7c295c539b/thunder/torch/__init__.py#L583-L586

thunder/torch/__init__.py
mruberry: More precisely, we need to get the default floating point dtype or default complex dtype if dtype is not specified and the inputing is a floating point or complex number (respectively)

kiya00:
Hi @kshitij12345 @mruberry , I found the check in https://github.com/Lightning-AI/lightning-thunder/blob/019557e1cbd2944a4d8d719954ee8a7c295c539b/thunder/core/prims.py#L2650-L2655 is not the same as the one used in `torch.full`
```
>>> torch.full((2,2),0.5,dtype=torch.bool)
tensor([[True, True],
        [True, True]])
>>> torch.full((2,2),0.5,dtype=torch.int)
tensor([[0, 0],
        [0, 0]], dtype=torch.int32)
>>> torch.full((2,2),-1,dtype=torch.uint32)
tensor([[4294967295, 4294967295],
        [4294967295, 4294967295]], dtype=torch.uint32)
>>> torch.full((2,2),sys.maxsize,dtype=torch.uint32)
  File "<stdin>", line 1, in <module>
RuntimeError: value cannot be converted to type uint32_t without overflow
>>> x=torch.full((1,2), 1j, dtype=torch.float)
  File "<stdin>", line 1, in <module>
RuntimeError: value cannot be converted to type float without overflow
```
And it seems the related check in torch is the series of `overflows` template starting from https://github.com/pytorch/pytorch/blob/f5e704a6f25939478f770f8980c344ab461f0113/c10/util/Half.h#L462
Do we want the same check in thunder as the overflow checks in torch?

kshitij12345:
IIUC, we error with different message than `torch` when `fill_value` can't be cast to the `dtype`, right? In that case, I like the message from `thunder` which mentions both `fill_value_dtype` and expected dtype.

thunder/torch/__init__.py
```diff
@@ -561,12 +561,31 @@ def arange(
...
+    fill_value_dtype = dtypes.numbertype_to_dtype(dtypes.to_dtype(fill_value))
+    if not dtype:
+        current_default_dtype = get_default_dtype()
+        if dtypes.is_boolean_dtype(fill_value_dtype):
+            return fill_value_dtype
+        elif dtypes.is_nonboolean_integer_dtype(fill_value_dtype):
+            return dtypes.int64
+        elif dtypes.is_complex_dtype(fill_value_dtype):
+            if current_default_dtype == torch.float64:
+                return dtypes.complex64
```
kshitij12345, line 575: float64 corresponds to complex128 (i.e. two doubles)
float32 -> complex64
float16 -> complex32

thunder/torch/__init__.py
```diff
@@ -583,7 +602,7 @@ def full_like(
 @torchsymbol(torch.ones)
 def ones(*shape: int, device: None | DeviceLike = None, dtype: None | dtypeLike = None) -> TensorLike:
     shape = utils.extract_shape_from_varargs(shape)
-    return full(shape, 1, device=device, dtype=dtype)
+    return full(shape, 1.0, device=device, dtype=dtype)
```
kshitij12345, line 605: I think we have to explicitly specify the `dtype` here as `maybe_get_default_dtype(dtype))`, so that `full` will not try to infer the value from `fill_value`. (similar for `zeros`)

```suggestion
    return full(shape, 1, device=device, dtype=maybe_get_default_dtype(dtype))
```

kiya00:
>we error with different message than torch when fill_value can't be cast to the dtype, right?
Thunder upcast check is different from torch overflow check, e.g. : `torch.full((1,), 0.5, device = "cuda", dtype=torch.bool)`
torch outputs: `tensor([True], device='cuda:0')` but thunder throws error `RuntimeError: Can't safely cast fill_value of numbertype <class 'float'> to dtype thunder.dtypes.bool8`.
I'm wondering if this different behavior is intentional or maybe we want the same behavior as Torch?

thunder/torch/__init__.py
kiya00: Thanks for pointing that out, a slight difference is that in Torch float16->complex64
```
>>> torch.set_default_dtype(torch.float16)
>>> torch.full((2,),1.2+3j,device="cuda").dtype
torch.complex64
>>> torch.set_default_dtype(torch.float32)
>>> torch.full((2,),1.2+3j,device="cuda").dtype
torch.complex64
>>> torch.set_default_dtype(torch.float64)
>>> torch.full((2,),1.2+3j,device="cuda").dtype
torch.complex128
```

thunder/core/transforms.py
```diff
@@ -1854,14 +1854,16 @@ def _vmap_call_metafunc(detached: bool, args, in_dims, out_dims, axis_size, func
...
+            result = ltorch.full(shape=(), fill_value=result, device=common_device)
             result = BatchedValue(result, not_mapped)
         elif (
             isinstance(result, BatchedValue)
             and isinstance(result.value, (Number, NumberProxy))
             and axis_size is not None
         ):
-            result = BatchedValue(full(shape=(), fill_value=result.value, device=common_device), result.batch_dim)
+            result = BatchedValue(
+                ltorch.full(shape=(), fill_value=result.value, device=common_device), result.batch_dim
```
kiya00, line 1865: Use ltorch.full instead of clang.full because the dtype infer is now done in ltorch function.

thunder/tests/test_core.py
```diff
@@ -1654,11 +1654,11 @@ def func(a, b, *, c=5):
 def test_transforms_vmap_axis_size(executor, device, _):
     from thunder.core.transforms import vmap
 
-    actual = executor.make_callable_legacy(vmap(lambda: 2, axis_size=4))()
+    actual = executor.make_callable(vmap(lambda: 2, axis_size=4), disable_torch_autograd=True)()
     expected = torch.full((4,), 2, device="cpu")
     assert_close(actual, expected)
 
-    actual = executor.make_callable_legacy(vmap(lambda x: x, axis_size=4))(2)
+    actual = executor.make_callable(vmap(lambda x: x, axis_size=4), disable_torch_autograd=True)(2)
```
kiya00, line 1661: The legacy `thunder.compile` doesn't have `cache_info`, so the `thunder.torch.get_default_dtype` fails, so we use jit instead

mruberry:
> Hi @kshitij12345 @mruberry , I found the check in
> 
> https://github.com/Lightning-AI/lightning-thunder/blob/019557e1cbd2944a4d8d719954ee8a7c295c539b/thunder/core/prims.py#L2650-L2655
> 
> is not the same as the one used in `torch.full`
> ```
> >>> torch.full((2,2),0.5,dtype=torch.bool)
> tensor([[True, True],
>         [True, True]])
> >>> torch.full((2,2),0.5,dtype=torch.int)
> tensor([[0, 0],
>         [0, 0]], dtype=torch.int32)
> >>> torch.full((2,2),-1,dtype=torch.uint32)
> tensor([[4294967295, 4294967295],
>         [4294967295, 4294967295]], dtype=torch.uint32)
> >>> torch.full((2,2),sys.maxsize,dtype=torch.uint32)
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
> RuntimeError: value cannot be converted to type uint32_t without overflow
> >>> x=torch.full((1,2), 1j, dtype=torch.float)
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
> RuntimeError: value cannot be converted to type float without overflow
> ```
> 
> And it seems the related check in torch is the series of `overflows` template starting from [pytorch/pytorch@`f5e704a`/c10/util/Half.h#L462](https://github.com/pytorch/pytorch/blob/f5e704a6f25939478f770f8980c344ab461f0113/c10/util/Half.h#L462) Do we want the same check in thunder as the overflow checks in torch?



> Hi @kshitij12345 @mruberry , I found the check in
> 
> https://github.com/Lightning-AI/lightning-thunder/blob/019557e1cbd2944a4d8d719954ee8a7c295c539b/thunder/core/prims.py#L2650-L2655
> 
> is not the same as the one used in `torch.full`
> ```
> >>> torch.full((2,2),0.5,dtype=torch.bool)
> tensor([[True, True],
>         [True, True]])
> >>> torch.full((2,2),0.5,dtype=torch.int)
> tensor([[0, 0],
>         [0, 0]], dtype=torch.int32)
> >>> torch.full((2,2),-1,dtype=torch.uint32)
> tensor([[4294967295, 4294967295],
>         [4294967295, 4294967295]], dtype=torch.uint32)
> >>> torch.full((2,2),sys.maxsize,dtype=torch.uint32)
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
> RuntimeError: value cannot be converted to type uint32_t without overflow
> >>> x=torch.full((1,2), 1j, dtype=torch.float)
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
> RuntimeError: value cannot be converted to type float without overflow
> ```
> 
> And it seems the related check in torch is the series of `overflows` template starting from [pytorch/pytorch@`f5e704a`/c10/util/Half.h#L462](https://github.com/pytorch/pytorch/blob/f5e704a6f25939478f770f8980c344ab461f0113/c10/util/Half.h#L462) Do we want the same check in thunder as the overflow checks in torch?

Would you please file this as a separate issue? I think this PR can proceed without it?

thunder/torch/__init__.py
mruberry: Great detailed look, @kiya00 -- let's follow the torch mapping for now (and add a comment).

What about the bfloat16 dtype?

mruberry:
> > we error with different message than torch when fill_value can't be cast to the dtype, right?
> 
> Thunder upcast check is different from torch overflow check, e.g. : `torch.full((1,), 0.5, device = "cuda", dtype=torch.bool)` torch outputs: `tensor([True], device='cuda:0')` but thunder throws error `RuntimeError: Can't safely cast fill_value of numbertype <class 'float'> to dtype thunder.dtypes.bool8`. I'm wondering if this different behavior is intentional or maybe we want the same behavior as Torch?

Let's file a separate issue for this behavior (I made the same comment elsewhere, too).

This is a great detail, but I think we can safely separate it from this PR

thunder/clang/__init__.py
```diff
@@ -272,9 +272,6 @@ def convolution(
 def full(
     shape: Sequence[int], fill_value: NumberLike, *, device: DeviceLike, dtype: None | dtypes.dtype = None
 ) -> TensorLike:
-    # Infers dtype from the fill_value when not explicitly provided
```
mruberry, line 275: Doesn't the prim require the dtype be specified? So what happens if `dtype` is None here?

@kshitij12345 (and at @IvanYashchuk who may have some thoughts), is the idea of the default dtype torch only, or is it a core language concept? 

If the default dtype is a core language concept, then I would update the core language operations to utilize it. 

If the default dtype is a torch language only concept, then I would make the clang operations require the dtype explicitly (like the primitives) or have them perform their own kind of type inference.

It seems like it'd be easier to have the default dtype be a system-wide concept, that would include the core language operations?

thunder/torch/__init__.py
kshitij12345: > a slight difference is that in Torch float16->complex64

Nice catch, this looks like a bug in PyTorch. Internally PyTorch has the following map from Complex to Real.
https://github.com/pytorch/pytorch/blob/947a446be4360994da3b11ef01a8bdfbfe812f49/c10/core/ScalarType.h#L492-L504

Also, I think we should use this helper function as it provides the same map -
https://github.com/Lightning-AI/lightning-thunder/blob/ffbebe07bdf003c3a60f2cab88298e96b80bdbba/thunder/core/dtypes.py#L394-L395

thunder/clang/__init__.py
kshitij12345: The idea of default dtype currently is torch only. The reason for that was, I believe that core lang can be used by other languages eg. numpy and it seems incorrect that details from PyTorch leak into it.

Though I must admit that I am still unclear on what semantics we have/want for the core language (and if it gets a default dtype, should it be same as pytorch?) Though as of now, since PyTorch is the main (and probably only supported higher language), it should probably be ok.

IvanYashchuk: `prims` should be strict and require all arguments to be provided with no default value. This pull request is about changing the `torch.full` behavior so I suggest reverting changes made to `clang.full` in this pull request and keep it as is until we start investing more in alternative languages for Thunder.

thunder/core/transforms.py
```diff
@@ -1854,14 +1854,16 @@ def _vmap_call_metafunc(detached: bool, args, in_dims, out_dims, axis_size, func
             return tree_unflatten(outs, spec)
         if isinstance(result, (Number, NumberProxy)) and axis_size is not None:
             # TODO: fetch the default device from the context
-            result = full(shape=(), fill_value=result, device=common_device)
+            result = ltorch.full(shape=(), fill_value=result, device=common_device)
```
kiya00, line 1857: Hi @IvanYashchuk , I'd like to confirm here: I change it to `ltorch.full` infers dtype considering the Torch's default dtype, `clang.full` infers dtype directly from `fill_value` when `dtype=None`, so here in the `vmap` we should use ltorch right?

thunder/torch/__init__.py
kiya00: I updated to use Thunder helper function `corresponding_complex_dtype` and added a comment about we have this single difference than Torch: float16->complex32 (Thunder) , complex64 (Torch). I hope that's OK @mruberry

thunder/torch/__init__.py
```diff
@@ -561,12 +561,31 @@ def arange(
     return clang.arange(start=start, step=step, stop=end, device=device, dtype=dtype)
 
 
+# Infers dtype from the fill_value and dtype
+def _infer_full_dtype(fill_value: NumberLike, dtype):
```
mruberry, line 565: Nit: type annotations would be nice

thunder/torch/__init__.py
```diff
@@ -561,12 +561,31 @@ def arange(
...
+    fill_value_dtype = dtypes.numbertype_to_dtype(dtypes.to_dtype(fill_value))
```
mruberry, line 566:
```
def _infer_full_dtype(fill_value: NumberLike, dtype):

  # Short-circuits if dtype is explicitly specified
  if dtype is not None:
    return to_dtype(dtype)

  # NOTE dtype is None
  fill_value_dtype = dtypes.numbertype_to_dtype(dtypes.to_dtype(fill_value))

  if dtypes.is_boolean_dtype(fill_value_dtype):
    # Is this correct? Seems like we can just the bool dtype here?
    return dtypes.bool8

  if dtypes.is_nonboolean_integer_dtype(fill_value_dtype):
    return dtypes.int64
  
  current_default_dtype = get_default_dtype()

  if dtypes.is_complex_dtype(fill_value_dtype):
    return dtypes.corresponding_complex_dtype(current_default_dtype)
  
  # NOTE fill_value_dtype is a non-complex floating-point type
  return to_dtype(current_default_dtype)
```

thunder/torch/__init__.py
```diff
@@ -561,12 +561,31 @@ def arange(
...
+    if not dtype:
+        current_default_dtype = get_default_dtype()
+        if dtypes.is_boolean_dtype(fill_value_dtype):
+            return fill_value_dtype
+        elif dtypes.is_nonboolean_integer_dtype(fill_value_dtype):
+            return dtypes.int64
+        # NOTE Thunder has slightly different behavior than Torch, Torch (2.5.0a0+git8927fc2):
```
mruberry, line 573: Would you elaborate on this comment about what is PyTorch's behavior, and what is meant by Thunder's behavior?

Is the behavior different in this function, or is the difference elsewhere?

thunder/torch/__init__.py
kiya00: > # Is this correct? Seems like we can just return the bool dtype here?

Yes and thanks for simplifying.

thunder/torch/__init__.py
kiya00: It's updated in the new commit: https://github.com/Lightning-AI/lightning-thunder/pull/949/commits/bc3358994e833eca15a571cdd44434211679fec7

mruberry (approved):
I think this looks pretty good! There may be more work to do on type inference in the future, and this is a good step forward.

@kshitij12345 Would you like to review again, too?
@IvanYashchuk Would you like to sanity check the transform updates?

@Borda and @t-vi It looks like the precommit's "large files" check may be incorrect

kshitij12345 (approved):
LGTM, thank you @kiya00

t-vi:
In `.pre-commit-config.yaml` : `args: ["--maxkb=300", "--enforce-all"]`

Can you bump this to 400, please?

PR #954, opened 2024-08-11
improve name assignment

t-vi:
One of the fundamentals of variables is that they are stored before they're loaded.
This can happen implicitly in the setup of frames or explicitly through `STORE_FAST` (for locals).

This PR does two things:
- it looks at assigning names when variables are set. This implicitly prefers outer scopes for variables that are passed to calls, which is good.
- as we also have occasions where the variable is not passed to calls but obtained otherwise, we avoid internal functions (e.g. when a variable is obtained though an attribute or item lookup).

Quite likely, it would be even better to make a stronger preference for outer scopes, maybe by overwriting name assignments if the stack depth of the current assignment is smaller than the previous one. I'll leave that for the next PR though.

### DIFF
--- a/thunder/core/jit_ext.py
+++ b/thunder/core/jit_ext.py
@@ -71,6 +71,7 @@ from thunder.core.proxies import (
 from thunder.core.trace import set_tracectx, reset_tracectx, tracectx, from_trace
 from thunder.core.interpreter import (
     InterpreterLogItem,
+    InterpreterFrame,
     interpret,
     _interpret_call,
     CapsuleType,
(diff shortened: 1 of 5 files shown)


IvanYashchuk:
Does this PR have any effect on the argument names for the computation trace from https://github.com/Lightning-AI/lightning-thunder/issues/657#issuecomment-2208317779? The expected variable names were suggested here: https://github.com/Lightning-AI/lightning-thunder/issues/657#issuecomment-2274264003.

t-vi:
Yes, it becomes
```
def computation(param, t_0_1, step_t, t_1_1):
```

Again, this isn't "the full solution", but to my mind, it's a strict improvement.

thunder/core/jit_ext.py
```diff
@@ -1127,16 +1128,27 @@ def _general_jit_const_callback(value: Any) -> WrappedValue:
 
 
 # TODO(nikitaved): maybe call it upon Frame creation
-def _maybe_update_proxy_name(orig_value: Any, name: str):
+def _maybe_update_proxy_name(orig_value: Any, name: str, is_internal: bool | None = None):
```
nikitaved, line 1131: what does `is_internal` mean?

nikitaved (approved):
LGTM! Thank you!

thunder/core/jit_ext.py
t-vi: ...to thunder.

Issue #872, opened 2024-07-27
Matmul lowering expands tensor input leading to 100x+ slowdown


parthmannan:
## 🐛 Bug

This was discovered while trying to debug why NeMo NeVA using Dynamo+Thunder was orders of magnitude slower than Eager mode.
The issue is that the Matmul lowering/decomposition broadcasts the weight tensor to match the batch size of the input, i.e., if the input tensor is `[b, s, h]` and weight tensor is `[h, h2]`, the weight tensor is expanded to `[b, h, h2]` leading to a much slower BMM. The issue exists outside of Dynamo integration as well.

### To Reproduce

This is a mini-repro script for one of the GEMMs inside NeVA model.
```
import torch
import torch.nn as nn
import thunder

class mm_Model(nn.Module):
    def __init__(self):
        super().__init__()
        self.weight = nn.Parameter(torch.ones(5120, 15360), requires_grad=True)

    def forward(self, x):
        return torch.matmul(x, self.weight)

mm_net = mm_Model().to(torch.bfloat16).cuda()
mm_net = thunder.jit(mm_net)

for x in range(10):
    input_tensor = torch.randn([384, 2, 5120], dtype=torch.bfloat16, device='cuda', requires_grad=True)
    target_tensor = torch.randn([384, 2, 15360], dtype=torch.bfloat16, device='cuda')

    if x == 5:
        torch.cuda.cudart().cudaProfilerStart()

    output = mm_net(input_tensor)
    output.backward(target_tensor)

    # print(thunder.last_traces(mm_net)[-1])

    if x == 5:
        torch.cuda.cudart().cudaProfilerStop()
```

Running the script directly with Thunder will lead to (which might be a separate issue on its own)
```
torch.OutOfMemoryError: CUDA out of memory. Tried to allocate 56.25 GiB. GPU 0 has a total capacity of 79.11 GiB of which 21.79 GiB is free. Including non-PyTorch memory, this process has 57.31 GiB memory in use. Of the allocated memory 439.00 MiB is allocated by PyTorch, and 56.24 GiB is reserved by PyTorch but unallocated. If reserved but unallocated memory is large try setting PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True to avoid fragmentation.  See documentation for Memory Management  (https://pytorch.org/docs/stable/notes/cuda.html#environment-variables)
```

[...]

PR #955, opened 2024-08-12
matmul: optimize cases (nd @ 2d) to be memory and compute efficient

nikitaved:
As per title. Resolves https://github.com/Lightning-AI/lightning-thunder/issues/872.

See https://github.com/Lightning-AI/lightning-thunder/issues/872#issuecomment-2283599147 for more details.

### DIFF
--- a/thunder/torch/__init__.py
+++ b/thunder/torch/__init__.py
@@ -3364,6 +3364,20 @@ def matmul(a: TensorLike, b: TensorLike, /) -> TensorLike:
     if a.ndim == 1 or b.ndim == 1:
         return prims.matmul(a, b)
 
+    # Case nd @ 2d --> reduce to a 2d gemm
+    if a.ndim > 2 and b.ndim == 2:
+        a_batch_dims = a.shape[:-2]
+
+        # a -> a_2d by flattening batch dims with the row space
+        a_2d = a.reshape(-1, a.shape[-1])
+
+        # 2d gemm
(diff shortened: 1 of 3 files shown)


thunder/torch/__init__.py
```diff
@@ -3364,6 +3364,20 @@ def matmul(a: TensorLike, b: TensorLike, /) -> TensorLike:
     if a.ndim == 1 or b.ndim == 1:
         return prims.matmul(a, b)
 
+    # Case nd @ 2d --> reduce to a 2d gemm
+    if a.ndim > 2 and b.ndim == 2:
+        a_batch_dims = a.shape[:-2]
+
+        # a -> a_2d by flattening batch dims with the row space
+        a_2d = a.reshape(-1, a.shape[-1])
```
t-vi, line 3372: This is OK because we know that PyTorch would do it, too, right?

t-vi (approved):
Thank you @nikitaved

thunder/torch/__init__.py
nikitaved: PyTorch will, our prim and lowering otherwise will not.

t-vi: Yeah, just that we don't introduce more copies when things are non-contiguous. Let's have it.

mruberry:
fyi @parthmannan, @tfogal

Issue #957, opened 2024-08-12
Support for reshape+in-place


IvanYashchuk:
## 🚀 Feature

`torch.reshape` (and all operations using `reshape` internally) is interesting because depending on PyTorch's internal conditions it may or may not return a tensor view. [PyTorch documentation](https://pytorch.org/docs/stable/tensor_view.html) says:
> [reshape()](https://pytorch.org/docs/stable/generated/torch.Tensor.reshape.html#torch.Tensor.reshape), [reshape_as()](https://pytorch.org/docs/stable/generated/torch.Tensor.reshape_as.html#torch.Tensor.reshape_as) and [flatten()](https://pytorch.org/docs/stable/generated/torch.Tensor.flatten.html#torch.Tensor.flatten) can return either a view or new tensor, user code shouldn’t rely on whether it’s view or not.

@jjsjann123 discovered that with the current functionalization enabled the behavior doesn't match with PyTorch silently producing unexpected results and matches when it's disabled (https://github.com/Lightning-AI/lightning-thunder/issues/657#issuecomment-2274293433).

Short-term solution: Thunder should raise an error if reshape+in-place occurs in the user script irrespective of whether the reshape is safely resolvable to be a non-viewing reshape or not. There should be an option of turning this error into a warning for developers. What do people think about this?

For a long-term solution, I see two paths:

1. Thunder acts as a linter for PyTorch code and raises an error when reshape+in-place results in a side effect observable outside the computation function. It should be suggested to users to change the code to use `torch.view` or `torch.reshape_copy`.
2. Thunder mimicks PyTorch's behavior. Even though it's discouraged to rely on reshape+in-place side effects it may smoothen Thunder adoption if Thunder behaves like PyTorch.

I'd like people to focus on what the user experience with Thunder should be and not how more technically challenging (or not) one solution is over another. Maybe others can come up with more alternatives?

cc @tfogal, @lantiga, @t-vi, @crcrpar

jjsjann123:
> user code shouldn’t rely on whether it’s view or not.

I'm wondering how this is perceived by users today? i.e. extreme cases are still tricky to resolve, I'm wondering with fake tensor, how reliable can torch's analysis resolve reshape-as-view vs reshape-as-copy? cc'ing @kiya00

IvanYashchuk:
Fake tensors do propagate the stride information but it may do it differently from Eager and then anyway we don't have stride information for the inputs so we cannot initialize correct fake tensors.

[...]

PR #958, opened 2024-08-12
error out if in-place operand is output of reshape/flatten

crcrpar:
rel: #957

### DIFF
--- a/thunder/tests/test_inplace_functionalization.py
+++ b/thunder/tests/test_inplace_functionalization.py
@@ -672,3 +672,24 @@ def test_inplace_to_alias_func_args(executor, device, dtype):
     out, out_ref = jitted_f(a, b, a), f(a_ref, b_ref, a_ref)
     torch.testing.assert_close(out, out_ref)
     torch.testing.assert_close((a, b), (a_ref, b_ref))
+
+
+@instantiate(dtypes=NOTHING)
+def test_reshape_flatten_error_out(executor, device, _):
+
+    def f(x):
+        y = x.reshape(6, 4)
+        y.add_(1)
+        return y
+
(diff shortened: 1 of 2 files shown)


thunder/tests/test_inplace_functionalization.py
```diff
@@ -672,3 +672,24 @@ def f(a, b, c):
...
     torch.testing.assert_close((a, b), (a_ref, b_ref))
+
+
+@instantiate(dtypes=NOTHING)
+def test_reshape_flatten_error_out(executor, device, _):
+
+    def f(x):
+        y = x.reshape(6, 4)
+        y.add_(1)
+        return y
```
IvanYashchuk, line 683: It should work if there's a clone between reshape and in-place operation. Currently, it fails with the same error because `clone` is not recorded on the trace:
https://github.com/Lightning-AI/lightning-thunder/blob/ffbebe07bdf003c3a60f2cab88298e96b80bdbba/thunder/torch/__init__.py#L2521-L2525
In the code review, it was decided that it's fine for the clone not to record itself in the trace, but now it seems we need this clarity to break the chain of view ops (https://github.com/Lightning-AI/lightning-thunder/pull/429#issuecomment-2121041926).
The code snippet I tried:
```py
import thunder
import torch

def f(x):
  y = x.reshape(6, 4).clone()
  y.add_(1)
  return y

x = torch.randn((4, 6))
jf = thunder.jit(f)
jf(x)
```

IvanYashchuk (approved):

t-vi (approved):
Thank you @crcrpar @IvanYashchuk . To my mind, recording .clone can be a follow-up, but I'll leave that to you.

PR #964, opened 2024-08-13
Fix type_as

### LINKED ISSUE #962: Different behavior in torch.Tensor.type_as when inputing a cpu and a cuda tensor

## 🐛 Bug

Different behavior in torch.Tensor.type_as when inputing a cpu and a cuda tensor

### To Reproduce

```
import torch
import thunder
a=torch.randn((20,128))
b=torch.randn((1,32,20,128),dtype=torch.float16).cuda()

def func(a,b):
    return a.type_as(b)

jfun = thunder.jit(func)
o = jfun(a,b)
print("thunder: ", o.device)

print("torch: ", func(a,b).device)
```
Output:
```
thunder:  cpu
torch:  cuda:0
```

### Expected behavior

Thunder behaves the same as Torch

### Additional context

https://github.com/pytorch/pytorch/blob/947a446be4360994da3b11ef01a8bdfbfe812f49/aten/src/ATen/native/TypeProperties.cpp#L66-L69

kiya00:
## What does this PR do?

Fixes #962.

Thunder behaves differently in torch.Tensor.type_as when inputting a cpu and a cuda tensor

### DIFF
--- a/thunder/tests/opinfos.py
+++ b/thunder/tests/opinfos.py
@@ -2959,12 +2959,11 @@ def type_as_sample_generator(op, device, dtype, requires_grad, **kwargs):
         (1, 2, 3),
     )
 
-    other_dev = "cpu" if device == "cuda" else "cuda"
     for a_shape, b_shape in itertools.product(shapes, shapes):
         yield SampleInput(make(a_shape), make(b_shape))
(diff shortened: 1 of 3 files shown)


t-vi (approved):
Thank you @kiya00

PR #968, opened 2024-08-14
initial version of state_dict() / load_state_dict() with override compatibility

t-vi:
This implements state_dict and load_state_dict considering overrides (by using named_parameters / named_buffers).
We don't yet support hooks etc.

### DIFF
--- a/thunder/core/module.py
+++ b/thunder/core/module.py
@@ -222,7 +222,8 @@ class ThunderModule(pytorch.nn.Module):
                     f"the shape in current model is {v.shape}."
                 )
                 continue
-            # check dtype?
+
+            # We don't check dtype because PyTorch also does dtype conversion on load.
             with pytorch.no_grad():
                 if assign:
                     if isinstance(v, pytorch.nn.Parameter):
--- a/thunder/core/module.py
+++ b/thunder/core/module.py
@@ -1,6 +1,7 @@
 from contextlib import contextmanager
 import itertools
-from typing import Any, Mapping
+from typing import Any
+from collections.abc import Mapping
 import collections
 
 import torch as pytorch
@@ -176,8 +177,6 @@ class ThunderModule(pytorch.nn.Module):
 
     def load_state_dict(self, state_dict: Mapping[str, Any], strict: bool = True, assign: bool = False):
         # this is much more simple than the original (hooks etc.)
-        if assign:
-            raise NotImplementedError("load_state_dict with assign=True")
 
         # non-persistent buffer overrides?
         non_persistent_buffers_set = set()
@@ -210,7 +209,18 @@ class ThunderModule(pytorch.nn.Module):
                     continue
                 # check dtype?
                 with pytorch.no_grad():
-                    v.copy_(sd_v)
+                    if assign:
+                        if isinstance(v, pytorch.nn.Parameter):
+                            if not isinstance(sd_v, pytorch.nn.Parameter):
+                                sd_v = pytorch.nn.Parameter(sd_v, requires_grad=v.requires_grad)
+                            else:
+                                sd_v.requires_grad_(v.requires_grad)
+                            self._overrides_parameters[name] = sd_v
+                        else:
+                            self._overrides_buffers[name] = sd_v
+
+                    else:
+                        v.copy_(sd_v)
         if strict:
             if keys_unused:
                 errors.insert(
--- a/thunder/core/module.py
+++ b/thunder/core/module.py
@@ -1,10 +1,11 @@
 from contextlib import contextmanager
 import itertools
-from typing import Any
+from typing import Any, Mapping
 import collections
 
 import torch as pytorch
 
+import thunder
 from thunder.core.compile_data import get_compile_data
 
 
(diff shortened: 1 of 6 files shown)


thunder/core/module.py
```diff
@@ -141,6 +143,101 @@ def load_original_state_dict(self, state_dict):
                 else:
                     raise NotImplementedError(f"don't know how to handle {full_k}")
 
+    def state_dict(self, *, destination=None, prefix="", keep_vars=False):
+        """
+        Save state dict of the (transformed) Thunder module.
```
nikitaved, line 148: nit: it actually returns a dict

thunder/core/module.py
```diff
@@ -141,6 +143,101 @@ def load_original_state_dict(self, state_dict):
...
+    def load_state_dict(self, state_dict: Mapping[str, Any], strict: bool = True, assign: bool = False):
+        # this is much more simple than the original (hooks etc.)
+
+        # non-persistent buffer overrides?
+        non_persistent_buffers_set = set()
+        for name, submodule in self._model.named_modules():
+            subprefix = f"{name}." if name else name
+            non_persistent_buffers_set.update(f"{subprefix}{bname}" for bname in submodule._non_persistent_buffers_set)
+
+        keys_unused = thunder.core.utils.OrderedSet(state_dict.keys())
```
nikitaved, line 187: Aren't keys already represent a proxy ordered set? Or could that be that the `state_dict` can be any dict?

thunder/core/module.py
```diff
@@ -141,6 +143,101 @@ def load_original_state_dict(self, state_dict):
...
+        keys_missing = thunder.core.utils.OrderedSet()
+        errors = []
+        for name, v in itertools.chain(self.named_parameters(), self.named_buffers()):
+            if name not in non_persistent_buffers_set:
+                if name not in state_dict:
```
nikitaved, line 192: super minor nit: maybe we can continue on `name in non_persistent_buffers_set` and save a level of indentation?

nikitaved, line 192: super minor nit, no action required: maybe we can `continue` on `name in non_persisten_buffers_set` and save a level of indentation?

thunder/tests/test_networks.py
```diff
@@ -336,17 +337,32 @@ def test_quantization():
     model_fp_reference.requires_grad_(False)
     model_fp_reference.eval()
 
+    model_fp_reference2.set_kv_cache(1, device="cuda", dtype=torch.bfloat16)
+    model_fp_reference2.max_seq_length = 20
+    model_fp_reference2.requires_grad_(False)
```
nikitaved, line 342: will it work if `requries_grad is True`?

thunder/core/module.py
t-vi: Right.

thunder/core/module.py
t-vi: I want to remove the keys from this when used, so I need a copy.

thunder/core/module.py
t-vi: Yeah, let's do that, though, it looks better.

thunder/tests/test_networks.py
t-vi: It doesn't mix with quantization, I'll add a separate test.

thunder/core/module.py
```diff
@@ -141,6 +143,101 @@ def load_original_state_dict(self, state_dict):
...
+            if (
+                getattr(self.__class__, "get_extra_state", pytorch.nn.Module.get_extra_state)
+                is not pytorch.nn.Module.get_extra_state
+            ):
+                destination[extra_state_key] = self.get_extra_state()
+        return destination
+
+    def load_state_dict(self, state_dict: Mapping[str, Any], strict: bool = True, assign: bool = False):
+        # this is much more simple than the original (hooks etc.)
+
```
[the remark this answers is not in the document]
t-vi: I'll add a docstring (come from PyTorch).

thunder/core/module.py
```diff
@@ -141,6 +143,116 @@ def load_original_state_dict(self, state_dict):
...
+        """
+        Returns the state dict of the (transformed) Thunder module.
+
+        Args:
+            destination: if given, use this mutuable mapping as the dict container.
+            prefix: a prefix for the keys.
+            keep_vars: do not detach
+
+        Note that this similar but rather more rudimentary than the original state_dict.
+        """
```
crcrpar, line 156: nice addition but this wouldn't be rendered unless we update https://github.com/Lightning-AI/lightning-thunder/blob/e88966f2f35433e167f40d26e1330684b8ee3fc1/docs/source/reference/thunder.rst?plain=1#L46, looking at https://lightning-thunder--968.org.readthedocs.build/en/968/reference/thunder.html#jited-model-wrapper

thunder/core/module.py
```diff
@@ -141,6 +143,116 @@ def load_original_state_dict(self, state_dict):
...
+        """Loads the state dict to a transformed module.
+
+        Args:
+            state_dict: the state dict to load.
+            strict: error on missing / unused state dict members
+            assign: assign the state dict tensors instead of copying the data
+
+        This is similar much more simple than the original load_state_dict.
+        (Regarding hooks, customization etc.)
+        """
```
crcrpar, line 194: ditto, as for the rendering

thunder/core/module.py
```diff
@@ -141,6 +143,116 @@ def load_original_state_dict(self, state_dict):
...
+    def state_dict(self, *, destination=None, prefix="", keep_vars=False):
+        """
+        Returns the state dict of the (transformed) Thunder module.
+
+        Args:
+            destination: if given, use this mutuable mapping as the dict container.
+            prefix: a prefix for the keys.
+            keep_vars: do not detach
+
+        Note that this similar but rather more rudimentary than the original state_dict.
```
crcrpar, line 155:
```suggestion
        Note that this is similar but rather more rudimentary than the original state_dict.
```

thunder/core/module.py
```diff
@@ -141,6 +143,116 @@ def load_original_state_dict(self, state_dict):
...
+                    f"received {type(sd_v)}"
+                )
+                continue
+            if sd_v.shape != v.shape:
+                errors.append(
+                    f"size mismatch for {name}: copying a param with shape {sd_v.shape} from checkpoint, "
+                    f"the shape in current model is {v.shape}."
+                )
+                continue
+            # check dtype?
```
crcrpar, line 225: is this a TODO or a NOTE?

t-vi: Probably a note that we don't because PyTorch does not.

thunder/core/module.py
t-vi: Right, thank you!

thunder/core/module.py
```diff
@@ -141,6 +143,117 @@ def load_original_state_dict(self, state_dict):
...
+        Args:
+            destination: if given, use this mutuable mapping as the dict container.
+            prefix: a prefix for the keys.
+            keep_vars: do not detach
+
+        Note that this is similar but rather more rudimentary than the original state_dict.
+        """
+        if destination is None:
+            destination = collections.OrderedDict()
+            destination._metadata = collections.OrderedDict()
```
crcrpar, line 159: why OrderedDict? following https://github.com/pytorch/pytorch/blob/8624a571b4eecd11547867591d70992843265e97/torch/nn/modules/module.py#L2190-L2192?

thunder/core/module.py
```diff
@@ -141,6 +143,117 @@ def load_original_state_dict(self, state_dict):
                 else:
                     raise NotImplementedError(f"don't know how to handle {full_k}")
 
+    def state_dict(self, *, destination=None, prefix="", keep_vars=False):
```
crcrpar, line 146: at glance, the major difference from `torch.nn.Module` seems to be that this implementation does not use `_state_dict_pre_hooks` and `_state_dict_hooks`.
I thought either of them could be used for `thunder.distributed.fsdp` to register its hook to create a state dict of unsharded params.

thunder/core/module.py
```diff
@@ -141,6 +143,117 @@ def load_original_state_dict(self, state_dict):
...
+        """
+        Returns the state dict of the (transformed) Thunder module.
+
+        Args:
+            destination: if given, use this mutuable mapping as the dict container.
+            prefix: a prefix for the keys.
+            keep_vars: do not detach
+
+        Note that this is similar but rather more rudimentary than the original state_dict.
```
crcrpar, line 155: how about mentioning the lack of the hooks?

thunder/core/module.py
t-vi: So the tricky thing is that in order to support the hooks, we would need to go to a module-by-module thing again and probably create a copy of the module with the overrides as parameters.
This would be a more elaborate than the current version, but should be doable.
We needed a thing that can save / load state dicts rather soon, so I suggest to start with this and refine from there.

But I wonder: What this does (and what we currently need) is a way to get the state of the thunder module. With FSDP, I would argue that that is the sharded state, and that we should make a function or method to get the unsharded state. Also, in my imagination, such a thing would do communication and I wonder if that is conceptually heavy-handed to do in hooks.

crcrpar: > We needed a thing that can save / load state dicts rather soon, so I suggest to start with this and refine from there.

agreed.

> With FSDP, I would argue that that is the sharded state, and that we should make a function or method to get the unsharded state.

yes.

> I wonder if that is conceptually heavy-handed to do in hooks.

I embarrassingly am not aware of any criteria to decide if a workload is fine to enclosed in a hook or not so I thought it's fair.

thunder/core/module.py
t-vi: Yes, and because otherwise you cannot assign `._metadata` (I tried at first).

thunder/core/module.py
t-vi: It's likely a bit of a taste thing and the thresholds differ for people, too.

But so in some sense hooks are hidden code, so to my mind, the answer is somewhere between "the hooks should not be used by library authors at all, only by end users" (I think I wrote something like that on the PyTorch forums a long time ago) and the hooks, as hidden code, should have only niche effects.

I think for loading state dicts, the hooks/customization is there (and used) in particular to facilitate loading state dicts from older versions.

lantiga (approved):

PR #970, opened 2024-08-14
bump max file size

t-vi:
...because opinfos hits this in #949


t-vi:
Merge to fix CI for #949

mruberry (approved):

Issue #972, opened 2024-08-15
Confusing error message when grad rule is registered incorrectly


kshitij12345:
Following snippet seems like it should work but errors with a confusing message
```python
import torch
import thunder

from thunder.extend import OperatorExecutor, register_executor
from thunder.core.proxies import AnyProxy, TensorProxy
from thunder.core.transforms import get_grad, put_grad

my_ex = OperatorExecutor("my_ex")
register_executor(my_ex)

forward_op = my_ex.register_operator("forward_op", meta=lambda x: (TensorProxy(like=x), AnyProxy(object())), fn=lambda x: (x, x.shape))

backward_op = my_ex.register_operator("backward_op", meta=lambda saved_meta, g: TensorProxy(like=g), fn=lambda saved_meta, g: g)

def forward_op_grad(x):
    out, meta = forward_op(x)
    g = get_grad(out)
    g_o = backward_op(meta, g)
    put_grad(x, g_o)
    return out

my_ex.register_implementation(forward_op, forward_op, grad_transform=forward_op_grad)

def foo(x):
    return forward_op(x)

jfoo = thunder.jit(foo, executors=[my_ex])

jfoo(torch.randn(3, 3, requires_grad=True))
```

Error
```
File "/home/kkalambarkar/lightning-thunder/thunder/core/transforms.py", line 1530, in eval_trace
    safe_map_flat(write, list(sequencify(symbol.output)), list(sequencify(result)))
  File "/home/kkalambarkar/lightning-thunder/thunder/core/utils.py", line 863, in safe_map_flat
    assert s == spec, f"argument layout mismatch: {args[0]} {args[i]}"
AssertionError: argument layout mismatch: [<TensorProxy(name="t0", dtype=thunder.dtypes.float32, shape=(3, 3))>, <AnyProxy '<object object at 0x7c7a38c65ee0>>'] [VJPDual(primal=<TensorProxy(name="t0", dtype=thunder.dtypes.float32, shape=(3, 3))>, residuals=[<AnyProxy '<object object at 0x7c7a38c66580>>'])]
```

One correct way to solve this would be to add a another operation which is then replaced by the forward and backward impls, it would be nice if error message printed something actionable.

<details>

<summary> Working Example </summary>

```python
import torch
import thunder

from thunder.extend import OperatorExecutor, register_executor
from thunder.core.proxies import AnyProxy, TensorProxy
from thunder.core.transforms import get_grad, put_grad

[...]

PR #973, opened 2024-08-15
Better error for the mismatch in the number of outputs between forward and an augmented forward

IvanYashchuk:
Fixes https://github.com/Lightning-AI/lightning-thunder/issues/972.

There's a requirement that the number of outputs in the forward function must match the number of outputs in the augmented forward function.
In the case of `grad_transform=` the augmented function is generated by Thunder but the same requirement still needs to be satisfied.


kshitij12345 (approved):
LGTM, thank you @IvanYashchuk

t-vi (approved):
Thank you @IvanYashchuk @kshitij12345 
What is not to love about better error messages

PR #974, opened 2024-08-15
Add Inductor (torch.compile) concatenation executor to the default list of executors

IvanYashchuk:
We should use an Inductor-based concatenation executor (through torch.compile) by default since it gives us a perf improvement for sections of the network that include the `torch.cat` operation (mainly RoPE). Previously we didn't enable it by default because there were memory leaks that were fixed (see https://github.com/Lightning-AI/lit-thunder-LEGACY/issues/2194 if you have access).

In addition, this PR also avoids hitting a problem with nvFuser (https://github.com/NVIDIA/Fuser/issues/2792). The backward of RoPE was generating a nvFuser fusion in between TorchCompile region because the torchcompile_cat executor wasn't marked to be able to execute `torch.sum`. I also added a test that verifies that only one fusion region is created.
In the (NVIDIA-internal) Mixology dashboard Llama-3-8B, Mistral-7B-v0.1, and stablecode-completion-alpha-3b models do not work with Thunder + Inductor concatenation executor and this problem should be fixed in this PR (cc @apaz-cli @mpatel31415).

### DIFF
--- a/thunder/__init__.py
+++ b/thunder/__init__.py
@@ -175,13 +175,17 @@ get_always_executors = extend.get_always_executors
 
 cudnn_executor: None | extend.Executor = extend.get_executor("cudnn")
 sdpa_executor: None | extend.Executor = extend.get_executor("sdpa")
+torchcompile_cat_executor: None | extend.Executor = extend.get_executor("torchcompile_cat")
 nvfuser_executor: None | extend.Executor = extend.get_executor("nvfuser")
 pytorch_executor: None | extend.Executor = extend.get_executor("torch")
 
-# Default executor list is [cudnn -> sdpa -> nvfuser -> torch -> python]
+# Default executor list is [cudnn -> sdpa -> torchcompile_cat -> nvfuser -> torch -> python]
 # Note that add_default_executor inserts executor at start of list, hence the reverse order below.
 if nvfuser_executor:
     add_default_executor(nvfuser_executor)
+
+if torchcompile_cat_executor:
+    add_default_executor(torchcompile_cat_executor)
 
 if sdpa_executor:
     add_default_executor(sdpa_executor)
--- a/thunder/__init__.py
+++ b/thunder/__init__.py
@@ -184,7 +184,7 @@ pytorch_executor: None | extend.Executor = extend.get_executor("torch")
 if nvfuser_executor:
     add_default_executor(nvfuser_executor)
 
-if torchcompile_cat_executor:
+if torchcompile_cat_executor and pytorch._dynamo.is_inductor_supported():
     add_default_executor(torchcompile_cat_executor)
 
 if sdpa_executor:
--- a/thunder/executors/torch_compile.py
+++ b/thunder/executors/torch_compile.py
@@ -200,8 +200,6 @@ from thunder.executors.torchex import ex as pytorch_ex
 required_ops = {
     "torch.cat",
     prims.cat.id,
-    prims.pad.id,
-    prims.slice_prim.id,
 }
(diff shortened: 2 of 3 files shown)


mruberry:
But the test failures look real

IvanYashchuk:
Need to fix:
```py
FAILED thunder/tests/test_torch_compile_executor.py::test_torch_compile_cat_rope_single_fusion - ValueError: An object of type <class 'NoneType'> is not a valid Executor. Executor list: [thunder.extend.OperatorExecutor('torchcompile_cat'), None]
FAILED thunder/tests/test_torch_compile_executor.py::test_supported_ops_are_in_pytorch_executor - NameError: name 'supported_ops' is not defined
FAILED thunder/tests/test_jit_general.py::test_litgpt - AssertionError: Tensor-likes are not close!
```

mruberry (approved):
Cool!

fyi @Borda, @t-vi, looks like something is up with the docs build

t-vi:
One thing we noticed when running thunder in containers is that torch.compile brings an implicit dependency on GCC. I wonder whether we could detect that and warn if it is not present.

IvanYashchuk:
> One thing we noticed when running thunder in containers is that torch.compile brings an implicit dependency on GCC. I wonder whether we could detect that and warn if it is not present.

Does `torch._dynamo.is_inductor_supported()` help here? I should add a check for this when adding it to the default executor list.

thunder/__init__.py
```diff
@@ -175,14 +175,18 @@ def __version__():
...
 nvfuser_executor: None | extend.Executor = extend.get_executor("nvfuser")
 pytorch_executor: None | extend.Executor = extend.get_executor("torch")
 
-# Default executor list is [cudnn -> sdpa -> nvfuser -> torch -> python]
+# Default executor list is [cudnn -> sdpa -> torchcompile_cat -> nvfuser -> torch -> python]
 # Note that add_default_executor inserts executor at start of list, hence the reverse order below.
 if nvfuser_executor:
     add_default_executor(nvfuser_executor)
 
+if torchcompile_cat_executor:
```
IvanYashchuk, line 187:
```suggestion
if torchcompile_cat_executor and pytorch._dynamo.is_inductor_supported():
```

mpatel31415:
Hi! I'm sorry for writing so late, but I was OOO. I guess we also want to add this executor in the benchmarking script [here](https://github.com/Lightning-AI/lightning-thunder/blob/main/thunder/benchmarks/benchmark_litgpt.py#L394), right? Or even use here only the default list of executors if "thunder" is given. Right now the changes won't be reflected in Mixology results.

IvanYashchuk:
> Hi! I'm sorry for writing so late, but I was OOO. I guess we also want to add this executor in the benchmarking script [here](https://github.com/Lightning-AI/lightning-thunder/blob/main/thunder/benchmarks/benchmark_litgpt.py#L394), right? Or even use here only the default list of executors if "thunder" is given. Right now the changes won't be reflected in Mixology results.

Thank you for pointing this out! We should change the benchmark script to start with the default executor list.

mpatel31415:
> > Hi! I'm sorry for writing so late, but I was OOO. I guess we also want to add this executor in the benchmarking script [here](https://github.com/Lightning-AI/lightning-thunder/blob/main/thunder/benchmarks/benchmark_litgpt.py#L394), right? Or even use here only the default list of executors if "thunder" is given. Right now the changes won't be reflected in Mixology results.
> 
> Thank you for pointing this out! We should change the benchmark script to start with the default executor list.

Sure, I can change the code of the benchmarking script.

PR #975, opened 2024-08-16
more pickling

t-vi:
devices and dtypes should be unique objects for each device / dtype, so we define `__reduce__` to look them up during unpickling rather than recreating them.

As optree < 0.12 has problems with segfaults and pickling (and we saw this happen with Thunder in the wild), we bump the dependency.


nikitaved (approved):
LGTM!

t-vi:
Thank you @nikitaved for reviewing.

PR #976, opened 2024-08-16
Add function to query the set of auto-registered torch operators on jitted function

kiya00:
<details>

</details>

## What does this PR do?

As a follow up to #811, this PR added a query function `print_auto_registered_torch_ops(fn) -> set[str] | None` to retrieve the set of auto-registered torch ops based on the symbols in the trace

Requirement: Fix the bug about `torch.special` operators found along the way (https://github.com/Lightning-AI/lightning-thunder/pull/979)

### DIFF
--- a/thunder/__init__.py
+++ b/thunder/__init__.py
@@ -1019,7 +1019,7 @@ def print_auto_registered_torch_ops(fn: Callable, /) -> set[str] | None:
     op_names = set()
     trc = last_traces(fn)[0]
     for bsym in trc.bound_symbols:
-        if (meta_func_name := getattr(bsym.sym.meta, "__name__", None)) and meta_func_name == "meta_func":
+        if getattr(bsym.sym.meta, "__name__", None) == "meta_func":
             op_names.add(bsym.sym.id)
     return op_names if op_names else None
 
--- a/thunder/__init__.py
+++ b/thunder/__init__.py
@@ -1014,6 +1014,16 @@ def last_compile_options(fn: Callable, /) -> None:
         print(f"\t{option}")
 
 
+def print_auto_registered_torch_ops(fn: Callable, /) -> set[str] | None:
+    """Returns a set of auto-registered Torch operator names present in the given JIT-compiled function."""
+    op_names = set()
+    trc = last_traces(fn)[0]
+    for bsym in trc.bound_symbols:
+        if (meta_func_name := getattr(bsym.sym.meta, "__name__", None)) and meta_func_name == "meta_func":
+            op_names.add(bsym.sym.id)
+    return op_names if op_names else None
+
+
 # TODO (mruberry) Update this
 def _grad_transform(trace):
     grad_fwd_trace = from_trace(trace)
--- a/thunder/__init__.py
+++ b/thunder/__init__.py
@@ -1015,7 +1015,7 @@ def last_compile_options(fn: Callable, /) -> None:
         print(f"\t{option}")
 
 
-def get_auto_registered_torch_ops(fn: Callable, /) -> set[str] | None:
+def get_auto_registered_torch_op_names(fn: Callable, /) -> set[str] | None:
     """Returns a set of auto-registered Torch operator names present in the given JIT-compiled function."""
     op_names = set()
     trc = last_traces(fn)[0]
--- a/thunder/__init__.py
+++ b/thunder/__init__.py
@@ -1017,10 +1017,8 @@ def last_compile_options(fn: Callable, /) -> None:
 
 def get_auto_registered_torch_op_names(fn: Callable, /) -> set[str] | None:
     """Returns a set of auto-registered Torch operator names present in the given JIT-compiled function."""
-    op_names = set()
     trc = last_traces(fn)[0]
-    tuple(op_names.add(bsym.sym.id) for bsym in trc.bound_symbols if has_tags(bsym, {prims.OpTags.AUTO_REGISTERED}))
-    return op_names if op_names else None
+    return {bsym.sym.id for bsym in trc.bound_symbols if has_tags(bsym, {prims.OpTags.AUTO_REGISTERED})}
 
 
 # TODO (mruberry) Update this
--- a/thunder/__init__.py
+++ b/thunder/__init__.py
@@ -54,6 +54,7 @@ from thunder.extend import Executor, add_default_executor
 from thunder.core.compile_data import compile_data_and_stats, get_compile_data
 from thunder.core.langctxs import LanguageContext
 import thunder.core.langctxs as langctxs
+from thunder.core.symbol import has_tags
 from thunder.core.baseutils import run_once, check
 from thunder.core.codeutils import Positions
 from thunder.core.proxies import (
@@ -1014,13 +1015,11 @@ def last_compile_options(fn: Callable, /) -> None:
         print(f"\t{option}")
 
 
-def print_auto_registered_torch_ops(fn: Callable, /) -> set[str] | None:
+def get_auto_registered_torch_ops(fn: Callable, /) -> set[str] | None:
     """Returns a set of auto-registered Torch operator names present in the given JIT-compiled function."""
     op_names = set()
     trc = last_traces(fn)[0]
-    for bsym in trc.bound_symbols:
-        if getattr(bsym.sym.meta, "__name__", None) == "meta_func":
-            op_names.add(bsym.sym.id)
+    tuple(op_names.add(bsym.sym.id) for bsym in trc.bound_symbols if has_tags(bsym, {prims.OpTags.AUTO_REGISTERED}))
     return op_names if op_names else None
 
 
--- a/thunder/common.py
+++ b/thunder/common.py
@@ -86,7 +86,7 @@ class CompileStats:
         last_computation_execution_start (int):
         last_computation_execution_stop (int):
         cache (dict):
-        interpreter_cashe (list):
+        interpreter_cache (list):
         calls (int):
         cache_hits (int):
         cache_misses (int):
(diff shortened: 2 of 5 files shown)


t-vi:
Thank you for the PR @kiya00 !
I can see how allowing the user to query for this is good, but I'm a bit hesitant about this approach. Could the same not be achieved by inspecting the last traces?
If so, I would prefer not collect another set of info.

t-vi:
I'm not quite understanding this. To my mind, a hacky way to get  the auto symbols in the trace would be `{bsym.sym for bsym in lt.bound_symbols if bsym.sym.meta.__code__ in thunder.torch.meta_adaptor.__code__.co_consts}`
for the forward.
And if we wanted, we could keep a set of all auto-registered symbols and compare `{bsym.sym in thunder.torch._auto_registered_symbols}`.

kiya00:
Yes, if we just look at the text format of the traces it's hard to tell, but when we get to the symbol.meta level we can always see the difference in whether the op is auto-registered, and also all the ops are listed in the `thunder/torch/default_torch_ops.py`.
Keeping the auto-registered ops in the CompileStats is more like for non-advanced users to get a direct look at what's going on. This depends on what kind of UX we want, what do you think @IvanYashchuk ?

thunder/common.py
```diff
@@ -88,11 +88,12 @@ class CompileStats:
         last_computation_execution_start (int):
         last_computation_execution_stop (int):
         cache (dict):
-        interpreter_cashe (list):
+        interpreter_cache (list):
```
mruberry, line 91: Nice spelling fix

thunder/torch/default_torch_ops.py
```diff
@@ -693,3 +694,11 @@
         torch.special.xlogy,
     ],
 }
+
+
+def _check_is_auto_registered_torch_op(op: Callable) -> str | None:
+    for k, vals in torch_auto_registered_ops.items():
+        if op in vals:
+            op_name = op.__name__.split("_")[1] if k is torch.special else op.__name__
```
mruberry, line 702: Since this handling comes up in a few places, maybe there should be a common function for the name resolution?

thunder/torch/default_torch_ops.py
```diff
@@ -693,3 +694,11 @@
         torch.special.xlogy,
     ],
 }
+
+
+def _check_is_auto_registered_torch_op(op: Callable) -> str | None:
```
mruberry, line 699: Would you add a docstring to this function describing what it returns and why?

thunder/core/compile_data.py
```diff
@@ -25,6 +25,13 @@ def get_compile_data() -> None | Any:
     return cd
 
 
+def get_compile_stats() -> None | Any:
```
mruberry, line 28: Nice extension

mruberry (commented):
Hey @kiya00! I just made a few requests for more comments/docs and a small suggestion about refactoring the handling of the torch.special name

mruberry:
> Thank you for the PR @kiya00 ! I can see how allowing the user to query for this is good, but I'm a bit hesitant about this approach. Could the same not be achieved by inspecting the last traces? If so, I would prefer not collect another set of info.

If we think that eagerly collecting statistics is burdensome, then we could add some architecture to support the lazy collection of statistics. This lazy collection could work by reviewing the traces associated with a statistics object. That sounds like a fine idea, but also an architectural change that is probably for another PR.

The cost of this particular collection seems very small (the call to check if an operator is auto-registered, and then the addition to a set).

t-vi:
>If we think that eagerly collecting statistics is burdensome, then we could add some architecture to support the lazy collection of statistics. This lazy collection could work by reviewing the traces associated with a statistics object. That sounds like a fine idea, but also an architectural change that is probably for another PR.

I have a strong preference for not adding collection of statistics that we can obtain without collecting.

mruberry:
> > If we think that eagerly collecting statistics is burdensome, then we could add some architecture to support the lazy collection of statistics. This lazy collection could work by reviewing the traces associated with a statistics object. That sounds like a fine idea, but also an architectural change that is probably for another PR.
> 
> I have a strong preference for not adding collection of statistics that we can obtain without collecting. I'm totally cool with providing users a convenient way access this, but I do think we should do this by having a function that grabs the traces and looks at them.

We can probably do that without too much of an issue. @kiya00, what would you think of lazily populating this statistic by inspecting the first trace? 

> 
> > The cost of this particular collection seems very small (the call to check if an operator is auto-registered, and then the addition to a set).
> 
> It also adds a context var lookup to every function call during tracing. We already have a dozen probably, but I think it is one part of what's slow (there are certainly more and many of them I have put in).

Would you file an issue for this? I wouldn't worry about the performance before we measure it.

kiya00:
>We can probably do that without too much of an issue. @kiya00, what would you think of lazily populating this statistic by inspecting the first trace?

Hi @t-vi @mruberry! Sure, I can instead create a function like `def get_auto_registered_torch_ops(fn) -> set[str] | None` to retrieve the set of auto-registered torch ops based on the symbols in the trace as mentioned by @t-vi

thunder/__init__.py
```diff
@@ -1014,6 +1014,16 @@ def last_compile_options(fn: Callable, /) -> None:
         print(f"\t{option}")
 
 
+def print_auto_registered_torch_ops(fn: Callable, /) -> set[str] | None:
+    """Returns a set of auto-registered Torch operator names present in the given JIT-compiled function."""
+    op_names = set()
+    trc = last_traces(fn)[0]
+    for bsym in trc.bound_symbols:
+        if (meta_func_name := getattr(bsym.sym.meta, "__name__", None)) and meta_func_name == "meta_func":
```
t-vi, line 1022: Looks great overall, thank you for revising.
Nit: Could this be
```suggestion
        if getattr(bsym.sym.meta, "__name__", None) == "meta_func":
```
or even better bsym.sym.meta directly to the meta func?

kiya00: `bsym.sym.meta` is the inner function returned by `thunder.torch.meta_adaptor`, I haven't found other checks can directly compare `bsym.sym.meta` to `thunder.torch.meta_adaptor.<locals>.meta_func`, probably checking the names is sufficient? @t-vi

thunder/__init__.py
```diff
@@ -1014,6 +1014,16 @@ def last_compile_options(fn: Callable, /) -> None:
         print(f"\t{option}")
 
 
+def print_auto_registered_torch_ops(fn: Callable, /) -> set[str] | None:
```
mruberry, line 1017: "print" suggest that the operation would actually print the names. Maybe a more accurate name would be `get_auto_registered_torch_op_names`?

thunder/__init__.py
```diff
@@ -1014,6 +1014,16 @@ def last_compile_options(fn: Callable, /) -> None:
...
+    """Returns a set of auto-registered Torch operator names present in the given JIT-compiled function."""
+    op_names = set()
+    trc = last_traces(fn)[0]
+    for bsym in trc.bound_symbols:
+        if getattr(bsym.sym.meta, "__name__", None) == "meta_func":
```
mruberry, line 1022: Instead of checking the name, can we add an attribute, tag, label, etc. to bound symbols that are auto-registered that can be queried, instead? That seems more robust and clear

thunder/tests/test_auto_register_torchops.py
```diff
@@ -206,3 +206,22 @@ def test_alexnet(self):
...
+    def fn_none(a):
+        return torch.nn.functional.relu(a)
+
+    expected = ({"torch.special.erf", "torch.special.gammaln"}, None)
+    for fn, expect in zip((fn, fn_none), expected):
+        cfn = executor.make_callable(fn)
+
+        a = make_tensor((2, 2), device=device, dtype=torch.float32)
+        cfn(a)
+        ops = thunder.get_auto_registered_torch_ops(cfn)
```
kiya00, line 226:
```suggestion
        ops = thunder.get_auto_registered_torch_op_names(cfn)
```

thunder/__init__.py
```diff
@@ -1014,6 +1015,14 @@ def last_compile_options(fn: Callable, /) -> None:
         print(f"\t{option}")
 
 
+def get_auto_registered_torch_ops(fn: Callable, /) -> set[str] | None:
```
kiya00, line 1018:
```suggestion
def get_auto_registered_torch_op_names(fn: Callable, /) -> set[str] | None:
```

thunder/__init__.py
```diff
@@ -1014,6 +1015,14 @@ def last_compile_options(fn: Callable, /) -> None:
         print(f"\t{option}")
 
 
+def get_auto_registered_torch_op_names(fn: Callable, /) -> set[str] | None:
+    """Returns a set of auto-registered Torch operator names present in the given JIT-compiled function."""
+    op_names = set()
```
mruberry, line 1020:
```
trc = last_traces(fn)[0]
return {bsym.sym.id for bsym in trc.bound_symbols if has_tags(bsym, {prims.OpTags.AUTO_REGISTERED}}
```

Use a set comprehension directly, like this. It's OK to return an empty set to indicate that there were no auto registered operations

mruberry (approved):
Hey @kiya00, I made one code cleanup suggestion for your review (it will also simplify the return annotation for the `get_auto_registered_torch_op_names` function)

@IvanYashchuk did you want to take a look?

IvanYashchuk (approved):

thunder/torch/__init__.py
```diff
@@ -5453,6 +5453,7 @@ def register_default_torch_op(torchfn: Callable, torch_module):
         name=torchfn_name,
         meta=_fn,
         id=f"{torch_module.__name__}.{torchfn_name}",
+        tags=(prims.OpTags.AUTO_REGISTERED,),
```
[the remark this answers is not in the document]
kiya00: I noticed that it is currently common to add the tags to the OpTags defined in prims.py, and the tags are also used on torch symbols.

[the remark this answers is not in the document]
t-vi: yeah, so for the ProxyTag I used a registration thing, but currently prims.OpTags is an enum.

PR #977, opened 2024-08-16
make cudagraphs a transform

t-vi:
This introduces thunder.transform.cudagraph.CUDAGraphTransform to replace `use_cudagraphs=True`.

For more detailed control, users can subclass the transform and e.g. override `can_fuse`.


thunder/transforms/cudagraph.py
```diff
@@ -230,7 +234,7 @@ def _can_fuse_node(n: Node):
                 fused_bsyms.extend(bsyms)
             else:
                 region = Region(producers, consumers, bsyms)
-                fusion_bsym: BoundSymbol = self.fuse(region, fusion_counter, num_static_inputs)
+                fusion_bsym: BoundSymbol = self.fuse(region, fusion_counter)
```
nikitaved, line 237: we are missing `num_static_inputs` here for a reason? If we get several non-isomorphic graphs, we should probably think about how to handle this parameter better. But that's not relevant now...

t-vi: I do think that this is one of the things where the default transform is not ideal, but it is not something that was used before: the parameter was there with a default, but there was no way of providing it.

nikitaved: It is used in the backward pass. We are not loosing it there, are we not?

t-vi: Ouch, right.

nikitaved: But yes, the design of this parameter is so-so as it was not expected to have graph breaks...

t-vi: but which were the args covered by it?

t-vi: So it seems that these were the saved for backwards of (???). Is the assumption here that they were either static in the forward (parameters) or copied to the input area of the forward cuda graph?
 
But with the old code:
```
import torch, thunder
with torch.device("cuda"):
    m = torch.nn.Linear(2, 3)
    inp = torch.randn(1, 2, requires_grad=True)
jm = thunder.jit(m, use_cudagraphs=True)

res = jm(inp)
grads = torch.autograd.grad(res.sum(), (inp, *m.parameters()))
```
the forward has no cudagraph, so in that case, having the input as a fixed is not really correct. (admittedly, a corner case).

t-vi: I guess the happy part is that we will get the parameters as static anyways and we will have to look into the buffers for our own good...

crcrpar (commented):
would there be a plan to add a test which demonstrates the composability with e.g. fsdp?

t-vi:
So as we talk about plans:
- refine the static inputs a bit,
- refine the caching, in particular don't use a global cache, but a per-transform one, to let things go out of scope,
- refine the operator selection for what is cuda-graphable,
- check composibility - I'm guessing the communication prims don't mix with cudagraphs?

To my mind, this greatly improves our starting position (similar to but smaller than Nik's work to make this a (then hardcoded) transform instead of a callable wrapper).

mruberry:
This is cool; how do you see cuda graphs as a transform? If we conceptually separate the transforms like this:

```
pre-execution transforms
execution transform
post-execution transforms
"destructive" transforms
```

Where would the cuda graphs transform go? If it's a post-execution (or destructive) transform, should we think about how executors might label their operations as being cuda graphs compatible (or not)? If it's a pre-execution transform, then does it preclude using certain executors in the execution transform?

lantiga (approved):
👍

t-vi:
I have merged this after coordination with @IvanYashchuk , we will address further review comments in a follow-up.

PR #980, opened 2024-08-16
materialization with reset_parameters

t-vi:
This implements materialization with reset_parameters (similar to how it is done in distributed) in the MaterializationTransform.

### DIFF
--- a/thunder/tests/test_transforms.py
+++ b/thunder/tests/test_transforms.py
@@ -231,14 +231,31 @@ def test_materialization_init():
     bitsandbytes_executor = get_bitsandbytes_executor()
 
     def get_model():
-        return torch.nn.Sequential(
-            torch.nn.Linear(2, 3),
-            torch.nn.GELU(),
-            torch.nn.Linear(3, 2),
+        m0 = torch.nn.Linear(2, 2)
+
+        # to not change the rng state
+        with torch.device("meta"):
+            m4 = torch.nn.Linear(2, 2)
+
+        m4.weight = m0.weight
+        m4.bias = m0.bias
+
+        return (
+            torch.nn.Sequential(
+                m0,
+                torch.nn.GELU(),
+                torch.nn.Linear(2, 2),
+                torch.nn.GELU(),
+                m4,
+            )
+            .eval()
+            .requires_grad_(False)
         )
 
     torch.manual_seed(1234)
-    m_ref = get_model()
+    with torch.device("cuda"):
+        m_ref = get_model()
+        inp = torch.randn(3, 2)
 
     jm_ref = thunder.jit(m_ref, transforms=[BitsAndBytesLinearQuant4bit()], executors=(bitsandbytes_executor,))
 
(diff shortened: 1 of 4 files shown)


t-vi:
This needs to handle shared parameters (the other init methods probably, too).

thunder/transforms/materialization.py
```diff
@@ -55,3 +70,40 @@ def module_init_from_transformed_state_dict(transform: MaterializationTransform,
...
+                # The tradeoff is that all of a module's direct parameters need to fit in device.
+                # Each module only initializes its own parameters and not those of its children (recurse=False)
+                if any(
+                    t.is_meta for t in chain(module_copy.parameters(recurse=False), module_copy.buffers(recurse=False))
+                ):
+                    # TODO: we could also support calling a "param_init_fn" argument like PyTorch
+                    module_copy.to_empty(device=transform.device, recurse=False)
+                    if not hasattr(module_copy, "reset_parameters"):
+                        raise TypeError(
+                            f"Materialization requires that the `{type(module).__name__}.reset_parameters` method is implemented."
```
crcrpar, line 95:
```suggestion
                            f"Materialization requires that the `{type(submodule).__name__}.reset_parameters` method is implemented."
```

thunder/transforms/materialization.py
```diff
@@ -55,3 +70,40 @@ def module_init_from_transformed_state_dict(transform: MaterializationTransform,
             model.load_state_dict(state_dict)
 
         return module_init_from_transformed_state_dict
+
+    @staticmethod
+    def init_from_original_module_init():
+        def module_init_from_original_module_init(transform: MaterializationTransform, tm: thunder.ThunderModule):
```
crcrpar, line 76:
```suggestion
        def module_init_from_original_module_init(transform: MaterializationTransform, tm: ThunderModule):
```

thunder/tests/test_transforms.py
```diff
@@ -221,3 +221,33 @@ def test_nvfuser_cse():
...
+
+    torch.manual_seed(1234)
+    init_from_module_init = MaterializationTransform.init_from_original_module_init()
+    with torch.device("meta"):
+        m = get_model()
+    jm = thunder.jit(
+        m,
+        transforms=[BitsAndBytesLinearQuant4bit(), MaterializationTransform("cuda", init=init_from_module_init)],
+        executors=(bitsandbytes_executor,),
+    )
```
crcrpar, line 253: wouldn't it make sense to compare the memory consumption of these two ways of initialization?

t-vi: I'm not entirely sure it is for this size of the model, for this, the quant metadata probably is larger than the weights themselves.
In general, yes. I am always a bit torn here because the transforms tests tend to be slow...

thunder/transforms/materialization.py
```diff
@@ -55,3 +83,60 @@ def module_init_from_transformed_state_dict(transform: MaterializationTransform,
...
+
+    @staticmethod
+    def init_from_original_module_init():
+        def module_init_from_original_module_init(transform: MaterializationTransform, tm: ThunderModule):
+
+            shared_names = tm._get_shared_names()
+            processed_names = set()
+
+            # Shared parameters in PyTorch eager are parameters of module which have different name but share the underlying tensor.
+            # For shared parameter, we replace all occurence shared parameter with it's corresponding `base` parameter.
```
lantiga, line 95:
```suggestion
            # For shared parameter, we replace all occurence shared parameter with its corresponding `base` parameter.
```

lantiga (approved):
Stamped!

PR #984, opened 2024-08-17
increase polygamma test_vjp_correctness tolerance

t-vi:
...seems to be acting up in the CI lately.


t-vi:
I will merge this as soon as the CI is through as the CI seems to need it.

PR #985, opened 2024-08-17
update nvfuser install instructions

t-vi:
This is what I like to use.
cc @xwang233 @jjsjann123 
also cc @nikitaved

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ pip install lightning-thunder
 Alternatively, you can install the latest version of Thunder directly from this GitHub repository as follows:
 
 ```
-# 1) Install nvFuser and PyTorch nightly dependencies:
+# 1) Install nvFuser and PyTorch dependencies:
 pip install --pre nvfuser-cu121-torch24
 ```
 
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ To use Thunder on your local machine:
 - install [nvFuser](https://github.com/NVIDIA/Fuser) and PyTorch stable together as follows:
 
 ```bash
-# install nvFuser which installs the matching nightly PyTorch
+# install nvFuser which installs the matching stable PyTorch
 pip install --pre nvfuser-cu121-torch24
 ```
 
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ To use Thunder on your local machine:
 
 ```bash
 # install nvFuser which installs the matching nightly PyTorch
-pip install --pre 'nvfuser-cu121[torch]' --extra-index-url https://pypi.nvidia.com
+pip install --pre nvfuser-cu121-torch24
 ```
 
 - install [cudnn](https://gitlab-master.nvidia.com/cudnn/cudnn_frontend) as follows:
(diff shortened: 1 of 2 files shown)


docs/source/fundamentals/installation.rst
```diff
@@ -6,11 +6,11 @@ Minimal dependencies
 
 Follow these instructions to install PyTorch, nvFuser, and finally Thunder.
 
-Install PyTorch and nvFuser with pip (command shown is for CUDA 12.1)::
+Install PyTorch and nvFuser with pip (command shown is for CUDA 12.1 and PyTorch 2.4.x)::
```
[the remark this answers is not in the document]
t-vi: My understanding is that 121 is for CUDA 12.x with x >= 1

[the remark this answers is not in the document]
nikitaved: These are not available, apparently. I tried to install wheels for CUDA 12.5

docs/source/fundamentals/installation.rst
```diff
@@ -6,11 +6,11 @@ Minimal dependencies
...
 
-  pip install --pre "nvfuser-cu121[torch]" --extra-index-url https://pypi.nvidia.com
+  pip install --pre nvfuser-cu121-torch24
 
-cu121 can be replaced with cu118 depending on your CUDA version.
+cu121 can be replaced with cu118 depending on your CUDA version. NVFuser builds typically support the latest point release of PyTorch stable versions.
```
nikitaved, line 13: I think we can point to `https://pypi.nvidia.com` as it has wheels for versions other than 11.8 and 12.1.

nikitaved: Also, if I follow the link, I see versions with torch and without torch. What is the difference? Which one should I choose?

README.md
```diff
@@ -79,7 +79,7 @@ To use Thunder on your local machine:
 
 ```bash
 # install nvFuser which installs the matching nightly PyTorch
-pip install --pre 'nvfuser-cu121[torch]' --extra-index-url https://pypi.nvidia.com
+pip install --pre nvfuser-cu121-torch24
```
xwang233, line 82: `nvfuser-...-torch24` builds against stable release of torch 2.4. I think we also need to update the line above "which installs the matching ~~nightly~~ stable"

docs/source/fundamentals/installation.rst
xwang233: torch 2.3 has cuda 11.8 and 12.1
torch 2.4 has cuda 11.8, 12.1, and 12.4

nvfuser nightly on https://pypi.nvidia.com has the cuda matrix (currently the same as torch 2.4)

t-vi:
```suggestion
cu121 can be replaced with cu118 depending on your CUDA version. NVFuser builds typically support the latest point release of PyTorch stable versions.
For torch 2.4, cu124 is also supported. For nightly versions and more detailed instructions, please see https://github.com/NVIDIA/Fuser/#installation
```

lantiga (approved):

README.md
t-vi: Yes, thank you!

docs/source/fundamentals/installation.rst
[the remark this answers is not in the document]
t-vi: Let's go with this until the next person improves further. :)

PR #986, opened 2024-08-18
Move FSDP module transformation to transform_module

t-vi:
This moves the modification of them Module inside `FSDPTransform.transform_module`.
There currently are no changes how it operates.

This is the first step towards a composable FSDP Transform, I filed #987 to discuss the broader plan.

cc @crcrpar @lantiga

### DIFF
--- a/thunder/distributed/__init__.py
+++ b/thunder/distributed/__init__.py
@@ -422,155 +422,6 @@ def get_extract_bucket_name_from_tensor_proxy(granularity: FSDPBucketingStrategy
     return f
 
 
-# When the user calls fsdp(jitted_module), this function does the following
-# - It transforms the ThunderModule jitted_module, materializing and sharding the parameters as `overrides`
-#   in the ThunderModule.
-# - While doing that, it leaves the original user module alone.
-# - It then registers an early transform (callback that runs before prologue is executed) that transforms the
-#   prologue and compute trace.
-#
-# Note that for doing so, there are a few constraints / caveats:
-# - We do not have prologues/compute traces when we transform the module.
-# - We need to record the info from the module transformations because a later transform might modify the module further.
-
-
-def fsdp_transform_module(
-    thunder_model: ThunderModule,
-    *,
-    device: torch.device | None = None,
-    broadcast_from: int | None = None,
-    sharding_strategy: FSDPType = FSDPType.ZERO2,
-    bucketing_strategy: FSDPBucketingStrategy = FSDPBucketingStrategy.NONE,
-) -> ThunderModule:
-    from thunder import compile_data as get_compile_data
-    from thunder.core.transforms import add_transform
-    from thunder.core.module import ThunderModule
-    from thunder.distributed.transforms.fsdp_v2 import FSDPTraceTransform
-
-    process_group = copy_default_process_group()
-    utils.check(process_group is not None, lambda: "The default process group is None")
-    global_rank = tdist.get_rank(group=process_group)
-    world_size = tdist.get_world_size(group=process_group)
-    if device is None:
-        local_rank = int(os.environ["LOCAL_RANK"])
-        device = torch.device("cuda", local_rank)
-
-    cd = get_compile_data(thunder_model)
-    # TODO: promote use_fsdp and use_ddp to public members of CompileData
-    cd.use_fsdp = True
-    orig_module: torch.nn.Module = cd.fn
-    utils.check(
-        isinstance(orig_module, torch.nn.Module) and not isinstance(orig_module, ThunderModule),
-        lambda: f"CompileData.fn expected to be `nn.Module` but {type(orig_module)}",
-    )
-    orig_module.use_fsdp = True
-    orig_module.process_group_for_ddp = process_group
-    orig_module.bucketing_strategy = bucketing_strategy
-    orig_module.sharding_strategy = sharding_strategy
-
-    # modify module
-    sharded_params = {}
-    device_adjustments = {}
-    # We use `shared_params` dictionary to track the shared parameters.
-    # Key to this dictionary is the original parameter from the user's Module.
-    # Values are the copied and sharded parameter for the thunder module and meta-data related to sharding.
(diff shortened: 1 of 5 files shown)


thunder/distributed/tensor_parallel/column_wise.py
```diff
@@ -284,6 +284,7 @@ def forward(self, tokens: torch.Tensor) -> torch.Tensor:
             chunked_param_name_to_layer_type=chunked_param_name_to_layer_type,
             process_group=process_group,
         ),
+        _legacy_copy_params=True,
```
crcrpar, line 287: I'd have tensor parallel transforms behave in the same way as fsdp v2 in regard to param handling; direct sharding w/o copies.

thunder/distributed/tensor_parallel/row_wise.py
```diff
@@ -291,6 +291,7 @@ def forward(self, tokens: torch.Tensor) -> torch.Tensor:
             chunked_param_name_to_layer_type=chunked_param_name_to_layer_type,
             process_group=process_group,
         ),
+        _legacy_copy_params=True,
```
crcrpar, line 294: same comment as column_wise.py

thunder/distributed/tensor_parallel/column_wise.py
crcrpar: also, if we go with this parameter being `True`, by default, which transforms want to copy params?

thunder/distributed/transforms/fsdp_v2.py
```diff
@@ -57,12 +65,166 @@ def __call__(self, bsym: BoundSymbol) -> VISIT_TYPE:
         return VISIT_TYPE.REPLACE
 
 
-@dataclass(frozen=True)
-class FSDPTraceTransform(Transform):
+# When the user calls fsdp(jitted_module), this function does the following
+# - It transforms the ThunderModule jitted_module, materializing and sharding the parameters as `overrides`
+#   in the ThunderModule.
+# - While doing that, it leaves the original user module alone.
```
crcrpar, line 71: is this comment still true?

thunder/distributed/transforms/fsdp_v2.py
```diff
@@ -57,12 +65,166 @@ def __call__(self, bsym: BoundSymbol) -> VISIT_TYPE:
...
+# - It then registers an early transform (callback that runs before prologue is executed) that transforms the
```
crcrpar, line 72: is it still an early transform or a vanilla transform?

thunder/distributed/transforms/fsdp_v2.py
```diff
@@ -57,12 +65,166 @@ def __call__(self, bsym: BoundSymbol) -> VISIT_TYPE:
...
+        self,
+        device: torch.device | None = None,
+        broadcast_from: int | None = None,
+        sharding_strategy: FSDPType = FSDPType.ZERO2,
+        bucketing_strategy: FSDPBucketingStrategy = FSDPBucketingStrategy.NONE,
+    ):
+        self.device = device
+        self.broadcast_from = broadcast_from
+        self.sharding_strategy = sharding_strategy
+        self.bucketing_strategy = bucketing_strategy
```
crcrpar, line 95: nit-picking: I might prefer explicit init of these three here, and check the size / type of them in their callsite.
```suggestion
        self.bucketing_strategy = bucketing_strategy
        self.sharded_params: dict[str, Any] = {}
        self.process_group: ProcessGroup | None = None
        self.shared_params_name: dict[str, str] = {}
```

thunder/distributed/tensor_parallel/column_wise.py
t-vi: Yes, this parameter is just a "let's keep Tensor Parallel without changing the transform in this PR", but we would want to update tensor parallel in the same way.

thunder/distributed/transforms/fsdp_v2.py
t-vi:
```suggestion
# - It then registers an transform (callback that runs before prologue is executed) that transforms the
```

thunder/distributed/transforms/fsdp_v2.py
t-vi: Briefly, #988 will change the original user module if requested to free the unsharded parameters.

thunder/distributed/tensor_parallel/column_wise.py
t-vi: I hope to get to the fourth part (delegate materialization to MaterializationTransform) ideally tomorrow, then we can see if we want to split the tensor parallel in the same parts or different ones.

thunder/distributed/transforms/fsdp_v2.py
t-vi: I updated this in #995

thunder/distributed/__init__.py
```diff
@@ -621,13 +472,19 @@ def fsdp(
     )
 
     if isinstance(model, thunder.ThunderModule):
-        return fsdp_transform_module(
+        from thunder.core.transforms import add_transform
+        from thunder.distributed.transforms.fsdp_v2 import FSDPTransform
+
+        new_model = add_transform(
```
crcrpar, line 478: nit-picking

```suggestion
        return add_transform(
```

thunder/distributed/__init__.py
```diff
@@ -621,13 +472,19 @@ def fsdp(
...
-            sharding_strategy=sharding_strategy,
-            bucketing_strategy=bucketing_strategy,
+            transform=FSDPTransform(
+                device=device,
+                broadcast_from=broadcast_from,
+                sharding_strategy=sharding_strategy,
+                bucketing_strategy=bucketing_strategy,
+            ),
         )
+        return new_model
```
crcrpar, line 487:
```suggestion
```

crcrpar (approved):

t-vi:
Thank you @crcrpar

PR #988, opened 2024-08-18
free orig parameters after sharding

### LINKED ISSUE #478: fsdp(jit(...)) transform can use more memory compared to jit(fsdp(...))

t-vi:
Fixes #478 

This is an alternative to #932 also covering the original #478 that leverages the Thunder meta tensor handling for the MaterializationTransform.
It builds on #986 (I need to change the target branch if that is merged.)

The test has been taken (and parametrized to cover cuda and meta) from #932 (all credit to @crcrpar ).

Thank you @kshitij12345 for the reproducer.

### DIFF
--- a/thunder/distributed/transforms/fsdp_v2.py
+++ b/thunder/distributed/transforms/fsdp_v2.py
@@ -24,7 +24,7 @@ from thunder.core.trace import TraceProvenance
 from thunder.core.transforms import VISIT_TYPE
 from thunder.core.transforms import visitor_transform
 from thunder.core.transform_common import Transform
-from thunder.distributed import copy_default_process_group, FSDPType, FSDPBucketingStrategy, _shard_param
+from thunder.distributed import copy_default_process_group, FSDPType, FSDPBucketingStrategy, _shard_param, _materialize
 
 if TYPE_CHECKING:
     from typing import Any
--- a/thunder/distributed/transforms/fsdp_v2.py
+++ b/thunder/distributed/transforms/fsdp_v2.py
@@ -88,11 +88,13 @@ class FSDPTransform(Transform):
         broadcast_from: int | None = None,
         sharding_strategy: FSDPType = FSDPType.ZERO2,
         bucketing_strategy: FSDPBucketingStrategy = FSDPBucketingStrategy.NONE,
+        release_original_parameters: bool = False,
     ):
         self.device = device
         self.broadcast_from = broadcast_from
         self.sharding_strategy = sharding_strategy
         self.bucketing_strategy = bucketing_strategy
+        self.release_original_parameters = release_original_parameters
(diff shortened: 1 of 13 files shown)


t-vi:
This is ready for review, but I'm marking as draft for the target branch.

thunder/distributed/transforms/fsdp_v2.py
```diff
@@ -212,6 +214,15 @@ def transform_module(
                 )
                 new_shape = thunder_model._overrides_parameters[pn].shape
                 self.sharded_params[pn] = (old_shape, new_shape, thunder_model._overrides_parameters[pn].device)
+                if self.release_original_parameters:
+                    base_pn = pn.rsplit(".", 1)[-1]
+                    p_orig = getattr(submodule, base_pn)
+                    if p_orig.device.type != "meta":
+                        p_meta = torch.nn.Parameter(p.to(device="meta"), requires_grad=p.requires_grad)
+                        p_meta._thunder_device = p_orig.device
+                        setattr(submodule, base_pn, p_meta)
```
crcrpar, line 223: Can't we use [`torch.nn.Module.register_parameter`](https://pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.register_parameter) or something like that, instead of `setattr`?

t-vi: We do know that it is a parameter, so setattr will do the right thing, but I'm happy to change it to register_parameter if you prefer.

t-vi:
```suggestion
                        submodule.register_parameter(base_pn, p_meta)
```

IvanYashchuk (approved):

crcrpar (approved):

t-vi:

Issue #987, opened 2024-08-18
Composable FSDP Transform


t-vi:
Make FSDP composable with other transforms.

Desired features:

- allow FSDPTransform to compose with other transforms (e.g. some Quantization or QLORA),
- allow FSDPTransform to work with Materialization:
   - sharding on meta,
   - sharding of state dict,
   - release memory after sharding if desired and called with a materialized model (in particular default for fsdp(jit(model))?), i.e. fix #478 
- as a side condition `fsdp(jit(model))` must continue to work as a convenience method for practitioners (thanks @crcrpar for the observation that this should stay simple). 

Steps include:

- [ ] move transformation of module to `FSDPTransform.transform_module` (PR #986 )
- [ ] do memory-releasing (PR #988 (also covering PR #932))
- [ ] implement `FSDPTransform.transform_state_dict_for_submodule` (PR #989 )
- [ ] delegate materialization (PR #995 )

For the side condition of keeping `fsdp(jit(model))` working as is, all of these should work while keeping the existing tests unchanged.

@crcrpar @IvanYashchuk (who asked about the plan)

cc @carmocca @crcrpar

PR #989, opened 2024-08-18
implement FSDPTransform.transform_state_dict_for_submodule

t-vi:
This implements state dict transformation, immediately enabling load_original_state_dict to load an unsharded dictionary into a sharded model.

This is a step in #987 .

(it's ready but relative to #988)

### DIFF
--- a/thunder/distributed/__init__.py
+++ b/thunder/distributed/__init__.py
@@ -422,155 +422,6 @@ def get_extract_bucket_name_from_tensor_proxy(granularity: FSDPBucketingStrategy
     return f
 
 
-# When the user calls fsdp(jitted_module), this function does the following
-# - It transforms the ThunderModule jitted_module, materializing and sharding the parameters as `overrides`
-#   in the ThunderModule.
-# - While doing that, it leaves the original user module alone.
-# - It then registers an early transform (callback that runs before prologue is executed) that transforms the
-#   prologue and compute trace.
-#
-# Note that for doing so, there are a few constraints / caveats:
-# - We do not have prologues/compute traces when we transform the module.
-# - We need to record the info from the module transformations because a later transform might modify the module further.
-
-
-def fsdp_transform_module(
-    thunder_model: ThunderModule,
-    *,
-    device: torch.device | None = None,
-    broadcast_from: int | None = None,
-    sharding_strategy: FSDPType = FSDPType.ZERO2,
-    bucketing_strategy: FSDPBucketingStrategy = FSDPBucketingStrategy.NONE,
-) -> ThunderModule:
-    from thunder import compile_data as get_compile_data
-    from thunder.core.transforms import add_transform
-    from thunder.core.module import ThunderModule
-    from thunder.distributed.transforms.fsdp_v2 import FSDPTraceTransform
-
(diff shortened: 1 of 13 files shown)


thunder/distributed/__init__.py
```diff
@@ -594,7 +594,29 @@ def _shard_param(
         # NOTE This could be a ShardTensor to indicate other parts of the code
         # that it's sharded and should be treated differently
         shard = param.data.narrow(dim_to_shard, chunk_size * rank, chunk_size).clone()
-        param.data = shard
+        return shard, None
+
+
+def _shard_param(
```
crcrpar, line 600: shouldn't this be decorated with `torch.no_grad()`?

thunder/distributed/__init__.py
```diff
@@ -594,7 +594,29 @@ def _shard_param(
...
+    shard, padding_size = _shard_tensor(
+        param,
+        rank,
+        world_size,
+        name,
+        allow_padding_for_fsdp=allow_padding_for_fsdp,
+        dim=dim,
+    )
+    param.data = shard
+    if allow_padding_for_fsdp:
```
crcrpar, line 618: nit-picking
```suggestion
    if allow_padding_for_fsdp is not None:
```

thunder/distributed/transforms/fsdp_v2.py
```diff
@@ -234,7 +243,14 @@ def transform_module(
     def transform_state_dict_for_submodule(
         self, model: thunder.ThunderModule, submodule_name: str, state_dict: dict
     ) -> dict:
-        raise NotImplementedError("cannot transform state dict yet")
+        prefix = submodule_name + ("." if submodule_name else "")
```
crcrpar, line 246: nit-picking
```suggestion
        prefix = ""
        if submodule_name:
            prefix = f"{submodule_name}."
```

crcrpar (commented):
so would this happen to indicate that all-gathering sharded state dict is user's responsibility?

thunder/distributed/__init__.py
t-vi: This is a (non-optional) bool and defaults to false, I don't think it should be None.

thunder/distributed/__init__.py
t-vi: Great idea!

thunder/distributed/__init__.py
```diff
@@ -555,15 +556,15 @@ def _shard_params(
             sharded_params[param] = True
 
 
-def _shard_param(
+def _shard_tensor(
```
t-vi, line 559:
```suggestion
@torch.no_grad()
def _shard_tensor(
```

thunder/distributed/__init__.py
crcrpar: my bad sorry I was confused with the other int | None padding var

thunder/distributed/__init__.py
t-vi: I added the decorator to `_shard_tensor`.

crcrpar (approved):
LGTM

t-vi:
Thank you @crcrpar

PR #936, opened 2024-08-07
Make return statement depend on the outputs of `prim.copy_`s

shino16:
Fixes #912 and #930. Partially reverts #920.

Now `prims.python_return` has a keyword-only parameter `hidden_dependencies`. It can contain TensorProxies that we are not returning but must be evaluated. We gather all the outputs of `prims.copy_` in it to make sure they do not get eliminated. Since we seemingly do not support `Tensor.copy_` directly, functionalization should be the only place `prims.copy_` can appear.

`TraceCtx.output` only cares the arguments of `prims.python_return` and not its keyword arguments, so I wonder how this affects backward computation (`thunder.core.transforms`, `thunder.core.vjp_utils`) and distributed computatoin (`thunder/distributed/`).

## Example

```py
import thunder
import torch

@thunder.jit
def f(x, y):
    x += 1
    y *= 2
    return x + y

z = f(torch.tensor(0), torch.tensor(1))
```
```

PR #990, opened 2024-08-19
Register `copy_`s as return's `BoundSymbol._additional_dependencies`

### LINKED ISSUE #912: `fuse_bound_symbols` (toposort on bsyms) can put `return` before `copy_` on arguments

### LINKED ISSUE #930: fuse_bound_symbols skip the side effect of Tensor.copy_

shino16:
Rewrites #936. Fixes #912. Fixes #930.

Based on a message from @IvanYashchuk

### DIFF
--- a/thunder/core/symbol.py
+++ b/thunder/core/symbol.py
@@ -366,9 +366,12 @@ class BoundSymbol(BoundSymbolInterface):
     _object_ctx: dict = field(default_factory=dict)
     _executor: None | Any = None
 
-    # Only used by prims.python_return
-    # prims.copy_'s must be placed before the return statement
-    _copy_outputs_before_return: Sequence[BoundSymbol] = ()
+    # BoundSymbols may later be reordered and grouped into regions for fusion.
+    # This reordering consideres bsym.flat_proxy_args as bsym's dependencies and place them before bsym itself.
+    # This includes bsym._additional_dependencies in addition to bsym.args and kwargs.
+    # e.g. functionalization registeres prims.copy_'s output as prims.python_return's additional dependency
+    # so that the copy happens before returning.
+    _additional_dependencies: Sequence[BoundSymbol] = ()
 
     # The line number of the bound symbol
     # NOTE This is only intended for internal use, and should be set explicitly on all BoundSymbols
@@ -399,7 +402,7 @@ class BoundSymbol(BoundSymbolInterface):
             "_import_ctx": self._import_ctx,
             "_object_ctx": self._object_ctx,
             "_executor": self._executor,
-            "_copy_outputs_before_return": self._copy_outputs_before_return,
+            "_additional_dependencies": self._additional_dependencies,
         }
 
         self_kwargs.update(kwargs)
--- a/thunder/core/symbol.py
+++ b/thunder/core/symbol.py
@@ -477,6 +477,7 @@ class BoundSymbol(BoundSymbolInterface):
 
         nargs = swap(self.args) if not skip_inputs else self.args
         nkwargs = swap(self.kwargs) if not skip_inputs else self.kwargs
+        new_additional_deps = swap(self._additional_dependencies) if not skip_inputs else self._additional_dependencies
 
         new_output = swap(self.output) if not skip_output else self.output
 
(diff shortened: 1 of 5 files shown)


IvanYashchuk:
Some tests fail with a common error: `AttributeError: 'BoundSymbol' object has no attribute '_hidden_dependencies'`
```
thunder/tests/test_inplace_functionalization.py::test_functionalization_remainder__torch_cpu_thunder.dtypes.float32 - AttributeError: 'BoundSymbol' object has no attribute '_hidden_dependencies'
```

IvanYashchuk:
Probably the `from_bsym_swap_proxies` method needs to be updated to copy the new attribute

thunder/core/symbol.py
```diff
@@ -366,6 +366,10 @@ class BoundSymbol(BoundSymbolInterface):
     _object_ctx: dict = field(default_factory=dict)
     _executor: None | Any = None
 
+    # Only used by prims.python_return
+    # prims.copy_'s must be placed before the return statement
+    _copy_outputs_before_return: Sequence[BoundSymbol] = ()
```
IvanYashchuk, line 371: This should be `_hidden_dependencies`, right?

IvanYashchuk: Could you please explain in the comment above for future readers why this attribute is useful and where it could be applicable in the future besides prims.python_return?

shino16: I updated the comment, but I wonder what it can be helpful for other than copy_ and return. Possibly something related to torch.distributed?

shino16:
Compared to adding hidden dependencies only to `prims.python_return`, it seems like a bit of pain to worry about maintaining additional dependencies on all BoundSymbols throughout transforms.

shino16:
Sorry for switching between draft and ready for review. `thunder.core.transforms.forward_and_backward_from_trace` [makes up a new trace](https://github.com/Lightning-AI/lightning-thunder/blob/main/thunder/core/transforms.py#L3684) which evaluates the computation trace (including `prims.copy_`) and returns auxiliary data, and I need to establish the dependencies between those operations. This did not cause test failure, but resulted in a failure of [this assertion](https://github.com/Lightning-AI/lightning-thunder/blob/main/thunder/executors/nvfuserex_impl.py#L724) during an experiment. Maybe this is not a good way to go.

IvanYashchuk:
@shino16 explained to me over Slack why this approach is far from ideal and we probably will not continue with this solution. He and Masaki have another idea that would work better and more suitable for Thunder. The linked doc currently doesn't have a solution to the reordering problem with inplace on inputs with no later usage. I have added an example and will write a section about this.

shino16:
* Having to worry about maintaining an additional attribute (`_additional_dependencies`) is a bit of trouble. We have to pass it around while manipulating bsyms, and it is conceptually not so intuitive as bsym.args/kwargs/output.
* `_additional_dependencies` has to be set up every time we create a trace, which may result in somewhat widespread changes.

PR #999, opened 2024-08-20
Use pytorch 2.3.1 based containers

crcrpar:
cc @shino16


t-vi (approved):
Thank you @crcrpar for covering us here!

Issue #981, opened 2024-08-16
CUDAGraphs in Thunder


t-vi:
#977 proposes to make CUDAGraphs a Transform with the cudagraphs being created in the `post_optimization` phase of creating the executable. This in particular enables people to subclass or copy the CUDAGraphs transform to have their own, bespoke thing.

We do have some ideas (and this is just a list from the top of my head to get things started), but this issue is to collect discussion and ideas about how you want CUDAGraphs to look like.

## static input detection

Currently we auto-mark parameters, but might be able to do more, e.g. buffers.
If we knew things about the things between fusions, we might know if we can re-use input buffers.

## caching

In particular we would not want to use a global cache, but a per-transform one, to let things go out of scope.
Also we want to enable users to potentially override the caching (we found that it can be interesting to re-use cudagraph
fusions in case we have some non-grapheable bit in a repeated block of a model).

## operator selection for what is cuda-graphable

I think it would be cool if symbols in general had a tag whether they are suitable for CUDAGraph inclusion.
After #977 we have the advantage that if people have particular ideas about this, they can just subclass and override can_fuse.   

## check composability

Test eg. fsdp + cudagraph
I'm guessing the communication prims don't mix well with cudagraphs? but that might be with operator tagging...

## (not immediate) future: memory information about ops

CUDAGraphs could enormously benefit by more information about the memory effects of operators (so similar to meta info but about what allocations a function makes, where the output is allocated etc.), but that is in the future.

Obiviously other people @nikitaved @IvanYashchuk @tfogal and more will be much more knowledgeable, but so here is a start.

mruberry:
triage review:

- let's set aside to talk about this more
- how do checks in the prologue extend to handle strides?

t-vi:
Outcomes of the design discussion: (https://docs.google.com/presentation/d/1JhmJqikAneYDv4S-g9rcAOS9xy35FztydE-PzxFhi3w/edit?usp=sharing)
- tag proxies as static memory location
- tag operators as safe or not
- bubble up constraints like strides? (maybe the dynamic shape work will show us the way) or update automagically (recently in cudagraphs)

PR #1001, opened 2024-08-20
attach cudagraph cache to the transform

t-vi:
This introduces a CUDAGraphRunner class to bundle building and caching cudagraphs and attach them to the CUDAGraphTransform introduced in #977 .

(Design issue #981 ) 

From my POV this is ready for review, but it builds on #977, so I'm labeling it draft.

### DIFF
--- a/thunder/transforms/cudagraph.py
+++ b/thunder/transforms/cudagraph.py
@@ -24,7 +24,6 @@ class ArgsDescriptor:
     sizes: tuple
     strides: tuple
     non_tensor_args: tuple
-    args: list = field(hash=False, repr=False, compare=False)
 
 
 def to_arg_descriptor(*args):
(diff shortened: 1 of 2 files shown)


t-vi:
@nikitaved @IvanYashchuk

thunder/transforms/cudagraph.py
```diff
@@ -100,44 +111,79 @@ def __call__(self, *args):
...
+            static_inputs_mask = None
+
+        x_fn_name = f"{fn_name}_{self.name_counter}"
+        self.name_counter += 1
+
+        self.python_callables[x_fn_name] = (
+            self.make_python_callable_from_symbols(fn_name, bsyms, inputs, outputs),
+            static_inputs_mask,
+        )
+        self.trace_symbols[x_fn_name] = (bsyms, inputs, outputs)
```
IvanYashchuk, line 164: How can I access these dictionaries in the regular user script? How should developers get to know that this information is saved and available for inspection?

thunder/transforms/cudagraph.py
```diff
@@ -35,63 +34,75 @@ def extract_descriptor(arg):
...
-    # PyTorch's Inductor: torch/_inductor/cudagraph_trees.py
-    # Design doc: https://docs.google.com/document/d/1ZrxLGWz7T45MSX6gPsL6Ln4t0eZCSfWewtJ_qLd_D0E/view
-    graph = torch.cuda.CUDAGraph()
-    with torch.cuda.graph(graph, stream=stream):
-        static_outputs = fn(*static_inputs)
-
-    return graph, static_inputs, static_outputs
+    def build_cuda_graph(
+        self, fn: Callable, args: list[any], static_args_mask: tuple[bool, ...]
+    ) -> tuple[torch.cuda.CUDAGraph, Sequence[torch.Tensor | Any], Sequence[torch.Tensor | Any]]:
```
IvanYashchuk, line 54: What is the benefit of having this as a method and not as a free function?

IvanYashchuk (approved):
The code is organized a bit differently now. Looking forward to where it will take us!

thunder/transforms/cudagraph.py
```diff
@@ -34,105 +34,155 @@ def extract_descriptor(arg):
...
-        self.num_static_inputs = num_static_inputs
+        graph.replay()
+        return static_outputs
 
-    def __call__(self, *args):
-        if self.num_static_inputs is not None:
-            static_inputs_mask = (True,) * self.num_static_inputs + (False,) * (len(args) - self.num_static_inputs)
-        else:
-            static_inputs_mask = tuple(isinstance(arg, torch.nn.Parameter) for arg in args)
+    def make_python_callable_from_symbols(
```
lantiga, line 114: This looks like something we could have in a transform building block library in the future

thunder/transforms/cudagraph.py
t-vi: Currently, you have to look at the source. I'm not yet sure that this is the exact information we want to keep, we should refine this.

thunder/transforms/cudagraph.py
t-vi: My tinking was that having the ability to override the buffer allocation (get_static_buffer) could get useful. We could have it as a callback instead but then it might just as well be a method.

lantiga (approved):
Looks great

thunder/transforms/cudagraph.py
t-vi: Indeed! The question is whether we should build the region before...

thunder/transforms/cudagraph.py
```diff
@@ -100,44 +111,79 @@ def __call__(self, *args):
...
+
+        return callable
 
 
 class CUDAGraphTransform(Transform):
     """
     Transform to fuse operations into CUDA graphs post optimization.
 
-    This class provides the basic infrastructure, but it is expected that you might subclass this transform
+    This class provides the basic infrastructure, but it is expected that you might subclass this transformm
```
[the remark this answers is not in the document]
t-vi: fixed, thank you!

thunder/transforms/cudagraph.py
t-vi: so in addition to everything else the `x_fn_name` is a bit stupid, the trick was that the transform is applied to forward and backward. If we are OK with cudagraph regions not starting from 1 in the backward, I would just generate the name in the runner and return it.
This would allow a link from trace->cache name here easily. Is that what you have in mind?

t-vi:
So I'm merging this now, I expect to do another PR or two in the next few days around the dev x of this.
(One other thing I want to do is perhaps pass in the static argument mask which we might obtain from analysing the graph(s) - parameters+buffers from the module are static etc.)

Issue #246, opened 2024-04-22
Not working stablecode-completion-alpha-3b model from LitGPT


mpatel31415:
## 🐛 Bug

With newest version of Docker image (tested on 2024-04-22 ) training with thunder.jit with additional inductor executor gives OOM error.

### To Reproduce

Before each testing each compilation method I restarted the container:
```
mkdir -p output
docker run --pull=always --gpus all --ipc=host --ulimit memlock=-1 --ulimit stack=67108864  -v $PWD/output:/output -it gitlab-master.nvidia.com:5005/dl/pytorch/update-scripts:pjnl-20240422
```
**Thunder inductor**
* OutOfMemoryError – CUDA out of memory. 
* command: 
```
python /opt/pytorch/lightning-thunder/thunder/benchmarks/benchmark_litgpt.py --model_name stablecode-completion-alpha-3b --compile  "thunder_inductor"
```

**Eager**
* Works fine
* command:
```
python /opt/pytorch/lightning-thunder/thunder/benchmarks/benchmark_litgpt.py --model_name stablecode-completion-alpha-3b --compile "eager"
```

**Thunder**
* Works fine
* command:
```
python /opt/pytorch/lightning-thunder/thunder/benchmarks/benchmark_litgpt.py --model_name stablecode-completion-alpha-3b --compile  "thunder"
```

**Inductor**
* Works fine
* command:
```
 python /opt/pytorch/lightning-thunder/thunder/benchmarks/benchmark_litgpt.py --model_name stablecode-completion-alpha-3b --compile  "inductor"
```

### Expected behavior

If we can run model in Eager mode, with Thunder + Pytorch default compilation we should be able to use it for Thunder + Inductor as well.

### Environment

As in the Docker image. This results come from single H100. 

nvidia-smi output:
![image](https://github.com/Lightning-AI/lightning-thunder/assets/149149379/720092ee-4ad1-4852-9e9c-ddcea53ecbb1)

mruberry:
triage review:

- @mpatel31415 do we know how much memory PyTorch eager and/or torch.compile alone uses when executing this model?
- we suspect the latest memory fix may not be in the docker, let's wait for the results with that fix

mpatel31415:
When running:
```
python /opt/pytorch/lightning-thunder/thunder/benchmarks/benchmark_litgpt.py \
--model_name stablecode-completion-alpha-3b \
--compile  "thunder_inductor"
```
[...]

PR #1003, opened 2024-08-20
Saved for backward tensors rematerialization 

### LINKED ISSUE #871: Recomputation of saved for backward tensors

riccardofelluga:
<details>

</details>

## What does this PR do?

This PR starts to fix #871 by introducing the recomputation logic for the "saved for backward" tensors. As discussed earlier, this PR acts after the creation of the backward trace and in essence moves all the tensors that would be saved in memory to the backward trace to be rematerialized as needed.

Also added the compile option to enable it: set `enable_saved_for_backward_recomputation=False`

From here sky is the limit really, there are still some open questions and that's the reason why this PR starts as a draft:
- [x] What should be the decision logic that identifies which tensors are better rematerialized and which are not? In particular it's gonna be interesting the integration with CPU offloading. 
- [x] Now the computation is simply appended to the start of the backward trace and later optimization steps take care of the reordering. Reordering computation is tricky at the stage where this feature acts, as all the optimizations that come after will change the order of the symbols anyway but it might help?

## PR review

Some early results, include #246 where `rematerialization + thunder_inductor_cat` gets the memory consumption downf from 98.49GB(OOM on H100) to 72.52GB. Which means 26.36% less memory at the cost of 21.02% more compute. 
Another example for the same model but with CUDNN shows that there are some situations where remat is not ideal: `remat + thunder_cudnn` resulted in ~5GB saved or more precisely 6.57% less memory for 29.58% more compute, which doesn't sound like a good deal for now.


thunder/core/transforms.py
```diff
@@ -3768,9 +3769,91 @@ def backward_fn(saved_for_backward, cotangents):
...
+    # Everything is now ready and we can start to build the new traces.
+    new_fw_trace = from_trace(fw_trace)
+    new_fw_trace.bound_symbols = fw_trace.bound_symbols.copy()
+    new_return = (fw_trace.output[0], (new_saved_for_backward, fw_trace.output[1][1]))
+    new_fw_trace.bound_symbols[-1] = prims.python_return.bind(new_return, output=())
+
+    new_bw_trace = from_trace(bw_trace)
+    new_bsym = bw_trace.bound_symbols.copy()
+
+    with tracectx(new_bw_trace):
```
kshitij12345, line 3836: Do we need `tracectx`, as we are explicitly passing the `name="C0"`?

thunder/core/transforms.py
```diff
@@ -3768,9 +3769,91 @@ def backward_fn(saved_for_backward, cotangents):
...
+        unpack_args = (CollectionProxy(new_saved_for_backward, name="C0"), len(new_saved_for_backward))
+
+    new_bsym[4] = prims.unpack_sequence.bind(*unpack_args, output=new_saved_for_backward)
```
kshitij12345, line 3839: We should assert that `new_bsym[4]` is `unpack_sequence` before replacing it. In case the position of this symbol changes in future.

thunder/core/transforms.py
```diff
@@ -3768,9 +3769,91 @@ def backward_fn(saved_for_backward, cotangents):
...
+    # Attach the producer symbols
+    new_bsym[6:6] = prod_symbols
```
kshitij12345, line 3841: Probably mention why we are adding the new symbols at position 6?

Also, I think here we are implicitly relying on the fact the these symbols will be re-ordered closer to their consumers in the later passes. (If that doesn't happen, we will just end up creating all these intermediate tensors early in the backward trace, nullifying the recomputation).

Adding a comment for the same would be really helpful.

kshitij12345:
Also, I am not sure how this will interact with TE executor - we wrap the forward function in `fp8_autocast` to enable the fp8 computation but we don't do it for backward pass as it is not [recommended](https://docs.nvidia.com/deeplearning/transformer-engine/user-guide/examples/fp8_primer.html#Handling-backward-pass). Also, we return a mock Context object and pull out saved for backward tensors. So I have a feeling that recomputation won't work correctly as is. 

For now, we should probably skip recomputing the intermediates generated by TE executor.

t-vi:
One thing I could imagine would work well is
- add a tag to proxies that they should be recomputed,
- have a transform that tags the proxies to which we want it to apply.
This way, we could give people a knob to turn until we have the perfect solution (or an evil person could make a lookaside to CheckpointingWrapper.forward that monkey-patches proxify to tag things in acquisition).

thunder/core/transforms.py
riccardofelluga: Unfortunately not because the proxy creation ends up checking if the trace contains the same name already here:

https://github.com/Lightning-AI/lightning-thunder/blob/34fed98e71957afb215cc1c5be6c9b3924e82d19/thunder/core/proxies.py#L66-L68

thunder/core/transforms.py
```diff
@@ -3768,9 +3769,91 @@ def backward_fn(saved_for_backward, cotangents):
...
+
+    Returns:
+        tuple[Trace, Trace]: A tuple containing the new forward and backward traces.
+    """
+    start_time_ns = time.perf_counter_ns()
+    # First of all we need to get the saved for backward tensors
+    saved_for_bw = get_saved_for_backward_tensors(fw_trace)
+
+    # Let's create these two maps that will help writing the logic
+    fw_trace_args = {j.name: j for j in fw_trace.args}
```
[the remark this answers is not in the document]
riccardofelluga: Thanks for this suggestion! Initially i tought that it would just become too ugly to use this interface all over but it turns out it can help create sets and it doesn't look that bad. Sets allow to do the operations we need to exclude/include symbols very easily

riccardofelluga:
As suggested the tagging system might be a good solution for giving the user a knob to turn. However, such remat decision happen before any user transform is executed and therefore as the code stands now we would need either an ugly patch or a big refactor and I don't think they are in the scope of this PR but it can be explored in another, follow-up issue

thunder/core/transforms.py
```diff
@@ -3736,9 +3737,99 @@ def backward_fn(saved_for_backward, cotangents):
...
+    all_rematerializable = old_saved_for_bwd - fwd_trace_args
+
+    # Here will come the selection logic for what to materialize.
+    from collections import Counter
+
+    tensor_counter = Counter({v: unvariableify(v).numel * unvariableify(v).dtype.bytes for v in all_rematerializable})
+    print(f"DEBUG: out of {len(saved_for_bw)} sfb, {len(all_rematerializable)} are recomputable")
+    print(f"DEBUG: can save up to {tensor_counter.total()*1e-6:.3f} MB")
+    top_one_third = len(saved_for_bw) // 3
+    print(f"DEBUG: remat {top_one_third} tensors (top one third)")
```
IvanYashchuk, line 3775: If you'd like to keep these messages they should be printed through a debug logger.
```suggestion
    top_one_third = len(saved_for_bw) // 3
```

thunder/core/transforms.py
```diff
@@ -3736,9 +3737,99 @@ def backward_fn(saved_for_backward, cotangents):
...
+
+    # Select the tensors that will be rematerialized
+    rematerializable = {k for k, v in tensor_counter.most_common(top_one_third)}
```
IvanYashchuk, line 3778: Have you thought of ways this could be exposed to developers to be made configurable?
What do you think about a callback passed with `get_compile_option`?

thunder/core/transforms.py
```diff
@@ -3736,9 +3737,99 @@ def backward_fn(saved_for_backward, cotangents):
...
+        if idx == 4:
+            new_unpack = prims.unpack_sequence.bind(*unpack_args, output=new_saved_for_backward)
+            new_bwd_trace.bound_symbols.append(new_unpack)
+        elif idx == 6:
+            new_bwd_trace.bound_symbols.extend(producers)
+            new_bwd_trace.bound_symbols.append(bsym)
+        else:
+            new_bwd_trace.bound_symbols.append(bsym)
+
+    new_bwd_trace.args = [(new_saved_for_backward, fwd_trace.output[1][1]), *bwd_trace.args[1:]]
```
IvanYashchuk, line 3820: How difficult was it to figure out the correct "preamble" of the backward trace? Is there any learning that you could add as a comment for future readers about why it's done as it is?

IvanYashchuk (changes requested):
While the pull request is marked as ready for review, it lacks tests for the newly introduced behavior.
It's nice that the new logic is isolated to one function `recompute_saved_for_backward`. However, to ensure the robustness of this change and prevent unexpected breakages, please add tests that verify the expected behavior of this function on a small test case.

thunder/core/transforms.py
riccardofelluga: Compile options are said to strictly be strings, I've added an external argument that sets the callable in the compile data

thunder/__init__.py
```diff
@@ -260,6 +261,7 @@ def jit(
     disable_torch_autograd: bool = False,  # TODO Revisit this UX for RC1
     transforms: list[Transform] | None = None,
     record_history: bool = False,
+    recomputation_policy: Callable[[set[Variable]], set[Variable]] | None = None,
```
t-vi, line 264: I still firmly object to passing callables to jit here. Sorry, it is terrible UX, and we can add features with better UX.

t-vi (changes requested):

thunder/core/transforms.py
IvanYashchuk: > Compile options are said to strictly be strings

What do you mean? Most of other existing compile option pass a boolean, not strings. What have you tried and what kind of error you encountered?

thunder/__init__.py
t-vi: My suggestion has been to have a transform that adds a proxytag `BACKWARD_REMATERIALIZE` (or what you want to call it) to proxies in `transform_pre_prologue` (note that it is fine to start with hardcoding the place where the rematerialization happens). Personally, I'd start with one that tags liberally, but if you want that to be with a callable that returns which to tag, cool.
I'm not trying to ask for a lot of work here, I would expect this 15-20 line addition to this PR.

This would immediately open the opportunity to add the tag for implementing checkpoint, too. Have pre-defined heuristics as separate transforms etc.

I have not been a good advocate of using transforms, but I would like to keep us working with traces and transforms (and executors) as much as possible and avoid side-channeling information through `jit(...)`.

IvanYashchuk: > I would expect this to be a 15-20 line addition to this PR.

If that's not a lot of work, could you please provide a concrete example by opening a pull request to Riccardo's branch. It would help both me and Riccardo to understand how to use what you're proposing here and in other future similar cases.

riccardofelluga: I see your point, however I don't think the UX is something that we should put before functionality at this stage. With that said, I'll get back on this and implement the requested changes.

lantiga: @riccardofelluga rather than pure UX (which we care about) it's about not letting special cases percolate into jit core, which will make things more gnarly down the road.

Also, making sure that the core has enough mechanisms to allow this PR to live in a transform + executor is the place we want to be. We should touch the core only to add enable such mechanisms and extension points.

t-vi: It turns out that the trace->callable->transform->trace of the autograd transforms drops the tags, so it's not easy and I'm withdrawing my objection. I do dislike the API, but it does not help.

riccardofelluga: to mitigate, I've hidden away the flag under compile options for now

t-vi:
@IvanYashchuk You still have your changes requested active, what is the state from your POV?

thunder/core/transforms.py
IvanYashchuk: This line is about keys, which are the names of options. A callback is a value here. I see that you've figured this out.

IvanYashchuk (commented):
Before merging we need to make sure the benchmark_litgpt.py file with `--checkpoint_activations=True` works.

Could you please apply the following patch:
```diff
diff --git a/thunder/benchmarks/benchmark_litgpt.py b/thunder/benchmarks/benchmark_litgpt.py
index f63a66f5..e65edf98 100644
--- a/thunder/benchmarks/benchmark_litgpt.py
+++ b/thunder/benchmarks/benchmark_litgpt.py
@@ -346,11 +346,6 @@ class Benchmark_litGPT:
                 self.global_batch_size % self.micro_batch_size * world_size == 0
             ), f"Global Batch Size {self.global_batch_size} should be a multiple Micro Batch Size {self.micro_batch_size} * World Size {world_size}."
 
-        if self.checkpoint_activations and "thunder" in self.compile:
-            warnings.warn(
-                "Activations checkpointing is configured, but Thunder does not support checkpointing. Checkpointing will be ignored."
-            )
-            self.checkpoint_activations = False
         self.skip_data_sync = skip_data_sync
 
         # Profiling Args
@@ -534,6 +529,9 @@ class Benchmark_litGPT:
         return model
 
     def setup_activation_checkpointing(self):
+        if "thunder" in self.compile:
+            # checkpointing is an option to thunder.jit
+            return
         if any(isinstance(mod, CheckpointWrapper) for mod in self.model.modules()):
             warnings.warn(
                 "FSDP checkpointing is configured, but the model already contains checkpointed layers."
@@ -569,6 +567,9 @@ class Benchmark_litGPT:
 
                 executors.insert(0, transformer_engine_ex)
 
+            jit_options = {
+                "enable_saved_for_backward_recomputation": self.checkpoint_activations,
+            }
             if "dynamo" in self.compile:
                 if self.distributed_mode == "fsdp2":
                     print("Resetting cache size for when fsdp2 and using thunder as backend torch.compile")
@@ -588,14 +589,14 @@ class Benchmark_litGPT:
                         raise ValueError(
                             "TransformerEngine executor cannot be used as an executor of Thunder when Thunder is used as torch.compile backend"
                         )
-                backend = ThunderCompiler(executors=executors)
+                backend = ThunderCompiler(executors=executors, **jit_options)
                 # Because Lightning Fabric is imported in this script it monkey patches the torch.compile function
                 # https://github.com/Lightning-AI/pytorch-lightning/blob/828fd998961f6a60f92c35254bb94d6e049ad069/src/lightning/fabric/wrappers.py#L421
                 # using __wrapped__ to access the original torch.compile function did not work
                 # so we are using the lower level torch._dynamo.optimize function
                 model = torch._dynamo.optimize(backend=backend)(model)
             else:
-                model = thunder.jit(model, executors=executors)
+                model = thunder.jit(model, executors=executors, **jit_options)
 
         elif self.compile != "eager":
             raise ValueError(f"Invalid compile option: {self.compile}")
diff --git a/thunder/core/rematerialization.py b/thunder/core/rematerialization.py
index 7db7b88f..3c64737e 100644
--- a/thunder/core/rematerialization.py
+++ b/thunder/core/rematerialization.py
@@ -163,6 +163,11 @@ def apply_rematerialization_for_consumer(
         filter(lambda x: x.name not in map(lambda x: x.name, new_consumer_args), consumer.args)
     )
 
+    # In the case where there are no tensors to rematerialize it is
+    # possible to terminate early and return the consumer as it was.
+    if not rematerialized_inputs:
+        return consumer
+
     # Construct a temporary Trace object with subsymbols from the producer.
     trace = TraceCtx(None)
     trace.bound_symbols = producer.subsymbols
```
and run
```
python thunder/benchmarks/benchmark_litgpt.py --model_name Llama-3-8B --compile dynamo+thunder --n_layers=2 --checkpoint_activations=True
```
Would you be able to reproduce the following assert?
```py
  File "/home/iyashchuk/dev/lightning-thunder/thunder/executors/torch_autograd.py", line 205, in split_forward_backward
    assert bw_trace.bound_symbols[4].args[0].name == "C0"
AssertionError
```
With this command I get an assertion from nvFuser, could you please try with the newest one and investigate what's causing the failure?
```
python thunder/benchmarks/benchmark_litgpt.py --model_name Llama-3-8B --compile thunder --n_layers=2 --checkpoint_activations=True
```

thunder/common.py
```diff
@@ -247,7 +247,6 @@ def __init__(
         self.disable_torch_autograd_support = disable_torch_autograd_support
         self.debug_log = debug_log
 
-        # TODO Consider validating that this dict has exclusively string keys
```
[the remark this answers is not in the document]
riccardofelluga: Regarding the addition to the rematerialization file in your patch above, it shouldn't be needed as the patch #1136 seems to have solved the problem stating Kshiteej.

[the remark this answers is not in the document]
riccardofelluga: On the latest image, both the runs you suggested result in an nvFuser crash. I'll be filing an issue there and I've added the patch in this PR

riccardofelluga:
```shell
  File "/opt/pytorch/lightning-thunder/thunder/__init__.py", line 624, in get_computation_and_inputs
    computation_trc, backward_trc = split_forward_backward(computation_trc, cd, cs, *inps)
  File "/opt/pytorch/lightning-thunder/thunder/executors/torch_autograd.py", line 205, in split_forward_backward
    assert bw_trace.bound_symbols[4].args[0].name == "C0"
AssertionError
```

@IvanYashchuk the above assertion error is cause by a name clash when registering the unpack operation. More specifically this does not always happen due to the fact that the C0 symbol is created here https://github.com/Lightning-AI/lightning-thunder/blob/0073731b8a7290fcf86fcd992c5763b353ef5d65/thunder/core/transforms.py#L2888-L2905 by tracing, creating a new trace, discarding trace meta and copying the bound symbols into the main backward trace. Doing this does not take into account the already existing names in the trace and does not add the new names to the `trace.name` set. 

From here, when registering the new unpack, by default "C0" name is available since it's not registered in the names. However, in the fusion rematerialization logic, the remat trace is created by a `from_trace` call that copies the trace meta(including names). https://github.com/Lightning-AI/lightning-thunder/blob/0073731b8a7290fcf86fcd992c5763b353ef5d65/thunder/core/rematerialization.py#L566 
 https://github.com/Lightning-AI/lightning-thunder/blob/0073731b8a7290fcf86fcd992c5763b353ef5d65/thunder/core/trace.py#L504-L511

 If we try to register the "C0" CollectionProxy after this pass, the assert fails as "C0" is found.

thunder/core/transforms.py
```diff
@@ -3065,4 +3068,102 @@ def backward_fn(saved_for_backward, cotangents):
...
+    new_fwd_trace.bound_symbols = fwd_trace.bound_symbols.copy()
+    new_return_args = (fwd_trace.output[0], (new_saved_for_backward, fwd_trace.output[1][1]))
+    new_fwd_trace.bound_symbols[-1] = prims.python_return.bind(*new_return_args, output=())
+
+    new_bwd_trace = from_trace(bwd_trace)
+
+    # In cases where C0 name is carried from previous trace it must be removed
+    # as the proxy needs to register with that specific name to follow the backward
+    # trace standard signature.
+    if "C0" in new_bwd_trace.names:
```
IvanYashchuk, line 3134: There's a special method for this use case: `.discard()`.

IvanYashchuk (approved):
Thank you for adding the tests and fixing the assert error!

thunder/core/transforms.py
```diff
@@ -3073,4 +3076,100 @@ def backward_fn(saved_for_backward, cotangents):
     _update_backward_with_new_saved_for_backward(backward_trace, only_used_bw_saved_for_backward)
     forward_trace.set_provenance(TraceProvenance("Augmented forward pass"))
     backward_trace.set_provenance(TraceProvenance("Backward pass"))
+
+    enable_saved_for_backward_recomputation: None | bool = get_compile_option(
```
crcrpar, line 3080: QQ: Could you tell me why this allows `None`?

thunder/core/transforms.py
```diff
@@ -3073,4 +3076,100 @@ def backward_fn(saved_for_backward, cotangents):
...
+
+    saved_for_bw = get_saved_for_backward_tensors(fwd_trace)
+    fwd_trace_args = {variableify(j) for j in fwd_trace.args}
+    old_saved_for_bwd = {variableify(j) for j in saved_for_bw}
+
+    all_rematerializable = old_saved_for_bwd - fwd_trace_args
+
+    remat_policy: None | Callable[[set[Variable]], set[Variable]] = get_compile_option(
+        "recomputation_policy",
+        "A callable that accepts a set of variables and returns a set of the variables that are allowed to be recomputed from the forward in the backward trace. The compile option `enable_saved_for_backward_recomputation` needs to be true for this policy to take effect.",
```
crcrpar, line 3109: How about clarifying the behavior when none provided as remat policy?

t-vi (approved):
Thank you @riccardofelluga and thank you @kshitij12345 @crcrpar @IvanYashchuk 
Thank you @IvanYashchuk for looking at this in your vacation.
So I still would want to fix this to move to tagging rather than the policy thing really soon (and I think it would work), but let's move this forward.

thunder/core/transforms.py
riccardofelluga: When the flag is not used this results in a `None` you're right tho that it could be initialized with a bool default value, say `False`

PR #1002, opened 2024-08-20
benchmark_litgpt.py: Start building executors list from the default ones

IvanYashchuk:
`thunder/benchmarks/benchmark_litgpt.py` was ignoring the default executor list in Thunder. With this PR the default list is respected and optional executors are added as first in the priority queue.

Thanks to @mpatel31415 for noticing this (https://github.com/Lightning-AI/lightning-thunder/pull/974#issuecomment-2296468811).

cc @crcrpar

PR #1006, opened 2024-08-21
benchmark_litgpt.py: fix executors to be a list; fix the order of executors for sdpa

IvanYashchuk:
A follow-up fix for https://github.com/Lightning-AI/lightning-thunder/pull/1002. `default_executors` returns a tuple that can't be modified. I'm sorry I haven't tested the code properly.

In addition, this PR removes "cudnn" reading from the compile option because it's part of the default executors.

cc @crcrpar

### DIFF
--- a/thunder/benchmarks/benchmark_litgpt.py
+++ b/thunder/benchmarks/benchmark_litgpt.py
@@ -401,14 +401,6 @@ class Benchmark_litGPT:
                 from thunder.executors.torch_compile import torch_compile_ex
 
                 executors.insert(0, torch_compile_ex)
-            if "cudnn" in self.compile:
-                from thunder.executors.cudnnex import cudnn_ex
-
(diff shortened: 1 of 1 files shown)


t-vi (approved):
Thank you @IvanYashchuk

thunder/benchmarks/benchmark_litgpt.py
```diff
@@ -401,14 +401,6 @@ def setup_compile(self, model):
                 from thunder.executors.torch_compile import torch_compile_ex
 
                 executors.insert(0, torch_compile_ex)
-            if "cudnn" in self.compile:
-                from thunder.executors.cudnnex import cudnn_ex
-
-                executors.insert(0, cudnn_ex)
-            else:
-                from thunder.executors.sdpaex import sdpa_ex
```
t-vi, line 409: So this is intentionally gone, too?

PR #1007, opened 2024-08-21
feat: add fullgraph argument for torch compile

### LINKED ISSUE #281: Expose `torch.compile` arguments as compile options
### Motivation

The `torch.compile` call is internal and has `fullgraph=True` hardcoded. It would be useful to allow customizing it, especially after #140 lands.

https://github.com/Lightning-AI/lightning-thunder/blob/3a2b27e9b38e639133913c42c03acdf0e136ec4f/thunder/executors/torch_compile.py#L84

### Pitch

```python

... = thunder.jit(..., executors=[torch_compile_ex], torch_compile_fullgraph=True, torch_compile_backend="reduce-overhead")
```

Use `get_compile_option` to allow customizing `fullgraph=` and `backend=`

https://github.com/Lightning-AI/lightning-thunder/blob/3a2b27e9b38e639133913c42c03acdf0e136ec4f/thunder/core/compile_data.py#L57

cc @apaz-cli

k223kim:
<details>

</details>

## What does this PR do?

Fixes #281. As this is a fairly old issue, I am not sure if this is needed. Also, I am not sure how to add a test case for this.

## PR review

## Did you have fun?

### DIFF
--- a/thunder/executors/torch_compile.py
+++ b/thunder/executors/torch_compile.py
@@ -16,6 +16,7 @@ from thunder.core.pytree import tree_flatten
 from thunder.executors.passes import update_fusion_call_ctx
 from thunder.executors.utils import Region
 from thunder.extend import FusionExecutor, register_executor, ImplInfo
+from thunder.core.compile_data import get_compile_option
 
 _TORCH_GREATER_EQUAL_2_3 = compare_version("torch", operator.ge, "2.3.0", use_base_version=True)
 
@@ -84,7 +85,12 @@ def make_compiled(
     # torch.compile executor"
     torch_trace = trace(inline_trace=False)(torch_interpreted_func, *sorted_unique_inputs)
     trace_callable = torch_trace.python_callable(include_decorators=False)
-    compiled_func = torch.compile(trace_callable, fullgraph=True)
+    torch_compile_fullgraph: None | bool = get_compile_option(
+        "torch_compile_fullgraph", "Whether to enable `fullgraph` from `torch.compile`. Defaults to `True`."
+    )
+    if torch_compile_fullgraph is None:
+        torch_compile_fullgraph = True
+    compiled_func = torch.compile(trace_callable, fullgraph=torch_compile_fullgraph)
     # For each of `@torch.no_grad(), and `torch.autocast(device_type="cpu"|"cuda")` torch.compile
     # create caches with a guard for the wrapped function. Since the torch.compile caches are per code object, not
     # frame, all the dynamic copies of these context managers share the same code cache.


t-vi (commented):
Looks good, would we want to add a test, say for llama2-like to run with this to test_networks or so?

k223kim:
I could, however, I don't think that will allow us to check if `fullgraph` has been modified. What do you think? Is there a way to check which parameter has been passed to `torch.compile` that I am not aware of?

t-vi:
Right, I had hoped we could get it from the compiler config or so, but apparently it is `None`...
```
import torch, thunder

def fn(a):
     return a * 2 + 1
jfn = thunder.jit(fn, executors=[thunder.executors.torch_compile.torch_compile_ex])
jfn(torch.randn(2, 2))

lt = thunder.last_traces(jfn)[-1]
tc = lt.python_ctx()['TorchCompile0']
cc = tc.__closure__[0].cell_contents.get_compiler_config()
``

So let's keep it as is.

**Comment by kshitij12345** [comment]
I think a simple way to test `fullgraph=True` would be to create a graph with graph break. So, compiling this should throw an exception, which we can verify with `pytest.raises`.

```python
import torch

def foo():
    print("This will lead to graph break")

torch.compile(foo, fullgraph=True)()  # This will throw `torch._dynamo.exc.Unsupported`

```

k223kim:
@kshitij12345 Hey! Thanks for the suggestion. I did try this but I don't think it'll will throw an error within `thunder`:
```
import torch
import thunder

def foo():
    print("This will lead to graph break")

torch.compile(foo, fullgraph=True)()  # This will throw `torch._dynamo.exc.Unsupported` 
jfn = thunder.jit(foo, executors=[torch_compile_ex], torch_compile_fullgraph=True, torch_compile_backend="reduce-overhead")
jfn()# this will not throw an error
```
Which makes it difficult to see what parameter has been passed to `torch.compile`.. Do you have a different suggestion?

kshitij12345:
> Which makes it difficult to see what parameter has been passed to torch.compile.. Do you have a different suggestion?

You are right, I forgot that the `torch.compile` is called on the trace from thunder, so it will receive only the parts it claims allowing it to compile the full graph it receives.

Thanks for looking trying this!

thunder/executors/torch_compile.py
```diff
@@ -84,7 +85,12 @@ def torch_interpreted_func(*args):
     # torch.compile executor"
     torch_trace = trace(inline_trace=False)(torch_interpreted_func, *sorted_unique_inputs)
     trace_callable = torch_trace.python_callable(include_decorators=False)
-    compiled_func = torch.compile(trace_callable, fullgraph=True)
+    torch_compile_fullgraph: None | bool = get_compile_option(
```
mruberry, line 88: Adding a compile option for this is a cool debugging idea.

Whether torch.compile generates a single graph or not for what we export to it may be considered an implementation detail, but this setting could also be useful to set for debugging.

thunder/executors/torch_compile.py
```diff
@@ -84,7 +85,12 @@ def torch_interpreted_func(*args):
...
+        "torch_compile_fullgraph", "Whether to enable `fullgraph` from `torch.compile`. Defaults to `True`."
+    )
+    if torch_compile_fullgraph is None:
+        torch_compile_fullgraph = True
```
mruberry, line 92: I would default it to False because the effect of setting this to True is just that some programs may throw an error instead of executing, but I think the historic value is True so this just preserves it?

k223kim: I do think it is better to default it to False, however, I am not sure if there were any specific reasons on why we kept this as True in the first place.

t-vi: The hypothesis of the executor and the default to True is to only send things to torch compile that do not cause graph breaks and if we do, we want to investigate and improve the fusion logic.

mruberry (approved):
Seems like a cool option

thunder/executors/torch_compile.py
mruberry: I don't know if causing a graph break is an issue or not, but passing `True` here doesn't seem to hurt anything

t-vi: That said, we might add a
```
try:
...
except torch._dynamo.exc.Unsupported:
    print("""The torch compile executor got a segment that it cannot process without graph breaks, 
you can enable graph breaks by passing `torch_compile_fullgraph=False` to thunder.jit""")
    raise
```

PR #1008, opened 2024-08-21
Save benchmark results in asv format

riccardofelluga:
## What does this PR do?

After discussing around some ways to implement a partial fix for #183 and #224, this option seems like a good compromise for the moment. Saving the results in this format allows for the benchmarks to be visually compared using the [Airspeed velocity framework](https://asv.readthedocs.io/en/v0.6.4/index.html).

I hardcoded the output path but let me know if it's better to make it depend on an input flag.


thunder/benchmarks/conftest.py
```diff
@@ -0,0 +1,52 @@
...
+        osType=f"{uname.system} {uname.release}",
+        pythonVer=python_version,
+        commitHash=commit_hash,
+        commitTime=commit_time,
+        cpuType=uname.processor,
+        arch=uname.machine,
+        ram=memory_size,
+    )
+
+    db = ASVDb(dbDir="./asv", repo=repo_name, branches=[current_branch])
```
crcrpar, line 41: I'm wondering if it'd do better with an option to specify `dbDir`, e.g., via an env var of `THUNDER_BENCH_DIR`

thunder/benchmarks/conftest.py
```diff
@@ -0,0 +1,52 @@
+import platform
+import psutil
+from typing import Any
+
+from asvdb import utils, BenchmarkInfo, BenchmarkResult, ASVDb
```
mruberry, line 5: Would you add a comment to this file explaining what it does and how it works, and then a comment for each function describing what it does

thunder/benchmarks/conftest.py
```diff
@@ -0,0 +1,52 @@
...
+    sane_params = []
+    for k, v in benchmark_params:
+        if k == "executor":
+            sane_params += [(k, str(v).split()[1])]
+        else:
+            sane_params += [(k, v)]
+    return sane_params
+
+
+def pytest_sessionfinish(session, exitstatus):
```
mruberry, line 18: How would a developer use this functionality? Can we provide some examples of disabling it, enabling it, and using its options (if any) in a comment?

thunder/benchmarks/conftest.py
mruberry: Maybe it could be a pytest command line option, too?

thunder/benchmarks/conftest.py
```diff
@@ -0,0 +1,52 @@
...
+    for bench in benchmarks:
+        name = bench.name.split("[")[0]
+        params_pairs = sanitize_params(bench.params.items())
+        result = BenchmarkResult(
+            funcName=name,
+            argNameValuePairs=params_pairs,
+            result=bench.stats.median * 1e6,
+            unit="µseconds",
+        )
+        db.addResult(bench_info, result)
```
mruberry, line 52: Can pytest print the location of written files when it completes, so developers can just click the link to see them?

thunder/benchmarks/conftest.py
xwang233: I think asv is an option to **store** benchmark results and is not required to **run** benchmarks. We should have some logic like

```python
try:
  from asvdb import ...
  HAS_ASVDB = True
except ImportError:
  print('asvdb is not imported, results wont be stored with asv')
  HAS_ASVDB = Fasle

# benchmark runs ...

if HAS_ASVDB:
  db = ASVDB(...)
  db.addResult(...)
```

thunder/benchmarks/conftest.py
riccardofelluga: Unfortunately it's pretty cumbersome to print stuff out of pytest. I've made it mandatory to specify the directory so the user knows where the files are saved. What is left to do there is to use `asv` cli tool to produce the visualization out of the folder which is pretty simple

IvanYashchuk:
Airspeed is a great tool and it can be handy for building open-source continuous benchmarking. For now, this PR enables only the manual part of the process. Could you please describe and provide an example of the workflow that this PR enables? I'd like to see what commands you would recommend people run and what kind of visual comparison is possible to get with this tool. I'm afraid without a clear guide it might be difficult to adopt developer-users for this feature.

riccardofelluga:
@IvanYashchuk That is true, let me clarify the usage.

## What is this about?
As of writing there is no easy way to discover if a pref regression has been introduced by the latest commit in Thunder. The option that developers have available at the moment is to manually benchmark `targets.py` and then look at the results in the terminal. The second option available at the moment is to extract the pytest-benchmark json file and look into the values manually or extract them with specialized code.

What this PR proposes is to mainstream Thunder performance tracking. In popular frameworks like numpy and scipy, performance is tracked using the [Airspeed Velocity framework](https://asv.readthedocs.io/en/v0.6.4/index.html). This framework includes a multitude of things that facilitate writing and tracking benchmarks. Not only it provides a way to write tests that are then used to track performance, but also it comprises of a visualization tool that given the test results creates a webpage that plots the results and allows for comparisons. Check out the numpy page [here](https://pv.github.io/numpy-bench/) and [here](https://pv.github.io/numpy-bench/#bench_reduce.ArgMax.time_argmax).

image

image

## Why this PR?

Now that we established what asv is about and why, it is possible to speak about this PR. The changes introduced here allow for integrating asv in the already existing ecosystem of benchmarks in Thunder. By manually setting the output path as env variable `THUNDER_BENCH_DIR` or by specifying it with the `--asv_bench_dir` command line flag, the logic in this PR will save the benchmark results and create a asv database in the specified directory. 

So far this is not that different from just dumping the json file out of pytest-benchmark. However, by using asv tooling it is now possible to generate the same website as numpys or scipy, with the Thunder benchmark results.

```shell
pytest thunder/benchmarks/targets.py --asv_bench_dir="./asv-bench" 

cd asv-bench
asv update
asv publish
```

Now in the asv-bench directory there will be an html folder with a static website displaying the results in the same form as the attached screenshots above.

t-vi (approved):
Thank you @riccardofelluga @IvanYashchuk @crcrpar @mruberry @xwang233

IvanYashchuk:
> As of writing there is no easy way to discover if a perf regression has been introduced by the latest commit in Thunder.

We run the benchmarks nightly and create NVIDIA-internal markdown reports. It's not accurate to say there's no easy way because this is how we discover performance regressions.

> Now in the asv-bench directory there will be an html folder with a static website displaying the results in the same form as the attached screenshots above.

Can you please edit your comment and include screenshots from Thunder benchmarks instead of SciPy? How should people use these generated web pages? And add a link to your comment in the PR description for better discovery.

PR #1011, opened 2024-08-21
Reverse transform of `state_dict` per submodule

crcrpar:
## What does this PR do?

Implements reverse transform to dump a state dict of unsharded parameters, for FSDP.

successor of https://github.com/Lightning-AI/lightning-thunder/pull/913

### DIFF
--- a/thunder/core/module.py
+++ b/thunder/core/module.py
@@ -244,11 +244,12 @@ class ThunderModule(pytorch.nn.Module):
 
         """
         module_names = {name for name, _ in self._model.named_modules()}
-        state_dict = self.state_dict(destination=destination, prefix=prefix, keep_vars=keep_vars)
-        state_dict_per_submodule = _convert_state_dict_to_per_module(state_dict, module_names)
+        state_dict_per_submodule = _convert_state_dict_to_per_module(self.state_dict(), module_names)
 
-        _state_dict_keys = list(state_dict.keys())
-        cur_idx: int = 0
+        if destination is None:
+            destination = collections.OrderedDict()
+            destination._metadata = collections.OrderedDict()
+
         transform: Transform
         for submodule_name, submodule_state_dict in state_dict_per_submodule.items():
             for transform in reversed(self._lc_transforms):
@@ -257,11 +258,8 @@ class ThunderModule(pytorch.nn.Module):
                     submodule_name,
                     submodule_state_dict,
                 )
-            state_dict_per_submodule[submodule_name] = submodule_state_dict
-            for v in submodule_state_dict.values():
-                state_dict[_state_dict_keys[cur_idx]] = v
-                cur_idx += 1
-        return state_dict
+            destination.update({f"{prefix}{submodule_name}.{k}": v for k, v in submodule_state_dict.items()})
+        return destination
 
     def load_state_dict(self, state_dict: Mapping[str, Any], strict: bool = True, assign: bool = False):
         """Loads the state dict to a transformed module.
--- a/thunder/core/module.py
+++ b/thunder/core/module.py
@@ -242,7 +242,11 @@ class ThunderModule(pytorch.nn.Module):
         prefix: str = "",
         keep_vars: bool = False,
     ) -> dict[str, Any]:
-        """Returns the state dict of the transformed :class:`ThunderModule`.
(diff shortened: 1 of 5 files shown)


thunder/core/module.py
```diff
@@ -215,6 +224,45 @@ def state_dict(self, *, destination=None, prefix="", keep_vars=False):
...
+            destination: if given, use this mutable mapping as the dict container.
+            prefix: a prefix for the keys.
+            keep_vars: do not detach
+
+        """
+        module_names = {name for name, _ in self._model.named_modules()}
+        state_dict = self.state_dict(destination=destination, prefix=prefix, keep_vars=keep_vars)
+        state_dict_per_submodule = _convert_state_dict_to_per_module(state_dict, module_names)
+
+        _state_dict_keys = list(state_dict.keys())
```
t-vi, line 250: I might be missing something, I think it would be simpler if we took/created the destination and then used `transformed_state_dict = self.state_dict(...)` (so not populate this into the destination) and then finally used something like
```
destination.update((f"{prefix}{submodule_prefix}{k}", v) for k, v in submodule_state_dict.items())
```
The background is that Transforms like Quantization, LoRA etc. will change what keys are in the state dict.
I don't think the preserving order in the state dict is important.

thunder/distributed/transforms/fsdp_v2.py
```diff
@@ -71,7 +69,7 @@ def __call__(self, bsym: BoundSymbol) -> VISIT_TYPE:
         return VISIT_TYPE.REPLACE
 
 
-# When the user calls fsdp(jitted_module), or applies this Transform direcly, it does the following
+# When the user calls fsdp(jitted_module), or applies this Transform directly, it does the following
```
t-vi, line 72: Thank you for fixing all these.

thunder/distributed/transforms/fsdp_v2.py
```diff
@@ -263,6 +259,36 @@ def transform_state_dict_for_submodule(
...
 
+    def reverse_transform_state_dict_for_submodule(
+        self,
+        model: ThunderModule,
+        submodule_name: str,
+        state_dict: dict[str, Any],
+    ) -> dict[str, Any]:
+        from thunder.executors.torchex import _all_gather_prim_impl
+
+        for name, tensor in state_dict.items():
```
t-vi, line 270: I wonder if we should make a new dict (never thought about the implications, though, so just a gut feeling).

thunder/core/module.py
```diff
@@ -215,6 +224,45 @@ def state_dict(self, *, destination=None, prefix="", keep_vars=False):
...
+        returns a state dict of unsharded tensors.
+
+        Args:
+            destination: if given, use this mutable mapping as the dict container.
+            prefix: a prefix for the keys.
+            keep_vars: do not detach
+
+        """
+        module_names = {name for name, _ in self._model.named_modules()}
+        state_dict = self.state_dict(destination=destination, prefix=prefix, keep_vars=keep_vars)
```
t-vi, line 247: I think we don't want the prefix here but only apply it when we collect the final state dict from the submodules.
I don't know if we can reasonably support keep vars in `original_state_dict`.

thunder/tests/distributed/test_fsdp.py
```diff
@@ -685,6 +685,26 @@ def test_load_original_state_dict(self):
 
         torch.testing.assert_close(y_1, y_2)
 
+    @pytest.mark.skipif(torch.cuda.device_count() < 2, reason="Requires 2 devices")
+    def test_original_state_dict(self):
```
t-vi, line 689: Great to have a test to go with it. :)

thunder/distributed/transforms/fsdp_v2.py
crcrpar: My take is that reverse transform is closed in `original_state_dict` so it wouldn't matter a lot whether creating a new dict or direct edit of a dict

t-vi: Fair enough, let's keep as is and see what happens.

t-vi (approved):
Exiting new aspect of transforms. Thank you @crcrpar !

PR #1019, opened 2024-08-22
Improve memory usage for stablecode-completion-alpha-3b

IvanYashchuk:
Running this model on DGX-H100 with FSDP was giving an OOM error. With the proposed change peak memory usage is now 67.79GB.

This works by fusing RoPE and GELU into one fusion region and allowing the rematerialization pass to recompute GELU in the backward fusion region.

More details in https://github.com/Lightning-AI/lightning-thunder/issues/246#issuecomment-2302121789.

The following command was used to check the OOM error.
```
torchrun --nproc_per_node=8 thunder/benchmarks/benchmark_litgpt.py --model_name stablecode-completion-alpha-3b --compile=thunder --distributed_mode=fsdp
```

cc @apaz-cli

### DIFF
--- a/thunder/executors/torch_compile.py
+++ b/thunder/executors/torch_compile.py
@@ -229,6 +229,10 @@ supported_ops = {
     prims.reshape.id,
     prims.slice_prim.id,
     prims.transpose.id,
+    # div and erf are used in GELU and are fused horizontally with RoPE when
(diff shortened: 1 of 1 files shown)


t-vi (approved):
Thank you @IvanYashchuk

PR #1020, opened 2024-08-22
Return arguments and swap them with outputs of copies on them

### LINKED ISSUE #912: `fuse_bound_symbols` (toposort on bsyms) can put `return` before `copy_` on arguments

### LINKED ISSUE #930: fuse_bound_symbols skip the side effect of Tensor.copy_

shino16:
Fixes #912. Fixes #930.

To guarantee that the topological sort on bsyms places the return statement after copies on the arguments, we return the outputs of those copies together with the real return value. To do this, we make a trace that directly returns all the arguments and swap them with their respective copy outputs.

```py
import torch
import thunder

@thunder.jit
def f(a, b):
  a += b
  return b

a = torch.tensor(0.0, device='cuda')
b = torch.tensor(0.0, device='cuda')
jitted(a, b)
print(*thunder.last_traces(jitted), sep="\n\n")
```
```py
def computation(a, b):
  t1 = ltorch.add_(a, b, alpha=None)  # t1: "cuda:0 f32[]"
  return b

# Constructed by Return arguments to track copies onto them
def computation(a, b):
  t1 = ltorch.add_(a, b, alpha=None)  # t1: "cuda:0 f32[]"
  return {'output': b, 'flat_args': [a, b]}

# Constructed by Functionalize in-place ops
def computation(a, b):
  t0 = ltorch.add(a, b, alpha=None)  # t0: "cuda:0 f32[]"
  t1 = prims.copy_(t0, a)  # t1: "cuda:0 f32[]"
  return {'output': b, 'flat_args': [t1, b]}
```

When autograd is enabled, we used to wrap the return value as `{'output': OUTPUT, 'flat_args': ..., 'flat_output' ...}`. Now we just add `flat_output` to the return value that is already augmented.

```py
import torch
import thunder

def f(a, b):
  a += b
  return b

a = torch.tensor(0.0, device='cuda', requires_grad=True)
b = torch.tensor(0.0, device='cuda')
jitted = thunder.jit(f)
jitted(a, b)
print(*thunder.last_traces(jitted), sep="\n\n")
```
```py
def computation(a, b):
  t1 = ltorch.add_(a, b, alpha=None)  # t1: "cuda:0 f32[]"
  return b

# Constructed by Return arguments to track copies onto them
def computation(a, b):
  t1 = ltorch.add_(a, b, alpha=None)  # t1: "cuda:0 f32[]"
  return {'output': b, 'flat_args': [a, b]}

# Constructed by Functionalize in-place ops
def computation(a, b):
  t0 = ltorch.add(a, b, alpha=None)  # t0: "cuda:0 f32[]"
  t1 = prims.copy_(t0, a)  # t1: "cuda:0 f32[]"
  return {'output': b, 'flat_args': [t1, b]}

# Constructed by Augmented forward pass
def augmented_forward_fn(a, b):
  t0 = ltorch.add(a, b, alpha=None)  # t0: "cuda:0 f32[]"
  t1 = prims.copy_(t0, a)  # t1: "cuda:0 f32[]"
  return {'output': b, 'flat_args': [t1, b], 'flat_output': (t1, b, b)}, ((), ())
```
It is a concern that `augmented_forward_fn` now returns `[t1, b]` as `flat_args` instead of `[a, b]`. Seemingly [what we pass here](https://github.com/Lightning-AI/lightning-thunder/blob/a59b4efb54ea69980fa7fb63ea78923290e93d5a/thunder/__init__.py#L771) [does not really matter](https://github.com/Lightning-AI/lightning-thunder/blob/a59b4efb54ea69980fa7fb63ea78923290e93d5a/thunder/executors/torch_autograd.py#L65-L75).


shino16:
I'll check the CI's results and update the documents tomorrow.

shino16:
The test failure is due to #1029

shino16:
Hi @IvanYashchuk, this is finally ready for review!

thunder/core/functionalization.py
```diff
@@ -566,7 +566,7 @@ def _reshape_bsym_ctor(src: TensorProxy, dst: TensorProxy, trace: Trace) -> tupl
         if bsym in bsym_to_copy_bsyms:
             functionalized_bsyms.extend(bsym_to_copy_bsyms[bsym])
             copy_bsym = functionalized_bsyms[-1]
-            swap_map_for_return[variableify(copy_bsym.flat_proxy_args[0])] = copy_bsym.flat_proxy_args[1]
+            swap_map_for_return[variableify(copy_bsym.flat_proxy_args[0])] = copy_bsym.flat_proxy_outs[0]
```
shino16, line 569: For example, consider a function
```py
def f(x):
  x += 1
  return x
```
On this line, the intermediate trace looks like
```py
t0 = ltorch.add(x, 1)
t1 = prims.copy_(t0, x)  # src=t0, dst=x
return {"output": t0, "flat_args": (t0,)}  # All occurrences of x became t0 due to functionalization
```
Before this PR, `swap_map_for_return` swapped `t0` with `x`, because the returned tensor should point to the same data segment in the memory as the argument `x`.

In the new code, `t0` now gets swapped with `t1`. This forces the topological sort algorithm to order the return statement after the copy. This is okay in terms of correctness, because the return value of `copy_` is exactly the same as the destination tensor.
https://github.com/Lightning-AI/lightning-thunder/blob/d70b10e44244317c462c6068a7b6a44656d5901d/thunder/executors/torchex.py#L2145-L2147

thunder/tests/test_inplace_copy.py
```diff
@@ -125,21 +125,21 @@ def func0(x, y):
 
     def func1(x, y):
         z = x * y
-        thunder.core.prims.copy_(z, x)
-        thunder.core.prims.copy_(y, x)
-        return x
+        o1 = thunder.core.prims.copy_(z, x)
+        o2 = thunder.core.prims.copy_(y, x)
+        return x, o1, o2
```
crcrpar, line 130: qq: why returning copy outputs in the input func? I couldn't see why at glance

shino16, line 130: These `copy_`'s bypasses the name swapping in functionalization, so we have to manually return their outputs so that they will be ordered after `copy_`.

shino16: These copy_'s bypasses the name swapping in functionalization, so we have to manually return their outputs so that they will be ordered after copy_.

IvanYashchuk (approved):
I think it's a good creative solution to the ordering problem. Thank you for working on this!
Keeping the argument variables live until the return statement also coincides with the implicit Python behavior and wouldn't introduce any additional memory overhead.

`trace.output["output"]` also in the non-autograd trace will definitely look weird to newcomers and more explanation in the documentation and tutorials would be needed.

@t-vi, I know you discussed this with @shino16 over Slack, but please review this pull request as it does change how computation traces are represented in the inference type workload.

thunder/__init__.py
```diff
@@ -760,18 +765,19 @@ def fn_(*args, **kwargs) -> Any:
...
                 saved_tensors,
                 saved_other,
-                data_for_autograd["flat_output"],
-                *data_for_autograd["flat_args"],
+                result["flat_output"],
+                *result["flat_args"],
             )
-            result = data_for_autograd["output"]
+
+        result = result["output"]
```
t-vi, line 780: Instead of this change, could we please add

```
if backward_trc is None:
      computation_trace = unwrap_return_value(computation_trace)
```

here:

https://github.com/Lightning-AI/lightning-thunder/blob/3770bf02240cdb6ec86a6d5224807bc6ad39d2a9/thunder/__init__.py#L670-L671

?
That would leave the inference cases alone.

t-vi (changes requested):
It would be cool if we could leave computation_fn return the output.
If the suggested change does run into issues, I'm happy to reconsider.

t-vi (approved):
Thank you. I think there are a few tests to fix yet.

shino16:
Oops, thank you!

thunder/tests/distributed/test_fsdp.py
```diff
@@ -82,17 +82,26 @@ def func(
         b = make_tensor((2, 2), device=device, dtype=torch.float32)
         process_group = c10d.new_group()
         _ = cfunc(a, b, process_group)
-        execution_trace = thunder.last_traces(cfunc)[-2]
-        sorted_execution_trace = sort_waits(execution_trace)
+        traces = thunder.last_traces(cfunc)
+
+        del_last_used_indices = [
+            i for i, trace in enumerate(traces) if "Delete Last Used" in traces.get_provenance().pss
```
shino16, line 88: `last_traces(cfunc)[-2]` used to be the trace just before `del_last_used` is applied. Now we have `Unwrap the return value` trace after `del_last_used`, so indexing by -2 no longer works.

Spelling out `"Delete Last Used"` is not very nice, but this makes the intention easily recognizable.

PR #1022, opened 2024-08-22
feat: rough draft of lora linear

k223kim:
<details>

</details>

## What does this PR do?

Fixes # (issue).

## PR review

## Did you have fun?


thunder/transforms/quantization.py
```diff
@@ -294,3 +295,214 @@ def transform_traces_pre_prologue(self, prologue_trace, computation_trace, epilo
 
         new_computation_trace.set_provenance(thunder.core.trace.TraceProvenance("quant pass"))
         return prologue_trace, new_computation_trace, epilogue_trace
+
+
+class LORATransform(Transform):
```
t-vi, line 300: Should we put LORATransform in its own file?

thunder/transforms/quantization.py
```diff
@@ -294,3 +295,214 @@ def transform_traces_pre_prologue(self, prologue_trace, computation_trace, epilo
...
+    def __init__(
+        self,
+        r: int = 0,
+        lora_alpha: int = 1,
+        lora_dropout: float = 0.0,
+        **kwargs,
```
t-vi, line 306: I don't think we want blanket kwargs here. We just use them in the transform methods because some transforms want additional keyword-args, but for the init it is better to error if we get arguments we don't know.
Maybe we should also make the other args kw-only (by inserting a `*, ` after `self`).
Also, I would likely want to support passing in a list of weight names which we want to create LoRA paths for.

thunder/transforms/quantization.py
```diff
@@ -294,3 +295,214 @@ def transform_traces_pre_prologue(self, prologue_trace, computation_trace, epilo
...
+    def transform_module(self, model: thunder.ThunderModule):
+        self.thunder_module = model
+        shared_names = model._get_shared_names()
+        processed_names = set()
+
+        def convert_linear_submodule(tm, name, r):
+            self.lora_linear_names.add(name)
+            weight_name = f"{name}.weight"
+            processed_copies = shared_names[weight_name] & processed_names
+            if processed_copies:
```
t-vi, line 323: Great handling shared params from the start!

thunder/transforms/quantization.py
```diff
@@ -294,3 +295,214 @@ def transform_traces_pre_prologue(self, prologue_trace, computation_trace, epilo
...
+                tm._overrides_parameters[weight_name] = tm._overrides_parameters[copy_name]
+                tm._overrides_parameters[f"{weight_name}.loraA"] = tm._overrides_parameters[f"{copy_name}.loraA"]
+                tm._overrides_parameters[f"{weight_name}.loraB"] = tm._overrides_parameters[f"{copy_name}.loraB"]
+                processed_names.add(weight_name)
+                return
+
+            w = tm.get_parameter(weight_name)
+            in_features, out_features = w.shape[0], w.shape[1]
+
+            loraA = torch.nn.Parameter(torch.empty((r, in_features)))
```
t-vi, line 335: The new params to be on the same device, of the same dtype (unless it's quantized already, then the one in the quantization records?) as `w` . The easiest way to do this is to use `w.new_empty((r, in_features))`.

thunder/transforms/quantization.py
```diff
@@ -294,3 +295,214 @@ def transform_traces_pre_prologue(self, prologue_trace, computation_trace, epilo
...
+            torch.nn.init.kaiming_uniform_(loraA, a=math.sqrt(5))
```
t-vi, line 336: I think we might move the initialization to a utility function as we will also need to do this when transforming the state dict.

thunder/transforms/quantization.py
```diff
@@ -294,3 +295,214 @@ def transform_traces_pre_prologue(self, prologue_trace, computation_trace, epilo
...
+            if isinstance(submodule, torch.nn.Linear):
+                convert_linear_submodule(model, n, self.r)
+
+    def transform_state_dict_for_submodule(
+        self, model: thunder.ThunderModule, submodule_name: str, state_dict: dict
+    ) -> dict:
+        if submodule_name not in self.lora_linear_names:
+            return state_dict
+
+        weight_name_full = f"{submodule_name}.weight"
```
t-vi, line 365: I don't think we need to do anything here with the weight, we just need to init the new loraA and loraB.
In particular w.to(w.device) is a no-op. (There is something funny in quantization because of the limitation described in  #1111 .)

t-vi: Does the original source code to subtract the lora initialization from the weight btw. (to leave the initial state unchanged in sum)? Then we should do this here and in transform module above.

thunder/transforms/quantization.py
```diff
@@ -294,3 +295,214 @@ def transform_traces_pre_prologue(self, prologue_trace, computation_trace, epilo
...
+
+        return state_dict
+
+    def transform_traces_pre_prologue(self, prologue_trace, computation_trace, epilogue_trace, **kwargs):
+        tm = self.thunder_module
+        from thunder.core.trace import tracectx
+
+        checks = get_checks(prologue_trace)
+
+        # do we need this?
```
t-vi, line 378: You are right, we don't need this as we are not modifying the tensor metadata (in particular dtype) of w.
We should copy the prologue trace

thunder/transforms/quantization.py
```diff
@@ -294,3 +295,214 @@ def transform_traces_pre_prologue(self, prologue_trace, computation_trace, epilo
...
+                    proxy_loraB = thunder.TensorProxy(
+                        name=f"{get_param.output.name}_loraB",
+                        shape=qs["loraB.shape"],
+                        dtype=thunder.dtypes.to_dtype(qs["loraB.dtype"]),
+                        device=thunder.devices.to_device(device),
+                        requires_grad=requires_grad,
+                    )
+                    new_bsyms.append(get_param.sym.bind(get_param.args[0], n_loraA, output=proxy_loraA))
+                    new_bsyms.append(get_param.sym.bind(get_param.args[0], n_loraB, output=proxy_loraB))
+
```
t-vi, line 429: I think I forgot this in quantization,  but it would be cool to add a shape, device, dtype check here.

thunder/transforms/quantization.py
```diff
@@ -294,3 +295,214 @@ def transform_traces_pre_prologue(self, prologue_trace, computation_trace, epilo
...
+            if bsym.sym != prims.unpack_trivial:
+                break
+            new_computation_trace.bound_symbols.append(bsym.from_bsym())
+        new_computation_trace.bound_symbols += new_bindings
+
+        for bsym in bound_symbols[idx:]:
+            if bsym.sym == thunder.torch.linear:
+                n = lora_linear_proxies[bsym.args[1].name]
+                with thunder.core.trace.tracectx(computation_trace):
+                    loraA_transpose_meta = prims.transpose.meta(additional_proxies[f"{n}.loraA"], (1, 0))
```
t-vi, line 472: does it work to call the symbol instead of the meta? 
```
loraA_transpose_meta = prims.transpose(additional_proxies[f"{n}.loraA"], (1, 0))
```
?

thunder/transforms/quantization.py
```diff
@@ -294,3 +295,214 @@ def transform_traces_pre_prologue(self, prologue_trace, computation_trace, epilo
...
+                    loraA_bsym = prims.transpose.bind(
```
t-vi, line 473: I think the bsym should be automatically recorded when you call prims.transpose, no binding needed.

thunder/transforms/quantization.py
```diff
@@ -294,3 +295,214 @@ def transform_traces_pre_prologue(self, prologue_trace, computation_trace, epilo
...
+
+        for idx, bsym in enumerate(bound_symbols):
+            if bsym.sym != prims.unpack_trivial:
+                break
+            new_computation_trace.bound_symbols.append(bsym.from_bsym())
+        new_computation_trace.bound_symbols += new_bindings
+
+        for bsym in bound_symbols[idx:]:
+            if bsym.sym == thunder.torch.linear:
+                n = lora_linear_proxies[bsym.args[1].name]
```
t-vi, line 470: I think we need to check whether the name is in the dict, in case people there are other linears.

thunder/transforms/quantization.py
```diff
@@ -294,3 +295,214 @@ def transform_traces_pre_prologue(self, prologue_trace, computation_trace, epilo
...
+                        (1, 0),
+                        output=(loraA_transpose_meta),
+                    )
+                    loraB_transpose_meta = prims.transpose.meta(additional_proxies[f"{n}.loraB"], (1, 0))
+                    loraB_bsym = prims.transpose.bind(
+                        additional_proxies[f"{n}.loraB"],
+                        (1, 0),
+                        output=(loraB_transpose_meta),
+                    )
+                    lora_meta_ = prims.matmul.meta(bsym.args[0], loraA_bsym.output)
```
t-vi, line 484: would it make sense to call linear instead of transpose and mamul btw?

thunder/transforms/quantization.py
```diff
@@ -294,3 +295,214 @@ def transform_traces_pre_prologue(self, prologue_trace, computation_trace, epilo
...
+                    lora_bsym_ = prims.matmul.bind(bsym.args[0], loraA_bsym.output, output=lora_meta_)
+                    lora_meta = prims.matmul.meta(lora_bsym_.output, loraB_bsym.output)
+                    lora_bsym = prims.matmul.bind(lora_bsym_.output, loraB_bsym.output, output=lora_meta)
+
+                    # Is there a better way?
+                    original_weight_meta = prims.linear.meta(*bsym.args[:3])
+                    original_weight = prims.linear.bind(*bsym.args[:3], output=original_weight_meta)
+
+                    lora_sum_bsym = bsym.from_bsym(
```
t-vi, line 493: I think this could be `prims.add.bind` so you can specify the output.

thunder/transforms/quantization.py
```diff
@@ -294,3 +295,214 @@ def transform_traces_pre_prologue(self, prologue_trace, computation_trace, epilo
...
+                new_computation_trace.bound_symbols.append(loraA_bsym)
+                new_computation_trace.bound_symbols.append(loraB_bsym)
+                new_computation_trace.bound_symbols.append(lora_bsym_)
+                new_computation_trace.bound_symbols.append(lora_bsym)
+                new_computation_trace.bound_symbols.append(original_weight)
+                new_computation_trace.bound_symbols.append(lora_sum_bsym)
+            else:
+                new_computation_trace.bound_symbols.append(bsym.from_bsym())
+
+        new_computation_trace.set_provenance(thunder.core.trace.TraceProvenance("lora linear pass"))
```
t-vi, line 507: Maybe we could also set the provenance for the (copied new) prologue trace.

thunder/transforms/quantization.py
```diff
@@ -294,3 +295,214 @@ def transform_traces_pre_prologue(self, prologue_trace, computation_trace, epilo
...
+                        (1, 0),
+                        output=(loraB_transpose_meta),
+                    )
+                    lora_meta_ = prims.matmul.meta(bsym.args[0], loraA_bsym.output)
+                    lora_bsym_ = prims.matmul.bind(bsym.args[0], loraA_bsym.output, output=lora_meta_)
+                    lora_meta = prims.matmul.meta(lora_bsym_.output, loraB_bsym.output)
+                    lora_bsym = prims.matmul.bind(lora_bsym_.output, loraB_bsym.output, output=lora_meta)
+
+                    # Is there a better way?
+                    original_weight_meta = prims.linear.meta(*bsym.args[:3])
```
t-vi, line 490: this is OK, I guess but it would be nicer to create a new `TensorProxy(like=bsym.output)` and then calling `bsym.from_bsym(output=...)` and append that into the trace to preserve the metadata (like source line info).

t-vi (commented):
Great job! Don't be discouraged by the number of comments, it is quite a complex operation.

thunder/transforms/quantization.py
k223kim: I agree to
- remove `**kwargs`
- make the arguments kw-only
Thanks for the suggestion!

Regarding the weight names, what would be the default that will be inserted?

thunder/transforms/quantization.py
k223kim: I was not aware of/familiar with `new_empty`. Will update this. Thanks Tom!

thunder/transforms/quantization.py
k223kim: Amazing, I think it is simpler to pass an empty map. Let me do that!

thunder/transforms/quantization.py
k223kim: I figured that the following does not work:
```
                    lora_a_bsym = prims.transpose(additional_proxies[f"{n}.lora_a"], (1, 0))
                    lora_b_bsym = prims.transpose(additional_proxies[f"{n}.lora_b"], (1, 0))
                    lora_meta_ = prims.matmul.meta(bsym.args[0], lora_a_bsym)
                    lora_bsym_ = prims.matmul.bind(bsym.args[0], lora_a_bsym, output=lora_meta_)
                    lora_meta = prims.matmul.meta(lora_bsym_.output, lora_b_bsym)
                    lora_bsym = prims.matmul.bind(lora_bsym_.output, lora_b_bsym, output=lora_meta)
                    original_weight_meta = prims.linear.meta(*bsym.args[:3])
                    original_weight = prims.linear.bind(*bsym.args[:3], output=original_weight_meta)

                    lora_sum_bsym = bsym.from_bsym(
                        sym=prims.add,
                        args=(original_weight.output, lora_bsym.output),
                    )

```
What I get from this is that, while I can remove the `.meta` by the simple `.transpose`, I don't think I can remove `.bind`. Or am I misinterpreting this somehow?

thunder/transforms/quantization.py
k223kim: Yes it does!

thunder/transforms/quantization.py
k223kim: Ok! Let me add that to the new prologue trace!

thunder/transforms/quantization.py
k223kim: Let me do the following:
```
                    original_weight = bsym.from_bsym(
                        sym=prims.linear,
                        args=(bsym.args[:3]),
                    )
                new_computation_trace.bound_symbols.append(original_weight)
```

thunder/transforms/quantization.py
k223kim: I guess so. Is this for the sake of simplicity?

thunder/transforms/quantization.py
k223kim: When checking [this](https://github.com/Lightning-AI/litgpt/blob/1d37f9a99bb4ba2b7373bc7fc5b8c5a457af48df/litgpt/lora.py#L564), I was not able to find where we subtract the initialization from the weight.

k223kim: Also, let me add lora_a and lora_b to state_dict!

thunder/transforms/quantization.py
k223kim: Let me get to this after resolving the other comments :)

thunder/transforms/quantization.py
t-vi: Yeah, with b initialized to zero, the contribution would initially be zero, so good.

thunder/transforms/quantization.py
t-vi: Probably just based on what I would do if I would implement LoRA, but probably does does not matter much, so let's keep matmul, but maybe make it thunder.torch.matmul instead to make it look more like "if you had traced matmul". (I must admit that I'm not sure whether I'd use torch or clang, but let's try torch).

thunder/transforms/quantization.py
```diff
@@ -294,3 +296,230 @@ def transform_traces_pre_prologue(self, prologue_trace, computation_trace, epilo
...
+                    )
+                    proxy_lora_b = thunder.TensorProxy(
+                        name=f"{get_param.output.name}_lora_b",
+                        shape=qs["lora_b.shape"],
+                        dtype=thunder.dtypes.to_dtype(qs["lora_b.dtype"]),
+                        device=thunder.devices.to_device(device),
+                        requires_grad=requires_grad,
+                    )
+                    new_bsyms.append(get_param.sym.bind(get_param.args[0], n_lora_a, output=proxy_lora_a))
+                    self.check_proxy(
```
t-vi, line 433: Obviously, I forgot this with the quantization transform you modeled this after, but instead of checking the proxy here (which should be always good because we just defined it), let's add checks to the prologue

https://github.com/Lightning-AI/lightning-thunder/commit/8d3561aae877f277bb820ce7cd74057399ccc7ac

thunder/tests/test_transforms.py
```diff
@@ -334,3 +334,41 @@ def get_model():
...
+    loratransform = LORATransform(r=16, lora_alpha=32)
+
+    jmodel = thunder.jit(
+        model,
+        transforms=[
+            loratransform,
+        ],
+    )
+    actual = jmodel(x)
+    original_jmodel = thunder.jit(model)
```
t-vi, line 368: I wonder if we could use LitGPT's LoraLinear as a reference, wdyt?

thunder/transforms/quantization.py
```diff
@@ -294,3 +296,230 @@ def transform_traces_pre_prologue(self, prologue_trace, computation_trace, epilo
...
+        for idx, bsym in enumerate(bound_symbols):
+            if bsym.sym != prims.unpack_trivial:
+                break
+            new_computation_trace.bound_symbols.append(bsym.from_bsym())
+        new_computation_trace.bound_symbols += new_bindings
+
+        for bsym in bound_symbols[idx:]:
+            if bsym.sym == thunder.torch.linear and bsym.args[1].name in lora_linear_proxies:
+                n = lora_linear_proxies[bsym.args[1].name]
+                with thunder.core.trace.tracectx(computation_trace):
```
t-vi, line 489: We should operate on the new computation trace here.
```
with tracectx(new_computation_trace):  # only need this if you don't have it outside somewhere
   lora_a_transpose = thunder.torch.transpose(...)  # this will automatically append a new bsym to the new_computation_trace

Comment (k223kim) [reply thunder/transforms/quantization.py]: Yes! Let me update this :)

File: thunder/tests/test_transforms.py (hunk 17 above)
Comment (k223kim) [reply thunder/tests/test_transforms.py:368]: I think that'll be much better than the current test case :) thanks for the suggestion!

Comment (k223kim) [reply thunder/tests/test_transforms.py:368]: However, as I try to add LitGPT's lora linear, i think there has to be a way to fix the seed right? (when initializing lora_a and lora_b)

File: thunder/transforms/quantization.py [hunk 19]
```diff
@@ -294,3 +296,221 @@ def transform_traces_pre_prologue(self, prologue_trace, computation_trace, epilo
...
+        new_computation_trace.bound_symbols += new_bindings
+
+        swap_map = dict()
+
+        for bsym in bound_symbols[idx:]:
+            if bsym.sym == thunder.torch.linear and bsym.args[1].name in lora_linear_proxies:
+                n = lora_linear_proxies[bsym.args[1].name]
+                with tracectx(new_computation_trace):
+                    new_computation_trace.push_scope(new_computation_trace.bound_symbols)
+                    lora_a_bsym = thunder.torch.transpose(additional_proxies[f"{n}.lora_a"], 1, 0)
```
t-vi, line 503: Could we rename the variable? `lora_a_t` or somesuch, it is not a bsym but a proxy. (Similar below).

thunder/transforms/quantization.py
```diff
@@ -294,3 +296,221 @@ def transform_traces_pre_prologue(self, prologue_trace, computation_trace, epilo
...
+                    lora_b_bsym = thunder.torch.transpose(additional_proxies[f"{n}.lora_b"], 1, 0)
+                    lora_bsym_1 = thunder.torch.matmul(bsym.args[0], lora_a_bsym)
+                    lora_bsym = thunder.torch.matmul(lora_bsym_1, lora_b_bsym)
+                    original_weight = thunder.torch.linear(*bsym.args[:3])
+                    lora_sum = thunder.torch.add(original_weight, lora_bsym)
```
t-vi, line 508: It's a bit of a taste thing, but if we use
```
new_computation_trace.bound_symbols.append(thunder.torch.add.bind(original_weight, lora_bsym, output=var_original_bsym_output)
```
I think we would not need the `swap_map` and swapping.
(So the rule of thumb is "use calls to the symbols for getting a new output, use `.bind` when you need to specify the an existing output.)

t-vi (commented):
Great updates, I think there are two small bits that could make it even more efficient.

thunder/transforms/quantization.py
k223kim: Sure thing!

thunder/transforms/quantization.py
k223kim: Ah thanks for explaining the rule of thumb (perhaps, the convention in Thunder?). Let me add that! However, when using thunder.torch.add, it was throwing an error:
```
E               NotImplementedError: VJP for torch.add is not implemented
```
However, I was able to resolve this by using prims.add. Please let me know if this can be an issue!

thunder/transforms/quantization.py
```diff
@@ -303,3 +305,219 @@ def transform_traces_pre_prologue(self, prologue_trace, computation_trace, epilo
...
+            )
+            for psym, csym in zip(prologue_trace.bound_symbols[-1].args[0][0], computation_trace.args)
+            if psym.shape != csym.shape or psym.dtype != csym.dtype
+        }
+
+        new_computation_trace = trace_with_replaced_proxy_metadata(computation_trace, computation_proxy_map)
+        bound_symbols = new_computation_trace.bound_symbols
+        new_computation_trace.bound_symbols = []
+
+        new_computation_trace.args = (*new_computation_trace.args, *new_compute_inputs)
```
t-vi, line 491:
```suggestion
        new_computation_trace.args = (*new_computation_trace.args, *new_compute_inputs)
        new_computation_trace.names.update(i.name for i in new_compute_inputs)
```

t-vi (approved):
Looks great, thank you @k223kim . Would you want to move the lora to its own file?

PR #1024, opened 2024-08-22
Trim '"_@_XXX_@_"' into 'XXX'

### LINKED ISSUE #1023: Fails to handle literal `complex('inf')` inside a collection
## 🐛 Bug

```py
import thunder

@thunder.jit
def f():
    return (complex("inf"),)

print(f())
```
```py
("_@_complex(float('inf'), 0.0)_@_",)
```

This does not happen with a finite value:

```py
import thunder

@thunder.jit
def f():
    return (complex(1.0),)

print(f())
```
```py
((1+0j),)
```

## Why this happens

`BoundSymbol.python` method generates a Python expression to be executed, and it prints literals by `thunder.core.codeutils.prettyprint`.

```py
def prettyprint(x, ...):
    ...
    if is_collection(x):
        flat, spec = tree_flatten(x)
        printed = tuple(
            prettyprint(x, with_type=False, literals_as_underscores=literals_as_underscores, _quote_markers=True)
            for x in flat
        )
        unflattened = tree_unflatten(printed, spec)
        unflattened_str = str(unflattened)
        # NOTE Collections of strings (so collections of names) print like this --
        #   ('a', 'b') -- but we want them to print like this -- (a, b) --
        #   so this just removes all the quotes -- this seems super hacky
        unflattened_str = unflattened_str.replace(f"{_quote_marker}'", "")
        unflattened_str = unflattened_str.replace(f"'{_quote_marker}", "")
        return unflattened_str
```
When `(complex(1.0),)` is the input, `unflattened_str` is `('_@_complex(1.0, 0.0)_@_',)`, and the single quotes are eliminated together with `_@_`. However, with `(complex('inf'),)`, `unflattened_str` is `("_@_complex(float('inf'), 0.0)_@_",)`, and the double quotes and `_@_` remain uneliminated.

shino16:
Fixes #1023. Adds support for the pattern `"_@_XXX_@_"` in addition to `'_@_XXX_@_'`.

### DIFF
--- a/thunder/core/codeutils.py
+++ b/thunder/core/codeutils.py
@@ -241,9 +241,11 @@ def prettyprint(
         unflattened_str = str(unflattened)
         # NOTE Collections of strings (so collections of names) print like this --
         #   ('a', 'b') -- but we want them to print like this -- (a, b) --
-        #   so this just removes all the single quotes -- this seems super hacky
+        #   so this just removes all the quotes -- this seems super hacky
         unflattened_str = unflattened_str.replace(f"{_quote_marker}'", "")
         unflattened_str = unflattened_str.replace(f"'{_quote_marker}", "")
+        unflattened_str = unflattened_str.replace(f'{_quote_marker}"', "")
+        unflattened_str = unflattened_str.replace(f'"{_quote_marker}', "")
         return unflattened_str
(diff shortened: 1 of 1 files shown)


mruberry (approved):
Sounds good!

PR #1026, opened 2024-08-22
Update fsdp_v2.py

crcrpar:
## What does this PR do?

Fixes the wrong `_overrides_buffers` update with `new_p` to use `new_b`


lantiga (approved):
Good catch!

PR #1027, opened 2024-08-23
allow NumberProxy in TensorProxy._shape for `symbolic values` cache policy

jjsjann123:
The main change in this PR is to:
* allows NumberProxy in TensorProxy.shape

Small related changes:
* adjust checks to not assert assuming all shapes are static;
* sorting inputs before prologue construction to unpack TensorProxy first;
* restore trace.scopes before transformation unpack;
* change `TensorProxy._numel` to be a deferred `reduce` method in order to query dynamic numel in trace;
* adding a test demonstrating compiled program re-use with cache=`symbolic values`


jjsjann123:
awesome~ CI is green now. Let me clean up this PR for review.

thunder/core/jit_ext.py
```diff
@@ -1203,6 +1205,46 @@ def propagate_constraints(ctx, inputs, intermediates, computation_trace):
                     front.append(variableify(inp))
 
 
+# lift shape computational logic into top level trace
+# Maybe: replace usage of shape NumberProxy with prim?
+def lift_shape_logic(computation_trace):
```
jjsjann123, line 1210: Note for myself. I don't think I needed this for now since we already have `prims.shape` I'll remove this and see how CI looks like.

thunder/core/prims.py
```diff
@@ -480,7 +480,7 @@ def _check_tensor_shape_and_metadata_meta(
 ) -> None:
     # Validates types
     baseutils.check_type(t, TensorProxy)
-    baseutils.check_valid_shape(shape)
+    # baseutils.check_valid_shape(shape)
```
jjsjann123, line 483: this doesn't look right.... :cry: revert this one.

thunder/core/proxies.py
```diff
@@ -1566,7 +1569,9 @@ def __call__(self):
                     return int(self)
 
             if attr == "numel":
-                return _Numel(self._numel)
+                if isinstance(self._numel, int):
+                    return _Numel(self._numel)
+                return method_or_value(self)
```
jjsjann123, line 1574: put a comment here as to *why would we want to defer the numel construction*

thunder/core/prims.py
jjsjann123: actually this one looks right....

jjsjann123:
This one is good for review now. tagging @t-vi @mruberry

thunder/core/prims.py
```diff
@@ -480,7 +480,7 @@ def _check_tensor_shape_and_metadata_meta(
 ) -> None:
     # Validates types
     baseutils.check_type(t, TensorProxy)
-    baseutils.check_valid_shape(shape)
+    # we don't need to validate shape here, because we use `-1` for dynamic shape
```
mruberry, line 483: Can't we still validate the numbers in the shape, and that the shape is composed of integers and number proxies?

mruberry: Also: what about changing the annotation for this function?

thunder/core/jit_ext.py
```diff
@@ -281,6 +279,9 @@ def proxify(self, value: WrappedValue) -> Any:
                 self.add_constraint((clang.check_tensor_shape_and_metadata, p_orig))
             elif co not in (CACHE_OPTIONS.SAME_INPUT, CACHE_OPTIONS.NO_CACHING):
                 raise NotImplementedError(f"Unsupported cache option {co}")
+
+            if p is not uvalue:
```
mruberry, line 283: Would you add a comment explaining this case?

thunder/core/jit_ext.py
```diff
@@ -1466,6 +1467,31 @@ def from_provenance(provenance, *, new_output=False):
                 assert n == "kwargs"
                 pro_kwargs_proxy = output
 
+    def sort_tensor_proxy_first(v: Variable | Proxy) -> Proxy:
```
mruberry, line 1470: Would you add a comment explaining what this function does?

thunder/core/proxies.py
```diff
@@ -1238,10 +1238,14 @@ def _infer_tensor_properties(
...
-    _shape = tuple(pyval(x) for x in _shape)
-
-    # Computes derived properties
-    _numel = reduce(operator.mul, _shape, 1)
+    if not using_symbolic_values():
+        _shape = tuple(pyval(x) for x in _shape)
+        # Computes derived properties
+        _numel = reduce(operator.mul, _shape, 1)
+    else:
+        # deferred computation of numel
```
mruberry, line 1246: Why defer the numel computation?

thunder/executors/pythonex.py
```diff
@@ -49,7 +49,11 @@ def _check_tensor_shape_and_metadata_impl(
 ) -> None:
     assert isinstance(t, torch.Tensor), f"expected Tensor, got {type(t).__name__}"
     assert (
-        tuple(t.shape) == shape and str(t.device) == device and t.dtype == dtype and t.requires_grad == requires_grad
+        len(t.shape) == len(shape)
+        and all(map(lambda x, ref: ref == -1 or x == ref, t.shape, shape))
```
mruberry, line 53: Why does the prim not take the actual number proxies for shape instead of converting them to -1?

thunder/core/prims.py
jjsjann123: I'll merge this conversation with the below one. Since I'm only allowing numbers, but replacing NumberProxy with `-1`.

thunder/core/proxies.py
jjsjann123: two issues:
1. Since this function is executed/traced before the TensorProxy is constructed. We'll be accessing the `shape` attributes. So deferring this naturally resolves that.
2. numel computation may or may not be used at all. So if we defer the use, we don't have to compute it in the trace.

Note that we'll be relying on CSE to remove duplicated `numel` computation across the board.

thunder/executors/pythonex.py
jjsjann123: This is just for simplification.
Say if we do take the actual number proxies here, what are we going to use it for? I assume it's serving the same purpose, it'll just be a wild card and we'll skip any checks on the entry.

 I don't hold a strong opinion on this. Do you think keeping a NumberProxy is easier to read, vs having a random `-1` holding a special meaning?

mruberry: I don't think it matters much today. I have a preference for passing the actual number proxies, if possible, as the number proxies may be augmented with data in the future that the prim could inspect

thunder/core/proxies.py
mruberry: This seems OK for now, but wouldn't precomputing numel just give another number proxy if the shape contains one or more number proxies? Then requests for numel would just return that proxy? That seems like nbd to compute at compiletime and not have to compute as part of the computation?

thunder/executors/pythonex.py
jjsjann123: Sounds good. Let me try to add it there as a placeholder

thunder/core/proxies.py
jjsjann123: That's true. It's just some extra things in the trace that could also be removed with DCE or something.

I can have a follow up PR try to clean this up a bit. I just need to figure out how/where to place the computation of numel in trace.

mruberry (approved):
Cool step forward, @jjsjann123. I look forward to seeing more!

mruberry:
Conflicts have to be resolved before this can be merged

t-vi:
Supercool ! Thank you  @jjsjann123, @mruberry

thunder/core/jit_ext.py
```diff
@@ -1474,6 +1498,12 @@ def from_provenance(provenance, *, new_output=False):
             for a in args:
                 if isinstance(a, Proxy):
                     unpack(a)
+            # unpacking Proxy in TensorProxy.shape which is used in `check_tensor_shape_and_metadata`
+            if prim == clang.check_tensor_shape_and_metadata:
```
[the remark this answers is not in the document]
jjsjann123: That's a great idea. I'm marking this one as resolved and I'll fix it up in another PR.

PR #1028, opened 2024-08-23
[tensor parallel] refactor sharding logic of tensor parallel

crcrpar:
## What does this PR do?

Move sharding logic which mostly is shared among column-wise and row-wise to `tensor_parallel/common`


t-vi (approved):
Look good, thank you @crcrpar

PR #1030, opened 2024-08-23
Set SDPBackend explicitly for validation

Priya2698:
NVFuser CI has failed `thunder/tests/test_sdpa.py` tests for H100.
This is because the backend picked for `torch.nn.scaled_dot_product_attention` in these cases is `cudnn_attention` whereas nvfuser is using flash attention. Explicitly setting the SDPBackend will allow validation against torch reference.

When checking if `sdpa` operator can be accepted by nvfuser, we use `utils.py/_fused_sdp_choice` (same as `sdpaex.py`). This function returns `FlashAttention` as the backend whenever flash attention is enabled and can be used for the given set of inputs (for `torch.__version__ > 2.2.0`). Hence, it is possible that for the reference computation,  `torch.nn.attention.scaled_dot_product_attention` actually uses a different backend.

### DIFF
--- a/thunder/tests/test_nvfuser.py
+++ b/thunder/tests/test_nvfuser.py
@@ -1056,10 +1056,11 @@ def test_sdpa(
         ref_inp = inp.clone().detach()
         ref_inp.requires_grad = True
         ref_tensor_inputs.append(ref_inp)
-    
+
     from torch.nn.attention import SDPBackend, sdpa_kernel
+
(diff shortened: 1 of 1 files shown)


Priya2698:
I have verified this locally on the GH node. @xwang233, is there a way to check if the nvfuser CI passes with this PR (only need to run `thunder/tests` for H100)?

wujingyue (approved):

t-vi (approved):
Stamping. Thank you @Priya2698 @wujingyue

PR #1031, opened 2024-08-23
[tensor parallel] move sharding into `Transform.transform_module`

crcrpar:
## What does this PR do?

As per title, this pr moves sharding to `Transform.transform_module` and have `MaterializationTransform` for materialization like fsdp


t-vi (approved):
Looks great, thank you @crcrpar

PR #1032, opened 2024-08-23
Take the interpreter holding references to the arguments into account

### LINKED ISSUE #1029: `get_alloc_memory` deallocates arguments before the function returns
## 🐛 Bug

```py
import torch
import thunder
from thunder.examine.memory_caculation import get_alloc_memory

def f(x):
    z = x + 1
    w = z + 1
    return w

jitted = thunder.jit(f, executors=[])
jitted(torch.zeros(8, dtype=torch.float32, device='cuda'))
print(get_alloc_memory(thunder.last_traces(jitted)[-1]))
```
#### Trace
```py
def computation(x):
  z = torch.add(x, 1)  # z: "cuda:0 f32[8]"
  del x
  w = torch.add(z, 1)  # w: "cuda:0 f32[8]"
  del z
  return w
```
#### Output:
```
(64, {'unpack_trivial x': 32, 'add z': 32, 'python_del x': -32, 'add w': 32, 'python_del z': -32})
```
64 bytes is equivalent to two `f32[8]` tensors, `(x, z)` or `(z, w)`.

#### Expected output:
```
(96, {'unpack_trivial x': 32, 'add z': 32, 'add w': 32, 'python_del z': -32})
```
Equivalent to three `f32[8]` tensors, `(x, z, w)`.

### Why the current output is incorrect

`get_alloc_memory` thinks that `del x` deallocates the tensor behind `x`, so it records `'python_del x': -32`. However, as Python holds the arguments in memory throughout the function call, it cannot be deallocated.

shino16:
Fixes #1029. The tests in `test_examine_memory.py` uses some magic numbers made with an incorrect assumption, and I am yet to update them.

### DIFF
--- a/thunder/examine/memory_caculation.py
+++ b/thunder/examine/memory_caculation.py
@@ -1,4 +1,5 @@
 from collections.abc import Callable, MutableSequence, MutableMapping, MutableSet
+from functools import partial
 
 from thunder.core.prims import PrimIDs
 from thunder.core.proxies import CollectionProxy, FutureTensorProxy, Proxy, TensorProxy
@@ -119,9 +120,16 @@ def del_op_memory(bsym: BoundSymbol, tensor_to_memory_data: ProxyDict, name_to_a
     return memory_size
 
 
+@register_memory_calculate_function("clear_mutable_collection")
 def clear_mutable_collection_argument_memory(
-    bsym: BoundSymbol, tensor_to_memory_data: ProxyDict, name_to_alloc_memory: dict[str, int]
+    bsym: BoundSymbol, tensor_to_memory_data: ProxyDict, name_to_alloc_memory: dict[str, int], is_argument: bool
 ) -> int:
+    # Clearing the collection forces the interpreter to release references to its elements,
+    # even if the collection was an argument.
+    # So we cancel the n += 1 (see get_alloc_memory) for tensors contained in such a collection
+    if not is_argument:
+        return 0
+
     collection_proxy = bsym.flat_proxy_args[0]
     if not isinstance(collection_proxy.collection(), (MutableSequence, MutableMapping, MutableSet)):
         return 0
@@ -175,11 +183,9 @@ def get_alloc_memory(trc: TraceCtx) -> tuple[int, dict[str, int]]:
             continue
 
         impl = memory_calculate_impls.get(bsym.sym.id, default_alloc_memory)
-        if bsym.sym.id == "clear_mutable_collection" and bsym.flat_proxy_args[0].name in arg_names:
-            # Clearing the collection forces the interpreter to release references to its elements,
-            # even if the collection was an argument.
-            # So we cancel the n += 1 for the tensors contained
-            impl = clear_mutable_collection_argument_memory
+        if impl is clear_mutable_collection_argument_memory:
+            is_argument = bsym.flat_proxy_args[0].name in arg_names
+            impl = partial(impl, is_argument=is_argument)
 
         allocated += impl(bsym, tensor_to_memory_data, name_to_alloc_memory)
         max_allocated = max(max_allocated, allocated)
--- a/thunder/examine/memory_caculation.py
+++ b/thunder/examine/memory_caculation.py
@@ -5,6 +5,7 @@ from thunder.core.proxies import FutureTensorProxy, Proxy, TensorProxy
 from thunder.core.symbol import BoundSymbol, Symbol
 from thunder.core.trace import TraceCtx
 from thunder.core.utils import check_type, ProxyDict
+from thunder.core.pytree import tree_iter
 
 memory_calculate_skip_list = (PrimIDs.RETURN, "clear_collection")
 
@@ -141,6 +142,10 @@ def get_alloc_memory(trc: TraceCtx) -> tuple[int, dict[str, int]]:
     allocated = 0
 
     tensor_to_memory_data = ProxyDict()
+    for arg in tree_iter((trc.args, trc.kwargs)):
+        # In addition to the arguments themselves, the interpreter holds references to the arguments
+        tensor_to_memory_data[arg] = MemoryData(n=2, proxy=arg)
+
     for bsym in trc.bound_symbols:
         if bsym.sym.id in memory_calculate_skip_list:
             continue
--- a/thunder/examine/memory_caculation.py
+++ b/thunder/examine/memory_caculation.py
@@ -21,6 +21,7 @@ thunder_alias_operator_list = (
     "torch_prims_reshape_impl",  # torchex implementation of prims.reshape.
     "permute",
     "contiguous",
+    "split",
     "torch_wait_prim_impl",
 )
 
@@ -96,11 +97,13 @@ def default_alloc_memory(
 def track_alias_op_memory(
     bsym: BoundSymbol, tensor_to_memory_data: ProxyDict, name_to_alloc_memory: dict[str, int]
 ) -> int:
+    import thunder
+
     inp = bsym.flat_proxy_args[0]
-    out = bsym.flat_proxy_outs[0]
     assert inp in tensor_to_memory_data
-    tensor_to_memory_data[inp].incr_ref()
-    tensor_to_memory_data[out] = tensor_to_memory_data[inp]
+    for out in bsym.flat_proxy_outs:
+        tensor_to_memory_data[inp].incr_ref()
+        tensor_to_memory_data[out] = tensor_to_memory_data[inp]
     return 0
 
 
--- a/thunder/examine/memory_caculation.py
+++ b/thunder/examine/memory_caculation.py
@@ -1,13 +1,14 @@
-from collections.abc import Callable
+from collections.abc import Callable, MutableSequence, MutableMapping, MutableSet
 
 from thunder.core.prims import PrimIDs
-from thunder.core.proxies import FutureTensorProxy, Proxy, TensorProxy
+from thunder.core.proxies import CollectionProxy, FutureTensorProxy, Proxy, TensorProxy
 from thunder.core.symbol import BoundSymbol, Symbol
 from thunder.core.trace import TraceCtx
 from thunder.core.utils import check_type, ProxyDict
 from thunder.core.pytree import tree_iter
 
-memory_calculate_skip_list = (PrimIDs.RETURN, "clear_collection")
+# Arguments are considered independently, so we ignore all unpacking operations on them
+memory_calculate_skip_list = (PrimIDs.RETURN, PrimIDs.UNPACK_TRIVIAL, PrimIDs.UNPACK_SEQUENCE)
 
 # List of operators that considered no memory changes occurred in Thunder.
 # NOTE: for the operators have different input and output shape, such as expand,
@@ -79,8 +80,6 @@ def default_alloc_memory(
     Returns:
         int: The size of memory change caused by the input bsym
     """
-    # Skip CollectionProxy(output of input unpacking operators, such as unpack_sequence)
-    # and other negligible scalar types
     tensor_outs = [x for x in bsym.flat_proxy_outs if isinstance(x, (TensorProxy, FutureTensorProxy))]
     result = sum(t.numel * t.dtype.bytes for t in tensor_outs)
     for x in tensor_outs:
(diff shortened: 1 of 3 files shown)


shino16:
When rewriting the test, I faced #1043. We previously did not test the backward for split.

shino16:
I tried to eliminate the "golden" values for tests, but this seems impossible when nvFuser is involved. For example, on the following test case, nvFuser returns as `result2` a tensor with `stride == (0, 0, 1)`.
```py
def bar(a, b):  # [4] [2,2]
    a_1 = torch.unsqueeze(a, 0)  # [1,4]
    a_2 = torch.unsqueeze(a_1, 1)  # [1,1,4]
    a_3 = a_2.expand(2, 3, 4)  # [2,3,4]

    b_1 = torch.reshape(b, (4,))  # [4]
    b_2 = torch.unsqueeze(b_1, 0)  # [1,4]
    b_3 = torch.unsqueeze(b_2, 1)  # [1,1,4]
    b_4 = b_3.expand(2, 3, 4)  # [2,3,4]

    result1 = a_2 + b_3
    result2 = b_4 + a_3
    return result1, result2
```

<details>
<summary>generated trace</summary>

```py
def augmented_forward_fn(a, b):
  # a: "cuda:0 f32[4]"
  # b: "cuda:0 f32[2, 2]"
  [t14, t15] = nvFusion0(a, b)
    # t0 = prims.broadcast_in_dim(a, [1, 4], [1])  # t0: "cuda:0 f32[1, 4]"
    # t1 = prims.broadcast_in_dim(t0, [1, 1, 4], [0, 2])  # t1: "cuda:0 f32[1, 1, 4]"
    # t5 = prims.broadcast_in_dim(t1, (2, 3, 4), (0, 1, 2))  # t5: "cuda:0 f32[2, 3, 4]"
    # t7 = prims.reshape(b, (4,))  # t7: "cuda:0 f32[4]"
    # t8 = prims.broadcast_in_dim(t7, [1, 4], [1])  # t8: "cuda:0 f32[1, 4]"
    # t9 = prims.broadcast_in_dim(t8, [1, 1, 4], [0, 2])  # t9: "cuda:0 f32[1, 1, 4]"
    # t13 = prims.broadcast_in_dim(t9, (2, 3, 4), (0, 1, 2))  # t13: "cuda:0 f32[2, 3, 4]"
    # t14 = prims.add(t1, t9)  # t14: "cuda:0 f32[1, 1, 4]"
    # t15 = prims.add(t13, t5)  # t15: "cuda:0 f32[2, 3, 4]"
  return {'output': (t14, t15), 'flat_args': [a, b], 'flat_output': (t14, t15)}, ((), ())
(144, {'unpack_trivial a': 16, 'unpack_trivial b': 16, 'nvFusion0 t14, t15': 112})
```

</details>

Can we skip `nvFuserExecutor` for such tests as this PR does? It still serves the purpose of checking if `get_alloc_memory` scans the trace correctly, and it does not make much sense to test something that gives untrue values.

IvanYashchuk:
> Can we skip nvFuserExecutor for such tests as this PR does? It still serves the purpose of checking if get_alloc_memory scans the trace correctly, and it does not make much sense to test something that gives untrue values.

Yes, let's skip nvFuser with a comment explaining why. In the future, we need to update `get_alloc_memory` to account for broadcasts in fused regions correctly.

thunder/examine/memory_caculation.py
```diff
@@ -95,11 +97,13 @@ def default_alloc_memory(
 def track_alias_op_memory(
     bsym: BoundSymbol, tensor_to_memory_data: ProxyDict, name_to_alloc_memory: dict[str, int]
 ) -> int:
+    import thunder
+
```
IvanYashchuk, line 101: This import seems unused.

thunder/examine/memory_caculation.py
```diff
@@ -141,6 +145,10 @@ def get_alloc_memory(trc: TraceCtx) -> tuple[int, dict[str, int]]:
     allocated = 0
 
     tensor_to_memory_data = ProxyDict()
+    for arg in tree_iter((trc.args, trc.kwargs)):
+        # In addition to the arguments themselves, the interpreter holds references to the arguments
+        tensor_to_memory_data[arg] = MemoryData(n=2, proxy=arg)
```
IvanYashchuk, line 150: Why is the counter initialized with 2?

thunder/examine/memory_caculation.py
```diff
@@ -141,6 +145,10 @@ def get_alloc_memory(trc: TraceCtx) -> tuple[int, dict[str, int]]:
     allocated = 0
 
     tensor_to_memory_data = ProxyDict()
+    for arg in tree_iter((trc.args, trc.kwargs)):
```
IvanYashchuk, line 148: Backward trace holds some tensors in a list on purpose because Python would hold onto the list variable, but not its content. Clearing a list allows `del` calls to free the memory. Here's an example:
```py
import torch

def f(list_of_tensors):
    print(f"0: {torch.cuda.memory_allocated()/1e6} MB")
    x0, x1 = list_of_tensors
    list_of_tensors.clear()
    print(f"1: {torch.cuda.memory_allocated()/1e6} MB")
    z = x0 + 1
    print(f"2: {torch.cuda.memory_allocated()/1e6} MB")
    del x0
    print(f"3: {torch.cuda.memory_allocated()/1e6} MB")
    w = x1 + z + 1
    print(f"4: {torch.cuda.memory_allocated()/1e6} MB")
    del x1, z
    print(f"5: {torch.cuda.memory_allocated()/1e6} MB")
    return w

l = [torch.ones(1000, device="cuda") for _ in range(2)]
out = f(l)
# prints
# 0: 0.008192 MB
# 1: 0.008192 MB
# 2: 0.012288 MB
# 3: 0.008192 MB
# 4: 0.012288 MB
# 5: 0.004096 MB
```
the same doesn't work with tuples because we can't modify tuple content:
```py
import torch

def f(tuple_of_tensors):
    print(f"0: {torch.cuda.memory_allocated()/1e6} MB")
    x0, x1 = tuple_of_tensors
    del tuple_of_tensors
    print(f"1: {torch.cuda.memory_allocated()/1e6} MB")
    z = x0 + 1
    print(f"2: {torch.cuda.memory_allocated()/1e6} MB")
    del x0
    print(f"3: {torch.cuda.memory_allocated()/1e6} MB")
    w = x1 + z + 1
    print(f"4: {torch.cuda.memory_allocated()/1e6} MB")
    del x1, z
    print(f"5: {torch.cuda.memory_allocated()/1e6} MB")
    return w

l = tuple(torch.ones(1000, device="cuda") for _ in range(2))
out = f(l)
# prints
# 0: 0.008192 MB
# 1: 0.008192 MB
# 2: 0.012288 MB
# 3: 0.012288 MB
# 4: 0.016384 MB
# 5: 0.012288 MB
```

How do you think we could support both immutable and mutable containers? Maybe it's better to avoid using `tree_iter` here and use plain iterations over args if args is a tuple?

thunder/tests/test_examine_memory.py
IvanYashchuk, line 1: Great test updates!

IvanYashchuk (approved):
Python has a really interesting behavior with respect to holding references to input variables for the duration of the function call. In order to make `get_alloc_memory` more accurately represent Python's memory deallocation we need to take into account the type of containers in which tensors are stored, they could be immutable and then Python would hold a reference to an immutable object and not let us modify its content so tensors will never be freed during the function call, but containers could be mutable and then it's possible to free the tensors if the container itself doesn't hold a reference to tensor anymore. We should take this into consideration in a follow-up.

thunder/examine/memory_caculation.py
shino16: I have just added a few more words.

thunder/examine/memory_caculation.py
shino16: Your point makes a lot of sense. Indeed, when I pass a list as an argument to `backward_fn`, the peak memory consumption decreased.

I suppose this can be simulated by interpreting `clear_mutable_collection` in the same way as `del`. This involves keeping track of how many lists contain which tensors, but this is apparently doable.

thunder/examine/memory_caculation.py
IvanYashchuk: Thank you!

jjsjann123:
> nvFuser returns as result2 a tensor with stride == (0, 0, 1).

I'm disappointed that eager doesn't do so by default. :cry:

shino16:
As Ivan pointed out, the previous code did not consider `clear_mutable_collection` that forces the interpreter to release references to its elements. Indeed, when I passed lists as arguments to `bw_trace` instead of tuples, it gave different numbers:
```py
def f(x, y, z):
    return x * y * z
```
`get_alloc_memory`'s output (commit 05c0a5da9508ff75119cd2c39e6e50547218340c):
```
{'unpack_trivial ': 0, 'unpack_sequence ': 0, 'clear_mutable_collection ': 0, 'unpack_sequence t2': 16, 'unpack_sequence t0, x, y, z': 64, 'mul t8': 16, 'mul t9': 16, 'mul t10': 16, 'mul t11': 16, 'python_del t8': -16}
peak = 144, after = 128
```
Actual memory usage:
```
peak = 96, after = 64
```

I slightly changed `thunder/core/transforms.py` so that `bw_trace.args` matches what `ThunderFunction.backward` actually passes to it, which is not tuples but lists. `get_alloc_memory` follows this change and takes `clear_mutable_collection` on an argument into account.

thunder/examine/memory_caculation.py
```diff
@@ -1,12 +1,15 @@
...
 
 from thunder.core.prims import PrimIDs
-from thunder.core.proxies import FutureTensorProxy, Proxy, TensorProxy
+from thunder.core.proxies import CollectionProxy, FutureTensorProxy, Proxy, TensorProxy
 from thunder.core.symbol import BoundSymbol, Symbol
 from thunder.core.trace import TraceCtx
 from thunder.core.utils import check_type, ProxyDict
+from thunder.core.pytree import tree_iter
 
-memory_calculate_skip_list = (PrimIDs.RETURN, "clear_collection")
```
shino16, line 9: `"clear_collection"` has changed its name to `"clear_mutable_collection"`, which should not be skipped.

thunder/examine/memory_caculation.py
```diff
@@ -118,6 +120,32 @@ def del_op_memory(bsym: BoundSymbol, tensor_to_memory_data: ProxyDict, name_to_a
...
+    bsym: BoundSymbol, tensor_to_memory_data: ProxyDict, name_to_alloc_memory: dict[str, int], is_argument: bool
+) -> int:
+    # Clearing the collection forces the interpreter to release references to its elements,
+    # even if the collection was an argument.
+    # So we cancel the n += 1 (see get_alloc_memory) for tensors contained in such a collection
+    if not is_argument:
+        return 0
+
+    collection_proxy = bsym.flat_proxy_args[0]
+    if not isinstance(collection_proxy.collection(), (MutableSequence, MutableMapping, MutableSet)):
```
shino16, line 134: This check works because the updated `transforms.py` now sets as `bw_trace.args` the same kinds of collections that `ThunderFunction.backward` actually passes.

shino16: The same check is in `thunder.executors.pythonex`.
https://github.com/Lightning-AI/lightning-thunder/blob/a89aa488aaa9220a03a82c982744529769f89b15/thunder/executors/pythonex.py#L252-L254

thunder/examine/memory_caculation.py
shino16: Thanks, updated in 136c7d16cf1a963279101272c8728517a9966b60

shino16:
@t-vi This is ready for review. Thank you!

t-vi:
> Yes, let's skip nvFuser with a comment explaining why.

Unless I missed it, we would want to add to add the comment.

shino16:
Thank you for pointing that out! I added the comment.

The current test failure is due to the breaking change in `actions/upload-artifact@v3`. I described the cause in https://github.com/Lightning-AI/utilities/issues/302 and made a fix in https://github.com/Lightning-AI/utilities/pull/303.

t-vi (approved):
Thank you @shino16 @IvanYashchuk @jjsjann123 @crcrpar

Issue #967, opened 2024-08-13
Record `Tensor.clone()` in trace


crcrpar:
It should work if there's a clone between reshape and in-place operation. Currently, it fails with the same error because `clone` is not recorded on the trace:
https://github.com/Lightning-AI/lightning-thunder/blob/ffbebe07bdf003c3a60f2cab88298e96b80bdbba/thunder/torch/__init__.py#L2521-L2525
In the code review, it was decided that it's fine for the clone not to record itself in the trace, but now it seems we need this clarity to break the chain of view ops (https://github.com/Lightning-AI/lightning-thunder/pull/429#issuecomment-2121041926).
The code snippet I tried:
```py
import thunder
import torch

def f(x):
  y = x.reshape(6, 4).clone()
  y.add_(1)
  return y

x = torch.randn((4, 6))
jf = thunder.jit(f)
jf(x)
```

_Originally posted by @IvanYashchuk in https://github.com/Lightning-AI/lightning-thunder/pull/958#discussion_r1715102637_
            

cc @apaz-cli

t-vi:
possibly fixed by https://github.com/Lightning-AI/lightning-thunder/pull/1046

PR #1046, opened 2024-08-26
record `torch.clone`

crcrpar:
## What does this PR do?

Fixes https://github.com/Lightning-AI/lightning-thunder/pull/958#discussion_r1715102637.

### DIFF
--- a/thunder/core/prims.py
+++ b/thunder/core/prims.py
@@ -151,6 +151,7 @@ class PrimIDs(Enum):
     RANDN = auto()
     EMPTY = auto()
     TENSOR_FROM_SEQUENCE = auto()
+    CLONE = auto()
     # Probability distribution-related ops
     MULTINOMIAL = auto()
     GET_AND_UPDATE_RNG_STATE = auto()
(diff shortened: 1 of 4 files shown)


thunder/core/prims.py
```diff
@@ -2814,6 +2815,14 @@ def _empty_meta(
 empty = make_prim(PrimIDs.EMPTY, "empty", meta=_empty_meta)
 
 
+# TODO(crcrpar): Cover `memory_format` kwarg
+def _clone_mta(a: TensorProxy, **kwargs) -> TensorProxy:
```
t-vi, line 2819: NIT: why mta not meta?

t-vi:
Should this be fixing https://github.com/Lightning-AI/lightning-thunder/issues/967 ?

thunder/core/prims.py
crcrpar: typo

thunder/core/prims.py
```diff
@@ -2814,6 +2815,14 @@ def _empty_meta(
 empty = make_prim(PrimIDs.EMPTY, "empty", meta=_empty_meta)
 
 
+# TODO(crcrpar): Cover `memory_format` kwarg
+def _clone_meta(a: TensorProxy, **kwargs) -> TensorProxy:
+    return TensorProxy(like=a, requires_grad=a.requires_grad)
```
IvanYashchuk, line 2820: requires_grad attribute is only for input TensorProxy and it's unused for intermediates. There's no harm in setting it, but this could be just `TensorProxy(like=a)`.
Was anything breaking without the requires_grad set?

PR #1048, opened 2024-08-26
add ProxyTag and static memory location tagging for cudagraphs

t-vi:
This is one of the outcomes of the design discussion re #981 .

We add a ProxyTag mechanism (with registration) and introduce STATIC_MEMORY_LOCATION as the first tag.

We then use this tag for knowing the static inputs in the CUDAGraph transform.

### DIFF
--- a/thunder/core/jit_ext.py
+++ b/thunder/core/jit_ext.py
@@ -57,6 +57,7 @@ from thunder.core.proxies import (
     DistParallelType,
     proxy,
     Proxy,
+    ProxyTag,
     AnyProxy,
     NumberProxy,
     StringProxy,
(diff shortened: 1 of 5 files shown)


lantiga (approved):
Looks great.

I'm thinking in terms of composition: how would we handle two different transforms registering the same tag (due to naming collisions)?

Maybe not a problem for now but at least a warning on register_tag if the tag is already registered could be useful for now.

t-vi:
I'll send a follow-up blocking double registration.

t-vi:
Don't know what the CI wants there. :/

riccardofelluga:
I guess I have a question here, what happens if the proxies are changed in some transform before they get a chance to be used? How do we intend to keep track of that?

CC some more people for visibility of this new feature
@IvanYashchuk @mruberry @kshitij12345 @kiya00 @crcrpar

mruberry:
> I guess I have a question here, what happens if the proxies are changed in some transform before they get a chance to be used? How do we intend to keep track of that?
> 
> CC some more people for visibility of this new feature @IvanYashchuk @mruberry @kshitij12345 @kiya00 @crcrpar

Would you file an issue for this so we can track the discussion?

PR #1049, opened 2024-08-26
followup to proxy tag pr

t-vi:
As mentioned in PR review, it would be good to error on duplicate registration.

Also fixes a typo in the QuantizationTransform.


lantiga (approved):
Thank you @t-vi for the follow-up!

Issue #1033, opened 2024-08-23
TE - Tests failing with latest main (v1.11)


kshitij12345:
```python
                    dgrad, _ = fp8_gemm(
>                       weight_fp8.transpose_2d(),
                        weight_fp8._scale_inv,
                        0,
                        weight_fp8._fp8_dtype,
                        grad_output_c,
                        ctx.fp8_meta["scaling_bwd"].scale_inv,
                        tex.FP8BwdTensors.GRAD_OUTPUT1,
                        fp8_dtype_backward,
                        output_dtype,
                        get_workspace(),
                        use_split_accumulator=_2X_ACC_DGRAD,
                        ub_algo=ub_algo if ctx.ub_overlap_ag else None,
                        ub=ctx.ub_obj_gradout if ctx.ub_overlap_ag else None,
                        out_index=out_index,
                        fp8_meta_tensor=meta_tensor,
                        D_dtype=output_te_dtype,
                    )
E                   AttributeError: 'NoneType' object has no attribute 'transpose_2d'

/usr/local/lib/python3.10/dist-packages/transformer_engine/pytorch/module/linear.py:467: AttributeError
===================================================================================================================================== short test summary info ======================================================================================================================================
FAILED thunder/tests/test_transformer_engine_executor.py::test_te_linear_forward_backward[None] - AttributeError: 'NoneType' object has no attribute 'transpose_2d'
FAILED thunder/tests/test_transformer_engine_executor.py::test_te_linear_forward_backward[fp8_recipe1] - AttributeError: 'NoneType' object has no attribute 'transpose_2d'
FAILED thunder/tests/test_transformer_engine_executor.py::test_te_linear_forward_backward_multiple_iteration[None] - AttributeError: 'NoneType' object has no attribute 'transpose_2d'
FAILED thunder/tests/test_transformer_engine_executor.py::test_te_linear_forward_backward_multiple_iteration[fp8_recipe1] - AttributeError: 'NoneType' object has no attribute 'transpose_2d'
FAILED thunder/tests/test_transformer_engine_executor.py::test_te_with_retain_graph - AttributeError: 'NoneType' object has no attribute 'transpose_2d'
============================================================================================================================= 5 failed, 2 passed, 8 warnings in 4.18s ==============================================================================================================================
```

PR #1052, opened 2024-08-26
TE - support v1.11 (current main)

kshitij12345:
Fixes https://github.com/Lightning-AI/lightning-thunder/issues/1033

TransformerEngine [PR](https://github.com/NVIDIA/TransformerEngine/pull/1083) updates the saved tensors order where position of tensor saved with scaling factor details is moved up. This leads to the errors mentioned in the issue.

Changes
* In this PR, we add yet another switch based on TE version based on which we correctly order the saved tensors.
* We remove an old version switch as it is not relevant anymore `TE_VERSION_1_6_PLUS` (current TE stable is v1.9).

Tests
* Tested with tests in thunder on TE v1.9 (current stable), v1.10 and v1.11 (current main).

Follow-up
* Clean-up related to switches for TE v1.8 version and making TE v1.9 (as the minimum required version).

### DIFF
--- a/thunder/executors/transformer_engineex.py
+++ b/thunder/executors/transformer_engineex.py
@@ -38,8 +38,8 @@ TE_AVAILABLE: bool = package_available("transformer_engine")
 # Ex. addition of a positional argument for cpu_offloading (not as the last argument)
 # between version 1.2 and 1.3.
 # Hence, we have these guards based on version.
-TE_VERSION_1_6_PLUS: bool = False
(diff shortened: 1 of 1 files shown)


kshitij12345:
Have also filed an issue to investigate integrating with a new public API - https://github.com/Lightning-AI/lightning-thunder/issues/1056

t-vi (approved):
let's have it. Thank you @kshitij12345

PR #1059, opened 2024-08-27
Auto-registered ops: Add ops returning views to `_syms_returning_views`

kiya00:
<details>

</details>

## What does this PR do?

As a follow-up to https://github.com/Lightning-AI/lightning-thunder/issues/957 (https://github.com/Lightning-AI/lightning-thunder/issues/957#issuecomment-2293521400)

When auto-registration is added there's no in-place support yet. As mentioned in #957, the operators that return views can interfere with in-place ops. This PR adds the relative auto-registered symbols to `_syms_returning_views` and `_syms_returning_runtime_dependently_views` and let the in-place pass deal with it.

### DIFF
--- a/thunder/torch/__init__.py
+++ b/thunder/torch/__init__.py
@@ -5635,9 +5635,11 @@ _torch_to_thunder_complete_map = {
     **{fn: fn for fn in _torch_noinline_functions},
 }
 
-
+# records the torch symbols that may return tensor views
 # ref: https://pytorch.org/docs/stable/tensor_view.html
-_syms_returning_runtime_dependently_views: set[Symbol] = {
+# NOTE Symbols that return tensor views can interfere with in-place operators
+# See :func:`thunder.core.functionalization.check_inplace_to_views` for the details.
+_syms_that_may_return_views: set[Symbol] = {
     reshape,
     contiguous,
     to,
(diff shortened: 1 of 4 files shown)


thunder/torch/__init__.py
```diff
@@ -5636,7 +5637,13 @@ def check_overlap_ops():
 
 
 # ref: https://pytorch.org/docs/stable/tensor_view.html
-_syms_returning_runtime_dependently_views: set[Symbol] = {reshape, contiguous, to, flatten}
+_syms_returning_runtime_dependently_views: set[Symbol] = {
```
mruberry, line 5640: `_syms_that_may_return_views` might be a clearer name

A comment explaining what these operations are would be helpful, as well as an explanation of how this set works

thunder/torch/default_torch_ops.py
```diff
@@ -756,3 +756,28 @@
         torch.fft.rfftn,
     ],
 }
+
+_auto_registered_operators_returning_views = [
```
mruberry, line 760: A comment here, like what this list does, what it means, and how it's generated, would be helpful

mruberry (commented):
Hey @kiya00! I just made a couple suggestions for comments to improve readability

thunder/torch/default_torch_ops.py
kiya00: Hi @mruberry , the list of ops that return or may return tensor views are used in the functionalization, I pointed it to the in-place check function that uses this list

mruberry (approved):
Cool!

@IvanYashchuk, would you like to take a look?

PR #706, opened 2024-07-04
quick patch on nvfuser getnv/reshape to handle sequence with NumberProxy

jjsjann123:
1. recursively go through `getnv` on sequence to retrieve dynamic scalar for nvfuser.
2. update reshape parsing in nvfuser executor to allow NumberProxy in new shape.

PR #1060, opened 2024-08-27
resnet18 failure with reshape is fixed

jjsjann123:
The old failure should be fixed in #706 
I'm not seeing a repro locally, so testing it with CI now.


crcrpar (approved):

kiya00 (approved):

t-vi (approved):
Thank you @jjsjann123 @crcrpar @kiya00

PR #1061, opened 2024-08-27
avoid renames proxies in forward trace during grad transform

jjsjann123:
With side branch (#1027 dynamic shape)
test_grad.py::test_update_forward_with_new_saved_for_backward_numberproxy_nvfuser_cuda_None

generates forward trace:
```
@torch.no_grad()
@no_autocast
def augmented_forward_fn(a, i0, i1, b):
  # a: "cuda:0 f32[[IntegerProxy name=i0, value=2, static=CONSTRAINT.CONSTRAINABLE], [IntegerProxy name=i1, value=2, static=CONSTRA
INT.CONSTRAINABLE]]"
  # i0: "int 2"
  # i1: "int 2"
  # b: "float 3.0"
  t4 = ltorch.mul(a, b)  # t4: "cuda:0 f32[[IntegerProxy name=i0, value=2, static=CONSTRAINT.CONSTRAINABLE], [IntegerProxy name=i1, value=2, static=CONSTRAINT.CONSTRAINABLE]]"
    # b0 = prims.eq(i1, 1)  # b0: "bool False"
    # b1 = prims.eq(i1, i1)  # b1: "bool True"
    # b2 = prims.eq(i0, 1)  # b2: "bool False"
    # b3 = prims.eq(i0, i0)  # b3: "bool True"
    # t4 = prims.mul(a, b)  # t4: "cuda:0 f32[[IntegerProxy name=i0, value=2, static=CONSTRAINT.CONSTRAINABLE], [IntegerProxy name=i1, value=2, static=CONSTRAINT.CONSTRAINABLE]]"
  return ({'output': t4, 'flat_args': [a, i0, i1, b], 'flat_output': (t4,)}, (((a, b), None, ([b],)),))
```

Without the change, forward trace will be constructed as
```
@torch.no_grad()
@no_autocast
def augmented_forward_fn(a, i0, i1, b):
  # a: "cuda:0 f32[[IntegerProxy name=i0, value=2, static=CONSTRAINT.CONSTRAINABLE], [IntegerProxy name=i1, value=2, static=CONSTRAINT.CONSTRAINABLE]]"
  # i4: "int 2"
  # i5: "int 2"
  # b: "float 3.0"
  t10 = ltorch.mul(a, b)  # t10: "cuda:0 f32[[IntegerProxy name=i0, value=2, static=CONSTRAINT.CONSTRAINABLE], [IntegerProxy name=i1, value=2, static=CONSTRAINT.CONSTRAINABLE]]"
    # b6 = prims.eq(i1, 1)  # b6: "bool False"
    # b7 = prims.eq(i1, i1)  # b7: "bool True"
    # b8 = prims.eq(i0, 1)  # b8: "bool False"
    # b9 = prims.eq(i0, i0)  # b9: "bool True"
    # t10 = prims.mul(a, b)  # t10: "cuda:0 f32[[IntegerProxy name=i0, value=2, static=CONSTRAINT.CONSTRAINABLE], [IntegerProxy nam
e=i1, value=2, static=CONSTRAINT.CONSTRAINABLE]]"
  return ({'output': t10, 'flat_args': [a, i4, i5, b], 'flat_output': (t10,)}, (((a, b), None, ([b],)),))
```

giving me an error on undefined proxy `i4` (and `i5` here as well)

### DIFF
--- a/thunder/core/transforms.py
+++ b/thunder/core/transforms.py
@@ -3657,9 +3657,7 @@ def forward_and_backward_from_trace(trace: Trace, torch_autograd=False) -> Forwa
         else:
             return None
 
-    forward_trace = construct_trace()(augmented_forward_fn, *trace.args, **trace.kwargs)
-    # We set forward trace to construct proxies because we need these proxies to
-    # have different names than the ones in the forward trace.
+    forward_trace = construct_trace(rename_proxies=False)(augmented_forward_fn, *trace.args, **trace.kwargs)
     try:
         tracectx_token = set_tracectx(forward_trace)
         # We don't want to record those ones_like calls in the forward trace.


IvanYashchuk:
I don't think it's a correct fix, it only masks the problem.
There are name clashes for symbolic integers on the shape of the tensor `a` which are named i0 and i1 and at the same time there are i0 and i1 arguments to the function.
The signature for this function is `<Signature (a, i0, i1, b)>` and it should be used as the basis for naming input proxies here:
https://github.com/Lightning-AI/lightning-thunder/blob/912fd01b73697ea8a8dcf9d11f62dbe590e288e9/thunder/common.py#L311-L315
TensorProxy `a` shouldn't use i0 and i1 as its integer dimension names because these names should be used for function inputs.

IvanYashchuk:
I don't think it's a correct fix, it only masks the problem.
There are name clashes for symbolic integers on the shape of the tensor `a` which are named i0 and i1 and at the same time there are i0 and i1 arguments to the function.
The signature for this function is `<Signature (a, i0, i1, b)>` and it should be used as the basis for naming input proxies here:
https://github.com/Lightning-AI/lightning-thunder/blob/912fd01b73697ea8a8dcf9d11f62dbe590e288e9/thunder/common.py#L311-L315
TensorProxy `a` shouldn't use i0 and i1 as its integer dimension names because these names should be used for function inputs.

jjsjann123:
> There are name clashes for symbolic integers on the shape of the tensor a which are named i0 and i1 and at the same time there are i0 and i1 arguments to the function.

Sorry for the lack of context here. Those are not name clashes and they are indeed the same number proxy.

I should have included the prologue as well, which has:
```
        prologue:
 import thunder
import torch

@torch.no_grad()
@no_autocast
  # args: "Any"
  prims.check_len(args, 2)
  # kwargs: "Any"
  prims.check_len(kwargs, 0)
  b: "float 3.0" = args[1]
  a: "cuda:0 f32[[IntegerProxy name=i0, value=2, static=CONSTRAINT.CONSTRAINABLE], [IntegerProxy name=i1, value=2, static=CONSTRAINT.CONSTRAINABLE]]" = args[0]
  obj: "Any" = a.shape
  i1: "int 2" = obj[1]
  i0: "int 2" = obj[0]
  # ...
  return ((a, i0, i1, b), ())
```

jjsjann123:
whether or not we think that prologue is the right behavior probably deserves its own thread for discussion and I'm happy to go through that as well.

Back to the old question, why do we want to rename proxies during forward trace in the first place?

t-vi:
@jjsjann123 We do have that sometimes thunder happens to re-use the tensor proxies between prologue and compute, or that they have the same name, but this should not relied upon (and the re-use is not that happy a thing). Is that what causes the problem?
We can rely on `prologue_output[0][i]` being fed into `computation_trace.args[i]` but that is about it.

I think the renaming might be part preparation for a revision of how forward and backward interact (though I'm not sure).

jjsjann123:
> thunder happens to re-use the tensor proxies between prologue and compute, or that they have the same name, but this should not relied upon (and the re-use is not that happy a thing)

That's super interesting. I'm wondering if that's considered a bug when thunder trace messes up the names... And the renaming is just covering that.

```
  a: "cuda:0 f32[[IntegerProxy name=i0, value=2, static=CONSTRAINT.CONSTRAINABLE], [IntegerProxy name=i1, value=2, static=CONSTRAINT.CONSTRAINABLE]]" = args[0]
  obj: "Any" = a.shape
  i1: "int 2" = obj[1]
  i0: "int 2" = obj[0]
  # ...
  return ((a, i0, i1, b), ())
```

Anyway, the issue here is that I'm passing both `i0, i1` and `a` as input to the compute trace and the renaming kicks in to rename the unpacking, but not the signature or uses. This is the trace I'm seeing without the changes. Looks like the renaming isn't doing the right thing neither. i.e. I would expect the rename to change both the unpack, as well as the signature.

```
@torch.no_grad()
@no_autocast
def augmented_forward_fn(a, i0, i1, b):
  # a: "cuda:0 f32[[IntegerProxy name=i0, value=2, static=CONSTRAINT.CONSTRAINABLE], [IntegerProxy name=i1, value=2, static=CONSTRAINT.CONSTRAINABLE]]"
  # i4: "int 2"
  # i5: "int 2"
  # b: "float 3.0"
  t10 = ltorch.mul(a, b)  # t10: "cuda:0 f32[[IntegerProxy name=i0, value=2, static=CONSTRAINT.CONSTRAINABLE], [IntegerProxy name=i1, value=2, static=CONSTRAINT.CONSTRAINABLE]]"
    # b6 = prims.eq(i1, 1)  # b6: "bool False"
    # b7 = prims.eq(i1, i1)  # b7: "bool True"
    # b8 = prims.eq(i0, 1)  # b8: "bool False"
    # b9 = prims.eq(i0, i0)  # b9: "bool True"
    # t10 = prims.mul(a, b)  # t10: "cuda:0 f32[[IntegerProxy name=i0, value=2, static=CONSTRAINT.CONSTRAINABLE], [IntegerProxy nam
e=i1, value=2, static=CONSTRAINT.CONSTRAINABLE]]"
  return ({'output': t10, 'flat_args': [a, i4, i5, b], 'flat_output': (t10,)}, (((a, b), None, ([b],)),))
```

But sounds like the philosophy is that we don't want duplicated input in the signature in the first place. `i0, i1` is passed in as direct argument, as well as a field in `a`. But I actually think that's exactly what's needed. Since unpacking is supposed to be done in prologue trace?!

t-vi:
```
def augmented_forward_fn(a, i0, i1, b):
  # a: "cuda:0 f32[[IntegerProxy name=i0, value=2, static=CONSTRAINT.CONSTRAINABLE], [IntegerProxy name=i1, value=2, static=CONSTRAINT.CONSTRAINABLE]]"
  # i4: "int 2"
  # i5: "int 2"
...
 ```
 this is clearly broken, but if `i0` and `i1` were also replaced by `i4` and `i5`.

jjsjann123:
> ```
> def augmented_forward_fn(a, i0, i1, b):
>   # a: "cuda:0 f32[[IntegerProxy name=i0, value=2, static=CONSTRAINT.CONSTRAINABLE], [IntegerProxy name=i1, value=2, static=CONSTRAINT.CONSTRAINABLE]]"
>   # i4: "int 2"
>   # i5: "int 2"
> ...
> ```
> 
> this is clearly broken, but if `i0` and `i1` were also replaced by `i4` and `i5` it would be fine even if they're `i0` and `i1` in the prologue.

I'm not sure what is expected for `a.shape` though. i.e. are we expecting the renaming to also reflect that as well?

i.e. `a.shape` was `[i0, i1]` before the renaming, and if we are updating the args to `i4`/`i5` respectively, we should also update the corresponding TensorProxy.shape. Which leaves me wonder, what is the use of passing them in as arguments to compute trace.

I think I'm going to change it to not passing the unpacked shape entries to compute trace for now. So the unpack would happen in compute trace directly. I think this is less intrusive to existing code. Closing this PR for now.

@mruberry might have a stronger opinion on this, I'll open an issue to discuss that on the side. Happy to revisit this decision.

jjsjann123:
> We do have that sometimes thunder happens to re-use the tensor proxies between prologue and compute, or that they have the same name

On a side note. Disabling the renaming didn't break any existing test. So this is a coverage issue that we should put a test to catch. Otherwise, it's hard to make an argument on justifying this as a necessary thing.

PR #1064, opened 2024-08-29
Add a bunch of benchmarks from NeMo's NeVA

tfogal:
This takes a number of the graphs from our "default" configuration of NeVA and copies them into benchmarks. This should mean most people do not have to get NeVA up and running to do something meaningful for this network.

The benchmarks are not perfect:

* There is a bug in the extraction logic that means we miss some graphs.
* Some graphs do not run under thunder at all; they are not included here. The common case for this happening is an unsupported communication op being buried in the graph.
* The extraction logic tends to dump graphs that are identical sans `requires_grad` flipping on some tensors. I only included 1 when this happens.
* I omitted "silly" graphs. One example: a graph that just casts its input tensor to ... 🥁🥁 🥁 ... the same type that it already is.

Thanks to Mike Ruberry for the idea on how to do this and Ivan Yashchuk for technical guidance.

cc @crcrpar @tfogal


tfogal:
cc @parthmannan these are from the "old" configuration. I can't run on the new config yet (still getting access to data...) but let's talk offline, I'd love if we could figure out a way to run the graph extraction code on the config you have going--I believe that config to be more representative of real users.

riccardofelluga:
Asking for context, the idea is to benchmark for NeVa with Thunder as a Dynamo backend right?

Another question I have is on why do we need to test these subgraphs? Is there an advantage of benchmarking the subgraphs over benchmarking the modules directly? 
Because my first thought is, what if NeVa implementation changes? will this benchmarks be obsolete?

IvanYashchuk:
Was a script used to generate this file? I think it would be better to store a script to create benchmark cases in Thunder instead of pushing this large autogenerated file to the repository.

tfogal:
> Asking for context, the idea is to benchmark for NeVa with Thunder as a Dynamo backend right?

Well, the idea is to *improve* NeVA's Dynamo+Thunder perf. This removes all the dynamo stuff to keep things actionable as purely torch / thunder code.

> Another question I have is on why do we need to test these subgraphs? Is there an advantage of benchmarking the subgraphs over benchmarking the modules directly?

If by "the modules" you meant the actual NeVA types such as `NevaModel` or `NevaWordEmbeddingMixin` those would be great too! As you say, that approach does have the great advantage that it adjusts as NeVA changes.

The point/advantage of *this* however is that people do not need to install NeMo and get it working to make meaningful progress on perf here.

> what if NeVa implementation changes? will this benchmarks be obsolete?

Yeah, these will need to be updated regularly. One of you/me/Kshiteej/Ivan (as I think that's the set of people that have NeVA running) would need to do that if we want help from others.

I envision doing updates soon and regularly (already have some ideas planned that should help), but certainly would be excited to see others doing that as well!

> Was a script used to generate this file? I think it would be better to store a script to create benchmark cases in Thunder instead of pushing this large autogenerated file to the repository.

I *do* intend to try to fit the script somewhere in thunder, but as above, I wonder if you and Riccardo were thinking this is aimed at e.g. you/Riccardo/Kshiteej/me. To repeat myself: the 4 of us don't really need this. This is about enabling people that don't care for the complicated NeMo setup (e.g. the nvFuser team) to  help us make progress on performance.

Sorry for the confusion--I could have explained this better!

parthmannan:
> I can't run on the new config yet (still getting access to data...)

I believe I was running with the same data Abhishree had provided for the old config. Is there new data we would need for the fine-tuning config?

tfogal:
> > I can't run on the new config yet (still getting access to data...)
> 
> I believe I was running with the same data Abhishree had provided for the old config. Is there new data we would need for the fine-tuning config?

There's definitely new data. I haven't yet been able to explore whether that changes things from our perspective.

Let's talk through internal channels, there are site-specific complexities as you may imagine. Open invite for other close collaborators reading this to ping me out of band on this as well!

tfogal:
ping all, please review, I'd love to start running these nightly.

@t-vi specifically: this is failing the 'large file' CI check, so it would need a manual override. What are your thoughts?

thunder/benchmarks/neva.py
```diff
@@ -0,0 +1,6275 @@
...
+                l_self_modules_vision_encoder_modules_vision_model_modules_encoder_modules_layers_modules_0_modules_self_attn_modules_v_proj_parameters_weight_,
+                l_self_modules_vision_encoder_modules_vision_model_modules_encoder_modules_layers_modules_0_modules_self_attn_modules_v_proj_parameters_bias_,
+            )
+            hidden_states_1 = (
+                l_self_modules_vision_encoder_modules_vision_model_modules_encoder_modules_layers_modules_0_modules_self_attn_modules_v_proj_parameters_weight_
+            ) = (
+                l_self_modules_vision_encoder_modules_vision_model_modules_encoder_modules_layers_modules_0_modules_self_attn_modules_v_proj_parameters_bias_
+            ) = None
+            view_1 = linear_2.view(2, -1, 16, 64)
+            linear_2 = None
```
IvanYashchuk, line 911: Please remove all these None assignments.

thunder/benchmarks/neva.py
```diff
@@ -0,0 +1,6275 @@
...
+@pytest.mark.benchmark(
+    group="neva",
+    warmup_iterations=1,  # takes a long time to run.
+)
+def g1():
+    class DynamoModule(torch.nn.Module):
+        def forward(
+            self,
+            L_vision_x_: torch.Tensor,
+            L_self_modules_vision_encoder_modules_vision_model_modules_embeddings_parameters_class_embedding_: torch.nn.parameter.Parameter,
```
IvanYashchuk, line 16: Using a shorter prefix instead of `L_self_modules_vision_encoder_modules_vision_model_modules` would make the code more readable.

thunder/benchmarks/neva.py
```diff
@@ -0,0 +1,6275 @@
...
+            transpose_42 = None
+            view_58 = query_states_16.view(2, 257, 16, 64)
+            query_states_16 = None
+            transpose_43 = view_58.transpose(1, 2)
+            view_58 = None
+            contiguous_26 = transpose_43.contiguous()
+            transpose_43 = None
+            query_states_17 = contiguous_26.view(32, -1, 64)
+            contiguous_26 = None
+            key_states_17 = k
```
IvanYashchuk, line 6217: This layout is the default one and can be omitted to save lines of code.

IvanYashchuk, line 6194: This is a very small graph. Do you know what part of the source code it relates to?
If it's as you say also relevant outside NeVA I suggest contributing this module separately to the collection of benchmarks we have in `thunder/benchmarks/__init__.py` for benchmark definitions and `thunder/benchmarks/targets.py` for benchmark invocations.

IvanYashchuk, line 6153: This code is unused and can be removed.

IvanYashchuk, line 5904: This code is unused.

IvanYashchuk (changes requested):
Pushing large script-generated files to the repository is not recommended.

Extracting pure PyTorch code out of NeVA using Dynamo is a clever approach, this is great. However, now it's important to make this code easier to work with and add some manual postprocessing. Another important aspect is to document how this file was generated so that other people could repeat the process if needed. The g1 function should be refactored using forloops to iterate over transformer blocks and initialize inputs for anyone to be able to make changes to the code and observe performance impact. Do you have a plan to maintain sync with future NeVA code changes?

Regarding NeMo access, could you elaborate on why Dynamo-generated scripts without direct NeMo/Megatron-LM dependencies are necessary? The [`nvcr.io/nvidia/nemo:dev` container from NGC](https://catalog.ngc.nvidia.com/orgs/nvidia/containers/nemo/tags) should provide a straightforward solution to setting up the environment.
If you want to run this in NVIDIA-internal nightly CI it doesn't have to be in the `main` branch. Furthermore, the frequency can be adjusted to weekly or on-demand runs instead of nightly.

thunder/benchmarks/neva.py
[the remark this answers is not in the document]
IvanYashchuk: GitHub is having trouble rendering correctly this large diff. This comment was referring to line 3946.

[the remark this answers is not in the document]
IvanYashchuk: GitHub is having trouble rendering correctly this large diff. This comment was referring to line 6231.

IvanYashchuk: GitHub is having trouble rendering correctly this large diff. This comment was referring to line 6194.

IvanYashchuk: GitHub is having trouble rendering correctly this large diff. This comment was referring to lines 6150-6153 with the pattern:
```py
function_ctx = torch.autograd.function.FunctionCtx()
function_ctx = None
```

t-vi:
If the goal is to make the benchmark / debugging more accessible, I think a better solution would be to have a Lightning Studio or Docker container that contains the things needed to get started.
In the end, this will be a maintenance headache and it does not scale well.

tfogal:
Ivan and I had a chat offline:

* I'll update this anyway (it's benchmarking compile time right now and I have a fix locally)
* We can set `num_layers=1` or 2 in the source NeVA model first, and that should create much smaller graphs while still keeping it worth people invest time optimizing.
* we're getting `requires_grad=False` because that's how the network is setup; IIRC the LLM is frozen and we're training the image analysis portions, but @parthmannan can correct me (this is the neva_pretrain.py situation).
* Tom to look at targets.py and see if we can create something that fits in thunder's existing "fwd", "fwdbwd" etc. scheme. might be a follow-up

tfogal:
we'll track these subgraphs elsewhere

PR #947, opened 2024-08-09
Add Thunder as a Dynamo backend

IvanYashchuk:
This PR adds `thunder.dynamo.ThunderCompiler` meant to be used as a `backend=` option to `torch.compile`/`torch._dynamo.optimize`. Only the most basic functionality is included in this PR. Thunder may not support the graphs generated by Dynamo, in this case, an (possibly scary) error will be thrown. 

Our main end-to-end benchmark script is modified to allow invoking Dynamo preprocessing with a `dynamo+` string passed to the `--compile` option. With Dynamo preprocessing PyTorch's FSDP and DDP are used for distributed training.

See https://github.com/Lightning-AI/lightning-thunder/issues/915 and the link to the design doc in the issue.

Issue #915, opened 2024-08-02
Enable Thunder as a `torch.compile` backend


IvanYashchuk:
## 🚀 Feature

Let's make Thunder applicable to a broader range of PyTorch model implementations by using Thunder together with `torch.compile`.

The detailed proposal can be found in the following Google doc:
https://docs.google.com/document/d/1zd45hmqMG2B9w10HrN7WnZ3d-50bd1fc5fW2SG5C41E/edit?usp=sharing.

Feel free to comment inline in the Google doc or leave general comments here.

### Additional context

- https://github.com/Lightning-AI/lightning-thunder/issues/539

- https://github.com/Lightning-AI/lit-thunder-LEGACY/issues/1904

PR #1066, opened 2024-08-29
Add graph-by-graph benchmarking of dynamo.ThunderCompiler

kiya00:
As a part of https://github.com/Lightning-AI/lightning-thunder/issues/915.
After https://github.com/Lightning-AI/lightning-thunder/pull/947 `torch.compile` can have `thunder.dynamo.ThunderCompiler` as a backend. This PR adds a graph-by-graph benchmarking of different executors(eager, inductor, thunder) working on each `torch.fx.GraphModule` after splitting and gives perf and peak allocated memory information. Moreover, a customized grouping option is supported `--benchmark-group-by='graph-by-graph:param:GraphID,param:SplitModuleName'` to group the test cases based on GraphID and SplitModuleName, allowing for performance comparison between different executors.(see examples in the comments below).

The added benchmarking structure is `thunder.benchmarks.ThunderCompilerGraphBenchmarking`, an example of how to use it is in `thunder.benchmarks.targets.test_dynamo_LlamaMLPBenchmark`

Example outputs in https://github.com/Lightning-AI/lightning-thunder/pull/1066#issuecomment-2317778178.


kiya00:
Run `pytest thunder/benchmarks/targets.py -k test_dynamo -vs`
outputs:
```
------------------------------------------------------------------------------------------------- benchmark: 3 tests -------------------------------------------------------------------------------------------------
Name (time in ms)                                          Min                Max               Mean            StdDev             Median               IQR            Outliers      OPS            Rounds  Iterations
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_dynamo_LlamaMLPBenchmark[torch]graph0             21.8733 (1.0)      23.3903 (1.0)      22.5793 (1.0)      0.3107 (1.0)      22.6010 (1.0)      0.4830 (1.0)          31;0  44.2883 (1.0)          72           1
test_dynamo_LlamaMLPBenchmark[torch.compile]graph0     22.2302 (1.02)     23.8240 (1.02)     22.9449 (1.02)     0.4052 (1.30)     22.8968 (1.01)     0.7638 (1.58)         29;0  43.5826 (0.98)         71           1
test_dynamo_LlamaMLPBenchmark[thunder]graph0           22.5432 (1.03)     24.0336 (1.03)     23.0256 (1.02)     0.3743 (1.20)     22.9507 (1.02)     0.7913 (1.64)         27;0  43.4298 (0.98)         73           1
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```
By using a function that is segmented into 2 graphs(e.g. `thunder.tests.test_dynamo.test_basic.<local>.func`), with `--benchmark-sort="name"`, outputs:
```
---------------------------------------------------------------------------------------------------- benchmark: 6 tests ---------------------------------------------------------------------------------------------------
Name (time in us)                                          Min                Max               Mean            StdDev             Median               IQR            Outliers  OPS (Kops/s)            Rounds  Iterations
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_dynamo_LlamaMLPBenchmark[thunder]graph0           63.6553 (12.11)    96.9659 (13.16)    67.5422 (12.49)    3.8962 (11.31)    66.4568 (12.47)    1.5790 (41.98)     125;148       14.8055 (0.08)       1558          10
test_dynamo_LlamaMLPBenchmark[thunder]graph1           60.5953 (11.52)    93.4610 (12.69)    64.1243 (11.86)    4.2762 (12.42)    63.0003 (11.82)    1.6009 (42.56)     112;141       15.5947 (0.08)       1650          10
test_dynamo_LlamaMLPBenchmark[torch.compile]graph0     15.3940 (2.93)     17.6217 (2.39)     15.6716 (2.90)     0.3665 (1.06)     15.5920 (2.93)     0.0959 (2.55)        34;42       63.8098 (0.35)        647         100
test_dynamo_LlamaMLPBenchmark[torch.compile]graph1      8.4523 (1.61)     10.7282 (1.46)      8.6604 (1.60)     0.3444 (1.0)       8.5798 (1.61)     0.0714 (1.90)        60;83      115.4680 (0.62)       1187         100
test_dynamo_LlamaMLPBenchmark[torch]graph0             11.9389 (2.27)     16.6382 (2.26)     12.1828 (2.25)     0.4257 (1.24)     12.0829 (2.27)     0.0876 (2.33)        46;56       82.0831 (0.44)        832         100
test_dynamo_LlamaMLPBenchmark[torch]graph1              5.2580 (1.0)       7.3666 (1.0)       5.4086 (1.0)      0.3455 (1.00)      5.3281 (1.0)      0.0376 (1.0)        93;122      184.8920 (1.0)        1910         100
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

```

thunder/benchmarks/__init__.py
```diff
@@ -2958,3 +2960,163 @@ def fn(self) -> Callable:
...
+                _concrete_shape(e_v),
+                dtype=e_v.dtype,
+                device=e_v.device,
+                requires_grad=e_v.requires_grad,
+            )
+            for e_v in example_value
+        )
+    else:
+        raise TypeError(
+            f"The 'example_value' in the placeholder node is expected to be either a Tensor or a Tuple of Tensors."
```
crcrpar, line 2995: This f does not look necessary
```suggestion
            "The 'example_value' in the placeholder node is expected to be either a Tensor or a Tuple of Tensors."
```

thunder/benchmarks/__init__.py
```diff
@@ -2958,3 +2960,163 @@ def fn(self) -> Callable:
...
+            ...     x = torch.ones(2, requires_grad=True).cuda()
+            ...     compiled(x)
+        Running the example with `pytest script.py  --benchmark-sort="name"` will produce benchmarking results.
+
+        # Dynamo segments the graph into two subgraphs, each identified by the 'GraphId[id]' field in the test name.
+        # Each subgraph contains a single split module, processed by the Thunder-defined splitter,
+        # which corresponds to the 'SplitModuleName[split_module_name]' field.
+        # The currently active executor is indicated by the '_eager' suffix.
+        ----------------------------------------------------------------------------------------------------- benchmark: 2 tests -----------------------------------------------------------------------------------------------------
+        Name (time in us)                                             Min                Max               Mean            StdDev             Median               IQR            Outliers  OPS (Kops/s)            Rounds  Iterations
```
crcrpar, line 3047: Looking at this example, I'm speculating that the memory foot print is not displayed even though we seem to record peak memory consumption.
Would it be possible to have that as well in the table or in a separate one?

thunder/benchmarks/targets.py
```diff
@@ -812,3 +814,29 @@ def test_resnet50(benchmark, executor: Callable, compute_type: ComputeType):
...
+#
+# Thunder dynamo benchmarks
+#
+
+
+def test_dynamo_LlamaMLPBenchmark(benchmark):
+    backend = ThunderCompilerGraphBenchmarking(benchmark)
+
+    bench: Benchmark = LlamaMLPBenchmark(
+        config="Llama-2-7b-hf",
```
crcrpar, line 828: Q: would there be a plan to use Llama-3-8B by default? I'm just wondering

thunder/benchmarks/__init__.py
```diff
@@ -2958,3 +2960,163 @@ def fn(self) -> Callable:
...
+    _executors = (
+        "eager",
+        "inductor",
+        "thunder",
+    )
+
+    def __init__(
+        self,
+        bench: BenchmarkFixture,
+        executors: Sequence[str] | None = None,
```
crcrpar, line 3009: Looking at the caller, this argument of `executors` does not seem to be used. Correct me if I'm wrong but what's the plan? Would we be going to parametrize executors, like torch executor, nvfuser executor, and torch compile executor?

crcrpar (commented):
overall LGTM

thunder/benchmarks/__init__.py
kiya00: From what I've investigated, the content of the table that is printed in pytest-benchmark seems to be fixed, the extensible interface is `extra_info` which is only shown in the json file. So I put the memory info there

thunder/benchmarks/targets.py
kiya00: At the moment, I have just randomly chosen this model as an example of how to use the graph-by-graph benchmark.

thunder/benchmarks/__init__.py
kiya00: the executors can be 'eager', 'inductor', and 'thunder', if we init it with `executors=['thunder']`, it only uses thunder, by default it runs all the three above executors on each graph

crcrpar: I personally want to avoid None if possible, would you think it possible to do so?

thunder/benchmarks/__init__.py
IvanYashchuk: Right, unfortunately, extra information can be printed only with some other custom tools. That was also the reason RAPIDS project [has forked pytest-benchmark](https://github.com/rapidsai/benchmark).

thunder/benchmarks/__init__.py
crcrpar: I mean to make the type hint non-optional and to remove the default value of None.

tfogal (approved):
Agh I realized I made comments last week and then never clicked 'submit', sorry. You may have addressed many of these already, i'll double check after meetings

thunder/benchmarks/__init__.py
```diff
@@ -2958,3 +2960,185 @@ def fn(self) -> Callable:
...
+    _executors = (
+        "eager",
+        "inductor",
+        "thunder",
+    )
+
+    def __init__(
+        self,
+        bench: BenchmarkFixture,
+        executors: Sequence[str] = None,
```
crcrpar, line 3011: can't we just 
```suggestion
        executors: Sequence[str],
```
?

thunder/benchmarks/__init__.py
```diff
@@ -9,6 +9,7 @@
 from functools import partial
 from numbers import Number
 from typing import Any
+from pytest_benchmark.fixture import BenchmarkFixture
```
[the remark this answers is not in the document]
kiya00: I think the purpose of using pytest-benchmark is probably to maintain compatibility with the current benchmarking setup (as it supports calibration and can generate both table-formatted output and JSON files used in GitLab's nightly benchmarks). so this PR is kind of fundamentally based on pytest-benchmark. If we decide to move away from pytest-benchmark, we would need to gather requirements and potentially develop a new tool, I'm not sure @IvanYashchuk .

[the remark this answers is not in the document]
IvanYashchuk: Yan started by using a custom benchmark code and I recommended to use pytest benchmark here. I'd like to understand better your concerns. We have used pytest-benchmark for quite some time in `thunder/benchmarks/targets.py` and it proved useful. nvFuser also uses it: https://github.com/NVIDIA/Fuser/blob/4cbd7a4978f23376f9d6dee2efcf981737880a56/benchmarks/python/core.py#L236 Have you communicated your concerns about pytest-benchmark to the nvFuser team?

> it doesn't know about GPU work and so it can make odd choices with how many times to run things

Can you provide an example where not knowing about GPU hurts?

> it runs things a variable number of times it can be really difficult to compare runs with different options to understand what's better

A variable number of runs is done to provide accurate statistics. This behavior can be overwritten with command line options.

> If you put nvtx around the benchmark(function, args) call, then you gave the range for all the calls, and what you really want is the range for one call. Inserting nvtx into the function you're running is difficult but could be an option.

It's straightforward to put the nvtx markers around the function call if that's what is needed, benchmark(add_nvtx(function), args).

thunder/benchmarks/__init__.py
kiya00: yes, I can rm it, I just thought there are https://github.com/Lightning-AI/lightning-thunder/blob/b6a8940cda2ba564eac3b1df6f2201dd954750f5/thunder/benchmarks/__init__.py#L3078-L3079, so it's not much harm to keep the None here

thunder/benchmarks/targets.py
```diff
@@ -812,3 +813,29 @@ def test_resnet50(benchmark, executor: Callable, compute_type: ComputeType):
...
+        batchdims=(2,),
+        device="cuda:0",
+        dtype=thunder.bfloat16,
+        requires_grad=True,
+    )
+
+    args, kwargs = bench.make_batch()
+    # Using torch.compile here fails with "TypeError: cannot pickle '_io.TextIOWrapper' object" in
+    # https://github.com/Lightning-AI/pytorch-lightning/blob/828fd998961f6a60f92c35254bb94d6e049ad069/src/lightning/fabric/wrappers.py#L421
```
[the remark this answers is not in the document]
IvanYashchuk: It's probably the correct link, the comment could use more details when the error could happen. The lightning package (for example imported in `benchmark_litgpt.py`) replaces `torch.compile` with their own function that can't handle some keyword arguments that can't be deepcopied
https://github.com/Lightning-AI/pytorch-lightning/blob/828fd998961f6a60f92c35254bb94d6e049ad069/src/lightning/fabric/wrappers.py#L415

thunder/benchmarks/targets.py
```diff
@@ -812,3 +814,29 @@ def test_resnet50(benchmark, executor: Callable, compute_type: ComputeType):
...
+    fn = torch._dynamo.optimize(backend=backend)(bench.fn())
```
IvanYashchuk, line 838: Does `torch.compile` fail here when used instead of `torch._dynamo.optimize`? Do we import lightning package somewhere that replaces real torch.compile with its own?

thunder/benchmarks/targets.py
```diff
@@ -812,3 +814,29 @@ def test_resnet50(benchmark, executor: Callable, compute_type: ComputeType):
...
     benchmark_for_compute_type(compute_type, benchmark, fn, args, kwargs)
+
+
+#
+# Thunder dynamo benchmarks
+#
+
+
+def test_dynamo_LlamaMLPBenchmark(benchmark):
+    backend = ThunderCompilerGraphBenchmarking(benchmark)
```
kiya00, line 825:
```suggestion
    backend = ThunderCompilerGraphBenchmarking(benchmark, executors=["thunder", "inductor", "eager"])
```

tfogal (approved):
again apologies for my slow replies

thunder/benchmarks/__init__.py
```diff
@@ -2958,3 +2960,156 @@ def fn(self) -> Callable:
...
+        for ex in self.executors:
+            # Uses the already compiled module if it is compiled with the expected executor
+            if name.startswith(ex):
+                fn = self.subgraph_infos[self.graph_idx].submodule_to_compiled_functions[gm].compiled_fn
+            else:
+                if ex == "thunder":
+                    fn = self._thunder_jit(gm)
+                elif ex == "inductor":
+                    fn = self._torch_compile(gm)
+                else:
```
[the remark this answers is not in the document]
kiya00: Hi @tfogal , this PR is related to what we planned in Q3, and I noticed there's another tool related to benchmarking on thunder+dynamo in Q4's plan, I guess you might be talking about the requirements on that tool, I'm also wondering how this 2 things are related, we can discuss more about it.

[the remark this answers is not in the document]
tfogal: Oh, interesting; I had missed that. I'm not sure there's much of a difference honestly 😄, other than these new ways to run things.
But anyway that's fine, feel free to throw #1258 at me if you'd like!

thunder/benchmarks/__init__.py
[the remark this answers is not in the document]
kiya00: Hi @tfogal , thank you so much for the long explanation, this PR currently benchmarks each dynamo subgraph (first segmented by dynamo, then split by the thunder splitter, but I think it could be just one subgraph for most litgpt models), on each subgraph(fx GraphModule) we do the compilation using 3 types of backends each(eager, torch.compile, thunder), then the compiled function is passed to the pybenchmark(no nvtx markers), and then the pybench outputs the statistics on each of the subgraphs of 3 backends for comparison, an example could be found [here](https://github.com/Lightning-AI/lightning-thunder/pull/1066#issuecomment-2317778178) 
IIUC, your main concern is that there's no nvtx marker, marker can be easily added in the granularity of the subgraph. And for the overhead of nvtx marker I think it could be small, let me add it and check on some real piece of model and give some data on Monday

thunder/benchmarks/targets.py
```diff
@@ -812,3 +813,29 @@ def test_resnet50(benchmark, executor: Callable, compute_type: ComputeType):
...
+        device="cuda:0",
+        dtype=thunder.bfloat16,
+        requires_grad=True,
+    )
+
+    args, kwargs = bench.make_batch()
+    # Using torch.compile here fails with "TypeError: cannot pickle '_io.TextIOWrapper' object" in
+    # https://github.com/Lightning-AI/pytorch-lightning/blob/828fd998961f6a60f92c35254bb94d6e049ad069/src/lightning/fabric/wrappers.py#L421
+    fn = torch._dynamo.optimize(backend=backend)(bench.fn())
+    fn(*args, **kwargs)
```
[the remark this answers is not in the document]
IvanYashchuk: I don't think mixing profiling and benchmarking is a good idea. This particular benchmark simply acts as an example of the new 'benchmarking' backend, adding nvtx markers around this line won't be useful, only raising the complexity of this example.

thunder/benchmarks/targets.py
```diff
@@ -109,7 +111,7 @@ def deco(old_timer):
         @functools.wraps(old_timer)
         def timer():
             ret = old_timer()
-            benchmark.extra_info["max_allocated_memory_MB"] = torch.cuda.max_memory_allocated() / (1024 * 1024.0)
+            benchmark.extra_info[MAX_ALLOCATED_MEMORY_KEYWORD] = torch.cuda.max_memory_allocated() / (1024 * 1024.0)
```
IvanYashchuk, line 114: Why is this change needed?

thunder/benchmarks/targets.py
IvanYashchuk: Most likely the whole lightning package is imported in this line:
https://github.com/Lightning-AI/lightning-thunder/blob/5dcbc55a0523a471fdfec26b6b098b1aeb3b1fd8/thunder/benchmarks/targets.py#L11

thunder/benchmarks/__init__.py
```diff
@@ -2958,3 +2960,185 @@ def fn(self) -> Callable:
...
+            ```
+
+            Note: Ensure the pytest configuration file (`thunder/benchmarks/conftest.py`) is present in the same directory as `script.py` to provide the grouping customization.
+            .
+            ├── script.py
+            ├── conftest.py
+
+        Usage:
+        To run the benchmark test and group the results by split module, execute the following command:
+        `pytest script.py  --benchmark-group-by='graph-by-graph:param:GraphID,param:SplitModuleName'`
```
IvanYashchuk, line 3053: Thank you for enabling the "--benchmark-group-by" option! This is great!

thunder/benchmarks/targets.py
```diff
@@ -812,3 +814,29 @@ def test_resnet50(benchmark, executor: Callable, compute_type: ComputeType):
...
 
     benchmark_for_compute_type(compute_type, benchmark, fn, args, kwargs)
+
+
+#
+# Thunder dynamo benchmarks
+#
+
+
+def test_dynamo_LlamaMLPBenchmark(benchmark):
```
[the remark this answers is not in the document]
IvanYashchuk: Is pytest-benchmark a test dependency to Thunder? I suggest putting this in the thunder/test directory because the main goal is testing `ThunderCompilerGraphBenchmarking`.

thunder/benchmarks/__init__.py
```diff
@@ -2958,3 +2960,185 @@ def fn(self) -> Callable:
...
+
+    def __init__(
+        self,
+        bench: BenchmarkFixture,
+        executors: Sequence[str],
+        **thunder_options,
+    ):
+        """
+        This class acts as a backend for the `torch.compile` function, facilitating the benchmarking of each `fx.GraphModule` produced by Thunder dynamo splitter.
+        Each `fx.GraphModule` instance is executed by the specified executors and benchmarked using `pytest_benchmark`.
```
IvanYashchuk, line 3016: The project name is "pytest-benchmark" but because Python doesn't allow "-" in the package name, in Python the import is "pytest_benchmark". In the documentation let's use the official project name.
```suggestion
        Each `fx.GraphModule` instance is executed by the specified executors and benchmarked using `pytest-benchmark`.
```

thunder/benchmarks/__init__.py
```diff
@@ -2958,3 +2960,185 @@ def fn(self) -> Callable:
...
+
+        Keyword arguments:
+            bench: the BenchmarkFixture created by `pytest_benchmark`
+            executors: List of executors to use. Supported executors include: 'eager', 'inductor', and 'thunder'. If None, defaults to all available executors.
```
IvanYashchuk, line 3020: Maybe "executors" should be changed to "backends". Since we're in the Thunder codebase the term executors means a set of executors to use within a single jitted function, but here the aim is to compare different execution modes or maybe backends: eager, inductor, thunder.
```suggestion
            executors: List of executors to compare. Supported executors include: 'eager', 'inductor', and 'thunder'. If None, defaults to all available executors.
```

thunder/benchmarks/__init__.py
```diff
@@ -2958,3 +2960,185 @@ def fn(self) -> Callable:
...
+from thunder.dynamo.compiler import ThunderCompiler
+from itertools import chain
+
+GRAPH_BY_GRAPH_BENCHMARK_PARAMS_KEYS = ("GraphID", "SplitModuleName", "executor")
+
+
+def _get_example_inputs_from_placeholder(node) -> tuple[torch.Tensor]:
+    check(node.op == "placeholder", lambda: f"The node must be placeholder type", ValueError)
+    check("example_value" in node.meta, lambda: "example_value does not exist in the meta of {node}", ValueError)
+    example_value = node.meta["example_value"]
```
IvanYashchuk, line 2974: Does it also store strides? If yes let's use it when making tensors below.

thunder/benchmarks/__init__.py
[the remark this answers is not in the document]
IvanYashchuk: > my concern is really that pyt-benchmark has made it hard to identify apples-to-apples comparisons.

Can you provide an example?

>And so to wrap that one would naturally do something like:
>```py
> def foo(args):
>   func = thunder.jit(Something())
>   func(*args)
>   torch.cuda.synchronize()
>```
> And then foo is the thing that gets passed to benchmark. 

Of course, if you set up the function-to-be-benchmarked this way you are not measuring the performance of just `func(*args)`. We use a different setup in Thunder "pytest-benchmark"-based benchmarks. `func = thunder.jit(Something())` should be created outside `foo`. `torch.cuda.synchronize()` shouldn't be in `foo`. And then all we're left with inside `foo` is `func(*args)`.

> What one really wants is for pytest-benchmark to make its decisions on the pgm(*args) line alone.

And this is what's happening with pytest-benchmark.

> before line 3103, execute fn(*sample_args) alone (without self.bench) to ensure it's compiled already when line 3103 executes the function.

Let's not do this. It's the benchmarking suite's responsibility to discard first executions that includes compilation as warmup iterations. We ask pytest-benchmark to drop the first iterations here: https://github.com/Lightning-AI/lightning-thunder/blob/952db42d71b2a2019484b05ae74ac828a6946f06/pyproject.toml#L34

> ensure we synchronize (torch.cuda.synchronize()) in whatever function we pass to the BenchmarkFixture. We may need to create a wrapper to do that? Or maybe that's happening already somewhere and it's just hidden from me?

It's hidden in the timer function. We use [`torch.utils.benchmark.timer`](https://github.com/pytorch/pytorch/blob/a063a82c8b166640268d4ebd43b5579c7bf66d01/torch/utils/benchmark/utils/timer.py#L17-L19) and we ask pytest-benchmark to use it here https://github.com/Lightning-AI/lightning-thunder/blob/952db42d71b2a2019484b05ae74ac828a6946f06/pyproject.toml#L33

IvanYashchuk (approved):

thunder/benchmarks/__init__.py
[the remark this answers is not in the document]
kiya00: Hi @tfogal 
>before line 3103, execute fn(*sample_args) alone (without self.bench) to ensure it's compiled already when line 3103 executes the function.

I think the pybenchmark warmup can take care of the first time you run fn so that when benchmarking the actually perf it's already compiled and warmed up.
 >ensure we synchronize (torch.cuda.synchronize()) in whatever function we pass to the BenchmarkFixture. We may need to create a wrapper to do that? Or maybe that's happening already somewhere and it's just hidden from me?

 the synchronize is added in https://github.com/Lightning-AI/lightning-thunder/blob/952db42d71b2a2019484b05ae74ac828a6946f06/pyproject.toml#L33-L34 the pytorch timer has the synchronize in it. But if we use it outside the thunder project the command-line option need to be explicitly passed `--benchmark-timer=torch.utils.benchmark.utils.timer.timer` cc: @IvanYashchuk 
 >ensure we get nvtx markers each time the fn is called

 I'll add an option to include the nvtx markers, like the following pattern:
 ```
def nvtx_func(args: list[tensor]):
    nvtx.push_range(...)
    func(*args)
    nvtx.pop_range(...)
benchmark(nvtx_func, args)
 ```

thunder/benchmarks/targets.py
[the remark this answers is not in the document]
kiya00: The `ThunderCompilerGraphBenchmarking` requires pytest to create the bench fixture. I'll move it to `thunder/test`

thunder/benchmarks/targets.py
kiya00: because in the `ThunderCompilerGraphBenchmarking` I query the key "max_allocated_memory_MB" in extra_info, so instead of using a hardcoded string, I change it to a constant str `MAX_ALLOCATED_MEMORY_KEYWORD`

thunder/benchmarks/__init__.py
kiya00: make_tensor probably can use the memory_format and noncontiguous

thunder/dynamo/compiler_graph_benchmark.py
```diff
@@ -0,0 +1,144 @@
...
+                                it accepts `torch_inductor_options` which are passed to :func:`torch.compile` if part of the graph
+                                is not supported by thunder.
+
+        Example:
+            .. code-block:: python
+
+                # script.py
+                import torch
+                import thunder
+                from thunder.benchmarks import ThunderCompilerGraphBenchmarking
```
crcrpar, line 48:
```suggestion
                from thunder.dynamo import ThunderCompilerGraphBenchmarking
```

kiya00:
Hi @tfogal @IvanYashchuk @crcrpar , can we merge this initial PR first, we can discuss the additional requirements in the new issue(https://github.com/Lightning-AI/lightning-thunder/issues/1258)

kiya00:
Hi @IvanYashchuk @t-vi , I think it's ready to merge, do you want to have another look?

t-vi (approved):
Thank you @kiya00 @IvanYashchuk @tfogal @crcrpar

Issue #1055, opened 2024-08-27
Task list to jit `Adam.step`


shino16:
This is a list of tasks to enable compiling `Adam.step`/`AdamW.step`.

## Goal

The expected usage is:

```py
optimizer = Adam(model.parameters())
optimizer_step = jit(optimizer.step)
# or torch.compile(optimizer.step, backend=<thunder>)

for X, y in data:
  optimizer.zero_grad()
  outputs = model(X)
  loss = loss_fn(X, y)
  loss.backward()
  optimizer_step()
```

## Task list for `torch.compile(optimizer.step, backend=<thunder>)`

We have to be able to trace through PyTorch's implementation (with `foreach=False`, `fused=False`)

- [ ] Add `TensorProxy.grad` attribute and track `Tensor.grad`
- [ ] Add `torch.lerp`

## Task list for `thunder.jit(optimizer.step)`

This can be easily achieved by applying an `OperatorExecutor` to symbolically replace `Adam.step`'s inner implementation with our own thunder-friendly implementation.

- [ ] Jit `Adam.step` with a custom operator executor

However, once we are able to `torch.compile` it with Thunder backend, I expect `thunder.jit` will be able to naturally trace through `Adam.step` without patching it.

There is a trouble, however, that `Adam.step` creates tensors as its inner state on the first call and use them on subsequent calls. This behavior cannot be tracked by the current implementation of `thunder.jit`, which traces through a given function once and takes the same path on every call.

- [ ] Workaround for `Adam.step`'s lazy initialization of its inner state

PR #1070, opened 2024-08-30
Add `TensorProxy.grad` attribute and proxify `Tensor.grad`

shino16:
Part of #1055. This is necessary to trace `Optimizer.step`.


thunder/core/proxies.py
```diff
@@ -1239,6 +1242,8 @@ def _infer_tensor_properties(
     _dtype = dtypes.numbertype_to_dtype(_dtype) if dtypes.is_numbertype(_dtype) else _dtype
     _requires_grad = requires_grad if requires_grad is not None else _requires_grad
     _requires_grad = False if not dtypes.is_inexact_dtype(_dtype) else _requires_grad
+    _grad = grad if grad is not None else _grad
+    _grad = None if not _requires_grad else _grad
```
crcrpar, line 1246: So `_grad` still can be `None`, which I think possible for multiple cases?

thunder/tests/test_core.py
```diff
@@ -238,6 +238,124 @@ def foo(a):
...
+
+    b = make_tensor((3, 3), device=device, dtype=tdtype, requires_grad=True)
+    b.grad = make_tensor((3, 3), device=device, dtype=tdtype)
+    cfoo(b)
+    assert thunder.cache_misses(cfoo) == 2
+
+
+@instantiate(dtypes=(thunder.float32,))
+def test_optimizer_unpack(executor, device, dtype):
+    class Optimizer:
```
crcrpar, line 323: Why not inheriting pytorch Optimizer class?

thunder/core/jit_ext.py
```diff
@@ -1576,7 +1576,7 @@ def from_opaque(provenance, *, new_output=False):
             raise NotImplementedError(f"unpacking from OPAQUE {fn.value} {provenance}")
 
         def from_provenance(provenance, *, new_output=False):
-            if hasattr(provenance, "proxy"):
```
mruberry, line 1579: The `hasattr` here seems fine? If you want to use `getattr`, a pattern like:

```
if (p := getattr(provenance, 'proxy', None)) is not None:
  return p
```

Might make more sense, as the `proxy` attribute is only acquired once

thunder/core/proxies.py
shino16: I did not find `TensorProxy.grad` being `None` to be problematic. For the purpose of tracing an optimizer's `_init_group`, Thunder needs to understand the idiom `if p.grad is not None:` ([`torch/optim/adam.py`](https://github.com/pytorch/pytorch/blob/30b98940b8c4f2903110fe497705ff62df5e968b/torch/optim/adam.py#L126)).

thunder/core/jit_ext.py
shino16: I changed it to `getattr` because we may later set `p.history.proxy = None` below on line 1621.

I have just learned the overhead of `getattr` is not negligible. Thanks for your suggestion!

thunder/core/proxies.py
```diff
@@ -1212,13 +1212,15 @@ def _infer_tensor_properties(
     device: devices.Device | None = None,
     dtype: dtypes.dtype | None = None,
     requires_grad: bool | None = None,
+    grad: TensorProxy | None = None,
```
IvanYashchuk, line 1215: I'm worried about having `grad` and `requires_grad` attributes on intermediate TensorProxies. Some meta function implementers may try setting these attributes but they are simply ignored by current forward-backward trace generation. An idea for the future is to have a separate TensorProxy subclass to represent input TensorProxies with `grad` and `requires_grad` attributes set.

thunder/core/proxies.py
```diff
@@ -1354,6 +1361,10 @@ def true_dtype(self):
     def requires_grad(self):
         return self._requires_grad
 
+    @property
+    def grad(self):
+        return None  # FutureTensorProxies never require grad
+
```
IvanYashchuk, line 1367: Let's raise an error for this case.

thunder/tests/test_core.py
IvanYashchuk, line 1: I know it's tested indirectly with the optimizer but let's add another simple test case with an inplace update of the `.grad` tensor.

IvanYashchuk (approved):

thunder/core/jit_ext.py
mruberry: @jjsjann123 and I were just talking about the cost of accessing an attribute of an object. It probably takes between .1us and 1us to do so

thunder/core/proxies.py
shino16: When making a `TensorProxy` or a `FutureTensorProxy` that reflects another `TensorLike`, we copy its metadata and `grad`.
https://github.com/Lightning-AI/lightning-thunder/pull/1070/files#diff-f25d3a604208b167d7f7882a364dd30b8bb9fe395a6cb05b9ed1ffc500ff8e87R1227-R1234
Although `FutureTensorProxy` simply ignores what `_infer_tensor_properties` infers as `grad` ([ignored here](https://github.com/Lightning-AI/lightning-thunder/pull/1070/files#diff-f25d3a604208b167d7f7882a364dd30b8bb9fe395a6cb05b9ed1ffc500ff8e87R1321)), setting `FutureTensorProxy.grad` as `None` keeps this logic easy.

This makes sense to me, because `FutureTensorProxy.requires_grad` is `False` anyway.

thunder/tests/test_core.py
shino16: This PR is not meant for supporting mutation on the `.grad` tensor. The new optimizer test only confirms that the interpreter can recognize the tensor `param.grad` and use it to update `param`.

In the future, we may be able to jit a wider region including both `loss.backward()` and `optimizer.step()`, but now Thunder cannot interpret the following:
```py
@thunder.jit
def f(x):
    x.grad = torch.zeros_like(x)
    #   File "/opt/pytorch/lightning-thunder/thunder/core/interpreter.py", line 6535, in _call_dispatch
    #     opaque_result: Any = fn(*args_, **kwargs_)
    # AttributeError: can't set attribute 'grad'
```

In-place operations on existing `.grad` tensors work fine, but mutating `.grad` does not seem like a complete feature subject to tests.

thunder/core/proxies.py
shino16: Indeed, the arguments of `TensorProxy.__init__` expose too much implementation details. It could be a good idea to provide a separate constructor or a parent class with less attributes for use in meta functions.

thunder/tests/test_core.py
IvanYashchuk: I'm sorry for the confusion by inplace update I meant inplace update of tensor elements, not assigning to the `x.grad` variable.

thunder/core/jit_ext.py
```diff
@@ -1254,8 +1254,8 @@ def from_opaque(provenance, *, new_output=False):
...
-                return provenance.proxy  # bind?
+            if (p := getattr(provenance, "proxy", None)) is not None:
```
t-vi, line 1257: Nit: Can we not use := just to save a vertical line of code, please? It's cool in elif changes, but here...

shino16: Sure. I saw other uses of `:=` like this in this repo, but this is not really necessary.

shino16:
I'm sorry for missing a comment! This is ready for review and approval (hopefully).

t-vi (approved):
Thank you @shino16 @IvanYashchuk @crcrpar @mruberry

PR #1073, opened 2024-08-30
use `torch.OP##_` if available not `torch.Tensor.OP##_`

### LINKED ISSUE #1072: `torch.exp_` is not suppurted, but `torch.Tensor.exp_` is

### To Reproduce

```python
import torch

def f(a):
    return torch.exp_(a)


def g(a):
    return a.exp_()


a = torch.randn(2)
examine(f, a)
examine(g, a)
```

then `python a.py` gives
```
Found 1 distinct operations, of which 0 (0.0%) are supported
Please file an issue requesting the following operators here: https://github.com/Lightning-AI/lightning-thunder/issues/new
_VariableFunctionsClass.exp_ of torch
Found 1 distinct operations, of which 1 (100.0%) are supported
The function appears to be working as expected
```

crcrpar:
## What does this PR do?

Fixes #1072.

```python
import torch

import thunder

def f(a):
    return torch.exp_(a)

def g(a):
    return a.exp_()

a = torch.randn(2)
b = thunder.jit(f)(a)
b = thunder.jit(g)(a)
```

Current main is not capable of functionalizing `f`, which should be unexpected.

### DIFF
--- a/thunder/torch/__init__.py
+++ b/thunder/torch/__init__.py
@@ -1330,7 +1330,7 @@ def abs(a: NumberLike | TensorLike, /) -> Number | TensorLike:
     return clang.abs(a)
 
 
-@torchsymbol(torch.Tensor.abs_, is_method=True, tags=(prims.OpTags.IN_PLACE,))
+@torchsymbol(torch.abs_, is_method=True, tags=(prims.OpTags.IN_PLACE,))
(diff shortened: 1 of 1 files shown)


t-vi (approved):
Thank you @crcrpar

PR #1090, opened 2024-09-03
Remove thunder+nvfuser+torch.compile executor from default list in targets.py

IvanYashchuk:
Since https://github.com/Lightning-AI/lightning-thunder/pull/974 torch.compile concatenation executor is part of the default executor list and it's the same as "Thunder" executor option in these benchmarks.

cc @crcrpar


t-vi (approved):
Thank you @IvanYashchuk

Issue #1091, opened 2024-09-03
AttributeError for Dynamo+Thunder+FP8 and Phi-3-mini-4k-instruct  'GraphModule' object has no attribute 'real_recompile'


mpatel31415:
## 🐛 Bug
When training Dynamo+Thunder + FP8 we get an attribute error:

> AttributeError: \'GraphModule\' object has no attribute \'real_recompile\'

I'm aware that it's not exected to work yet, but I don't see similar issue, so I'm adding it to track the progress.

Models affected:
'stablecode-completion-alpha-3b', 'Phi-3-mini-4k-instruct', 'pythia-14m', 'Mistral-7B-v0.1', 'dolly-v2-3b', 'falcon-7b', 'tiny-llama-1.1b'

### To Reproduce

Start interactive job on a cluster:
srun -A YOUR_SLURM_ACCOUNT -J YOUR_SLURM_ACCOUNT-thunder.lit-gpt -N1 -p batch --container-image=INTERNAL_IMAGE:pjnl-20240830 --pty bash

Then execute:
torchrun --standalone --max-restarts=0 --no-python --nproc-per-node=8 python /opt/pytorch/lightning-thunder/thunder/benchmarks/benchmark_litgpt.py \
    --model_name Mistral-7B-v0.1 \
    --distributed_mode ddp \
    --shard_mode None \
    --compile dynamo_thunder \
    --checkpoint_activations False \
    --low_precision_mode fp8-delayed-te  \
    --micro_batch_size 1

### Expected behavior

We should be able to run the training or get OOM error. 

### Environment

system.device_product_name                      DGXH100
system.gpu_driver_version                    535.129.03
libraries.cuda                               12.6.1.005
libraries.pip.lightning               2.4.0.dev20240728
libraries.pip.lightning-thunder              0.2.0.dev0
libraries.pip.lightning-utilities                0.11.6
libraries.pip.litgpt                             0.4.11
libraries.pip.nvfuser                 0.2.10+git58dfdc1
libraries.pip.pytorch-lightning                   2.4.0
libraries.pip.torch                  2.5.0a0+git578b8d7
libraries.pip.torchmetrics                        1.4.1
libraries.pip.torchvision              0.19.0a0+d23a6e1

IvanYashchuk:
[...]

PR #1094, opened 2024-09-03
thunder + dynamo : use recompile instead of real_recompile

kshitij12345:
Fixes: https://github.com/Lightning-AI/lightning-thunder/issues/1091

Depending on the setting, Dynamo's output_graph maybe `GraphModule` or it's derived class `_LazyGraphModule` (only this subclass provides `real_recompile` method).  The difference between the two classes is that calling `recompile` on `_LazyGraphModule` just marks it for recompilation and it actually does recompilation when a code that will observe it is called like `forward` method or `code` attribute.

In this PR, we call `recompile` followed by `code` attribute which is supported by both classes.

Doc for `_LazyGraphModule` -
https://github.com/pytorch/pytorch/blob/39935e0fdef02c67ba808175dcc800d0695bfe1b/torch/fx/_lazy_graph_module.py#L38-L62
Point where dynamo may create `GraphModule` or `_LazyGraphModule` -
https://github.com/pytorch/pytorch/blob/39935e0fdef02c67ba808175dcc800d0695bfe1b/torch/_dynamo/output_graph.py#L1331

Test - new test in `thunder/tests/test_dynamo.py` and `thunder/tests/test_networks.py -k Dynamo`.

### DIFF
--- a/thunder/dynamo/compiler.py
+++ b/thunder/dynamo/compiler.py
@@ -12,6 +12,14 @@ def _warn_thunder_compiler():
     )
 
 
+def recompile_graph(gm: torch.fx.GraphModule):
+    # NOTE - `gm` could also be the `_LazyGraphModule`, in which case calling `recompile` is not enough as it marks the `GraphModule`
+    # and actual recompilation happens when either `forward` or `code` (or when user tries to observe the GraphModule).
+    # See for more details - https://github.com/pytorch/pytorch/blob/39935e0fdef02c67ba808175dcc800d0695bfe1b/torch/fx/_lazy_graph_module.py#L65-L89
+    gm.recompile()
+    gm.code
+
+
 class ThunderCompiler:
     def __init__(self, **thunder_options):
         """
@@ -63,7 +71,7 @@ class ThunderCompiler:
 
         # Dynamo uses lazy generation of the underlying Python code, so we need to
         # force recompilation of the GraphModule before passing it to Thunder.
-        gm.real_recompile()
+        recompile_graph(gm)
 
         # Here in the future we could add some logic to check if the GraphModule
         # is executable by Thunder, but for now we simply compile it and return
--- a/thunder/dynamo/compiler.py
+++ b/thunder/dynamo/compiler.py
@@ -1,4 +1,5 @@
 import torch
+from torch.fx._lazy_graph_module import _LazyGraphModule
 import warnings
 
 from thunder.core.baseutils import run_once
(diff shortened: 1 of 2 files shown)


thunder/dynamo/compiler.py
```diff
@@ -12,6 +12,14 @@ def _warn_thunder_compiler():
     )
 
 
+def recompile_graph(gm: torch.fx.GraphModule):
+    # NOTE - `gm` could also be the `_LazyGraphModule`, in which case calling `recompile` is not enough as it marks the `GraphModule`
+    # and actual recompilation happens when either `forward` or `code` (or when user tries to observe the GraphModule).
+    # See for more details - https://github.com/pytorch/pytorch/blob/39935e0fdef02c67ba808175dcc800d0695bfe1b/torch/fx/_lazy_graph_module.py#L65-L89
+    gm.recompile()
+    gm.code
```
kshitij12345, line 20: Other option is 

```python
if isinstance(gm, torch.fx._lazy_graph_module._LazyGraphModule):
     return gm.real_recompile()
return gm.recompile()
```

IvanYashchuk:
What are the conditions for Dynamo to generate LazyGraphModule and plain GraphModule?

thunder/dynamo/compiler.py
IvanYashchuk: And one more variant to choose from:
```py
from torch.fx._lazy_graph_module import _LazyGraphModule
_LazyGraphModule.force_recompile(gm)
```

thunder/tests/test_dynamo.py
```diff
@@ -55,3 +56,32 @@ def func(x):
 
     # This shouldn't be empty
     assert last_traces(thunder_func)
+
+
+@instantiate(
+    dtypes=NOTHING,
+    executors=[DynamoThunderExecutor],
+)
+def test_recompile(executor, device: str, dtype: dtypes.dtype):
```
IvanYashchuk, line 65: "recompile" is too ambiguous, how about "test_force_skip_lazy_graph_module"?

thunder/dynamo/compiler.py
IvanYashchuk: Accessing code is a clever solution because real_recompile is triggered in _LazyGraphModule 
https://github.com/pytorch/pytorch/blob/39935e0fdef02c67ba808175dcc800d0695bfe1b/torch/fx/_lazy_graph_module.py#L176 but it's better to rely on explicit calls to do recompilation so it's better to choose the code snippet you posted or the one that I posted.

IvanYashchuk (approved):

kshitij12345:
> What are the conditions for Dynamo to generate LazyGraphModule and plain GraphModule?

For the original issue in question because of using DDP, we were going via this path below which returns a GraphModule instead of _LazyGraphModule.

https://github.com/pytorch/pytorch/blob/39935e0fdef02c67ba808175dcc800d0695bfe1b/torch/_dynamo/backends/distributed.py#L514-L516

thunder/tests/test_dynamo.py
```diff
@@ -6,6 +6,7 @@
 
 import torch
 import pytest
+from unittest.mock import patch
```
crcrpar, line 9: noob question: where is this `patch` utilized?

kshitij12345: Good catch, I forgot to remove this. I had initially used it to disable `LazyGraphModule` usage but later found that there was a ctx manager. Thanks!

crcrpar (approved):

kshitij12345:
Ping @t-vi for approval.

mruberry (approved):

PR #1095, opened 2024-09-03
Add allow_cpu_scalar_tensors for shape,device check and broadcast functions

### LINKED ISSUE #941: Devices were expected to be the same, but got ... when Benchmarking Gemma-7b with Micro Batch Size 1
## 🐛 Bug

When using `benchmark_litgpt` Following error occurs:

```
Exception: Unexpected error occurred for {\'--micro_batch_size\': 1} due to [W803 20:39:12.165884696 socket.cpp:752] [c10d] The client socket has failed to connect to [eos0157.eos.clusters.nvidia.com]:59504 (errno: 22 - Invalid argument).
[rank4]: RuntimeError: Devices were expected to be the same, but got devices thunder.devices.Device(type=\'cuda\', index=4) and thunder.devices.Device(type=\'cpu\')!
```

### To Reproduce

Start interactive job on a cluster:

```
srun -A YOUR_SLURM_ACCOUNT -J YOUR_SLURM_ACCOUNT-thunder.lit-gpt -N1 -p batch --container-image=INTERNAL_IMAGE:pjnl-20240801 --pty bash
```

Then execute:

```
torchrun --standalone --max-restarts=0 --no-python --nproc-per-node=8 python /opt/pytorch/lightning-thunder/thunder/benchmarks/benchmark_litgpt.py \
    --model_name Gemma-7b \
    --distributed_mode fsdp \
    --shard_mode zero2 \
    --compile thunder_cudnn \
    --checkpoint_activations False \
    --low_precision_mode fp8-delayed-te \
    --save_logs_for_all_batches True \
    --micro_batch_size 1
```

### Expected Behavior
When running the benchmark for the Gemma-7b model with the specified configurations, the benchmarking scripts should successfully execute without any errors. 

### Environment

```
system.device_product_name                      DGXH100
system.gpu_driver_version                    535.129.03
libraries.cuda                               12.6.0.021
libraries.pip.lightning               2.4.0.dev20240728
libraries.pip.lightning-thunder              0.2.0.dev0
libraries.pip.lightning-utilities                0.11.6
libraries.pip.litgpt                              0.4.7
libraries.pip.nvfuser                  0.2.8+git671171f
libraries.pip.pytorch-lightning                   2.3.3
libraries.pip.torch                  2.5.0a0+gita94e507
libraries.pip.torchmetrics                  1.4.0.post0
libraries.pip.torchvision              0.19.0a0+d23a6e1
```

### Additional context

Comment: seems like a thunder bug somewhere; something is somehow ending up on the CPU even though it shouldn’t be.

kiya00:
<details>

</details>

## What does this PR do?

Fixes #941.

## PR review

## Did you have fun?

### DIFF
--- a/thunder/core/transforms.py
+++ b/thunder/core/transforms.py
@@ -2362,6 +2362,8 @@ if torch.distributed.is_available():
 
 
 def sum_to(a: TensorProxy, shape: Sequence[int]) -> TensorProxy:
+    if utils.same_shape(a.shape, shape):
+        return a
     if not shape:
         return a.sum()
     leading_dims = a.ndim - len(shape)
@@ -2745,6 +2747,14 @@ def vjp(func):
         flat_func, flat_args, spec = flatten_func(func, primals, kwargs)
         trace = construct_trace()(flat_func, *flat_args)
         result, vjp_result = vjp_call(flat_args, cotangents, trace=trace)
+        vjp_result = tuple(
+            (
+                sum_to(grad, arg.shape)
+                if (grad is not None and isinstance(arg, TensorProxy) and arg.device.type == "cpu")
+                else grad
+            )
+            for grad, arg in zip(vjp_result, flat_args)
+        )
         gprimals, gkwargs = tree_unflatten(vjp_result, spec)
         grads = gprimals + (gkwargs,) if len(gkwargs) != 0 else gprimals
         return result, grads
--- a/thunder/core/transforms.py
+++ b/thunder/core/transforms.py
@@ -1633,6 +1633,9 @@ def zeta_backward(x, y, g):
     # Therefore, we compute only the derivative wrt the second argument
     gy = g * -x * prims.zeta(x + 1.0, y)
 
+    # if x is CUDA tensor and y is CPU scalar tensor, gy should be on CPU
+    if gy is not None and gy.device.type != y.device.type:
+        gy = gy.to(device=y.device)
     # Return a mappping from the forward arguments to the gradients
     return {"y": gy}
 
@@ -2747,6 +2750,7 @@ def vjp(func):
         flat_func, flat_args, spec = flatten_func(func, primals, kwargs)
         trace = construct_trace()(flat_func, *flat_args)
         result, vjp_result = vjp_call(flat_args, cotangents, trace=trace)
+        # If the argument is a CPU scalar tensor, its gradient needs to be summed into a scalar tensor.
         vjp_result = tuple(
             (
                 sum_to(grad, arg.shape)
--- a/thunder/core/utils.py
+++ b/thunder/core/utils.py
@@ -562,8 +562,10 @@ def same_shape(a: Sequence[int], b: Sequence[int], /) -> bool:
 
 
 # TODO: improve error message
-def check_same_shape(*args):
+def check_same_shape(*args, allow_cpu_scalar_tensors=False):
     shapes = tuple(x.shape for x in args if isinstance(x, TensorProxy))
+    if allow_cpu_scalar_tensors:
+        shapes = tuple(x.shape for x in args if isinstance(x, TensorProxy) and not is_cpu_scalar_tensor(x))
     if len(shapes) > 1:
         shape = shapes[0]
         for othershape in shapes[1:]:
@@ -658,6 +660,10 @@ def check_no_duplicates(dims: Sequence):
     check(len(dims) == len(set(dims)), lambda: f"Duplicate value in list of dimensions {dims}!")
 
 
+def is_cpu_scalar_tensor(a: Any) -> bool:
+    return isinstance(a, TensorProxyInterface) and a.ndim == 0 and a.device.devicetype == devices.DeviceType.CPU
+
+
 #
 # Device-related functions
 #
@@ -665,9 +671,14 @@ def check_no_duplicates(dims: Sequence):
 
 # TODO: improve device handling by canonicalizing devices and expressing them per langctx
 # TODO: should the comparison between devices be ==?
-def check_same_device(*args):
+def check_same_device(*args, allow_cpu_scalar_tensors=False):
     devices = tuple(x.device for x in args if isinstance(x, TensorProxyInterface) and x.device.type != "meta")
-
+    if allow_cpu_scalar_tensors:
+        devices = tuple(
+            x.device
+            for x in args
+            if isinstance(x, TensorProxyInterface) and x.device.type != "meta" and not is_cpu_scalar_tensor(x)
+        )
     if len(devices) > 1:
         device = devices[0]
         for otherdevice in devices[1:]:
--- a/thunder/core/utils.py
+++ b/thunder/core/utils.py
@@ -662,8 +662,11 @@ def check_no_duplicates(dims: Sequence):
     check(len(dims) == len(set(dims)), lambda: f"Duplicate value in list of dimensions {dims}!")
 
 
-def is_cpu_scalar_tensor(a: Any) -> bool:
-    return isinstance(a, TensorProxyInterface) and a.ndim == 0 and a.device.devicetype == devices.DeviceType.CPU
+def is_cpu_scalar_tensor(a: Any, /) -> bool:
+    """
+    Returns True if the input is a Thunder TensorProxy or torch.tensor and is a scalar tensor on the CPU.
+    """
+    return isinstance(a, (TensorProxyInterface, torch.Tensor)) and a.ndim == 0 and a.device.type == "cpu"
 
 
 #
--- a/thunder/core/utils.py
+++ b/thunder/core/utils.py
@@ -562,9 +562,9 @@ def same_shape(a: Sequence[int], b: Sequence[int], /) -> bool:
 
 
 # TODO: improve error message
-# NOTE: the tensors in args have different shapes is only allowed if the tensor is a CPU scalar tensor and allow_cpu_scalar_tensors=True
-def check_same_shape(*args, allow_cpu_scalar_tensors=False):
-    if allow_cpu_scalar_tensors:
+# NOTE: the tensors in args have different shapes is only allowed if the tensor is a CPU scalar tensor and treat_cpu_scalar_tensors_as_numbers=True
+def check_same_shape(*args, treat_cpu_scalar_tensors_as_numbers=True):
+    if treat_cpu_scalar_tensors_as_numbers:
         shapes = tuple(x.shape for x in args if isinstance(x, TensorProxy) and not is_cpu_scalar_tensor(x))
     else:
         shapes = tuple(x.shape for x in args if isinstance(x, TensorProxy))
@@ -673,9 +673,9 @@ def is_cpu_scalar_tensor(a: Any) -> bool:
 
 # TODO: improve device handling by canonicalizing devices and expressing them per langctx
 # TODO: should the comparison between devices be ==?
-# NOTE: the tensors in args have different devices is only allowed if the tensor is a CPU scalar tensor and allow_cpu_scalar_tensors=True
-def check_same_device(*args, allow_cpu_scalar_tensors=False):
-    if allow_cpu_scalar_tensors:
+# NOTE: the tensors in args have different devices is only allowed if the tensor is a CPU scalar tensor and treat_cpu_scalar_tensors_as_numbers=True
+def check_same_device(*args, treat_cpu_scalar_tensors_as_numbers=True):
+    if treat_cpu_scalar_tensors_as_numbers:
         devices = tuple(
             x.device
             for x in args
--- a/thunder/core/utils.py
+++ b/thunder/core/utils.py
@@ -562,10 +562,12 @@ def same_shape(a: Sequence[int], b: Sequence[int], /) -> bool:
 
 
 # TODO: improve error message
+# NOTE: the tensors in args have different shapes is only allowed if the tensor is a CPU scalar tensor and allow_cpu_scalar_tensors=True
 def check_same_shape(*args, allow_cpu_scalar_tensors=False):
-    shapes = tuple(x.shape for x in args if isinstance(x, TensorProxy))
     if allow_cpu_scalar_tensors:
         shapes = tuple(x.shape for x in args if isinstance(x, TensorProxy) and not is_cpu_scalar_tensor(x))
+    else:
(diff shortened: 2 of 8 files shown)


t-vi:
Thank you, @kiya00 , for looking into this! I think having a `allow_cpu_scalar_tensors` is a good ingredient for the solution, but in addition we need to take `scalar.item()` and then process treating it as a number.
The tricky part will be that this needs to either
- be *inside* the prims because we cannot leak the .item() value into the trace, or
- building on  @jjsjann123 's number proxy work, make the .item() return a number proxy and make that work.

I imagine that that either option will be not be terribly easy.
  
The alternative (quite inferior to my mind) is that we need to move it to GPU and broadcast to treat it as a tensor (to my mind not desirable).

jjsjann123:
> we cannot leak the .item() value into the trace

QQ: what's the issue with having item in the trace? I thought we do have PrimIDs.ITEM, which does exactly that?

t-vi:
> QQ: what's the issue with having item in the trace? I thought we do have PrimIDs.ITEM, which does exactly that?

Sorry, I was too terse: we can have  `.item()` in the trace, we cannot have the returned value _as a constant value_ in the trace.

jjsjann123:
> > QQ: what's the issue with having item in the trace? I thought we do have PrimIDs.ITEM, which does exactly that?
> 
> Sorry, I was too terse: we can have `.item()` in the trace, we cannot have the returned value _as a constant value_ in the trace. So if we want to have the `.item()` in the trace, the return value needs to be a number proxy.

Got'ya.

Do we *need* to lift cpu scalar tensor into numberproxy? If that's going to be treated as a dynamic thing, why not just leave it as-is?
Looking at the changes in this draft PR, why would we have an extra field to *allow* cpu scalar? Are we seeing any errors when we stopped broadcasting cpu scalar tensors?

My naive understanding is that the changes here should be the default behavior. PyTorch has been treating cpu scalar tensors as *scalars*. So I would expect backends to play well with that already. cc'ing @Priya2698 for visibility.

t-vi:
> Do we need to lift cpu scalar tensor into numberproxy? If that's going to be treated as a dynamic thing, why not just leave it as-is?

To my mind, PyTorch's handling of cpu scalar->scalar is similar to upcasting or broadcasting of tensors and implicit.
It then takes the overloads with Scalar args, not the Tensor ones:
So for the `torch.add(a, b)`, PyTorch will call a 2-tensor `vectorized_elementwise` kernel if b is a number or a cpu scalar tensor and a 3-tensor kernel if b is a cuda tensor (different ones depending on whether there is broadcasting).

kiya00:
>Do we need to lift cpu scalar tensor into numberproxy? If that's going to be treated as a dynamic thing, why not just leave it as-is?
Looking at the changes in this draft PR, why would we have an extra field to allow cpu scalar? Are we seeing any errors when we stopped broadcasting cpu scalar tensors?

Before the changes in this draft, the binary operator broadcast the cpu scalar tensor and check the same device and same shape, the changes try to skip those for cpu scalar tensor in the meta function propagation, and in the runtime since the torch/nvfuser backend can take care of the binary ops using the mix of cpu scalar tensor and cuda tensor(is that right?). Then maybe we can leave it as-is instead of take out the value as numberproxy? what do you think? @t-vi @jjsjann123

mruberry:
Hey @kiya00!

It seems like we have two options for dealing with this issue:

1. We can treat cpu tensors with zero dimensions like numbers in most cases
2. We can convert scalar cpu tensors to numbers in torch operations

The first option requires the primitives be able to understand some cpu tensors as numbers. This is not a totally unreasonable request to make of them. Instead of saying "allow_cpu_scalar_tensors", an argument like "treat_cpu_scalar_tensors_as_numbers" might be more accurate. We should be careful to note that type promotion is different for cpu tensors with zero dimensions and numbers, because the former has an actual tensor dtype. Our current type promotion probably handles this incorrectly, but maybe we could update

https://github.com/Lightning-AI/lightning-thunder/blob/f945409580e555b010f5272eb19dcba744ae3196/thunder/core/dtypes.py#L311

so that it returns a weak dtype for any cpu scalar tensor? We want the behavior to mimic PyTorch's in cases like:

```
a = torch.randn(1, dtype=torch.float16).squeeze()
b = torch.randn((2, 2), dtype=torch.bfloat16)

# The tensor dtype of b is used because a is a cpu scalar tensor
a + b
tensor([[-0.5312, -3.5781],
        [-1.3906,  0.1211]], dtype=torch.bfloat16)

# The dtype of a is the result, not float32 as would be the case if 1. + 2 were turned into a tensor
a + 2
tensor(1.1211, dtype=torch.float16)
```

Alternatively, we could change the torch operations or maybe the clang operations to convert scalar cpu tensors into number proxies. There are a couple challenges with this, however. First, it requires adding code to a lot of operations. Second, the conversion to a number would lose the tensor dtype information, because our number proxies currently only have number types (fyi @jjsjann123, it could be interesting to think about changing this). Converting tensors to number proxies might also require updating the conversion from thunder objects to torch objects, because a numberproxy with a tensor dtype would have to be wrapped in a cpu tensor before being returned to the practitioner.

I think either approach is possible. I guess my question is how easy would it be for executors, like nvFuser, to use a scalar cpu tensor as a number? @kevinstephano @jjsjann123. If that is easy, then using the approach in this PR, with `treat_cpu_scalar_tensors_as_numbers`, is probably the simpler approach. We should add tests to ensure it works and file an issue for nvfuser support. Maybe there are some custom tests we can write and extend some OpInfo sample inputs?

If we take this approach, then will operations like adding a cuda tensor and a cpu scalar tensor always be dispatched to the PyTorch executor and not the nvFuser executor, because one of the inputs is a cpu scalar tensor?

jjsjann123:
> I guess my question is how easy would it be for executors, like nvFuser, to use a scalar cpu tensor as a number?

nvfuser today already support scalar cpu tensor. Hence my earlier suggestion to just leave it as-is and just update the broadcast/type promotion in proxy to properly handle cpu scalar tensors. :)

another question I have is re: `allow_cpu_scalar_tensors` vs `treat_cpu_scalar_tensors_as_numbers`

I don't see the reason of having these field in the first place. Shouldn't it be a generic feature for promotion to handle? IMHO. if there are operations that doesn't allow cpu scalar tensor, the assert should happen inside the specific operation instead.

> do like PyTorch and make it explicit in the trace" has worked well for us with broadcasting and upcasting

Re: @t-vi 's earlier comment, I don't disagree with this. However, in the cpu scalar specific things, I don't see a clear win by explicitly converting it to a number proxy in the trace. I expect that's a lot more work up front to tackle, comparing with fixing broadcast/type promotion, which can be done incrementally.

thunder/core/utils.py
```diff
@@ -562,8 +562,12 @@ def same_shape(a: Sequence[int], b: Sequence[int], /) -> bool:
 
 
 # TODO: improve error message
-def check_same_shape(*args):
-    shapes = tuple(x.shape for x in args if isinstance(x, TensorProxy))
+# NOTE: the tensors in args have different shapes is only allowed if the tensor is a CPU scalar tensor and allow_cpu_scalar_tensors=True
+def check_same_shape(*args, allow_cpu_scalar_tensors=False):
```
jjsjann123, line 566: same question here, when would we want to assert same device for cpu scalar tensors?

jjsjann123 (commented):
A nitpick, do we want to keep the flags around in the utility or maybe rather having that special handling on the call site instead?

thunder/core/utils.py
kiya00: Well, I'm not sure, this function is already used in many places, and before this change they all assert the same device for cpu scalar tensors, I saw your comments that this should be the default behavior, but I don't know if I set it to default if any current test cases in Thunder will fail(currently this change only affects binary ops). I'll try it.

mruberry:
> Thanks a lot for the awesome summary and listing out options @t-vi @mruberry 👏
> 
> > I guess my question is how easy would it be for executors, like nvFuser, to use a scalar cpu tensor as a number?
> 
> nvfuser today already support scalar cpu tensor. Hence my earlier suggestion to just leave it as-is and just update the broadcast/type promotion in proxy to properly handle cpu scalar tensors. :)

Great!

> another question I have is re: `allow_cpu_scalar_tensors` vs `treat_cpu_scalar_tensors_as_numbers`
> 
> I don't see the reason of having these field in the first place. Shouldn't it be a generic feature for promotion to handle? IMHO. if there are operations that doesn't allow cpu scalar tensor, the assert should happen inside the specific operation instead.

I agree; having no option (or defaulting the option to on if it's required for some reason) sounds good.

thunder/core/transforms.py
```diff
@@ -3478,6 +3480,14 @@ def _vjp(primals, cotangents, **kwargs):
         flat_func, flat_args, spec = flatten_func(func, primals, kwargs)
         trace = construct_trace()(flat_func, *flat_args)
         result, vjp_result = vjp_call(flat_args, cotangents, trace=trace)
+        vjp_result = tuple(
```
kiya00, line 3483: Hi @IvanYashchuk , when calculating `binary_ops(cpu_scalar_tensor, cuda_tensor)`, we want to skip the broadcast for cpu_scalar_tensor, and in the runtime backend(nvfuser/torch) can take the mix of cpu_scalar_tensor and cuda_tensor to do the calculation. The problem is in the backward, the bwd of broadcast will generate a sum for the case like cpu_scalar_tensor, now the broadcast of cpu_scalar_tensor is ignored in forward, this sum doesn't appear in the bwd anymore so the grad of scalar cpu tensor is wrong.
I tried to do the sum here for grad of cpu scalar tensor

thunder/core/transforms.py
```diff
@@ -2373,6 +2373,9 @@ def zeta_backward(x, y, g):
     # Therefore, we compute only the derivative wrt the second argument
     gy = g * -x * prims.zeta(x + 1.0, y)
 
+    # if x is CUDA tensor and y is CPU scalar tensor, gy should be on CPU
+    if gy is not None and gy.device.type != y.device.type:
```
mruberry, line 2377: fyi @IvanYashchuk -- an interesting special case, @kiya00

thunder/core/utils.py
```diff
@@ -658,16 +662,27 @@ def _reify(x):
     check(len(dims) == len(set(dims)), lambda: f"Duplicate value in list of dimensions {dims}!")
 
 
+def is_cpu_scalar_tensor(a: Any) -> bool:
```
mruberry, line 665: @kiya00, would you add a doc string explaining this function?

nit: let's require `a` be a position argument

```
def is_cpu_scalar_tensor(a: Any, /) -> bool:
```

thunder/tests/opinfos.py
```diff
@@ -1818,6 +1818,16 @@ def elementwise_binary_prims_generator(op, device, dtype, requires_grad, **kwarg
 
     yield SampleInput(a, b)
 
+    a = make_tensor((4, 4), device=device, dtype=dtype, requires_grad=requires_grad, **kwargs)
```
mruberry, line 1821: Would you add a comment explaining what's being tested in each of these new cases?

thunder/tests/test_grad.py
```diff
@@ -258,15 +258,29 @@ def _replace_none_with_zero(x, y):
...
 
 
+def _is_cpu_scalar_tensor(x):
+    return isinstance(x, torch.Tensor) and x.device.type == "cpu" and x.ndim == 0
+
+
+# If one tensor is a CPU scalar tensor and the other is on CUDA, move the scalar tensor to CUDA
+# Then do the ravel and dot operation
+def _tensor_dot(x, y):
+    if _is_cpu_scalar_tensor(x) and y.is_cuda:
```
mruberry, line 277: Can this call the function defined in utils instead of defining a new one?

mruberry:
I think the test error is unrelated:

```
b = tensor([[                  0,                   0,                   0,
                           0],
        [      ...       [                  0,                   0,                   0,
                           0]], device='cuda:0')

>       lambda a, b: comp(a, b, equal_nan=True),
    )
E   AssertionError: Tensor-likes are not equal!
E   
E   Mismatched elements: 1 / 16 (6.2%)
E   Greatest absolute difference: 9223372036854775807 at index (2, 3)
E   Greatest relative difference: 1.0 at index (2, 3)

thunder/tests/test_ops.py:84: AssertionError
```

That's quite a big error! I wonder what's going on with that test. I filed https://github.com/Lightning-AI/lightning-thunder/issues/1130 to track it.

mruberry (approved):
LGTM! Nice focus on the details, @kiya00. 

@jjsjann123, @IvanYashchuk would you like to take a look?

thunder/tests/test_grad.py
kiya00: I modified the function in utils to detect cpu scalar tensor for both types torch.Tensor and TensorProxy (I think it's no harm to also check torch.Tensor in utils in this case?), so then we can also use it here

thunder/core/transforms.py
jjsjann123: good catch!

jjsjann123 (approved):
LGTM

kiya00:
Hi @IvanYashchuk , could you take a look if it's ready to merge?

thunder/core/transforms.py
IvanYashchuk: Usually, the gradient tensors mirror their tensor properties. Seems correct that `gy` is moved to `y.device`.

thunder/core/transforms.py
IvanYashchuk: Thank you for the explanation! It makes sense and the approach is correct.

thunder/tests/test_ops.py
```diff
@@ -34,6 +34,7 @@ def test_errors(op, device, dtype, executor, comp):
 # TODO: should snippets be able to access the original opinfo? -- No?
 # TODO: revisit atol/rtol, maybe be more selective about which ops need a more permissive check
 def snippet_torch_consistency(op: OpInfo, torch_op, sample: SampleInput, comp: Callable):
+    print("-----", sample.args)
```
IvanYashchuk, line 37:
```suggestion
```

IvanYashchuk (approved):
Thank you for updating all the code parts that broke when moving to allow scalar CPU tensors.
Looks great!

thunder/tests/test_ops.py
kiya00: Thank you!! I forgot to remove this

PR #1096, opened 2024-09-04
nvfuserex to leverage static shape in trace

jjsjann123:
Updating `nvfuserex_impl.py::compute_symbolic_shape`

This function translate tensor inputs to nvfuser's FusionDefinition. The old logic blindly converts non-1 value as dynamic sized in fusion, despite Thunder trace keeps static size TensorProxy.

The updated logic takes TensorProxy.shape as well as Tensor.size into consideration. It now only translate a dimension to be dynamic shape when: 1. TensorProxy.shape entry is a NumberProxy; 2. Tensor.size entry is not `1`.

Impact from this change:
* Pros
1. Better performance in generated kernels, since analysis/indexing can be simplified with static shape.
2. Compatible with dynamic shape support.
* Cons
1. FusionDefinition cache hit is going to be impacted. Even though thunder today doesn't support dynamic shape, nvfuser's Fusion cache does and could re-use compiled FusionDefinition if we choose to compile for dynamic shape.

TODO:

end-2-end testing is not possible yet, unless we plumb through some toy example with dynamic shapes. I have locally verified the behavior. I will add end-2-end test after #1027


thunder/executors/nvfuserex_impl.py
```diff
@@ -289,21 +288,39 @@ def translate_bound_symbol(bsym: BoundSymbol) -> Any:
...
         shape (Union[torch.Size, Sequence[int]]): The shape of the tensor.
 
     Returns:
-        Tuple[int, ...]: The symbolic shape of the tensor.
+        Tuple[int, ...]: The shape of the tensor for FusionDefinition.
     """
-    return tuple(1 if l == 1 else -1 for l in shape)
+
+    # TODO: what's the expected behavior for the cache?
+    # if we see conflicting proxy_shape and shape, should we raise exception or quietly re-compile?
```
mruberry, line 322: Maybe start with an exception until the behavior is better understood?

mruberry (approved):

Issue #604, opened 2024-06-17
randn meta function should be able to accept shape of type list


IvanYashchuk:
## 🐛 Bug

```py
import thunder
import torch

seed = 12345
torch.cuda.manual_seed(seed)
state = torch.cuda.get_rng_state("cuda:0") # For some reason this line is required for the reproducibility

@thunder.jit
def uniform(shape, device):
    return torch.randn(shape, device=device)

shape = [1, 4, 40, 84, 84]
t1 = uniform(shape, device="cuda:0")
print(t1)
```
traceback:
```py
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[1], line 13
     10     return torch.randn(shape, device=device)
     12 shape = [1, 4, 40, 84, 84]
---> 13 t1 = uniform(shape, device="cuda:0")
     14 print(t1)

File ~/dev/lightning-thunder/thunder/__init__.py:660, in jit.<locals>.fn_(*args, **kwargs)
    657 cs.last_trace_host_start = time.time_ns()
    658 cs.calls += 1
--> 660 cache_entry, inps, pro_to_epi = get_computation_and_inputs(*args, **kwargs)
    661 cs.last_trace_host_execution_start = time.time_ns()
    663 result = cache_entry.computation_fn(*inps)

File ~/dev/lightning-thunder/thunder/__init__.py:217, in _with_cache_info_ctx.<locals>.cache_info_wrapper(*args, **kwargs)
    215 tok = _cache_info_ctx.set({})
    216 try:
--> 217     res = fn(*args, **kwargs)
    218 finally:
    219     _cache_info_ctx.reset(tok)

File ~/dev/lightning-thunder/thunder/__init__.py:496, in jit.<locals>.get_computation_and_inputs(*args, **kwargs)
    494 prologue_trc: TraceCtx
    495 computation_trc: TraceCtx
--> 496 jit_results: TraceResults = interpreter(
    497     fn, args, kwargs, record_history=record_history, sharp_edges=cd.sharp_edges
    498 )
    499 prologue_trc = jit_results.prologue_trace
    500 computation_trc = jit_results.computation_trace

File ~/dev/lightning-thunder/thunder/__init__.py:205, in _general_frontend(fn, args, kwargs, record_history, sharp_edges)
    196 def _general_frontend(
    197     fn: Callable,
    198     args: tuple[Any, ...],
   (...)
[...]

PR #1101, opened 2024-09-04
support list[int] as shape for torch.randn

kshitij12345:
Fixes https://github.com/Lightning-AI/lightning-thunder/issues/604

We also update prims.full to only accept `tuple[int]` as shape as that is better for CSE.


t-vi (approved):
Thank you @kshitij12345

Issue #1017, opened 2024-08-22
backwards for autograd.Function


t-vi:
We currently "fudge" autograd.Function by running through the forward as if it was the function.
(This is not good when there is `.detach()` or some such in the forward.)

Given that
- we do know the call args (in the sense of "meta") of the backward once we have traced through the forward,

PR #1106, opened 2024-09-05
Define `torchsymbol` for `torch.ops.higher_order.autograd_function_apply`

crcrpar:
## What does this PR do?

This could be useful when we see `torch.ops.higher_order.autograd_function_apply` which torch.compile introduces into a GraphModule if the graph includes a custom `torch.autograd.Function`.


t-vi:
Given that this is dynamo-specific, this should go into the dynamo section rather than jit_ext, please.

IvanYashchuk:
> Given that this is dynamo-specific, this should go into the dynamo section rather than jit_ext, please.

I think you're confusing Dynamo and Functorch. It's a Functorch-specific thing. Functorch concepts integrate nicely with Thunder because it's oriented at providing functional style (no side-effects) equivalents of object-oriented PyTorch concepts. It's not difficult to convert `torch.autograd.Function` subclasses into a call to `torch._functorch.autograd_function.autograd_function_apply` carrying explicitly forward and backward definitions that in turn can be used by Thunder's reverse differentiation pass.

No Dynamo is involved in the following code:
```py
import torch
from torch._functorch.autograd_function import autograd_function_apply

def forward(ctx, input):
    saved_for_backward = (input,)
    return input.sin(), saved_for_backward

def backward(ctx, grad_output, *saved_tensors):
    input, = saved_tensors
    return grad_output * input.cos()

def my_sin(input):
    return autograd_function_apply(forward, backward, input, args_tensor_mask = [True])

a = torch.randn(3, 4, requires_grad=True)
g = torch.ones_like(a)
my_sin(a).backward(g)
torch.testing.assert_close(a.grad, g * a.cos())
```

Do you have other ideas on how to support using `torch.autograd.Function.backward` definitions properly in Thunder (https://github.com/Lightning-AI/lightning-thunder/issues/1017)?

thunder/tests/test_jit_general.py
```diff
@@ -1200,6 +1200,45 @@ def forward(self, x):
...
+
+    def my_sin(x):
+        return torch.ops.higher_order.autograd_function_apply(
+            forward,
+            backward,
+            x,
+            args_tensor_mask=[True],
+        )
+
+    jitted = thunder.jit(my_sin)
```
crcrpar, line 1221: `thunder.last_traces(jitted)[0]` is as follows:

```python
import torch

@torch.no_grad()
@no_autocast
def computation(x):
  # x: "cpu f32[2, 2]"
  # output of fwd_body: <TensorProxy(name="t0", dtype=thunder.dtypes.float32, shape=(2, 2))>, saved_values from fwd_body: ['t_0']
  t0 = ltorch.autograd_function_apply_cegfp(x)  # t0: "cpu f32[2, 2]"
    # t0 = ltorch.sin(x)  # t0: "cpu f32[2, 2]"
      # t0 = prims.sin(x)  # t0: "cpu f32[2, 2]"

  # /path/to/lightning-thunder/thunder/tests/test_jit_general.py:1211:                 return grad_output * x.cos()
  return t0
```

IvanYashchuk: Cool! In the lowering passes this `autograd_function_apply_cegfp` is decomposed using subsymbols because no pass has a rule to handle `autograd_function_apply_cegfp` as a composite, right?

The backward definition is currently ignored. I think it would be better if the backward function in this test raises an error with a comment to be clear about it.

thunder/core/jit_ext.py
```diff
@@ -609,6 +609,120 @@ def _general_jit_torch_autograd_function_apply_lookaside(obj: Any, *args, **kwar
...
+    # TODO(crcrpar): Think about making use of `non_differentiable_idx`
+    # note that this key is quite new: https://github.com/pytorch/pytorch/pull/134087
+    # non_differentiable_idx = fwd_kwargs.get("non_differentiable_idx")
+    length_of_tensor_args = sum(args_tensor_mask)
+    new_fwd_args = (wrap_const(None),) + fwd_args[:length_of_tensor_args]
+    old_scope = jit_ctx.computation_trace.scopes
+    fwd_bsyms = []
+    jit_ctx.computation_trace.scopes = [fwd_bsyms]
+
+    result = _interpret_call(fwd, *new_fwd_args)
```
IvanYashchuk, line 639: @t-vi, should this be made a public and documented function? It seems a generally useful utility for developers to convert PyTorch callables into Thunder ones.

thunder/tests/test_jit_general.py
```diff
@@ -1200,6 +1200,46 @@ def forward(self, x):
...
+    assert any(
+        bsym.sym.id.startswith("autograd_function_apply")
+        for bsym in initial_computation_trace.bound_symbols
+        if isinstance(bsym.sym.id, str)
+    )
+
+    grad = torch.rand_like(y)
+    actual_grad = torch.autograd.grad(y, x, grad)
+    expect_grad = torch.autograd.grad(y_ref, x_ref, grad)
+    torch.testing.assert_close(actual_grad, expect_grad)
```
IvanYashchuk, line 1240: There's a chance that the provided backward function is ignored but the test would still pass. In addition to checking that the `torch.autograd.grad` works as expected, I suggest you define nonsense backward, for example, use `exp` instead of `cos` and check the grad you receive back `grad * x.exp()`.

thunder/core/jit_ext.py
IvanYashchuk, line 639:
```suggestion
    result = _interpret_call(fwd, *new_fwd_args, **fwd_kwargs)
```

IvanYashchuk: Oh, I see that in the PyTorch code, this fwd_kwargs is unused for calling fwd. Is it a bug?

thunder/core/jit_ext.py
```diff
@@ -609,6 +609,120 @@ def _general_jit_torch_autograd_function_apply_lookaside(obj: Any, *args, **kwar
...
+
+    producer_map = utils.producers(fwd_bsyms)
+    tensor_to_prod_bsym: dict[Variable, BoundSymbol] = {}
+    for p in tree_flatten((output, saved_values))[0]:
+        if not isinstance(p, TensorProxy):
+            continue
+        if p in producer_map:
+            prod_bsym = producer_map[p]
+            tensor_to_prod_bsym[variableify(p)] = prod_bsym
+    prod_bsym_to_tensor = {v: unvariableify(k) for k, v in tensor_to_prod_bsym.items()}
```
crcrpar, line 653: audit the necessity

thunder/core/jit_ext.py
crcrpar: I was speculating `fwd_kwargs` is for "args_tensor_mask" and "non_differentiable_idx"

IvanYashchuk: Ok, `torch.autograd.Function.apply` can't accept keyword arguments.

t-vi: So yes, this can and should be a function that is available to developers of lookasides. (and more or less that only)

t-vi:
So I'm good with merging this, incremental progress and all, but would we want it to share the infra with #1125 ?
We have a variation of this in #1123 , too.

thunder/core/jit_ext.py
t-vi: Also: please don't hesitate to add it, maybe guarded by an `is_jitting` to get a reasonable error message.

crcrpar:
> would we want it to share the infra with #1125 ?

could you elaborate on this? I'm not seeing things to share with it.

t-vi:
>> would we want it to share the infra with https://github.com/Lightning-AI/lightning-thunder/pull/1125 ?

> could you elaborate on this?

> torch.ops.higher_order.autograd_function_apply however is quite free from it and we could just naively use fwd and bwd.
> On top of this point, IMHO having a logic in thunder/torch/__init__.py would be more readable than one in
> thunder/core/jit_ext.py but it feels quite hard for the logic for torch.autograd.Function to be in thunder.torch

Yeah, so the advantage of having it here as it is seems to be that it lives in thunder.torch easily, the advantage of integrating it in the jit more would seem to be that one could trace into fwd and bwd, similar to what we do for torch.autograd.Function.

You are totally right about the function logic in `jit_ext.py` not being ideal. Maybe we could have a more structured JIT hooking facility that would take care of the wrapping bits, wraps the `_interpet_call` things (including returning error checking) and move the "purely autograd.Function" things out of `jit_ext.py`.

crcrpar:
I'm not following why you maintain jit_ext would trace into fwd and bwd as I think the current implementation would also do. My original implementation was basically in jit_ext and I now think that was unnecessarily complicated due to the logic to create bsyms and traces

t-vi:
I am not set on it. Either way works for me.

crcrpar:
@t-vi @IvanYashchuk friendly ping

IvanYashchuk (approved):
Other reviewers, please note that the approach taken here is the same as in https://github.com/Lightning-AI/lightning-thunder/pull/1127 and the current implementation is limited to functions that work both in PyTorch and Thunder without any additional processing. More general support requires implementing ideas for making https://github.com/Lightning-AI/lightning-thunder/issues/1134 possible and maybe resolving bugs that can be triggered today as described in https://github.com/Lightning-AI/lightning-thunder/issues/1126.

For the Dynamo path, we have the input function representation as an FX graph that can be straightforwardly transformed to point to the `thunder.torch` functions instead of `torch`.

t-vi (approved):

PR #1108, opened 2024-09-05
[benchmark_litgpt] always `destroy_process_group` with `try-except-finally`

crcrpar:
## What does this PR do?

As per title. Currently if an exception is raised, `destroy_process_group` is not called.


t-vi (approved):
Thank you @crcrpar

PR #1110, opened 2024-09-05
patching inplace update nvfuser translation

### LINKED ISSUE #1109: nvfuser translation with inplace update is not correct
## 🐛 Bug

Thanks to @IvanYashchuk for the example in an earlier presentation.

### To Reproduce

```
import thunder
import torch
dtype = torch.float32

def foo(T0):
    T1 = torch.sin(T0)
    T0.copy_(T1) # destination.copy_(source)
    T2 = torch.cos(T1)
    T0.copy_(T2)
    return T1, T2

jfoo = thunder.jit(foo, disable_inplace_copy_check=True)


a = torch.randn(2, 3, 4, device="cuda")
a_ref = a.clone()
out = jfoo(a)

out_ref = foo(a_ref)

print("====================")
print("\n\tcompute final:\n", thunder.last_traces(jfoo)[-1])

for o1, o2 in zip(out, out_ref):
    assert o1.allclose(o2)

assert a.allclose(a_ref)
```

This example gives a thunder trace:
```
def computation(T0):
  # T0: "cuda:0 f32[2, 3, 4]"
  [T1, T2] = nvFusion0(T0)
    # T1 = prims.sin(T0)  # T1: "cuda:0 f32[2, 3, 4]"
    # prims.copy_(T1, T0)
    # T2 = prims.cos(T1)  # T2: "cuda:0 f32[2, 3, 4]"
    # prims.copy_(T2, T0)
  return (T1, T2)
```

But the translated nvfuser program looks very different
```
import torch

def nvfuser_fusion_id0(fd : FusionDefinition) -> None :
    T0 = fd.define_tensor(shape=[-1, -1, -1], contiguity=[True, True, True], dtype=DataType.Float, is_cpu=False, stride_order=[2, 1, 0])
    T1 = fd.ops.sin(T0)
    fd.add_output(T1, T0)
    T2 = fd.ops.cos(T1)
    fd.add_output(T2, T0)
    fd.add_output(T1)
    fd.add_output(T2)

    nvfuser_fusion_id0(fd)

inputs = [
    torch.randn(24, dtype=torch.float32, device='cuda:0').as_strided((2, 3, 4), (12, 4, 1)),
]
fd.execute(inputs)
```

we are translating `copy_` to alias destination directly to source, this is wrong. In the example above, the program is expected to return a copy after the update. we need to add a one line copy here https://github.com/Lightning-AI/lightning-thunder/blob/e64d347def39bb47101efafe4177adf9f77a63ec/thunder/executors/nvfuserex_impl.py#L2019

jjsjann123:
Fixes #1109 

`copy_` shouldn't lead to `copy_from` aliasing `copy_to`.

### DIFF
--- a/thunder/executors/nvfuserex_impl.py
+++ b/thunder/executors/nvfuserex_impl.py
@@ -2016,7 +2016,8 @@ def copy_(
 ) -> Any:
     nvcopy_from = getnv(copy_from, fd, lc_to_nv_map)
     nvcopy_to = getnv(copy_to, fd, lc_to_nv_map)
-    fd.add_output(nvcopy_from, alias_input=nvcopy_to)
+    alias_output = fd.ops.set(nvcopy_from)
+    fd.add_output(alias_output, alias_input=nvcopy_to)
     return nvcopy_to
 
 
--- a/thunder/tests/test_inplace_copy.py
+++ b/thunder/tests/test_inplace_copy.py
@@ -158,7 +158,7 @@ def test_inplace_copy_sanity_check(executor, device, dtype):
 def test_inplace_copy_dst_copy_returned_issue_1109(executor, device, dtype):
     def func(T0):
         T1 = torch.sin(T0)
-        T0.copy_(T1) # destination.copy_(source)
+        T0.copy_(T1)  # destination.copy_(source)
         T2 = torch.cos(T1)
         T0.copy_(T2)
         # T1 & T2 should be returned as separate buffer, instead of sharing
--- a/thunder/tests/test_inplace_copy.py
+++ b/thunder/tests/test_inplace_copy.py
@@ -161,13 +161,18 @@ def test_inplace_copy_dst_copy_returned_issue_1109(executor, device, dtype):
         T0.copy_(T1) # destination.copy_(source)
         T2 = torch.cos(T1)
         T0.copy_(T2)
+        # T1 & T2 should be returned as separate buffer, instead of sharing
+        # storage with T0
         return T1, T2
 
-    traced_foo = executor.make_callable(func)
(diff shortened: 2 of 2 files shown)


jjsjann123:
> With this change, would a nvfuser region that has only copies work?

Does that not work? Do you happen to have an example for that?

```
import thunder
import torch
dtype = torch.float32

def foo(T0, T1):
    T0.copy_(T1) # destination.copy_(source)
    return T0, T1

jfoo = thunder.jit(foo, disable_inplace_copy_check=True)


a = torch.randn(2, 3, 4, device="cuda")
b = torch.randn(2, 3, 4, device="cuda")
a_ref = a.clone()
b_ref = b.clone()
out = jfoo(a, b)

out_ref = foo(a_ref, b_ref)

print("====================")
print("\n\tcompute final:\n", thunder.last_traces(jfoo)[-1])

for o1, o2 in zip(out, out_ref):
    assert o1.allclose(o2)

assert a.allclose(a_ref)
assert b.allclose(b_ref)
print(out[0].data_ptr())
print(out[1].data_ptr())
```

I tried this example above and it seems to be working fine.

crcrpar:
I could be wrong but previously didn't we add a change to avoid such one as copy in nvfuser isn't an op?
If it works now then it's great. I just was curious

t-vi (approved):
Thank you @jjsjann123 @crcrpar

jjsjann123:
> I could be wrong but previously didn't we add a change to avoid such one as copy in nvfuser isn't an op? If it works now then it's great. I just was curious

You are right. I remember someone (likely also you :laughing:) complaining about that and we have patched it already on nvfuser side https://github.com/NVIDIA/Fuser/pull/2638

PR #1118, opened 2024-09-07
Move autocast from thunder.core.transforms to thunder.transforms

IvanYashchuk:
This moves the interpreter-based autocast implementation to its own file under `thunder.transforms`.


t-vi (approved):
Thank you, @IvanYashchuk

PR #1121, opened 2024-09-07
make BoundSymbol.from_bsym call bind_postprocess

t-vi:
The need to call it manually from transforms is easy to forget and annoying to debug.


lantiga (approved):
Great, big UX improvement

PR #1125, opened 2024-09-09
Proxify `FunctionCtx`, define `BoundSymbol`, and define and register augmented rule for `torch.autograd.Function`

crcrpar:
## What does this PR do?

Address #1017.

- ~~[ ] Translate lines involving `ctx` into bsyms, especially in backward to correctly use non-tensor saved intermediates.~~ It seems that non-tensor attributes of `ctx` are folded into a trace.

### DIFF
--- a/thunder/core/baseutils.py
+++ b/thunder/core/baseutils.py
@@ -75,6 +75,10 @@ class TensorProxyInterface:
     pass
 
 
+class TorchAutogradFunctionXtxProxyInterface:
+    pass
+
+
 class SymbolInterface:
     name: str
     is_prim: bool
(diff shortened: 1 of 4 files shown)


IvanYashchuk (approved):

t-vi (approved):
Very nifty!

thunder/core/baseutils.py
```diff
@@ -75,6 +75,10 @@ class TensorProxyInterface:
     pass
 
 
+class TorchAutogradFunctionXtxProxyInterface:
```
lantiga, line 78: Typo here, should be `TorchAutogradFunctionCtxProxyInterface`

crcrpar: Great catch, thank you Luca! opened 1147

Issue #1126, opened 2024-09-09
Calling general_thunder_jit inside lookasides doesn't work


IvanYashchuk:
## 🐛 Bug

I want to convert a Python function that might contain PyTorch calls into a Thunder function inside the lookaside function.
I wasn't successful at using `thunder.core.interpreter.interpret` so I resorted to `thunder_general_jit`. The inner function `interpreted_fn` does the correct thing. However, something stands on the way of correct nested usage of `thunder_general_jit` and I see the following error:
```py
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
Cell In[1], line 60
     58 x = torch.randn(3, 4, requires_grad=True)
     59 jf = thunder.jit(f)
---> 60 out = jf(x)

File ~/dev/lightning-thunder/thunder/__init__.py:704, in jit.<locals>.fn_(*args, **kwargs)
    701 cs.last_trace_host_start = time.perf_counter_ns()
    702 cs.calls += 1
--> 704 cache_entry, inps, pro_to_epi = get_computation_and_inputs(*args, **kwargs)
    705 cs.last_trace_host_execution_start = time.perf_counter_ns()
    707 if cache_entry.vanilla_tensor_args:

File ~/dev/lightning-thunder/thunder/core/langctxs.py:136, in langctx.__call__.<locals>._fn(*args, **kwargs)
    134 try:
    135     tok = set_langctx(self.langctx)
--> 136     result = fn(*args, **kwargs)
    137     return result
    138 finally:

File ~/dev/lightning-thunder/thunder/__init__.py:213, in _with_cache_info_ctx.<locals>.cache_info_wrapper(*args, **kwargs)
    211 tok = _cache_info_ctx.set({})
    212 try:
--> 213     res = fn(*args, **kwargs)
    214 finally:
    215     _cache_info_ctx.reset(tok)

File ~/dev/lightning-thunder/thunder/__init__.py:500, in jit.<locals>.get_computation_and_inputs(*args, **kwargs)
    498 prologue_trc: TraceCtx
    499 computation_trc: TraceCtx
--> 500 jit_results: TraceResults = thunder_general_jit(
    501     fn, args, kwargs, record_history=record_history, sharp_edges=cd.sharp_edges
    502 )
    503 prologue_trc = jit_results.prologue_trace
    504 computation_trc = jit_results.computation_trace

[...]

PR #1127, opened 2024-09-09
Add initial support for torch.utils.checkpoint

IvanYashchuk:
A checkpointed function doesn't save any intermediates from forward to backward. Instead, all required values are recomputed during the backward pass. Because less intermediates are saved, peak memory usage is usually decreased.

This PR introduces the support of recognizing `torch.utils.checkpoint.checkpoint` calls and inserting a new bound symbol in the initial trace. Then in the forward-backward generation pass this bound symbol is converted into augmented forward and backward parts of the computation. This step requires the function argument to `thunder.torch.checkpoint` be a Thunder function. Currently, there's no conversion PyTorch->Thunder implemented and this works only for simple functions that are both recognized by Thunder and PyTorch, for example when only methods are used.

The PyTorch function needs to be converted to a Thunder function in Thunder's JIT. Previously we could simply use `thunder.preprocess` which is not available today. When I attempted implementing a redispatching/reinterpretation of PyTorch functions using `general_thunder_jit` I hit the following bug: https://github.com/Lightning-AI/lightning-thunder/issues/1126.

Example:
```py
import thunder
import torch

def f(x):
    return torch.utils.checkpoint.checkpoint(lambda x: x.sin().cos().exp(), x)

jf = thunder.jit(f)
x = torch.randn(3, 4, device="cuda", requires_grad=True)
jf(x).backward(x)
print(thunder.last_backward_traces(jf)[-1])
```
Forward execution trace:
```py
def augmented_forward_fn(x):
  # x: "cuda:0 f32[3, 4]"
  [t2] = nvFusion0(x)
    # t0 = prims.sin(x)  # t0: "cuda:0 f32[3, 4]"
    # t1 = prims.cos(t0)  # t1: "cuda:0 f32[3, 4]"
    # t2 = prims.exp(t1)  # t2: "cuda:0 f32[3, 4]"
  return {'output': t2, 'flat_args': [x], 'flat_output': (t2,)}, ((x,), ())
```
Backward execution trace:
```py
  # cotangents: "Collection"
  C0, _, = saved_for_backward
  del saved_for_backward
  t3, = cotangents
  del cotangents
  x, = C0
  clear_mutable_collection(C0)
  del C0
  [t12] = nvFusion0(x, t3)
    # t4 = prims.sin(x)  # t4: "cuda:0 f32[3, 4]"
    # t11 = prims.cos(x)  # t11: "cuda:0 f32[3, 4]"
    # t5 = prims.cos(t4)  # t5: "cuda:0 f32[3, 4]"
    # t8 = prims.sin(t4)  # t8: "cuda:0 f32[3, 4]"
    # t6 = prims.exp(t5)  # t6: "cuda:0 f32[3, 4]"
    # t7 = prims.mul(t3, t6)  # t7: "cuda:0 f32[3, 4]"
    # t9 = prims.neg(t8)  # t9: "cuda:0 f32[3, 4]"
    # t10 = prims.mul(t7, t9)  # t10: "cuda:0 f32[3, 4]"
    # t12 = prims.mul(t10, t11)  # t12: "cuda:0 f32[3, 4]"
  del x, t3
  return (t12,)
```

### DIFF
--- a/thunder/core/jit_ext.py
+++ b/thunder/core/jit_ext.py
@@ -53,6 +53,7 @@ from types import (
 )
 
 import torch
+import torch.utils.checkpoint
 from thunder.core.proxies import (
     DistParallelType,
     proxy,
@@ -615,6 +616,25 @@ def _general_jit_torch_finfo_lookaside(dtype: thunder.dtypes.dtype):
     torch_dtype = thunder.dtypes.to_torch_dtype(dtype)
     res = torch.finfo(torch_dtype)
     return res
+
+
+@register_general_jit_lookaside(torch.utils.checkpoint.checkpoint)
+def _general_jit_torch_checkpoint_lookaside(
+    function: Callable,
+    *args,
+    **kwargs: Any,
+):
+    from thunder.torch import checkpoint
+
+    # It should be possible to call the general_thunder_jit here to handle the
+    # conversion from torch to thunder but it doesn't work now
+    # See https://github.com/Lightning-AI/lightning-thunder/issues/1126
+    # TODO: Convert the function to a Thunder function
+    def thunder_function(*args, **kwargs):
+        return unwrap(function)(*args, **kwargs)
+
+    wrapped_thunder_function = wrap_const(thunder_function)
+    return interpreter_needs_wrap(checkpoint)(wrapped_thunder_function, *args, **kwargs)
 
 
 # Adds proxy methods
--- a/thunder/core/jit_ext.py
+++ b/thunder/core/jit_ext.py
@@ -632,8 +632,9 @@ def _general_jit_torch_checkpoint_lookaside(
     function that can handle only Thunder functions as input.
 
     Args:
-        function: The function to be checkpointed. *args: Arguments to the
-        function. **kwargs: Keyword arguments to the function.
+        function: The function to be checkpointed.
+        args: Arguments to the function.
+        kwargs: Keyword arguments to the function.
 
     Returns:
         The result of calling `thunder.torch.checkpoint` with the preprocessed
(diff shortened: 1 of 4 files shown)


thunder/core/jit_ext.py
```diff
@@ -617,6 +618,25 @@ def _general_jit_torch_finfo_lookaside(dtype: thunder.dtypes.dtype):
     return res
 
 
+@register_general_jit_lookaside(torch.utils.checkpoint.checkpoint)
+def _general_jit_torch_checkpoint_lookaside(
```
mruberry, line 622: Would you add a comment / docstring explaining this function and its intended use?

mruberry (approved):
Cool! @t-vi, do you want to take a look?

thunder/core/jit_ext.py
IvanYashchuk: Yes, will do!

t-vi (changes requested):
I'm not convinced of the design.
Why would we not just let the function be any function and have a state "currently checkpointing" that informs Thunder to add a tag to the proxies that are generated during the checkpointing instead?
We would need to clear that tag on the outputs, but that would be easier than reentrant jit.

IvanYashchuk:
> Why would we not just let the function be any function and have a state "currently checkpointing" that informs Thunder to add a tag to the proxies that are generated during the checkpointing instead? We would need to clear that tag on the outputs, but that would be easier than reentrant jit.

Do you have ideas about how the "currently checkpointing" approach would generalize to supporting, for example, `torch.cond`? Please continue in the issue https://github.com/Lightning-AI/lightning-thunder/issues/1134.

t-vi:
I don't have immediate ideas, but I don't see that we should be having higher order functions right now.

syed-ahmed:
@IvanYashchuk You might wanna checkout selective activation checkpointing available in PyTorch nightlies: https://pytorch.org/docs/main/checkpoint.html#torch.utils.checkpoint.create_selective_checkpoint_contexts to specify which activations to save for backward.

Issue #1074, opened 2024-08-30
Memory Leak when using nn.Module hooks and thunder.jit


kshitij12345:
Using nn.Module hook with thunder.jit leads to memory leak. Disabling the hook, doesn't leak the memory.

This affects usage with Megatron-LM's DistributedOptimizer which uses hooks - https://github.com/NVIDIA/Megatron-LM/blob/73ca2e34ab58e65191a1329189720e1520eef263/megatron/core/optimizer/distrib_optimizer.py#L535-L542

Repro - 
```python
import weakref
import torch
import thunder
import refcycle  # pip install refcycle

class Identity(torch.nn.Module):
    def forward(self, x):
        return x

def main():
    with torch.device("cuda"):
        # model = torch.nn.Sequential(Identity())  # PyTorch eager doesn't leak

        model = torch.nn.Sequential(thunder.jit(Identity()))
        # More tensors the model has, more it leaks
        # model = torch.nn.Sequential(torch.nn.Linear(16, 16), thunder.jit(Identity()), torch.nn.Linear(16, 16))
        
        def my_noop_hook(module, unsued):
            for p in module.parameters():
                pass

        # Disabling the hook -> doesn't leak.
        remove_handle = torch.nn.modules.module.register_module_forward_pre_hook(my_noop_hook)
        x = torch.randn(16, 16)

    model(x)
    return weakref.ref(x)

weak_x = main()

snap = refcycle.snapshot()

for t in snap.find_by(lambda t: isinstance(t, torch.Tensor)):
    print(t.shape)  # Prints - torch.Size([16, 16])

# Code to find and save cycles
# for idx, anc in enumerate(snap.ancestors(weak_x())):
#     print(idx)
#     try:
#         cycle = snap.shortest_cycle(anc)
#         # cycle.export_image(f"cycles/cycle_{idx}.png")
#         print(cycle)
#     except:
#         pass
```

There are a lot of cycles but I am hoping resolving this first one might take care of others.

![cycle_120](https://github.com/user-attachments/assets/da26d53f-4faf-4335-af2f-99015c99f7bc)

I could check the cycle in debugger with
```python
anc.f_locals["python_frame"].f_back.f_back is anc
```

And the path from frame (which owns the tensor and is part of the cycle) to the tensor is

![path](https://github.com/user-attachments/assets/92531151-ea62-4243-9cf2-c1127b435ce6)

```python
[...]

PR #1145, opened 2024-09-12
clear runtimectx on exit to avoid refcycles 

t-vi:
...in particular through frame->runtimectx->framstack->frame
and CURRENT_FRAME->caught exception->frame->f_back...->CURRENT_FRAME

Fixes: #1074 

All analysis credit to @kshitij12345 and @lantiga , screwing up the fix is me.


t-vi:
So it seems this is overambitious. Maybe we try to clear the things holding the ctx instead.

t-vi:
Superseded by #1170

PR #1152, opened 2024-09-16
thunder + dynamo : splitter - fix support for cat

kshitij12345:
Repro - 
```python
import torch
import thunder

def my_cat(x):
    y = torch.cat(tensors=[x, x])
    return y.sin()

backend = ThunderCompiler()
jfn = torch.compile(my_cat, backend=backend)

o = jfn(torch.randn(3, 3))
print(backend.subgraph_infos[0].split_reasons)
```

Output
```python
[SplitReason(type=<SplitReasonType.EXCEPTION_META_THUNDER_OP: 4>, info='Failed while running meta for node with name: y and target: <built-in method cat of type object at 0x7296b46e2ea0>, see exception field', exception=ValueError("l_x_ had an unexpected type <class 'torch.fx.node.Node'>. Supported types are (<class 'thunder.core.proxies.TensorProxy'>, <class 'numbers.Number'>, <class 'thunder.core.proxies.NumberProxy'>)"))]
```

It happens because the `node.args` for `torch.cat` node is `<class 'torch.fx.immutable_collections.immutable_list'>` and we don't handle it correctly.

Solution - We use `map_arg` from `torch.fx` to apply `make_tensor_proxy` correctly to all input `fx.Node`.

`map_arg` - https://github.com/pytorch/pytorch/blob/bbc3fdbbde3f35467b40cf06ec4926adca1876d4/torch/fx/node.py#L777-L784

### DIFF
--- a/thunder/dynamo/utils.py
+++ b/thunder/dynamo/utils.py
@@ -9,7 +9,6 @@ import torch
 from thunder.torch.default_torch_ops import torch_auto_registered_ops
 from thunder.torch import _torch_to_thunder_function_map
 
-
 auto_register_ops = set(itertools.chain(*torch_auto_registered_ops.values()))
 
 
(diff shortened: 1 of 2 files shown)


mruberry:
The test failure appears unrelated. I filed https://github.com/Lightning-AI/lightning-thunder/issues/1153 to track it

mruberry (approved):
Fixed!

PR #1158, opened 2024-09-17
Add Torch.Tensor.exponential_

ali-alshaar7:
<details>

</details>

## What does this PR do?

Adds support for Torch.Tensor.exponential_ which Is used in token sampling.

## PR review

## Did you have fun?

### DIFF
--- a/thunder/tests/test_ops.py
+++ b/thunder/tests/test_ops.py
@@ -256,31 +256,29 @@ def test_exponential():
     def fn(a):
         return a.exponential_(1)
 
-    size = 1000000
+    size = 1000
+    seed = 1234
+    with torch.device("cuda"):
+        a_ref = torch.ones(size)
+        torch.manual_seed(seed)
+        a_ref = fn(a_ref)
 
-    a_ref = torch.ones(size)
-    out_ref = a_ref.geometric_(0.4)
-    a = torch.ones(size)
-    jf = thunder.jit(fn)
-    out = jf(a)
-    print(a, a_ref)
+        a = torch.ones(size)
+        torch.manual_seed(seed)
 
+        # nbfuser fuses prims.uniform, which is used by exponential resulting in differing numerics.
+        executors = thunder.get_default_executors()
+        fuser_i = None
+        for i, e in enumerate(executors):
+            if e.name == "nvfuser":
+                fuser_i = i
+                break
 
-    import torch.nn.functional as F
-    def js_divergence(p, q):
-        # Convert tensors to probabilities using softmax
-        p_prob = F.softmax(p, dim=0)
-        q_prob = F.softmax(q, dim=0)
-        
-        # Average of the two distributions
-        m = 0.5 * (p_prob + q_prob)
-        
-        # Compute KL divergence for each part and average them
-        jsd = 0.5 * (F.kl_div(F.log_softmax(p, dim=0), m, reduction='batchmean') +
-                    F.kl_div(F.log_softmax(q, dim=0), m, reduction='batchmean'))
-        return jsd
+        executors = executors[:fuser_i]+ executors[fuser_i+1:] if fuser_i is not None else executors
+        jf = thunder.jit(fn, executors=executors)
+        a = jf(a)
 
-    print(js_divergence(a, a_ref))
+        assert_close(a, a_ref)
 
 
 test_exponential()
--- a/thunder/tests/test_ops.py
+++ b/thunder/tests/test_ops.py
@@ -252,6 +252,7 @@ def test_setitem():
     assert_close(a, a_ref)
     assert_close(out, out_ref)
 
+
 # TODO: Add random operator support to OpInfo
 # https://github.com/Lightning-AI/lightning-thunder/issues/1163
 @requiresCUDA
@@ -281,4 +282,3 @@ def test_exponential():
 
         assert_close(a, a_ref)
         assert_close(b, b_ref)
-
--- a/thunder/tests/test_ops.py
+++ b/thunder/tests/test_ops.py
@@ -266,19 +266,9 @@ def test_exponential():
         a = torch.ones(size)
         torch.manual_seed(seed)
 
-        # nbfuser fuses prims.uniform, which is used by exponential resulting in differing numerics.
-        executors = thunder.get_default_executors()
-        fuser_i = None
-        for i, e in enumerate(executors):
-            if e.name == "nvfuser":
-                fuser_i = i
-                break
-
-        executors = executors[:fuser_i]+ executors[fuser_i+1:] if fuser_i is not None else executors
-        jf = thunder.jit(fn, executors=executors)
+        # nbfuser fuses prims.uniform, which is used by our exponential resulting in differing numerics.
+        jf = thunder.jit(fn, executors={})
         a = jf(a)
 
         assert_close(a, a_ref)
 
-
-test_exponential()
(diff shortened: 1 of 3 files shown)


thunder/tests/test_ops.py
```diff
@@ -251,3 +251,32 @@ def fn(a):
...
+        # nbfuser fuses prims.uniform, which is used by exponential resulting in differing numerics.
+        executors = thunder.get_default_executors()
+        fuser_i = None
+        for i, e in enumerate(executors):
+            if e.name == "nvfuser":
+                fuser_i = i
+                break
+
+        executors = executors[:fuser_i]+ executors[fuser_i+1:] if fuser_i is not None else executors
+        jf = thunder.jit(fn, executors=executors)
```
t-vi:
```suggestion
        # nvfuser fuses prims.uniform and implements it slightly differently to PyTorch, leading to different random samples
        jf = thunder.jit(fn, executors=())
```

I think we can have it easy here.

thunder/tests/test_ops.py
```diff
@@ -251,3 +251,27 @@ def fn(a):
...
+    # on cpu, aten.exponential_ is not decomposed to ops used in Thunder exponential_.
+    with torch.device("cuda"):
+        a_ref = torch.ones(size)
+        torch.manual_seed(seed)
+        a_ref = fn(a_ref)
+
+        a = torch.ones(size)
+        torch.manual_seed(seed)
+
+        # nbfuser fuses prims.uniform, which is used by our exponential resulting in differing numerics.
```
crcrpar, line 273:
```suggestion
        # nvfuser fuses prims.uniform, which is used by our exponential resulting in differing numerics.
```

thunder/torch/__init__.py
```diff
@@ -1503,6 +1503,39 @@ def exp2_(a):
     return prims.copy_(exp2(a), a)
 
 
+# fake out of place variant
+@torchsymbol(id="exponential")
+def exponential(a, rate=1, generator=None):
```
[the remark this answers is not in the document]
mruberry: Would you add type annotations? Also, we should probably require the first two arguments be specified positionally, and the last argument be specified using a keyword:

```
(a, rate=1, /, *, generator=None)
```

thunder/torch/__init__.py
```diff
@@ -1503,6 +1503,39 @@ def exp2_(a):
...
+    assert generator is None
```
mruberry, line 1509: Instead of asserting, check that the generator is None. That will create a clearer error message.

Also, Python executed in optimized mode won't check asserts, which are intended for debugging and testing

thunder/torch/__init__.py
```diff
@@ -1503,6 +1503,39 @@ def exp2_(a):
...
+    utils.check(
+        not thunder.dtypes.is_complex_dtype(a.dtype)
```
mruberry, line 1511: This could affirmatively check that `a` is a floating point type instead of asserting it's not any other type, which would probably be more straightforward

thunder/torch/__init__.py
```diff
@@ -1503,6 +1503,39 @@ def exp2_(a):
...
+        and not thunder.dtypes.is_integer_dtype(a.dtype)
+        and not thunder.dtypes.is_boolean_dtype(a.dtype),
+        lambda: f"Exponential distribution is a continuous probability distribution. \
+        dtype must be a floating point but you specified {self.dtype}",
```
mruberry, line 1515: `self` is undefined here

thunder/torch/__init__.py
```diff
@@ -1503,6 +1503,39 @@ def exp2_(a):
...
+        lambda: f"exponential_ expects lambda > 0.0, but found lambda={rate}",
+    )
+    uniform_val = uniform_like(a)
+
+    # copying numerics of transformation::exponential see comment:
+    # curand_uniform has (0,1] bounds. log(1) is 0 and exponential excludes 0.
+    # we need log to be not 0, and not underflow when converted to half
+    # fast __logf approximation can underflow, so set log to -epsilon/2 for 1 or close to 1 args
+    epsilon = torch.finfo(thunder.dtypes.to_torch_dtype(a.dtype)).eps / 2
+    condition = uniform_val >= 1.0 - epsilon
```
mruberry, line 1528: Parentheses would make this line more legible

thunder/torch/__init__.py
```diff
@@ -1503,6 +1503,39 @@ def exp2_(a):
...
+    log_uniform = where(condition, -epsilon, log(uniform_val))
+
+    return -1 / rate * log_uniform
```
mruberry, line 1531: Parentheses would help make this line more readable, too

thunder/torch/__init__.py
```diff
@@ -1503,6 +1503,39 @@ def exp2_(a):
...
+
+
+@torchsymbol(torch.Tensor.exponential_, id="exponential_", is_method=True, tags=(prims.OpTags.IN_PLACE,))
+def exponential_(a, rate=1, generator=None):
```
mruberry, line 1535: Signature updates here, too, please (type annotations and positional-only and keyword-only arguments)

thunder/tests/test_ops.py
mruberry, line 273: "nbfuser -> nvfuser"

thunder/tests/test_ops.py
```diff
@@ -251,3 +251,27 @@ def fn(a):
...
+
+@requiresCUDA
+def test_exponential():
+    def fn(a):
+        return a.exponential_(1)
+
+    size = 1000
+    seed = 1234
+
+    # on cpu, aten.exponential_ is not decomposed to ops used in Thunder exponential_.
```
mruberry, line 264: What about adding a direct lowering to the PyTorch executor?

How different is PyTorch's Tensor.exponential_ on CPU and CUDA? Are they so different that we should refuse to execute exponential on the CPU for now (and file an issue to follow-up)?

mruberry (commented):
Good start, @ali-alshaar7!

I made some comments inline for your review

thunder/torch/__init__.py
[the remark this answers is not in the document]
mruberry: Hm, I guess the problem with an OpInfo for exponential is that it's a random operation. I don't think we support random operations in OpInfos yet, although it'd be nice to do so in the future. Would you file an issue (and comment a link to it next to the hand written exponential test) to support random operations and inplace operations using OpInfos?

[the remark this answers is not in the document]
crcrpar: I have maladroit logic to have opinfo around https://github.com/Lightning-AI/lightning-thunder/blob/d08c2715f611145f7687b680e4af281b565a2908/thunder/tests/test_inplace_functionalization.py#L60-L84

t-vi (approved):
Thank you @ali-alshaar7 @crcrpar @mruberry

thunder/tests/test_ops.py
ali-alshaar7: Pardon the naive question, but how would that work given that on CPU, PyTorch lowers to inplace `exponential_` ATen kernel and on GPU, it lowers to the decomposition shown here, and we don't support inplace ops in thunder atm? As in how would we handle the cpu case with just PyTorch. Thanks.

t-vi: We can create a `prim.exponential` and then have that call `torch.ops.aten.exponential`, which will - by definition - do the same thing as PyTorch, and instead of decomposing lower to that prim.

However, when we don't require NVFuser to give us the same random numbers as PyTorch, why would we require this. In practice, we'll use NVFuser's random numbers.

PR #1170, opened 2024-09-18
Fix leak related to stop iteration

### LINKED ISSUE #1074: Memory Leak when using nn.Module hooks and thunder.jit

lantiga:
## What does this PR do?

We forgot to clean up the exception after handling `StopIteration` in the `for_iter` lookaside. This PR fixes it.
It also adds two tests for testing exception-related leaks.

Fixes #1074 

## Limitations

We still leak when we raise from inside a jitted function and catch on the outside. I'll open a separate issue, but this is unrelated to #1074

### DIFF
--- a/thunder/core/interpreter.py
+++ b/thunder/core/interpreter.py
@@ -4246,6 +4246,7 @@ def _for_iter_handler(
     if r is INTERPRETER_SIGNALS.EXCEPTION_RAISED:
         ctx = get_interpreterruntimectx()
         if isinstance(ctx.curexc, StopIteration):
+            ctx.curexc = None
             if sys.version_info >= (3, 12):
                 # 3.12 uses jumps relative to the next instruction offset and does not pop here
                 #      instead it pushes a fake value?!
--- a/thunder/core/interpreter.py
+++ b/thunder/core/interpreter.py
@@ -1175,8 +1175,6 @@ class InterpreterFrame:
             return self.code._varname_from_oparg(idx)  # type: ignore
 
     def get_or_make_python_frame(self) -> FrameType:
-        def fn():
-            pass
 
         assert self.positions is not None
         lineno = self.positions.lineno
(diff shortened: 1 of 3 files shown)


thunder/tests/test_interpreter.py
```diff
@@ -779,6 +779,53 @@ def cross_function_exceptions():
...
+
+def test_exception_no_leak(jit):
+
+    class Identity(torch.nn.Module):
+        def raises():
+            raise RuntimeError("Exc")
+
+        def forward(self, x):
+            try:
+                raises()
```
t-vi, line 811:
```suggestion
                self.raises()
```
?

thunder/tests/test_interpreter.py
```diff
@@ -779,6 +779,53 @@ def cross_function_exceptions():
...
+        model = thunder.jit(Identity())
+        x = torch.randn(16, 16)
+
+        model(x)
+
+        return weakref.ref(x)
+
+    weak_x = foo()
+
+    assert weak_x is None
```
t-vi, line 800:
```suggestion
    assert weak_x() is None
```
?

thunder/tests/test_interpreter.py
```diff
@@ -779,6 +779,53 @@ def cross_function_exceptions():
...
+
+    weak_x = foo()
+
+    assert weak_x() is None
+
+
+def test_exception_no_leak(jit):
+
+    class Identity(torch.nn.Module):
+        def raises():
```
t-vi, line 806:
```suggestion
        @staticmethod
        def raises():
```

thunder/core/interpreter.py
```diff
@@ -1175,40 +1175,55 @@ def get_localsplus_name(self, idx: int) -> str:
...
+        # we run the getting of the frame in a separate thread because
+        # we want to avoid having f_back pointing to the function
+        # handling the error
+        import _thread
+
+        result_container = []
+        _thread.start_new_thread(get_frame, (result_container, rel_lineno, filename, firstlineno, name, qualname))
+        while not result_container:
+            pass
+        return result_container[0]
```
t-vi, line 1226: So this is needed to remove this cycle (image by @kshitij12345 ) that apparently causes trouble for us.
Before this, I could reproduce the problem by running the (`-k leak`) tests ~5-10 times, not I can run 50 iterations without anything happening.

image

lantiga: Interesting, this may solve the cross-boundary exception leak too

PR #1177, opened 2024-09-20
(Experimental) Make nvFuser executor treat `ltorch.copy_` differently from `prims.copy_`

### LINKED ISSUE #1173: PR #1110 nearly doubles the compilation & execution time of a copy-heavy program
I took a benchmark of `torch.compile`'d `Adam.step` with Thunder backend. Surprisingly, the compiled `Adam.step` was even slower than the eager mode.

||compilation (s)|execution (ms)|
|:-|-:|-:|
|eager|0.0|10.87|
|`torch.compile(adam.step)`|21.0|5.83|
|`torch.compile(adam.step, backend=thunder)`|95.0|11.59|
|`torch.compile(adam.step, backend=thunder)`, [#1110](#1110) reverted|46.0|6.24|

#1110 adds `fd.ops.set` on every `prims.copy_` ([diff](https://github.com/Lightning-AI/lightning-thunder/pull/1110/files#diff-8588eeba7400704dd5b82554f037a9d89ea99a66cd82d81bf4ed029c6f04eb38L2019-R2020)). What is `fd.ops.set`? Can we avoid using this op?

shino16:
Fixes #1173. I implemented the idea mentioned in https://github.com/Lightning-AI/lightning-thunder/issues/1173#issuecomment-2362895907. This PR makes the nvFuser executor absorb `ltorch.copy_` separately from `prims.copy_` (which appears within in-place ops).

This PR is experimental. It works (if the tests pass), but it is not a good solution in that some transform may walk in and replace `ltorch.copy_` with something equivalent. A complete solution would be to track all `fd.add_output` generated so far and insert `fd.ops.set` where needed.

### DIFF
--- a/thunder/clang/__init__.py
+++ b/thunder/clang/__init__.py
@@ -1861,6 +1861,11 @@ def copysign(a, b):
     return maybe_convert_to_dtype(result, result_dtype)
 
 
+@clangop()
+def copy_(src, dst):
+    return prims.copy_(src, dst)
+
+
 @clangop(method_name="eq")
 def eq(a, b):
     return _elementwise_binary_wrapper(
--- a/thunder/executors/nvfuserex_impl.py
+++ b/thunder/executors/nvfuserex_impl.py
@@ -2048,9 +2048,27 @@ def _copy__check(
     return are_supported_tensors(copy_from, copy_to)
 
 
+# prims.copy_(copy_from, copy_to)
 def copy_(
     copy_from: TensorProxy,
     copy_to: TensorProxy,
+    *,
+    fd: FusionDefinition,
+    lc_to_nv_map: dict,
+) -> Any:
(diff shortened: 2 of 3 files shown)


t-vi:
Would it not be cleaner to have a separate prim for the inplace-resolving copy?

shino16:
Yes, it must be possible to prepare a separate primitive or give `prims.copy_` a flag indicating whether it comes from an inplace op. When I was working on #912, Ivan told me that all our `prims` functions should be considered as part of public API). So I am not sure how much implementation details we can expose under `prims`.

t-vi:
Yeah, I totally agree that we should not mess with prims too much, but to my mind, having a specialized, separate prim for the inplace is something very straightforward and likely better than playing tricks with the levels of decomposition.

PR #1181, opened 2024-09-20
Clean up and slightly optimize `TraceCtx::python()`

apaz-cli:


apaz-cli:
Context: We're working on litgpt container startup times. We have a lot of traces, and spend a solid 10 seconds building callables from them. This PR is unlikely to make much of a difference in that, but it's the obvious low hanging fruit.

t-vi:
Let's not. Optmizing without measurable impact is not a habit we want to get into.

PR #1189, opened 2024-09-23
Preserve tags when swapping proxies

ali-alshaar7:
<details>

</details>

## What does this PR do?

While testing LitGPT Llama 3 8b, we ran into an issue where one of the kv-cache symbols had its name updated without carrying the `ProxyTag.STATIC_MEMORY_LOCATION` tag. 

This causes `CUDAGraphsTransform` to avoid updating the kv-cache of that layer after each inference step, resulting in bad outputs.

This PR makes cloning a Proxy perform a reference assignment on tags instead of copying.

## PR review

## Did you have fun?

### DIFF
--- a/thunder/core/jit_ext.py
+++ b/thunder/core/jit_ext.py
@@ -1576,9 +1576,11 @@ def _get_process_group_from(*fn_and_args) -> Optional["ProcessGroup"]:
             raise NotImplementedError("jitting modules with different ProcessGroup is not supported currently.")
     return found_pg
 
+
 def update_tags(proxy_swapmap: dict[Variable, Proxy]) -> None:
     for old, new in proxy_swapmap.items():
         new.tags.update(unvariableify(old).tags)
+
 
 def thunder_general_jit(
     fn: Callable,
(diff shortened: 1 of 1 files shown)


t-vi:
Hmm. I think I would prefer to fix this at the level of proxy creation.
proxy.replace_name (via proxy.replace) should actually preserve proxy tags, so I'm not entirely sure where the untagged proxies are created...

ali-alshaar7:
> Hmm. I think I would prefer to fix this at the level of proxy creation. proxy.replace_name (via proxy.replace) should actually preserve proxy tags, so I'm not entirely sure where the untagged proxies are created...

So the renaming map is actually created in an interpreter callback. This happens before the tags are added to the inputs. We could pass the proxy renaming map to `unpack_inputs` so that tags being added to one adds it to the other, or make the `tag` `set()` a pointer or something. I think the point is that tagging one proxy should also tag all the others its connected to.

ali-alshaar7:
> Hmm. I think I would prefer to fix this at the level of proxy creation. proxy.replace_name (via proxy.replace) should actually preserve proxy tags, so I'm not entirely sure where the untagged proxies are created...

So the tagging actually happens after the swap map is cerated, and since we are currently copying tag sets, we end up not propagating tagging. 

Simplest fix is make tags a pointer when cloning proxies. I think it makes sense since memory management should be handled by GC, so no memory safety issues. In terms of correctness, do we want tags of a cloned proxy to be different ever? Probably no, but im not sure.

lantiga (approved):
Great catch!

t-vi:
Thank you for revising the mechanism, great work @ali-alshaar7

PR #1193, opened 2024-09-24
fix leak with uncaught exceptions

### LINKED ISSUE #1171: Memory leak when raising an exception in jitted fn and catching it outside
From the analysis of #1170, there's one case where we still leak memory related to exceptions.

When we don't handle the exception within the interpreter but do it outside we leak.

```python
class Identity(torch.nn.Module):
    def forward(self, x):
        raise RuntimeError("Error")
        return x

def foo():
    model = thunder.jit(Identity())
    x = torch.randn(16, 16)

    model(x)

    return weakref.ref(x)

weak_x = foo()

assert weak_x() is None  # this is false, x leaks
```

The leak seems mainly related to the exception keeping a reference to the function in the traceback:

cycle_9

Full repro:

```python
import weakref
import torch
import thunder
import refcycle  # pip install refcycle


class Identity(torch.nn.Module):
    def forward(self, x):
        raise RuntimeError("FOOBAR")
        return x

def main():
    with torch.device("cpu"):
        model = thunder.jit(Identity())
        x = torch.randn(16, 16)

    try:
        model(x)
    except:
        pass
    return weakref.ref(x)


weak_x = main()

if weak_x() is not None:
    snap = refcycle.snapshot()

    for t in snap.find_by(lambda t: isinstance(t, torch.Tensor)):
        print(t.shape)  # Prints - torch.Size([16, 16])

    # Code to find and save cycles
    for idx, anc in enumerate(snap.ancestors(weak_x())):
        print(idx)
        try:
            cycle = snap.shortest_cycle(anc)
            cycle.export_image(f"cycles/cycle_{idx}.png")
            print(cycle)
            break
        except:
            pass
else:
    print("No leaks!")
```

t-vi:
Fixes #1171

Using plain
```python
    raise e
```
necessarily has a reference to the exception, which has a reference to the current function in the traceback which has a reference to `e`, leading to a ref cycle.

To break this, we switch to manually deleting e in this pattern
```python
    try:
        raise e
    except Exception:
        del e
        raise
```

### DIFF
--- a/thunder/core/interpreter.py
+++ b/thunder/core/interpreter.py
@@ -1342,7 +1342,12 @@ def is_jitting_with_raise():
 
     # Guard against opaque functions which interrupt jitting.
     if (ctx := get_interpretercompilectx_if_available()) is not None:
-        raise InterpreterError(f"Lookaside was not triggered, but there is an active compile context: {ctx}")
+        # nested try to delete ctx from locals
+        try:
+            raise InterpreterError(f"Lookaside was not triggered, but there is an active compile context: {ctx}")
+        except InterpreterError:
+            del ctx
+            raise
 
     return False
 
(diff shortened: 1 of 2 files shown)


lantiga (approved):
Great pattern! Stamped

PR #1194, opened 2024-09-24
Add `prims.copy_to_out_`

### LINKED ISSUE #1173: PR #1110 nearly doubles the compilation & execution time of a copy-heavy program

shino16:
Fixes #1173. This adds a new primitive `prims.copy_to_out_(computed, *, out)`, which is used instead of `prims.copy_` for the update in in-place ops. Unlike `prims.copy_`, `prims.copy_to_out_(computed, *, out)` assumes that `computed` is not used by subsequent ops, so `out` can simply alias `computed`.

main: 63887b3d020932d5a45ade4cac96f6376e59d602
#1193: 2781a201044ec3671f685a493052111a18c45e53
||compilation (s)|execution (ms)|
|:-|-:|-:|
|eager|0.0|10.86|
|`torch.compile(adam.step, backend=thunder)` `main`|94.0|11.52|
|`torch.compile(adam.step, backend=thunder)`, #1193|48.0|6.00|

The rule with `prims.copy_to_out_` is ([link](https://github.com/shino16/lightning-thunder/blob/eb61e6a6636ce0e4fc46c213e7d86ea0aed10491/thunder/clang/__init__.py#L1889-L1891)):
```py
# WARN: `computed` must be an intermediate tensor used solely for this `copy_to_out_` call,
# e.g. copy_to_out_(add(a, b), out=a). Thunder does not guarantee that `computed` remains to have
# the correct values after copy_to_out_ returns. For general-purpose copy, use prims.copy_ instead
```

This rule comes from the fact that any copies onto `out` will be propagated to its alias, `computed`.

To prevent users from using `prims.copy_to_out_` inappropriately, I made the [sanity check](https://github.com/shino16/lightning-thunder/blob/eb61e6a6636ce0e4fc46c213e7d86ea0aed10491/thunder/core/transform_common.py#L63) on `prims.copy_to_out_` rather conservative. When enabled, it raises an error when `computed` is
* used as an input to another ops within the nvFuser region
* defined outside of the region (because it may be modified in-place), or
* used outside of the region (because it may not have the correct value).

See [tests](https://github.com/shino16/lightning-thunder/blob/eb61e6a6636ce0e4fc46c213e7d86ea0aed10491/thunder/tests/test_inplace_copy.py#L175) for examples.

### DIFF
--- a/thunder/core/functionalization.py
+++ b/thunder/core/functionalization.py
@@ -90,7 +90,9 @@ def is_functionalizable(bsym: BoundSymbol) -> bool:
     from thunder.torch import _inplace_to_out_of_place
 
     return (
-        bsym.sym in _inplace_to_out_of_place and bsym.subsymbols and bsym.subsymbols[-1].sym.id == prims.PrimIDs.COPY_
+        bsym.sym in _inplace_to_out_of_place
+        and bsym.subsymbols
+        and bsym.subsymbols[-1].sym.id == prims.PrimIDs.COPY_TO_OUT_
     )
 
 
@@ -252,15 +254,15 @@ def canonicalize_bsym_args(
             new_bsym = new_bsym.from_bsym_swap_proxies(cur_orig_to_view_swap_map, skip_output=True)
 
         # non in-place bsyms would only need to update its outputs.
-        # but in-place bsyms would, as it needs to make sure they return `prims.copy_`'s return value.
+        # but in-place bsyms would, as it needs to make sure they return `prims.copy_to_out_`'s return value.
         if not is_functionalizable(new_bsym):
             bsyms.append(new_bsym)
         else:
             copy_bsym = bsym.subsymbols[-1]
-            copy_out = copy_bsym.flat_proxy_outs[0]
-            copy_dst = copy_bsym.flat_proxy_args[1]
-            swap_map[variableify(copy_dst)] = copy_out
-            # make sure an in-place bsym returns `prims.copy_` output
+            copy_return = copy_bsym.flat_proxy_outs[0]
+            out = copy_bsym.flat_proxy_args[1]
+            swap_map[variableify(out)] = copy_return
+            # make sure an in-place bsym returns `prims.copy_to_out_` output
             new_bsym = new_bsym.from_bsym_swap_proxies(swap_map, skip_inputs=True, skip_subsymbols=True)
             bsyms.append(new_bsym)
         if cur_orig_to_view_swap_map:
@@ -304,26 +306,11 @@ def create_functional_bsym_from(inplace_bsym: BoundSymbol) -> BoundSymbol:
         subsymbols=inplace_bsym.subsymbols,
         _call_ctx=inplace_bsym._call_ctx,
     )
-    if functional_bsym.subsymbols[-1].sym.id == prims.PrimIDs.COPY_:
+    if functional_bsym.subsymbols[-1].sym.id == prims.PrimIDs.COPY_TO_OUT_:
         functional_bsym.subsymbols = functional_bsym.subsymbols[:-1]
     if len(functional_bsym.subsymbols) == 1 and functional_bsym.rhs == functional_bsym.subsymbols[0].rhs:
         functional_bsym.subsymbols = functional_bsym.subsymbols[0].subsymbols
     return functional_bsym
-
-
-def needs_replace_of_sibling_views(
-    var_copy_to: VariableInterface,
-    bsym: BoundSymbol,
-    reverse_swap_map_for_canonicalization: dict[BoundSymbol, dict[VariableInterface, TensorProxy]],
-) -> dict[VariableInterface, TensorProxy]:
-    if (
-        (swap_map_from_canonicalization := reverse_swap_map_for_canonicalization.get(bsym, None))
-        and swap_map_from_canonicalization is not None
-        and var_copy_to in swap_map_from_canonicalization
-    ):
-        return swap_map_from_canonicalization
-    else:
-        return {}
 
 
 def apply_functionalization_to_canonicalized_trace(
@@ -364,18 +351,18 @@ def apply_functionalization_to_canonicalized_trace(
     bsym_to_idx = {bsym: i for i, bsym in enumerate(canonicalized_trace.bound_symbols)}
     consumer_map_of_intermediate_trace = consumers(canonicalized_trace)
     producer_map_of_intermediate_trace = producers(canonicalized_trace)
-    copy_from_to_copy_bsyms: dict[VariableInterface, list[BoundSymbol]] = {}
+    computed_to_copy_bsyms: dict[VariableInterface, list[BoundSymbol]] = {}
     base_to_views: dict[VariableInterface, list[TensorProxy]] = defaultdict(list)
     view_to_base: dict[VariableInterface, TensorProxy] = {}
     bsym_to_trigger_inplace_propagation: dict[BoundSymbol, tuple[TensorProxy, ...]] = {}
 
-    arg_to_copy_froms: dict[VariableInterface, list[TensorProxy]] = {}
+    arg_to_computeds: dict[VariableInterface, list[TensorProxy]] = {}
     swap_map_for_tensor_alias_child: dict[VariableInterface, TensorProxy] = {}
 
     new_bsyms: list[BoundSymbol] = []
     for idx, bsym in enumerate(canonicalized_trace.bound_symbols):
         # `new_bsym` is new in the sense that its args/kwargs do not use a tensor proxy
-        # returned from `prims.copy_`, at this point.
+        # returned from `prims.copy_to_out_`, at this point.
         new_bsym = bsym.from_bsym_swap_proxies(swap_map, skip_output=True)
 
         if new_bsym.sym in _syms_returning_views:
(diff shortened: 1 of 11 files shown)


t-vi (commented):
Thank you @shino16 .
I think the overall approach is good and to my mind, we can merge, but I want to let @crcrpar and/or @IvanYashchuk to take a look.

thunder/torch/__init__.py
crcrpar, line 1: could you educate me why would we want to prefer `clang.copy_to_out_` to `prims.copy_to_out_`?
To allow `out` having a different dtype from `computed`? If so, why wouldn't the prim allow it?

t-vi: Traditionally, we have resolved typing before the prim level when decomposing, so having  `clang.copy_to_out_` decompose to (optionally) `prims.convert_element_type(result_tensor, out_dtype)` plus `prims.copy_out_to_` seems matching the patterns we have.

thunder/core/functionalization.py
```diff
@@ -569,12 +554,13 @@ def _reshape_bsym_ctor(src: TensorProxy, dst: TensorProxy, trace: Trace) -> tupl
...
-            # We swap these arguments in the return value with the outputs of copies onto them
-            # This prevents subsequent transforms from ordering the return statement before those copies
-            swap_map_for_return[variableify(copy_bsym.flat_proxy_args[0])] = copy_bsym.flat_proxy_outs[0]
+            for copy_bsyms in bsym_to_copy_bsyms[bsym]:
+                functionalized_bsyms.extend(copy_bsyms)
+                copy_bsym = functionalized_bsyms[-1]
+                # wrap_return_value_together_with_argments places all the arguments in the return value
+                # We swap these arguments in the return value with the outputs of copies onto them
+                # This prevents subsequent transforms from ordering the return statement before those copies
+                swap_map_for_return[variableify(copy_bsym.flat_proxy_args[0])] = copy_bsym.flat_proxy_outs[0]
```
shino16, line 563: All the changes in this file are just variable renaming, except this loop and [line 547](https://github.com/Lightning-AI/lightning-thunder/pull/1194/files#diff-d894c4085b6c5cb78bcd61db411e3da8b28c0a6688814662b6a05eabc3509fa5L562-R547) brought in commit 47f35d984768e9c482774f50655238e8ec31426f.

This fixes a bug which caused some of the test failures mentioned in https://github.com/Lightning-AI/lightning-thunder/pull/1194#pullrequestreview-2324963202. When `bsym` (`key_bsym` in [line 547](https://github.com/Lightning-AI/lightning-thunder/pull/1194/files#diff-d894c4085b6c5cb78bcd61db411e3da8b28c0a6688814662b6a05eabc3509fa5L562-R547)) is associated with multiple `copy_bsyms`, `bsym_to_copy_bsyms[bsym]` previously looked like `[(reshape,) copy, (reshape,) copy, ...]`. Now it looks like `[[(reshape,) copy], [(reshape,) copy], ...]`, and we iterate through all copies.

shino16:
As of now, the test `thunder/tests/test_jit_general.py::test_litgpt_variants_kvcache[cuda-llama1-like]` does NOT pass.

Minimal reproducible example:
```py
import torch
import thunder

@partial(thunder.jit, disable_inplace_copy_check=True)
def f(q, k, v, mask, idx, src):
    q.index_copy_(2, idx, src)
    k.index_copy_(2, idx, src)
    return torch.nn.functional.scaled_dot_product_attention(q, k, v, mask)

q = torch.randn((1, 4, 2, 16), device='cuda', dtype=torch.float32)
k = torch.randn((1, 4, 3, 16), device='cuda', dtype=torch.float32)
v = torch.randn((1, 4, 3, 16), device='cuda', dtype=torch.float32)
mask = torch.ones((1, 1, 2, 3), device='cuda', dtype=torch.bool)
idx = torch.arange(2).to(device='cuda')
src = torch.randn((1, 4, 2, 16), device='cuda', dtype=torch.float32)

f(q, k, v, mask, idx, src)
```
Execution trace:
```py
def computation(q, k, v, mask, idx, src):
  # q: "cuda:0 f32[1, 4, 2, 16]"
  # k: "cuda:0 f32[1, 4, 3, 16]"
  # v: "cuda:0 f32[1, 4, 3, 16]"
  # mask: "cuda:0 b8[1, 1, 2, 3]"
  # idx: "cuda:0 i64[2]"
  # src: "cuda:0 f32[1, 4, 2, 16]"
  t0 = torch.index_copy(q, 2, idx, src)  # t0: "cuda:0 f32[1, 4, 2, 16]"
    # t0 = ltorch.index_copy(q, 2, idx, src)  # t0: "cuda:0 f32[1, 4, 2, 16]"
      # t0 = prims.index_copy(q, idx, src, 2)  # t0: "cuda:0 f32[1, 4, 2, 16]"
  t2 = torch.index_copy(k, 2, idx, src)  # t2: "cuda:0 f32[1, 4, 3, 16]"
    # t2 = ltorch.index_copy(k, 2, idx, src)  # t2: "cuda:0 f32[1, 4, 3, 16]"
      # t2 = prims.index_copy(k, idx, src, 2)  # t2: "cuda:0 f32[1, 4, 3, 16]"
  [t1, t3] = nvFusion0(t0, q, t2, k)
    # t1 = prims.copy_to_out_(t0, out=q)  # t1: "cuda:0 f32[1, 4, 2, 16]"
    # t3 = prims.copy_to_out_(t2, out=k)  # t3: "cuda:0 f32[1, 4, 3, 16]"
  del q, k
  (t19, _, _, _) = sdpaex_grad_forward_scaled_dot_product_efficient_attention(t0, t2, v, mask, 0.0, False, None)
  del t0, t2
  return t19
```
[`_inplace_copy_sanity_check`](https://github.com/shino16/lightning-thunder/blob/9f005ca59170fa06d86effd11484c0a42e06589f/thunder/core/transform_common.py#L63) raises an error for this, because 
* `t0` is passed to the sdpa operator, and
* if `nvFusion0` had a copy onto `q` in the form of `prims.copy_(XX, copy_to=q)`, `XX` is propagated to `t0`.

Note that, before passing the trace to the nvFuser executor, `prims.copy_to_out_(t0, out=q)` is put after the sdpaex operator thanks to functionalization.

<details>
<summary>Trace just before nvFuser</summary>

```py
def computation(q, k, v, mask, idx, src):
  # q: "cuda:0 f32[1, 4, 2, 16]"
  # k: "cuda:0 f32[1, 4, 3, 16]"
  # v: "cuda:0 f32[1, 4, 3, 16]"
  # mask: "cuda:0 b8[1, 1, 2, 3]"
  # idx: "cuda:0 i64[2]"
  # src: "cuda:0 f32[1, 4, 2, 16]"
  # Functionalized from `t1 = index_copy_(q,2,idx,src)`
  t0 = ltorch.index_copy(q, 2, idx, src)  # t0: "cuda:0 f32[1, 4, 2, 16]"
    # t0 = prims.index_copy(q, idx, src, 2)  # t0: "cuda:0 f32[1, 4, 2, 16]"
  # Functionalized from `t3 = index_copy_(k,2,idx,src)`
  t2 = ltorch.index_copy(k, 2, idx, src)  # t2: "cuda:0 f32[1, 4, 3, 16]"
    # t2 = prims.index_copy(k, idx, src, 2)  # t2: "cuda:0 f32[1, 4, 3, 16]"

  # /opt/pytorch/lightning-thunder/mshinokawa/sandbox/debug.py:60:          return torch.nn.functional.scaled_dot_product_attention(q, k, v, mask)
  # ['t1', 't3'] are replaced by ['t0', 't2'], respectively
  t19 = ltorch.scaled_dot_product_attention(t0, t2, v, mask, 0.0, False, scale=None)  # t19: "cuda:0 f32[1, 4, 2, 16]"
    # subsymbols omitted
  # /opt/pytorch/lightning-thunder/mshinokawa/sandbox/debug.py:58:          q.index_copy_(2, idx, src)
  t1 = prims.copy_to_out_(t0, out=q)  # t1: "cuda:0 f32[1, 4, 2, 16]"

  # /opt/pytorch/lightning-thunder/mshinokawa/sandbox/debug.py:59:          k.index_copy_(2, idx, src)
  t3 = prims.copy_to_out_(t2, out=k)  # t3: "cuda:0 f32[1, 4, 3, 16]"

  # /opt/pytorch/lightning-thunder/mshinokawa/sandbox/debug.py:60:          return torch.nn.functional.scaled_dot_product_attention(q, k, v, mask)
  return {'output': t19, 'flat_args': [t1, t3, v, mask, idx, src]}
```

</details>

#### Possible solutions

* Functionalize `prims.copy_` in a way such that there will never be multiple copies onto the same tensor. If we assume this, we can be sure that `q` never changes after `prims.copy_to_out_(t0, out=q)` in the previous example.
* Preserve order between operations involving `t0` and `prims.copy_to_out_(t0, out=q)`. This order is enforced by functionalization ([link](https://github.com/Lightning-AI/lightning-thunder/blob/314e748cdd27e437065434c821ac4857f25d28cc/thunder/core/functionalization.py#L561)), but they do not establish dependency relationship in terms of outputs and inputs. This would lead to exposing fixes only for nvFuser.

shino16:
I disabled the sanity check for `test_litgpt_variants_kvcache`.

t-vi:
> Note that, before passing the trace to the nvFuser executor, prims.copy_to_out_(t0, out=q) is put after the sdpaex operator thanks to functionalization.

Ugh.

Could it be that `inplace_copy_` is particular here?

I wonder if something along the lines of @IvanYashchuk 's planned primitive for dataflow healing would be useful.

shino16:
The same happens when we use `Tensor.add_` instead of `Tensor.index_copy_` too, so this is a generic issue.

My internship period is about to end, so I can no longer spend much time on this issue. Maybe we can close this PR for now and wait for functionalization of `copy_` or dependency establishment to mature, and merge #1177 or take other strategies if the performance regression in #1173 is urgent.

t-vi:
closing for now @crcrpar please reopen as you see fit.

PR #1195, opened 2024-09-24
skip when `bitsandbytes` is unavailable

crcrpar:


t-vi (changes requested):
Great idea, but can we
```
bitsandbytes_executor = get_bitsandbytes_executor()
```
at module level and use
```
pytest.mark.skipif(bitsandbytes_executor is None, reason="bitsandbytes not available")
```
please?

lantiga (approved):
👍

t-vi (approved):
Thank you @crcrpar

PR #1196, opened 2024-09-24
Skip thunder/tests/test_examine_memory.py:test_nanogpt_block on Hopper

kiya00:
The test fails on H100 due to the changes of actual peak memory changes:
```py
=================================== FAILURES ===================================
______________________________ test_nanogpt_block ______________________________

    @requiresCUDA
    def test_nanogpt_block():
        import thunder.tests.nanogpt_model as nanogpt_model
    
        config = nanogpt_model.GPTConfig(dropout=0)
        block = nanogpt_model.Block(config).to(dtype=torch.float32, device="cuda")
        cblock = TorchExecutor.make_callable(block)
        inp = make_tensor((2, config.block_size, config.n_embd), dtype=torch.float32, device="cuda", requires_grad=True)
        cblock(inp)
    
        fw_trace = thunder.last_traces(cblock)[-1]
        bw_trace = thunder.last_backward_traces(cblock)[-1]
        max_mem_fw = get_alloc_memory(fw_trace)
        max_mem_bw = get_alloc_memory(bw_trace)
    
        result = measure_fw_and_bw_memory_usage(fw_trace, bw_trace)
>       assert max_mem_fw[0] == result["fw_peak"]
E       assert 381754368 == 409017344

thunder/tests/test_examine_memory.py:115: AssertionError
```
The estimated memory usage is not the same as actual peak memory usage, some changes in cuda/cublas that are unique to Hopper that change the actual memory consumption.

Since the estimated memory usage is a fixed value and has been tested on other hardware types, here is a fix to skip this test case on H100


kiya00:
Hi @crcrpar @kshitij12345 , I tried to skip this test, is that the right way to skip it on Hopper?

thunder/tests/test_examine_memory.py
```diff
@@ -98,6 +98,14 @@ def bar2(a, b):  # [5,2], [2,2]
 
 @requiresCUDA
 def test_nanogpt_block():
+    from thunder.tests.test_sdpaex_executor import device_version_support, CudaVersion
+
+    torch_device = torch.device("cuda")
+    # The estimated memory usage is not the same as actual peak memory usage on Hopper
+    if device_version_support(torch_device, CudaVersion(9, 0), CudaVersion(9, 0)):
```
kshitij12345, line 105: Would we want to run the test on device with compute capability more than 9.0? If not then following should do the trick

```suggestion
    if torch.cuda.get_device_capability() >= (9, 0):
```

thunder/tests/test_examine_memory.py
```diff
@@ -98,6 +98,14 @@ def bar2(a, b):  # [5,2], [2,2]
 
 @requiresCUDA
 def test_nanogpt_block():
+    from thunder.tests.test_sdpaex_executor import device_version_support, CudaVersion
+
+    torch_device = torch.device("cuda")
```
kiya00, line 103:
```suggestion
```

kiya00:
The reason is "PyTorch NGC container uses a 32MB workspace for cuBLAS for Hopper for performance reasons."
Related cherry-pick: https://gitlab-master.nvidia.com/dl/pytorch/pytorch/-/commit/6ca9eab1e748338a061bb99651141b32f0575c44
I think it's hard to get the cublas workspace size in python level and not necessary, the estimated memory is a fixed value. We can either skip this test case or modify it to assert equal to a fixed value, wdyt @crcrpar @kshitij12345

kiya00:
Hi @IvanYashchuk @kshitij12345 @crcrpar , do we merge this to fix the container nvbug, or modify it to assert the estimated memory to equal to a fixed value

IvanYashchuk (approved):
Let's merge this now and modify the assert later.

```
>       assert max_mem_fw[0] == result["fw_peak"]
E       assert 381754368 == 409017344
```
The difference here is interesting, exactly 26 MiB.

t-vi (approved):
Thank you @kiya00 @kshitij12345 @IvanYashchuk

PR #1197, opened 2024-09-25
Define nvFuser `reshape` operations with an inline definition.

kevinstephano:
This is a debugability improvement to nvFuser definitions from Thunder!

`reshape`s were being defined from Thunder as follows:
```python
    S40 = fd.define_scalar(4, dtype=DataType.Int)
    S41 = fd.define_scalar(64, dtype=DataType.Int)
    S42 = fd.define_scalar(6, dtype=DataType.Int)
    S43 = fd.define_scalar(128, dtype=DataType.Int)
    T45 = fd.ops.reshape(T26, new_shape=[S40, S41, S42, S43])
 ```
 This PR changes it to:
 ```python
    T45 = fd.ops.reshape(T26, new_shape=[4, 64, 6, 128])
 ```
 
 This API has been in nvFuser for a long time but was never updated in Thunder as `broadcast_in_dim` already took advantage of this feature.  The following [nvFuser PR](https://github.com/NVIDIA/Fuser/pull/2643) also allowed for inline printing of definitions that greatly reduces the repro size when printing from Thunder definitions.

This is a one line change to the `reshape` operation in the nvFuser Executor in Thunder!

### DIFF
--- a/thunder/executors/nvfuserex_impl.py
+++ b/thunder/executors/nvfuserex_impl.py
@@ -1224,7 +1224,7 @@ def reshape(a: TensorProxy, shape: list[int, NumberProxy, ...], *, fd: FusionDef
     nv_a = getnv(a, fd, lc_to_nv_map)
     if any(map(lambda x: isinstance(x, NumberProxy), shape)):
         nv_shape = getnv(shape, fd, lc_to_nv_map)
-    else: 
+    else:
         nv_shape = shape
 
     return fd.ops.reshape(nv_a, nv_shape)
--- a/thunder/executors/nvfuserex_impl.py
+++ b/thunder/executors/nvfuserex_impl.py
@@ -1220,10 +1220,14 @@ def _reshape_check(a: TensorProxy, shape: list[int]) -> bool:
     return is_supported_tensor(a)
 
 
-def reshape(a: TensorProxy, shape: list[int], *, fd: FusionDefinition, lc_to_nv_map: dict) -> Any:
+def reshape(a: TensorProxy, shape: list[int, NumberProxy, ...], *, fd: FusionDefinition, lc_to_nv_map: dict) -> Any:
     nv_a = getnv(a, fd, lc_to_nv_map)
+    if any(map(lambda x: isinstance(x, NumberProxy), shape)):
+        nv_shape = getnv(shape, fd, lc_to_nv_map)
+    else: 
+        nv_shape = shape
 
-    return fd.ops.reshape(nv_a, shape)
+    return fd.ops.reshape(nv_a, nv_shape)
 
 
 register_supported(PrimIDs.RESHAPE, reshape, _reshape_check)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: thunder/executors/nvfuserex_impl.py [hunk 1]
```diff
@@ -1214,9 +1214,8 @@ def _reshape_check(a: TensorProxy, shape: list[int]) -> bool:
 
 def reshape(a: TensorProxy, shape: list[int], *, fd: FusionDefinition, lc_to_nv_map: dict) -> Any:
     nv_a = getnv(a, fd, lc_to_nv_map)
-    nv_shape = getnv(shape, fd, lc_to_nv_map)
```
jjsjann123, line 1217: I see what you are trying to do here. I think we need to check that `shape` does not contain `NumberProxy` and only do this simplification then.

CI failures looks strange, seems like an infra issue?

tfogal (approved):
The effect looks great to me! but Jie would know more about the dynamic shapes thing than I would.

I wonder if we could just special case inside `getnv`: `if isinstance(x, Sequence) and all(isinstance(y, int) for y in x):` ? That might help other places, too, then.

thunder/executors/nvfuserex_impl.py
wujingyue: I'm probably missing something. Isn't shape of type list[int] as defined in the signature?

jjsjann123: yeah, that annotation needs to be updated. 
But if we are to support dynamic shape, we'll update it as `shape: list[int, NumberProxy]`.

jjsjann123: to be specific, I'm saying that we should do this instead:
```
    if any(map(lambda x: isinstance(x, Proxy), shape)):
        nv_shape = getnv(shape, fd, lc_to_nv_map)
    else: 
        nv_shape = shape 
```

extra credit if we also fix the annotation like the actual prim: https://github.com/Lightning-AI/lightning-thunder/blob/f70b0fe22d27a2e2a0a6ac31cf2c52c9d4e4e35d/thunder/core/prims.py#L3158

jjsjann123: could also use a comment as, why we would want to use `shape` directly in the code.

kevinstephano: Do we need to make a change now for dynamic shapes? `broadcast_in_dim` should have identical `shape` usage.

kevinstephano: The only reason shape was not originally used, directly, was because nvFuser previously required using `define_scalar()` to add a scalar constant.  That behavior was changed a long time ago and did not get propagated to `reshape`, for whatever reason, but did make it to the `broadcast_in_dim` definition.

jjsjann123: As long as you are passing CIs, you don't have to. I don't mind clean up afterwards.

But I'm a bit suspicious that you are not going to pass CI, because we happen to have a reshape test case with symbolic values: https://github.com/Lightning-AI/lightning-thunder/blob/f70b0fe22d27a2e2a0a6ac31cf2c52c9d4e4e35d/thunder/tests/test_jit_general.py#L1124-L1125

jjsjann123: actually, that one just checks for cache hit... so even though nvfuser recompiles I think you are still good.

jjsjann123 (approved):
Feel free to merge it if CI is green. I'll clean up afterwards.

jjsjann123:
Thanks for updating the PR. cc'ing @t-vi for the merge review. :pray:

t-vi (approved):
Thank you @kevinstephano @jjsjann123 @tfogal @wujingyue

PR #1201, opened 2024-09-25
Numberproxy slice

### LINKED ISSUE #1182: python `slice` is not represented properly in thunder
## 🚀 Feature

`slice` is currently showing up in trace differently from list/tuple, where NumberProxy is printed explicitly, resulting in an invalid python program.

e.g. in the script below:

```
import thunder
import torch
dtype = torch.float32
 
def foo(a):
    return a[..., : (a.shape[-1] // 2)]

a = torch.randn(2, 2, device="cuda")
out = jfoo(a)
```
On PR #1027, commit `4d260aed2b0939cebdeeeb4f04cf47358d3d9c8b`.

We have a trace like this:
```
def computation(a, i1):
  # a: "cuda:0 f32[[IntegerProxy name=i0, value=2, static=CONSTRAINT.CONSTRAINABLE], [IntegerProxy name=i1, value=2, static=CONSTRAINT.CONSTRAINABLE]]"
  # i1: "int 2"

  # /volume/thunder_dynamic/t.py:7:         return a[..., : (a.shape[-1] // 2)]
  b2 = prims.signbit(i1)  # b2: "bool False"
  _ = prims.signbit(2)
  b3 = prims.ne(b2, False)  # b3: "bool False"
  f4 = prims.fmod(i1, 2)  # f4: "float 0.0"
  i5 = prims.convert_element_type(f4, int)  # i5: "int 0"
  b6 = prims.ne(i5, 0)  # b6: "bool False"
  b7 = prims.bitwise_and(b3, b6)  # b7: "bool False"
  i8 = prims.div(i1, 2)  # i8: "int 1"
  i9 = prims.convert_element_type(b7, int)  # i9: "int 0"
  i10 = prims.sub(i8, i9)  # i10: "int 1"
  t37 = ltorch.getitem(a, (..., slice(None, [IntegerProxy name=i10, value=1, static=CONSTRAINT.CONSTRAINABLE], None)))  # t37: "cuda:0 f32[[IntegerProxy name=i29, value=2, static=CONSTRAINT.CONSTRAINABLE], [IntegerProxy name=i36, value=1, static=CONSTRAINT.CONSTRAINABLE]]"
```

This line is not a valid python program.
```
t37 = ltorch.getitem(a, (..., slice(None, [IntegerProxy name=i10, value=1, static=CONSTRAINT.CONSTRAINABLE], None)))
```
We should instead print out something like `slice(None, i10, None)`.

We plan on just having `slice` print out being handled inline, similar to how list/tuple. Thanks to suggestion by @t-vi & @mruberry .

### Alternative

I initial thought is that we needed `SliceProxy`, like `ListProxy` / `TupleProxy`. I thought that's needed, but as @t-vi pointed out that 

> I don't think that TupleProxy, DictProxy, and ListProxy are currently not used at all, they are relicts of the functional JIT which tried to proxy everything. I think CollectionProxy is used exclusively for autograd.
Having a slice proxy is certainly possible and may be needed if we run into trouble putting numberproxies into slices.
(but I wondered why you have lists in your slices above until... ).

I don't think we need `SliceProxy` for now, so I'll proceed with the less invasive approach first to unblock myself.

jjsjann123:
Fixes #1182

Highlighting impactful changes:
1. updated tree_flatten to use `OPTREE_NAMESPACE` as the default namespace. This allows flattening operations to unpack custom pytree node, (including `slice`). I have to undo this in code printing to avoid breaking dataclasses.

Changes in this PR:
1. fixing `_unpack_inputs` used in `thunder_trace` to register proxy names used by `NumberProxies` in `TensorProxy.shape`, this avoids name conflicts in the new trace.
2. updated `slice` print function, so NumberProxy in `slice` only shows name, rather than the full `[IntegerProxy ...]`, which results in illegal python.
3. update `slice` unpack, to treat it as a CollectionProxy
4. register `slice` as a pytree node, so fatten calls would work on `slice` instances with NumberProxy as slice args.


jjsjann123:
err seem to cause some issue with grad transformation... need to debug this one.

jjsjann123:
Interesting.... I see now that after this change, `trace` now takes slice inputs as direct input and converts that into a list now...

```
def foo(a, key):
    return thunder.clang.getitem(a, key)
a = torch.randn(2, 2, 9, device="cuda")
tracer = thunder.trace(inline_trace=False)
foo_trace = tracer(foo, a, (..., slice(None, 8, None)))
print(foo_trace)
```

```
import thunder
import torch

@torch.no_grad()
@no_autocast
def foo(a, key):
  # a: "cuda:0 f32[2, 2, 9]"
  # key: "Collection"
  _, _, = key
  t12 = prims.slice_prim(a, [0, 0, 0], [2, 2, i0], [1, 1, 1])  # t12: "cuda:0 f32[2, 2, 8]"
  return t12
```

And this is causing issue in autograd... :face_exhaling: Looks like I must have missed something...

thunder/core/pytree.py
```diff
@@ -21,7 +21,15 @@
...
-def tree_flatten(args, namespace=""):
+optree.register_pytree_node(
+    slice,
+    lambda s: ([s.start, s.stop, s.step], None, None),
+    lambda _, children: slice(*children),
+    namespace=OPTREE_NAMESPACE,
+)
+
+
+def tree_flatten(args, namespace=OPTREE_NAMESPACE):
```
jjsjann123, line 32: following the earlier question, when would we want to avoid flattening without custom node?

t-vi (approved):
Looks great, thank you @jjsjann123

PR #1207, opened 2024-09-27
Auto-registered ops: Make torch functions with the same name in torch and torch.Tensor namespace share the same symbol

### LINKED ISSUE #1140: duplicate symbols from autoregistration  (torch.foo vs. torch.Tensor.foo)
Autoregistration registers duplicate symbols under the same name, which is not ideal, and we should go the same way as for manually registered symbols (see below).

Repro:

```python
import thunder, torch
def fn(a):
     return torch.positive(a.positive())
jf = thunder.jit(fn)
jf(torch.randn(1))
lt = thunder.last_traces(jf)[0]
print(lt)
s1 = lt.bound_symbols[1].sym
s2 = lt.bound_symbols[2].sym
assert s1 == s2, f"{s1} != {s2}"
```

```python
import thunder
import torch

@torch.no_grad()
@no_autocast
def computation(a):
  # a: "cpu f32[1]"

  # <ipython-input-2-d90f2ca6b8df>:2: 	    return torch.positive(a.positive())
  t0 = ltorch.positive(a)  # t0: "cpu f32[1]"
  t1 = ltorch.positive(t0)  # t1: "cpu f32[1]"
  return t1

AssertionError                            Traceback (most recent call last)
----> 1 assert s1 == s2, f"{s1} != {s2}"

AssertionError: [Symbol name=positive] != [Symbol name=positive]
```

By contrast, manually registered functions register both with a single symbol:
https://github.com/Lightning-AI/lightning-thunder/blob/3770bf02240cdb6ec86a6d5224807bc6ad39d2a9/thunder/torch/__init__.py#L251-L256

(credit for discovery and analysis: @k223kim , thank you)

kiya00:
<details>

</details>

## What does this PR do?

Fixes #1140 .

## PR review

## Did you have fun?


thunder/torch/__init__.py
```diff
@@ -2912,7 +2912,7 @@ def scatter_add_(a: TensorLike, /, dim: int, index: TensorLike, src: TensorLike)
     return prims.copy_(scatter_add(a, dim, index, src), a)
 
 
-@torchsymbol(torch.take_along_dim)
+@torchsymbol(torch.take_along_dim, torch.Tensor.take_along_dim, is_method=True)
```
crcrpar, line 2915: Question (for my understanding): I thought `is_method=True` would be sufficient here due to this logic: https://github.com/Lightning-AI/lightning-thunder/blob/59467aa5e401ea1c4d2c4e6f9e94d1dbe147a19e/thunder/torch/__init__.py#L183-L188 and I'm speculating we register `torch.Tensor.take_along_dim` to `_torch_to_thunder_function_map` twice, which I wouldn't find harmful but I'm just curious

crcrpar (commented):
I have one question but this looks neat

thunder/torch/__init__.py
kiya00: Thank you I missed this, let me remove the duplicates

kiya00:
Hi @t-vi @k223kim , this PR let the functions with the same name in `torch` and `torch.Tensor` namespace to use the same symbol.
I noticed there's a comment:
https://github.com/Lightning-AI/lightning-thunder/blob/59467aa5e401ea1c4d2c4e6f9e94d1dbe147a19e/thunder/torch/__init__.py#L5534-L5536
Do we want the same name in all the torch namespaces (now we have `torch`,`torch.Tensor`,`torch.nn.functional`,`torch.special`,`torch.linalg`) to use the same symbol? Otherwise this can't be converted to assert. In the manual registration, it's decided manually like `@torchsymbol(torch.sigmoid, torch.nn.functional.sigmoid, torch.special.expit, is_method=True)`

t-vi:
Thank you for working on this @kiya00 ! So the two things that we would want is
- unique symbol names (i.e. no two symbols with the same name),
- all symbols are discoverable (can be looked up either through `f"{module}.{name}"` or through the executor) for serialization,
- ideally we would want a given torch op to have only one symbol for the benefit of transform writers. 

I would be happy with any way to achieve this (separate namespaces or other disambiguation is totally fine).

t-vi (approved):
Thank you @kiya00 @crcrpar

k223kim:
Thank you @kiya00 @crcrpar ! Appreciate the help :)

PR #1063, opened 2024-08-28
`all_reduce` & `wait_tensor` of `torch.ops._c10d_functional`

crcrpar:
## What does this PR do?

Update all_reduce `torchsymbol` to cover https://github.com/pytorch/pytorch/blob/bb22132c8dd4647f6b75481dabfbe5d2717e7b4d/torch/distributed/nn/functional.py#L205

PR #1209, opened 2024-09-27
try removing `ltorch.copy_`

crcrpar:
It seems that I've added `ltorch.copy_` in #1063 but I don't remember exactly why I did so.
I even am speculating that was an unwanted change with some cost.
So in this PR, I just want to try reverting the change and see if things can be simplified.

### DIFF
--- a/thunder/executors/nvfuserex_impl.py
+++ b/thunder/executors/nvfuserex_impl.py
@@ -2071,7 +2071,8 @@ def copy_(
 ) -> Any:
     nvcopy_from = getnv(copy_from, fd, lc_to_nv_map)
     nvcopy_to = getnv(copy_to, fd, lc_to_nv_map)
-    fd.add_output(nvcopy_from, alias_input=nvcopy_to)
+    alias_output = fd.ops.set(nvcopy_from)
+    fd.add_output(alias_output, alias_input=nvcopy_to)
     return nvcopy_to
 
 
--- a/thunder/executors/nvfuserex_impl.py
+++ b/thunder/executors/nvfuserex_impl.py
@@ -2071,8 +2071,7 @@ def copy_(
 ) -> Any:
     nvcopy_from = getnv(copy_from, fd, lc_to_nv_map)
     nvcopy_to = getnv(copy_to, fd, lc_to_nv_map)
-    alias_output = fd.ops.set(nvcopy_from)
-    fd.add_output(alias_output, alias_input=nvcopy_to)
+    fd.add_output(nvcopy_from, alias_input=nvcopy_to)
     return nvcopy_to
 
 
--- a/thunder/tests/test_inplace_copy.py
+++ b/thunder/tests/test_inplace_copy.py
@@ -170,12 +170,21 @@ def test_inplace_copy_dst_copy_returned_issue_1109(executor, device, dtype):
     # This pattern is unsafe in general. Disabling sanity check to silence
     # exception for testing
     traced_foo = executor.make_callable(func, disable_inplace_copy_check=True)
-    a = make_tensor((4, 4), device=device, dtype=tdtype)
-    a_ref = a.clone()
+    t0 = make_tensor((4, 4), device=device, dtype=tdtype)
+    t0_ref = t0.clone()
 
-    o_thunder = traced_foo(a)
-    o_eager = func(a_ref)
+    actual_t1, actual_t2 = traced_foo(t0)
 
-    assert_close(a_ref, a)
-    for o, o_ref in zip(o_thunder, o_eager):
-        assert_close(o, o_ref)
+    expected = t0_ref.sin().cos()
+    expected_t1 = t0_ref.sin()
+    expected_t2 = expected_t1.cos()
+    expected = expected_t2
+
+    assert_close(t0, expected)
+    assert_close(actual_t2, expected_t2)
+    # FIXME(crcrpar): Since there's no `ltorch.Tensor.copy_`, functions like `func` would not
+    # be observed and executed with pytorch eager mode. Though there should be either an audit of
+    # `prims.copy_` in a nvfuser region and/or what #1110 did.
+    assert actual_t1.data_ptr() == actual_t2.data_ptr()
+    with pytest.raises(AssertionError):
+        assert_close(actual_t1, expected_t1)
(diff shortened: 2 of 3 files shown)


t-vi:
isn't torch.Tensor.copy_ a legit method?

crcrpar:
We don't seem to have solid way to support the op with decent complexity in implementation and cost of compile time and performance while it doesn't seem that we have many use at the moment.

thunder/tests/test_inplace_copy.py
```diff
@@ -158,9 +159,9 @@ def func3(x, y):
 def test_inplace_copy_dst_copy_returned_issue_1109(executor, device, dtype):
     def func(T0):
         T1 = torch.sin(T0)
-        T0.copy_(T1)  # destination.copy_(source)
+        prims.copy_(T1, T0)
```
crcrpar, line 162: https://dev.azure.com/Lightning-AI/lightning/_build/results?buildId=219314&view=logs&j=3f274fac-2e11-54ca-487e-194c91f3ae9f&t=7d03ab59-8b52-555c-a216-77e14d28deec&l=15179

crcrpar: @IvanYashchuk  I thought this would satisfy what you recommended in our offline chat. With the removal, we're to explicitly disallow `copy_` in a callable

jjsjann123:
How's this branch going?

crcrpar:
still we could write an equivalent function even without `ltorch.copy_` using `prims.copy_` but that function would not be executed in eager mode.

thunder/executors/nvfuserex_impl.py
```diff
@@ -2071,8 +2071,7 @@ def copy_(
 ) -> Any:
     nvcopy_from = getnv(copy_from, fd, lc_to_nv_map)
     nvcopy_to = getnv(copy_to, fd, lc_to_nv_map)
-    alias_output = fd.ops.set(nvcopy_from)
-    fd.add_output(alias_output, alias_input=nvcopy_to)
+    fd.add_output(nvcopy_from, alias_input=nvcopy_to)
```
jjsjann123, line 2074: Are we supposed to do this? I'm worried that this will give us wrong program.

thunder/tests/test_inplace_copy.py
```diff
@@ -169,12 +170,21 @@ def func(T0):
...
+    expected_t1 = t0_ref.sin()
+    expected_t2 = expected_t1.cos()
+    expected = expected_t2
+
+    assert_close(t0, expected)
+    assert_close(actual_t2, expected_t2)
+    # FIXME(crcrpar): Since there's no `ltorch.Tensor.copy_`, functions like `func` would not
+    # be observed and executed with pytorch eager mode. Though there should be either an audit of
+    # `prims.copy_` in a nvfuser region and/or what #1110 did.
+    assert actual_t1.data_ptr() == actual_t2.data_ptr()
```
jjsjann123, line 188: Thanks for putting a FIXME there. but this is the case that nvfuser is producing wrong result in `func`?

one of the output will still be aliasing the input, and that's not what the program should be translated. i.e. I don't think an audit that removes redundant `prims.copy_` would be able to solve that.

jjsjann123: For the record, I'm all in for having a `remove redundant inplace copy`, since that's a net improvement. But I'm not sure it would resolve the issue with performance regression that we are seeing right now.

thunder/executors/nvfuserex_impl.py
crcrpar: This PR is composed with the revert of https://github.com/Lightning-AI/lightning-thunder/pull/1209 and the test update. 
I thought `fd.ops.set` is something ideally we want to avoid as per https://github.com/Lightning-AI/lightning-thunder/issues/1173

jjsjann123: sorry about not being more clear in #1173 

We still wanted to have the set for `copy_`. I think the root-cause with issue #1173 is that we are returning `nvcopy_from` in the fusion region.

crcrpar: why isn't there a copy op?

jjsjann123: There is a copy op, the difference is, whether we are returning a different buffer, or the aliased source.

I think the example below would help.

The thunder program below behaves differently, depends on what's the return value.
```
import thunder
import torch
from thunder.core import prims

def foo(x):
    x_relu = prims.abs(x)
    old_x = prims.copy_(x_relu, x)
    return old_x  # this returns x, the aliased source from the program.
    #return x_relu  # this returns the x_relu, which should be in a different buffer.

jfoo = thunder.jit(foo)

x = torch.randn(2, 4, device="cuda")
o = jfoo(x)

print(x)
print(o)
print(x.data_ptr())
print(o.data_ptr())
```

They are translated as nvfuser programs correspondingly.
```
import torch

def nvfuser_fusion_id4(fd : FusionDefinition) -> None :
    T0 = fd.define_tensor(shape=[2, 4], contiguity=[True, True], dtype=DataType.Float, is_cpu=False, stride_order=[1, 0])
    T1 = fd.ops.neg(T0)
    T2 = fd.ops.set(T1)
    fd.add_output(T2, T0)
    fd.add_output(T2)
    # fd.add_output(T1)  # doing this would cause us to return an extra buffer.
```
The part of `copy_` is translated to
```
T2 = fd.ops.set(T1)  # this is the extra copy we added.
fd.add_output(T2, T0)  # the extra copy is used to overwrite T0.
```
When we return T2 as output of the fusion, since T2 is marked as alias to T0 (sharing storage), we are not returning an extra buffer here.

However, the story changes when we  **return T1** from the fusion. Since `T1` is a new buffer in the program. You can play with this using vvv:
```
    nvfuser_fusion_id4(fd)

inputs = [
    torch.testing.make_tensor((2, 4), dtype=torch.float32, device='cuda:0'),
]

print(inputs)
o = fd.execute(inputs)
print(inputs)

print(o[0].data_ptr())
print(inputs[0].data_ptr())
```

You'll notice that without the extra `fd.ops.set` in `copy_`, nvfuser won't be able to faithfully represent the intended thunder behavior.

crcrpar: okay, then I don't think this PR deserves a merge.

f0eac737e2a32792f2b0a4101a3075b730329b9f: compile 54s, 11.07ms
82dc7a7c0f4739604f7a32b98defd3158abf5ac9: compile 97s, 14.28ms

thunder/tests/test_inplace_copy.py
crcrpar: If a callable has multiple in-place ops, then the redundant copies would be cleaned up. We just don't have an equivalent for multiple raw `prims.copy_`s

PR #1210, opened 2024-09-27
access `.is_leaf` before `.grad`

### LINKED ISSUE #1205: Warning - The .grad attribute of a Tensor that is not a leaf Tensor is being accessed
```python
import torch
import thunder

def foo(x):
    return x

x = torch.randn(3, requires_grad=True) + 1
thunder.jit(foo)(x)

```

Output
```python
thunder/core/proxies.py:1954: UserWarning: The .grad attribute of a Tensor that is not a leaf Tensor is being accessed. Its .grad attribute won't be populated during autograd.backward(). If you indeed want the .grad field to be populated for a non-leaf Tensor, use .retain_grad() on the non-leaf Tensor. If you access the non-leaf Tensor by mistake, make sure you access the leaf Tensor instead. See github.com/pytorch/pytorch/pull/30531 for more informations. (Triggered internally at /home/kkalambarkar/git/pytorch/build/aten/src/ATen/core/TensorBody.h:489.)
  if t.grad is not None:
```

This is also triggered when running `benchmark_litgpt.py` with dynamo_thunder as few `thunder` subgraphs receive non-leaf tensors as input.

Related : #1070

crcrpar:
## What does this PR do?

Fixes #1205

### DIFF
--- a/thunder/core/proxies.py
+++ b/thunder/core/proxies.py
@@ -1951,7 +1951,7 @@ def tensorproxy(t: torch.Tensor, /, *, name: None | str, history: None | tuple =
     dtype = dtypes.to_dtype(t.dtype)
 
     grad = None
-    if t.grad is not None:
+    if t.is_leaf and t.grad is not None:
         grad_pr = None
         if history is not None:
             attr_pr = ProvenanceRecord(inst=PseudoInst.CONSTANT, inputs=[], value="grad")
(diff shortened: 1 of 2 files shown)


kshitij12345 (approved):
LGTM, thank you!

t-vi:
Great find. Should we add a test case?

t-vi (approved):
Thank you @crcrpar @kshitij12345

Issue #1043, opened 2024-08-26
Cannot backward when the return values of `torch.split` are partially unused


shino16:
## 🐛 Bug

```py
import torch
import thunder

@thunder.jit
def f(a):
    a_1, a_2, a_3 = torch.split(a, 2)
    return a_1

a_1 = f(torch.zeros((5, 2), dtype=torch.float32, requires_grad=True))
a_1.sum().backward()
```
```
  ...
  File "/opt/pytorch/lightning-thunder/thunder/clang/__init__.py", line 1292, in cat
    _, output_dtype = utils.elementwise_type_promotion(
  File "/opt/pytorch/lightning-thunder/thunder/core/utils.py", line 474, in elementwise_type_promotion
    check_type(a, (TensorProxy, Number, NumberProxy))
  File "/opt/pytorch/lightning-thunder/thunder/core/baseutils.py", line 107, in check_type
    check(
  File "/opt/pytorch/lightning-thunder/thunder/core/baseutils.py", line 103, in check
    raise exception_type(s())
ValueError: None had an unexpected type <class 'NoneType'>. Supported types are (<class 'thunder.core.proxies.TensorProxy'>, <class 'numbers.Number'>, <class 'thunder.core.proxies.NumberProxy'>)
```

## Explanation

`a_1, a_2, a_3 = torch.split(a, 2)` turns into a bsym
```py
[a_1, _, _] = ltorch.split(a, 2, 0)
  # a_1 = prims.slice_prim(a, [0, 0], [2, 2], [1, 1])  # a_1: "cpu f32[2, 2]"
  # a_2 = prims.slice_prim(a, [2, 0], [4, 2], [1, 1])  # a_2: "cpu f32[2, 2]"
  # a_3 = prims.slice_prim(a, [4, 0], [5, 2], [1, 1])  # a_3: "cpu f32[1, 2]"
```
Its output (accessed [here](https://github.com/Lightning-AI/lightning-thunder/blob/7c1f94ab64e0103ced14696da8fb1652e830ab6d/thunder/core/transforms.py#L3310)) is `[<TensorProxy(name="a_1", dtype=thunder.dtypes.float32, shape=(2, 2))>, None, None]`. Its corresponding backward function, `ltorch.cat`, cannot concatenate a tensor and `None`s.

cc @apaz-cli

shino16:
Is this something out of Thunder's official support? But the following seems to work.
```py
@thunder.jit
def f(a):
[...]

PR #1211, opened 2024-09-27
split backward : fix for unused outputs

kshitij12345:
Fixes https://github.com/Lightning-AI/lightning-thunder/issues/1043

```python
import torch
import thunder

@thunder.jit
def f(a):
    a_1, a_2, a_3 = torch.split(a, 2)
    return a_1

a_1 = f(torch.zeros((5, 2), dtype=torch.float32, requires_grad=True))
a_1.sum().backward()
```

Error
```
ValueError: None had an unexpected type <class 'NoneType'>. Supported types are (<class 'thunder.core.proxies.TensorProxy'>, <class 'numbers.Number'>, <class 'thunder.core.proxies.NumberProxy'>)
```

Script above failed as for the unused outputs, we get None as incoming gradient. However, the rule doesn't handle this case.

https://github.com/Lightning-AI/lightning-thunder/blob/59467aa5e401ea1c4d2c4e6f9e94d1dbe147a19e/thunder/core/transforms.py#L2146-L2150

### DIFF
--- a/thunder/core/transforms.py
+++ b/thunder/core/transforms.py
@@ -2138,14 +2138,26 @@ def nll_loss_backward(input, target, weight, reduction, ignore_index, total_weig
 def split_aug_fwd(a: TensorProxy, split_size_or_sections: int | Sequence[int], dim: int = 0) -> VJPDual:
     from thunder.torch import split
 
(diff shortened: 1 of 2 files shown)


t-vi (approved):
Thank you @kshitij12345

PR #1212, opened 2024-09-27
Fix method_name for torch.Tensor.all/any/max

kiya00:
<details>

</details>

## What does this PR do?

When debugging #1140, I found that `@torchsymbol(torch.any, is_method=True, id="torch.any")def any_tensor(...)` should register `torch.Tensor.any` in the `_torch_to_thunder_function_map`, but it doesn't, the reason is that
https://github.com/Lightning-AI/lightning-thunder/blob/91e9f31db423492cc5d42e3622fc5d5c775d6c80/thunder/torch/__init__.py#L187-L192
uses `fn.__name__` which is `any_tensor`, so I think the fix is to give the `method_name` explicitly for this kind of operators

## PR review

## Did you have fun?


crcrpar (approved):
clinical, great investigation

t-vi (approved):
Thank you @kiya00 @crcrpar

Issue #1025, opened 2024-08-22
Drop dependency


IvanYashchuk:
Drop the igraph dependency for mincut, one less dep of something large.

PR #1244, opened 2024-10-01
Remove igraph in favor of networkx

lantiga:
This PR fixes a licensing issue we have had since the early days.

igraph (https://igraph.org) used to compute mincut during rematerialization) is GPL licensed, so it's incompatible with our license.

On the other hand, NetworkX (https://networkx.org) is BSD licensed and serves the purpose well.

This PR replaces one with the other to sanitize the situation.

Fixes https://github.com/Lightning-AI/lightning-thunder/issues/1025.

### DIFF
--- a/thunder/core/rematerialization.py
+++ b/thunder/core/rematerialization.py
@@ -369,13 +369,11 @@ def find_cut(
     g = nx.DiGraph()
     g.add_edges_from(edges)
 
-    _, (set_a, set_b) = nx.minimum_cut(g, "source", "sink")
+    _, (reachable, non_reachable) = nx.minimum_cut(g, "source", "sink")
 
-    cut_edges = []
-    for u in set_a:
-        for v in g.successors(u):
-            if v in set_b:
-                cut_edges.append((u, v))
+    cut_edges = set()
+    for u, nbrs in ((n, g[n]) for n in reachable):
+        cut_edges.update((u, v) for v in nbrs if v in non_reachable)
 
     cut_nodes = set()
     for node_in, node_out in cut_edges:
--- a/thunder/core/rematerialization.py
+++ b/thunder/core/rematerialization.py
@@ -7,7 +7,7 @@ from collections.abc import Sequence
 from collections import defaultdict
 import time
 
-from igraph import Graph
+import networkx as nx
 
 from thunder.core import prims, utils
 from thunder.core.baseutils import BoundSymbolInterface, ProxyInterface
@@ -312,12 +312,9 @@ def find_cut(
 
     # Create a graph
     edges = []
-    name_to_id = {}
-    capacities = []
 
     def add_edge(src, dst, capacity):
-        edges.append((name_to_id.setdefault(src, len(name_to_id)), name_to_id.setdefault(dst, len(name_to_id))))
-        capacities.append(capacity)
+        edges.append((src, dst, {"capacity": capacity}))
 
     utils.check(
         len(required_consumer_vars) > 0,
@@ -369,23 +366,19 @@ def find_cut(
         for var in symbol.flat_proxy_outs:
             add_edges(var)
 
-    g = Graph(
-        n=len(name_to_id),
-        edges=edges,
-        directed=True,
-        edge_attrs={"capacity": capacities},
-    )
-    source = name_to_id["source"]
-    sink = name_to_id["sink"]
+    g = nx.DiGraph()
+    g.add_edges_from(edges)
 
-    id_to_name = dict(map(reversed, name_to_id.items()))
+    _, (set_a, set_b) = nx.minimum_cut(g, "source", "sink")
 
-    g_edges = g.get_edgelist()
-    cut = g.mincut(source, sink, "capacity").cut
+    cut_edges = []
+    for u in set_a:
+        for v in g.successors(u):
+            if v in set_b:
+                cut_edges.append((u, v))
+
     cut_nodes = set()
-    for cut_edge_id in cut:
-        u, v = g_edges[cut_edge_id]
-        node_in, node_out = id_to_name[u], id_to_name[v]
+    for node_in, node_out in cut_edges:
         if node_out == "sink":
             continue
         assert node_in.endswith("_in"), node_in
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -2,7 +2,7 @@ torch >=2.3.0
 looseversion ==1.3.0
 lightning-utilities >=0.7.0
 numpy >=1.23.0,<2  # not yet ready for numpy 2
-igraph >=0.10.4
+networkx >= 3.3
 optree >=0.12.1
 opt_einsum >= 3.3.0
 mpmath <1.4.0  # todo: teporarl pin for `NameError: name '_C' is not defined`
--- a/thunder/__init__.py
+++ b/thunder/__init__.py
@@ -310,6 +310,8 @@ def jit(
 
     # Resolve names of executors
     executors = resolve_executors(executors)
+    ad_hoc_executor = extend.AdHocExecutor()
+    executors = (*executors, ad_hoc_executor)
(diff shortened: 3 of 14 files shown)


t-vi (approved):
Stamped. Thank you @lantiga

thunder/core/rematerialization.py
```diff
@@ -369,23 +366,19 @@ def add_edges(var):
...
+    g.add_edges_from(edges)
+
+    _, (set_a, set_b) = nx.minimum_cut(g, "source", "sink")
 
-    id_to_name = dict(map(reversed, name_to_id.items()))
+    cut_edges = []
+    for u in set_a:
+        for v in g.successors(u):
+            if v in set_b:
+                cut_edges.append((u, v))
```
nikitaved, line 378: Fine by now, but `minimum_cut` is a simple composition of other functions that compute the cut as an intermediary: https://networkx.org/documentation/stable/_modules/networkx/algorithms/flow/maxflow.html#minimum_cut.

IvanYashchuk:
Thank you, Riccardo, for pinging me here. I don't think this straightforward change requires my code review.

I'd only advise using the example from the [documentation](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.flow.minimum_cut.html) on constructing the cutset (named `cut_edges` here).

> ‘partition’ here is a tuple with the two sets of nodes that define the minimum cut. You can compute the cut set of edges that induce the minimum cut as follows:
> ```py
> cutset = set()
> for u, nbrs in ((n, G[n]) for n in reachable):
>     cutset.update((u, v) for v in nbrs if v in non_reachable)
> print(sorted(cutset))
> [('c', 'y'), ('x', 'b')]
> ```

thunder/core/rematerialization.py
lantiga: oh nice

lantiga: turns out `cutset` in there is not the droid we're looking for :)
that is the set of saturated edges that need to be removed from the float graph to obtain the residual flow graph, but we still need to look for unreachable nodes in order to get the actual mincut

nikitaved:
A faster way is to use the implementation of the minimum_cut. They call a public function that finds the optimal flow, and then find the cutset through filtering saturated edges. We can stop there. There is no need in finding the partition of the edges to later use it again to find the cutest from the previous step.

t-vi:
This is related to this PR, the other failures should go away by the next rebuild (even without merging main).
```
FAILED thunder/tests/test_nvfuser.py::test_cse_subsymbol_removal_nvfuser_cuda_None - assert 2 == 5
 +  where 2 = len((t0 = prims.gt(x, 0.0)  # t0: "cuda:0 b8[5, 5]", t1 = prims.where(t0, x, 0.0)  # t1: "cuda:0 f16[5, 5]"))
 +    where (t0 = prims.gt(x, 0.0)  # t0: "cuda:0 b8[5, 5]", t1 = prims.where(t0, x, 0.0)  # t1: "cuda:0 f16[5, 5]") = [t1] = nvFusion0(x)\n  # t0 = prims.gt(x, 0.0)  # t0: "cuda:0 b8[5, 5]"\n  # t1 = prims.where(t0, x, 0.0)  # t1: "cuda:0 f16[5, 5]".subsymbols
```

thunder/core/rematerialization.py
nikitaved: I am not sure I understand. The flow function is defined on the edges, not on the vertices. So we should be able to extract what we need from this set, modulo removing irrelevant edges from the sink and to the source.

lantiga:
Interesting, this is what I get in the failing test

```
def func(x):
    t0 = x.relu()
    t1 = t0 + 5
    t2 = t0 + 5
    t3 = t0 @ t0
    t4 = torch.where(t3 > t1, t1, t2)
    return t4
```

```
# networkx
def computation(x):
  # x: "cuda:0 f16[5, 5]"
  [t1] = nvFusion0(x)
    # t0 = prims.gt(x, 0.0)  # t0: "cuda:0 b8[5, 5]"
    # t1 = prims.where(t0, x, 0.0)  # t1: "cuda:0 f16[5, 5]"
  t8 = torch.matmul(t1, t1)  # t8: "cuda:0 f16[5, 5]"
    # t8 = ltorch.matmul(t1, t1)  # t8: "cuda:0 f16[5, 5]"
      # t8 = prims.matmul(t1, t1)  # t8: "cuda:0 f16[5, 5]"
  [t10] = nvFusion1(t1, t8)
    # t2 = prims.convert_element_type(t1, dtypes.float32)  # t2: "cuda:0 f32[5, 5]"
    # t3 = prims.add(t2, 5.0)  # t3: "cuda:0 f32[5, 5]"
    # t4 = prims.convert_element_type(t3, dtypes.float16)  # t4: "cuda:0 f16[5, 5]"
    # t9 = prims.gt(t8, t4)  # t9: "cuda:0 b8[5, 5]"
    # t10 = prims.where(t9, t4, t4)  # t10: "cuda:0 f16[5, 5]"
  del t1, t8
  return t10
```

```
# igraph (main)
def computation(x):
  # x: "cuda:0 f16[5, 5]"
  [t1, t4] = nvFusion0(x)
    # t0 = prims.gt(x, 0.0)  # t0: "cuda:0 b8[5, 5]"
    # t1 = prims.where(t0, x, 0.0)  # t1: "cuda:0 f16[5, 5]"
    # t2 = prims.convert_element_type(t1, dtypes.float32)  # t2: "cuda:0 f32[5, 5]"
    # t3 = prims.add(t2, 5.0)  # t3: "cuda:0 f32[5, 5]"
    # t4 = prims.convert_element_type(t3, dtypes.float16)  # t4: "cuda:0 f16[5, 5]"
  t8 = torch.matmul(t1, t1)  # t8: "cuda:0 f16[5, 5]"
    # t8 = ltorch.matmul(t1, t1)  # t8: "cuda:0 f16[5, 5]"
      # t8 = prims.matmul(t1, t1)  # t8: "cuda:0 f16[5, 5]"
  del t1
  [t10] = nvFusion1(t8, t4)
    # t9 = prims.gt(t8, t4)  # t9: "cuda:0 b8[5, 5]"
    # t10 = prims.where(t9, t4, t4)  # t10: "cuda:0 f16[5, 5]"
  del t8, t4
  return t10
```

The two computations are equivalent, but with igraph we get to fuse more ops before the matmul, while with networkx the other way around.

t-vi:
Both solutions look valid to me, maybe we should check that
- there are two fusions,
- that the total length of the fusion subsymbols is 7
- that only matmul + del + trivial bind are non-fusion symbols at the top level.
WDYT?

lantiga:
makes sense @t-vi 

@IvanYashchuk mentioned that maybe just selecting the same underlying algorithm in networkx as the one used in igraph will bring back the same partition, but in fact both graphs are valid and we may want to improve the test so that it's not too hung up on the details

nikitaved:
We construct a graph with many edges having the same capacity, so it is not common to have a subgraph which is a path, where we can cut along any edge to the same effect.

thunder/core/rematerialization.py
nikitaved: yeah, since we have many edges with the  same capacity, we can saturate subgraphs which are paths, and there any single edge can be included in the cut. We do not need all of them, just one will do, might as well be the one closed to the source or the sink. Agree!

lantiga:
> We construct a graph with many edges having the same capacity, so it is not uncommon to have a subgraph which is a path, where we can cut along any edge to the same effect. The solution which we choose here with the new routine is biased to pick edges closest to the source. I wonder what igraph is doing here? Maybe it biases towards edges closest to the sink?

Yeah I believe so, and they're equally good. I'll get back to this later today and adapt the tests. But overall it's quite interesting to get a definite answer, it's good learning.

PR #1249, opened 2024-10-02
Add ThunderFX as a benchmark executor

riccardofelluga:
<details>

</details>

## What does this PR do?

This PR adds ThunderFX in the list of executors in the benchmark suite.


thunder/benchmarks/targets.py
```diff
@@ -155,11 +156,13 @@ def interpreter_fwd(module: Callable):
     torch_executor,
     torch_compile_executor,
     thunder_executor,
+    thunderfx_executor,
```
IvanYashchuk, line 159: What do we get by using this executor for all benchmarks in this file? All existing benchmarks work fine with plain Thunder.

riccardofelluga: My idea is to add more information to the benchmarks, that then if not needed can be filtered out by using the `-k` filters. In particular this is useful for the torchbench benchmarks, but I thought that it is also useful for the other existing benchmarks don't you think so?

riccardofelluga: Updated and splitted the executors to save on time on the execution of the benchmarks.

t-vi (approved):

thunder/benchmarks/targets.py
```diff
@@ -151,17 +152,19 @@ def interpreter_fwd(module: Callable):
     return fn_
 
 
-executors = (
-    torch_executor,
-    torch_compile_executor,
-    thunder_executor,
-)
+executors = (torch_executor, torch_compile_executor, thunder_executor)
```
IvanYashchuk, line 155: What formatter applied this change?

riccardofelluga: @IvanYashchuk ruff configured by this https://github.com/Lightning-AI/lightning-thunder/blob/40904c762f6b1c2eb1a0c3a75a7b4b1382108b84/pyproject.toml#L84 . Let me know if in your case this doesn't match expectation.

Issue #1257, opened 2024-10-03
symbolic values: error in infer_tensor_properties, through reshape or numel


t-vi:
Running (with lates litgpt to get Llama 3.2, but you can also use llama2-like)
```python
with torch.device('cuda'):
    m = litgpt.GPT.from_name('Llama-3.2-1B').bfloat16().requires_grad_(False)
    m.set_kv_cache(1)
inp1 = torch.ones(1, 16, device="cuda", dtype=torch.int32)
inp_pos1 = torch.arange(16, device="cuda")
jm = thunder.jit(m, cache=thunder.core.options.CACHE_OPTIONS.SYMBOLIC_VALUES)
jm(inp1 inp_pos1)
```

gives

```
File ~/data/firma/grid/thunder/lightning-thunder/thunder/torch/__init__.py:4578, in embedding(a, weight, padding_idx, max_norm, norm_type, scale_grad_by_freq, sparse)
   4575     return clang.take(weight, a, 0)
   4577 output_shape = list(a.shape) + list(weight.shape[1:])
-> 4578 flatten_indices = reshape(a, [a.numel()])
   4579 flatten_output = clang.take(weight, flatten_indices, 0)
   4580 return reshape(flatten_output, output_shape)

TypeError: _infer_tensor_properties.<locals>.<lambda>() missing 1 required positional argument: 'tp'
```

@jjsjann123

jjsjann123:
mark here that the old assert is gone with #1260 

I'm hitting a new assert here https://github.com/Lightning-AI/lightning-thunder/blob/994377895e96b221cb361a636a616d2b1ccaef80/thunder/core/jit_ext.py#L1376
 I think it's coming from using `numel` in code logic.

One way to resolve is to add static constraints, which we should do. Meanwhile, I'm unsure why we needed that numel call in the first place, I'll try to remove it in #1451

jjsjann123:
With #1260 #1450 #1451 

This script seems to be *functional*, but we don't have a cache hit during second iteration, which is a bit odd. Looks like there's some input with dtype change during second iteration. But I did confirm that the same behavior happens even for static input shapes. So that's just something the model does I guess?!

But cache hit did happen afterwards. :partying_face:

jjsjann123:
```
import torch

configs = [
    dict(
        name="llama2-like",
        vocab_size=320,
        padding_multiple=64,
        n_layer=2,
        n_head=4,
        n_embd=64,
[...]

Issue #1253, opened 2024-10-03
TensorProxy.shape should be unpacked automatically


jjsjann123:
## 🐛 Bug

`TensorProxy.shape` remains as an attribute, hence accessing it won't leave an unpack in trace. This causes issues when we have `NumberProxy` in `TensorProxy.shape`.

In #1201 commit 26f883e33f2231c0f04fa5443a5790e892e93c78. I have to rely on [this hack](https://github.com/Lightning-AI/lightning-thunder/blob/26f883e33f2231c0f04fa5443a5790e892e93c78/thunder/clang/__init__.py#L978-L979). Otherwise, grad transform would see an invalid trace,

e.g. in a trivial slice:
```
def foo(a):
  return a[..., : a.shape[-1]]

thunder.jit(foo, cache="symbolic values")
```

```
def computation(a, i1):
  # a: "cpu f32 [IntegerProxy name=i0], [IntegerProxy name=i1]
  # i1: "int 8"
  a = ltorch.getitem(a, (..., slice(None, i1, None))
      # (i0, i1) = prims.shape(a)  # THIS IS WHAT THE HACK DOES
      # b2 = prims.lt(i0, 0)
      # ...
```

Without the explicit unpack of a.shape, the subsymbols in ltorch.getitem would access `i0`, which is implicitly carried by a.shape but not explicitly in the trace.

### Alternative

This problem can also be properly resolved in prologue trace. i.e. here `i1` is unpacked in prologue, because it is consumed by the top level symbol `ltorch.getitem`. Unfortunately the uses of subsymbol is not considered as consumed by computation trace today, see [code](https://github.com/Lightning-AI/lightning-thunder/blob/d6455982d5ca6815efd3d7dc0341b4f945f99be5/thunder/core/jit_ext.py#L1691), so `i0` isn't getting unpacked in prologue yet.

So for input TensorProxy, I think prologue unpacking is the right choice here. For intermediate tensor, it might be a mixed solution, which goes back to the conversation we have in #1133 .

t-vi:
[...]

Issue #1133, opened 2024-09-10
Modeling of shape queries


jjsjann123:
## 🚀 Feature

Modeling of accessing shape attribute of Tensor/TensorProxy has raising up to discussion in separate PRs & discussions.

We are trying to debate on whether we would want to lift shape inference logic into prologue trace in Thunder in general.

i.e. For a program

```
   def foo(...):
      # assume we computed a tensor `t0`
      t1 = t0.reshape(t0.size(0), -1)
```

If we leave all shape logic in the computation, it should be simplified as below:

```
    # produce t0 from some earlier trace
    (i0, i1) = prims.shape(t0)

    i2 = prims.mul(1, i0)
    i3 = prims.mul(i2, i1)  # this is t0.numel

    i4 = prims.mul(1, i0)
    i5 = prims.div(i3, i4)  # this is the simplified logic in clang.reshape with `-1` in the entry
    t1 = prims.reshape(t0, (i0, i5))
```

One alternative is to lift all shape logic in the prologue trace, so we'll have
```
@prologue trace:
foo(...):
    # all the shape logic to compute i0 & i1 from original input

    i2 = prims.mul(1, i0)
    i3 = prims.mul(i2, i1)  # this is t0.numel

    i4 = prims.mul(1, i0)
    i5 = prims.div(i3, i4)  # this is the simplified logic in clang.reshape with `-1` in the entry  
    return (..., i0, i5, ...)  # NOTE: we are not necessarily passing i0 / i5, it could be any equivalent symbols.

@compute trace:
foo(..., i0, i5, ...):
    # compute t0
    t1 = prims.reshape(t0, (i0, i5))  # here we do not seeing that i0 is equivalent to t0.shape[0]
```

jjsjann123:
Logging offline suggestions by @mruberry:

* It's OK to have prims.shape appear in the trace, which allows overlapping cpu and gpu computation, and if it's in the prologue that won't happen;

I think this justifies #1113 . cc'ing @t-vi 

* It'd be nice if the appearance of shape queries happened once per tensor it was needed for, at the global scope, and in a place that didn't disrupt fusions or require fusion executors like nvFuser to reason about the presence of the shape calls

I think this just requires a decent DCE to aggregate the shape query. Shape query on python is not trivial and takes ~ us.
Meanwhile, I'm uncertain if we would want to hide `shape` calls from fusion executors. We need to evaluate its impact on generated kernels as well as overhead of cache.

[...]

Issue #925, opened 2024-08-05
Make `numel` a method of TensorProxy not a property attribute


IvanYashchuk:
## 🐛 Bug

PyTorch uses `tensor.numel()` while Thunder's TensorProxy and in the codebase it's used as a property `tensor.numel`. The property attribute makes more sense, but it conflicts with setting PyTorch as the default language context (https://github.com/Lightning-AI/lightning-thunder/pull/924).

Related PR: https://github.com/Lightning-AI/lightning-thunder/pull/278

We should decide what to do with `numel` and how to avoid the discrepancy with PyTorch.

PR #1260, opened 2024-10-04
Patching tensor proxy shape in trace

jjsjann123:
Working on #1253

1. Updating deferred `_numel`'s deferred computation, so we can handle call without argument. (This is an existing technical debt, see #925 for more context).
This change fixes the error reported in #1257, but let's keep that issue open since there likely will be other more errors coming from that repro script.

2. Updating `TensorProxy.shape`, so shape query within trace context is recorded with `prims.shape`.
This doesn't contradict #1133. The explicit trace ensures correctness of the trace. I'll follow up with transformation to optimize shape queries.
There are occasions where we don't want to leave a `prims.shape` in the log while query shape of a TensorProxy. e.g. [printing a trace](https://github.com/Lightning-AI/lightning-thunder/blob/b4ca020b6e194ac9cd6ce47a28c04df963110a83/thunder/core/trace.py#L359). I'm working around those cases via explicitly accessing `TensorProxy._shape`.

3. add `prims.shape` support in python executor, since `prims.shape` could show up in prologue trace, where torch executor isn't available.

4. Updating dce pass to choose the first producer instead of the last one, when multiple producer exists for a single proxy.

TODO:
Need to update DCE pass to remove duplicated producer in a given scope. I intend to handle that in a separate PR. See [comment](https://github.com/Lightning-AI/lightning-thunder/issues/1253#issuecomment-2400638135) for details.

### DIFF
--- a/thunder/core/proxies.py
+++ b/thunder/core/proxies.py
@@ -1464,6 +1464,7 @@ class TensorProxy(Proxy, TensorProxyInterface):
             return self._shape
         else:
             from thunder.core.prims import shape
+
             return shape(self)
 
     @property
--- a/thunder/core/proxies.py
+++ b/thunder/core/proxies.py
@@ -16,7 +16,12 @@ import torch
 
 from thunder.core.compile_data import using_symbolic_values, using_jit
 from thunder.core.interpreter import is_jitting, ProvenanceRecord, PseudoInst
-from thunder.core.trace import VariableInterface, get_tracectx, TraceCtx
+from thunder.core.trace import (
+    VariableInterface,
+    get_tracectx,
+    is_tracing,
+    TraceCtx,
+)
 from thunder.core.baseutils import (
     ProxyInterface,
     NumberProxyInterface,
@@ -1459,8 +1464,7 @@ class TensorProxy(Proxy, TensorProxyInterface):
     #   outside of a trace or language context
     @property
     def shape(self):
-        trace: None | TraceCtx = get_tracectx()
-        if trace is None:
+        if not using_symbolic_values() or not is_tracing():
             return self._shape
         else:
             from thunder.core.prims import shape
--- a/thunder/core/proxies.py
+++ b/thunder/core/proxies.py
@@ -1247,8 +1247,7 @@ def _infer_tensor_properties(
         thunder_fsdp_padding_size if thunder_fsdp_padding_size is not None else _thunder_fsdp_padding_size
     )
 
-    # dynamic shape not yet enabled, otherwise, the bake in should be guarded with if not using_symbolic_values():
-    # dynamic shape support is currently block by #471 https://github.com/Lightning-AI/lightning-thunder/issues/471
+    baseutils.check(_shape is not None, lambda: f"_shape cannot be None when creating TensorProxy")
     if not using_symbolic_values():
         _shape = tuple(pyval(x) for x in _shape)
         # Computes derived properties
--- a/thunder/core/proxies.py
+++ b/thunder/core/proxies.py
@@ -1251,7 +1251,7 @@ def _infer_tensor_properties(
     else:
         # deferred computation of numel
         # TODO: similar to how `shape` is handled, this should be CSE or lifted for efficiency
-        _numel = lambda tp: reduce(operator.mul, tp.shape, 1)
+        _numel = lambda *args: reduce(operator.mul, _shape, 1)
 
     # TODO Alias rank to ndim?
     _ndim = len(_shape)
@@ -1459,7 +1459,12 @@ class TensorProxy(Proxy, TensorProxyInterface):
     #   outside of a trace or language context
     @property
     def shape(self):
-        return self._shape
+        trace: None | TraceCtx = get_tracectx()
+        if trace is None:
+            return self._shape
+        else:
+            from thunder.core.prims import shape
+            return shape(self)
 
     @property
     def ndim(self):
--- a/thunder/core/proxies.py
+++ b/thunder/core/proxies.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import copy
 from enum import auto, Enum
 from numbers import Number
 from typing import Type, Optional, Any, Tuple, List, Union
@@ -1962,7 +1963,7 @@ def tensorproxy(t: torch.Tensor, /, *, name: None | str, history: None | tuple =
     distparallel_type = getattr(t, "distparallel_type", None)
     _thunder_fsdp_padding_size = getattr(t, "_thunder_fsdp_padding_size", None)
     if using_symbolic_values():
-        shape_attr = ProvenanceRecord(PseudoInst.LOAD_ATTR, inputs=[history, wrap_const("shape").provenance])
+        shape_attr = ProvenanceRecord(PseudoInst.LOAD_ATTR, inputs=[copy.copy(history), wrap_const("shape").provenance])
         shape = tuple(
             IntegerProxy(
                 None,
--- a/thunder/core/proxies.py
+++ b/thunder/core/proxies.py
@@ -1554,10 +1554,10 @@ class TensorProxy(Proxy, TensorProxyInterface):
         )
 
     def __repr__(self):
-        return f'<{type(self).__name__}(name="{self.name}", dtype={self.dtype}, shape={self.shape})>'
+        return f'<{type(self).__name__}(name="{self.name}", dtype={self.dtype}, shape={self._shape})>'
 
     def type_string(self):
-        return f"{self.device.device_str()} {self.dtype.shortname()}{list(self.shape)}"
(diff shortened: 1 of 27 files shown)


jjsjann123:
Note for myself: failing cases looks real, since I don't see them in other PRs. But strangely I'm not seeing the repro on my local non-cuda environment.... 😕 , even though the failing test does run on CPU as well.
e.g. thunder/tests/test_jit_general.py::test_cache_symbolic_values_reshape[cpu]

Let me try to grab a node with an actual GPU on it.

jjsjann123:
errr. I don't know what's going on with that failing test. but looks like CI on main is green, so I must have done something else here...

jjsjann123:
Still don't see what's wrong with this failing test and I can't repro it locally: https://dev.azure.com/Lightning-AI/lightning/_build/results?buildId=216738&view=logs&j=2840892e-91ab-5245-da62-77ec9923516a&t=444f4171-6797-5730-4229-41427ed3bdc9&l=15058

checking my luck again with CI.

jjsjann123:
🤯 Nope. CI failed.

jjsjann123:
kkk. Looks like indeed a real issue here.

```
t_0 = tensor([[ 8,  0,  8,  6],
        [-2,  4,  6, -4],
        [-5, -8,  8,  9],
        [-1, -5,  5, -8]], device='cuda:0', dtype=torch.int8)
t_1 = tensor(-1, dtype=torch.int8)

torch.pow(t_0, t_1)
```

Strangely this doesn't repro locally with that test. That's a bit concerning.

jjsjann123:
Now I know where the failure is coming from, I'll try to get a repro on this one and verify if it's caused in this PR (likely maybe?!) But is it really an issue introduced by this PR?!

Though I'm not sure why others didn't hit this one.

jjsjann123:
Note for myself:

FAILED thunder/tests/test_ops.py::test_core_vs_torch_consistency_pow_torch_cuda_thunder.dtypes.int8 - RuntimeError: "reciprocal_cuda" not implemented for 'Char' 
Something very strange here. In https://github.com/Lightning-AI/lightning-thunder/pull/1260#issuecomment-2397690721

Opinfo test for pow runs through some random inputs. specifically I think this one https://github.com/Lightning-AI/lightning-thunder/blob/fceb64efc93a80a27d38b8e84f0e2b5f132f3d2f/thunder/tests/opinfos.py#L1841-L1843
    # Tests the inputs are a CPU scalar tensor and a CUDA tensor
    a = make_tensor((4, 4), device=device, dtype=dtype, requires_grad=requires_grad, **kwargs)
    b = make_tensor((), device="cpu", dtype=dtype, requires_grad=requires_grad, **kwargs)
So for int8 type, pytorch implementation cannot handle where b == torch.tensor(-1) -> https://github.com/pytorch/pytorch/blob/fe44b6a67f32b562c88701b630e65b62ce1b63ba/aten/src/ATen/native/cuda/PowKernel.cu#L178

Somehow this is reliably failing on CI for me in my PR, but not on my local runs.

Anyway, I'm trying to avoid this by `exclude_zero=True` 🤞

jjsjann123:
@t-vi This PR is ready for review now. 🙇 

cc'ing @tfogal @kevinstephano

thunder/core/proxies.py
```diff
@@ -1251,7 +1256,7 @@ def _infer_tensor_properties(
     else:
         # deferred computation of numel
         # TODO: similar to how `shape` is handled, this should be CSE or lifted for efficiency
-        _numel = lambda tp: reduce(operator.mul, tp.shape, 1)
+        _numel = lambda *args: reduce(operator.mul, _shape, 1)
```
tfogal, line 1259: AFAICT `_shape` can be `None` here. Do we need to guard against that?

thunder/core/proxies.py
```diff
@@ -1251,7 +1256,7 @@ def _infer_tensor_properties(
...
 
     # TODO Alias rank to ndim?
     _ndim = len(_shape)
```
tfogal, line 1262: hrm, unrelated to this commit, but to rehash the previous point: when `_shape` is `None` this is going to break.

thunder/tests/test_jit_general.py
```diff
@@ -1439,3 +1439,47 @@ def foo(a):
...
+    if device == "cuda" and not torch.cuda.is_available():
+        pytest.skip("CUDA not available")
+
+    def foo(a):
+        a = torch.reshape(a, [a.numel()])
+        return a.relu()
+
+    jfoo = thunder.jit(foo, cache="symbolic values")
+
+    a = torch.randn(2, 3, 8, requires_grad=True, device=device)
```
tfogal, line 1458: nit: is the `requires_grad` critical here? seems like we'd still need to compute the shape even w/o grad, right?

thunder/tests/test_jit_general.py
```diff
@@ -1116,7 +1116,7 @@ def forward(self, x):
     ("cpu", "cuda"),
 )
 def test_cache_symbolic_values_reshape(device):
-    if not torch.cuda.is_available():
+    if device == "cuda" and not torch.cuda.is_available():
```
tfogal, line 1119: do we get anything out of running this test on multiple devices?

I am wondering if it makes more sense to just not parameterize the test and run it once on the CPU.

thunder/tests/test_jit_general.py
```diff
@@ -1439,3 +1439,47 @@ def foo(a):
...
+    "device",
+    ("cpu", "cuda"),
+)
+def test_cache_symbolic_values_slice(device):
+    if device == "cuda" and not torch.cuda.is_available():
+        pytest.skip("CUDA not available")
+
+    def foo(a):
+        a = a[..., : a.shape[-1]]
+        return a.relu()
```
tfogal, line 1476: nit: do we even need the `relu`? would returning the new `a`, or `a.shape` be enough to tickle this code path?

tfogal (approved):
I'd recommend that TomV (or Mike? both?) take a look at this before it goes in, but my naive view is that this seems good.

thunder/core/proxies.py
jjsjann123: Good call. Let me throw a meaningful error message for that.

thunder/tests/test_jit_general.py
jjsjann123: That's a good call!
At this point since nvfuser isn't taking shape operations at all, GPU test doesn't do anything. Let me clean it up.

thunder/tests/test_jit_general.py
jjsjann123: Yes you are right.

This is just for test coverage, with `requires_grad`, we'll run grad transform on top of the original trace, which is where things tend to break. 😢

thunder/tests/test_jit_general.py
jjsjann123: The shape unpacking inside `relu` is what triggered the small code change in `producers` (called inside DCE during autograd).

I would like to keep this test as my own guard rail, so I won't accidentally break things in follow up PRs. Should I put a comment to remove them at a later point if you prefer to have leaner tests?

thunder/tests/opinfos.py
```diff
@@ -2246,7 +2246,11 @@ def pow_sample_input_generator(op, device, dtype, requires_grad, *, no_rhs_numbe
         ),
         # NOTE: PyTorch fails with RuntimeError: "reciprocal_cuda" not implemented for 'Long' occasionally when the exponent is CPU scalar tensor
         # e.g.: x=torch.tensor([[ 6,  5,  1, -8],], device='cuda:0');y=torch.tensor(-1);torch.pow(x,y)
-        DecorateInfo(pytest.mark.xfail, "test_core_vs_torch_consistency", dtypes=(datatypes.int32, datatypes.int64)),
+        DecorateInfo(
+            pytest.mark.xfail,
+            "test_core_vs_torch_consistency",
+            dtypes=(datatypes.int8, datatypes.int16, datatypes.int32, datatypes.int64),
```
jjsjann123, line 2252: Want to bring this up to @t-vi 's attention for review. I believe this is the right fix, though I'm unsure how we are getting away from this in the existing code.

thunder/tests/test_jit_general.py
tfogal: nah it's fine, i just didn't realize the necessity. thanks for explaining

thunder/tests/test_jit_general.py
tfogal: ack, okay

thunder/tests/opinfos.py
t-vi: Don't know why this is happening exactly, but let's have it for the other int dtypes as well.

thunder/core/proxies.py
t-vi: Seems to be there now.

PR #1261, opened 2024-10-04
A converter for FXGraph with Torch calls -> FXGraph with Thunder calls

kiya00:
Dynamo represents the `function` argument of `torch.utils.checkpoint.checkpoint(function, args...)`  as FX Graph, this FX graph has PyTorch operators in it. This PR creates a converter to replace the torch operators with the thunder equivalents, so it can be traced in the thunder `checkpoint` symbol. With this change native PyTorch activation checkpointing can be supported.


thunder/dynamo/utils.py
```diff
@@ -163,6 +163,10 @@ def try_execute_thunder_symbol(thunder_symbol: "Symbol", node: torch.fx.Node) ->
     import thunder
     from thunder.core.trace import TraceCtx
 
+    # `tag_activation_checkpoint` has function input that get_proxy_inputs_from_node can not handle
```
IvanYashchuk, line 166: We need to support `tag_activation_checkpoint`. What's missing to through get_proxy_inputs_from_node?

thunder/tests/test_grad.py
```diff
@@ -1700,6 +1700,82 @@ def func(a, b):
         get_saved_for_backward_tensors(execution_trace)
 
 
+def test_torch_checkpoint():
+    import torch.utils.checkpoint
+    import torch._higher_order_ops.wrap
+
+    def fn_to_checkpoint(x):
+        return x.sin().cos().exp()
```
IvanYashchuk, line 1708: When this PR is ready the test should be working with this
```suggestion
        return torch.exp(torch.cos(torch.sin(x)))
```

thunder/torch/__init__.py
```diff
@@ -5190,6 +5192,96 @@ def _unwrap_if_dead(tensor):
...
+    if determinism_check is not None:
+        warnings.warn("torch.checkpoint: determinism_check is not supported in Thunder and will be ignored")
+    if preserve_rng_state is not None:
+        warnings.warn("torch.checkpoint: preserve_rng_state is not supported in Thunder and will be ignored")
+    if isinstance(function, torch.fx.GraphModule):
+        _FXGraph_converter(function)
+    return function(*args, **kwargs)
+
+
+def _FXGraph_converter(g):
```
IvanYashchuk, line 5229: I know it's still a work in progress. Could you please put this conversion inside `thunder.dynamo.ThunderCompiler` before the FX graph is passed to `thunder.jit`.
We need to isolate Thunder-only and Dynamo-only changes from each other.

thunder/dynamo/utils.py
kiya00: This is a short cut for now to let the splitter not split for `tag_activation_checkpoint(func, ...)` op (always return True and the splitReason is None), we need to extend this to return the result that whether the `func` is supported by Thunder or not

thunder/dynamo/splitter.py
```diff
@@ -140,20 +144,40 @@ def is_thunder_supported_partition(node: torch.fx.Node) -> bool:
     # Call compile on the split region/s.
     thunder_compiled_fns = []
     submodule_to_compiled_fns = {}
+
+    # If the split_gm contains Thunder-traceable checkpoint operators, the checkpointed submodules are replaced, so we need to record the original split_gm
+    origin_split_gm = split_gm
+    if has_thunder_tracable_checkpoint:
+        import copy
+
+        split_gm = copy.deepcopy(split_gm)
```
kiya00, line 153: How it works with splitter:
when checkpoint operator is contained in `submod_1`, the `split_gm` is represented as follows:
```
GraphModule(
  (wrap_body_0): GraphModule() # the module checkpoint calls
  (submod_1): GraphModule() # the module contains the checkpoint operator
)
```
The converter will keep the original split gm as `original_split_gm` and deepcopy the `split_gm`, then replace the torch operaters in the submodule `wrap_body_0` of `split_gm`, after that the `split_gm` is passed to the futher steps of splitter, like replacing the `submod_1` with jitted `thunder_1`. At last the `original_split_gm` is stored in the `SubgraphInfo` and the key of `submodule_to_compiled_fns` is the original `submod_1`, so that these information can be used by other API, like benchmarking (the graph by graph benchmark needs the original GraphModule that before any changes are made by the converter)
So in the end:
split_gm in `SubgraphInfo`:
```
GraphModule(
  (wrap_body_0): GraphModule() # the module updated by converter(has thunder symbols)
  (thunder_1): GraphModule() # the module updated by spliter
)
```
original_split_gm in `SubgraphInfo`:
```
GraphModule(
  (wrap_body_0): GraphModule() # the module checkpoint calls(has torch function)
  (submod_1): GraphModule() # not jitted
)
```
If no checkpoint op exists, no deepcopy happens, the original_split_gm and split_gm are identical

kshitij12345: Thanks for the extra context, I think it would be helpful to have a summarized comment in code

thunder/dynamo/splitter.py
```diff
@@ -140,20 +144,40 @@ def is_thunder_supported_partition(node: torch.fx.Node) -> bool:
...
+            # Replace the torch operators within the function called by activation checkpoint with the corresponding Thunder symbols
+            checkpoint_converter(split_gm, graph_module)
             jit_fn = thunder_jit(graph_module)
             # Update the node name from "submod_*" to "thunder_*" for more user-friendly names
             update_node_and_submodule(split_gm, node, node.name.replace("submod", "thunder"), jit_fn)
             thunder_compiled_fns.append(jit_fn)
-            submodule_to_compiled_fns[graph_module] = CompiledFunction(jit_fn, CompilerType.THUNDER)
+            if has_thunder_tracable_checkpoint:
+                submodule_to_compiled_fns[getattr(origin_split_gm, node.name.replace("thunder", "submod"))] = (
+                    CompiledFunction(jit_fn, CompilerType.THUNDER)
```
kshitij12345, line 166: Here the assumption is user will try to index with graph from `original_split_gm`. I assume it will fail with KeyError if graph came from `split_graph_module`.

I was thinking adding `original_split_gm` to `SubgraphInfo` could be confusing as to whether we should look at `split_graph_module` or `original_split_gm`. So maybe we should keep only one and add a boolean flag to denote if checkpointing occured. Wdyt?

thunder/dynamo/utils.py
```diff
@@ -304,6 +331,14 @@ def is_node_supported_by_thunder(node: torch.fx.Node) -> tuple[bool, SplitReason
         )
         return False, split_reason
 
+    # If the operation is higher order function, check whether the submodule is supported by Thunder
```
kshitij12345, line 334: I think we should rewrite `If the operation is higher order function` -> `If the operation is higher order function for checkpointing`, as I am not sure this will work with other operations in higher order ops.

kshitij12345 (commented):
Overall looks good, I just have one main comment related to information being saved in `SubgraphInfo`. Curious to know your thoughts on the same. Thanks @kiya00!

thunder/dynamo/splitter.py
kshitij12345: > Do we want to save the original split graph or the split graph after compile or both? Maybe we keep them both?
(for the graph by graph benchmarking we need the original module to compile for different backend)

So IIUC, we would need the `original_split_graph_module` for benchmarking utility. In that case, do you think we should rename `original_split_graph_module` -> `split_graph_module` (this will always be present in SubgraphInfo and be output of `fx.split_module`) and current `split_graph_module` -> `compiled_split_graph_module` (this will have thunder/inductor modules patched into the `split_graph_module`). 

With this the SubgraphInfo will look a bit like

```python
original_graph_module: torch.fx.GraphModule
split_graph_module: torch.fx.GraphModule
compiled_split_graph_module: torch.fx.GraphModule
thunder_compiled_fns: list[Callable] | None
submodule_to_compiled_functions: dict[torch.fx.GraphModule, CompiledFunction]
split_reasons: list | None = None
```

I think currently, my main confusion was while debugging which one of `original_split_graph_module` and `split_graph_module` should be referred. As both maybe same or different depending on whether `checkpointing` was present. (And one has to look at either graphs to determine whether it was present or not).

kiya00: Thank you @kshitij12345 , I'll get the `split_graph_module` as the original split module and the `compiled_split_graph_module` as the compiled one (with jitted func and converter changes if there are any), then in this way the naming is consistent w/ or w/o checkpointing.

thunder/dynamo/splitter.py
```diff
@@ -91,11 +92,12 @@ def forward(self, l_x_: "f32[2]", y: "f32[2]"):
     partition_cnt = 0
     supported_partitions: set[int] = set()
     split_reasons: list[SplitReason] = []
+    has_thunder_tracable_checkpoint = False
```
IvanYashchuk, line 95:
```suggestion
    has_thunder_traceable_checkpoint = False
```

thunder/dynamo/splitter.py
```diff
@@ -140,20 +144,40 @@ def is_thunder_supported_partition(node: torch.fx.Node) -> bool:
     # Call compile on the split region/s.
     thunder_compiled_fns = []
     submodule_to_compiled_fns = {}
+
+    # If the split_gm contains Thunder-traceable checkpoint operators, the checkpointed submodules are replaced, so we need to record the original split_gm
+    origin_split_gm = split_gm
```
IvanYashchuk, line 149:
```suggestion
    original_split_gm = split_gm
```

kiya00:
With the patch on this PR, the benchmark is tested to confirm the native pytorch checkpoint can work. 
`python thunder/benchmarks/benchmark_litgpt.py --model_name longchat-7b-16k --compile thunder+dynamo --n_layers=2 --max_iters=5 --warmup_iters=3 --checkpoint_activations=True --dump_thunder_traces=True` will print out the first and last trace
<details>
<summary>patch</summary>

```
index afb9d79f..beef7623 100644
@@ -532,7 +532,7 @@ class Benchmark_litGPT:
         return model
 
     def setup_activation_checkpointing(self):
-        if "thunder" in self.compile:
+        if "thunder" in self.compile and "dynamo" not in self.compile:
             # checkpointing is an option to thunder.jit
             return
 
@@ -571,11 +571,6 @@ class Benchmark_litGPT:
 
-            jit_options = {
-                "enable_saved_for_backward_recomputation": self.checkpoint_activations,
-                "recomputation_policy": None,
-            }
-
             if "dynamo" in self.compile:
                 if self.distributed_mode == "fsdp2":
                     print("Resetting cache size for when fsdp2 and using thunder as backend torch.compile")
@@ -583,13 +578,16 @@ class Benchmark_litGPT:
 
                     dynamo_config.cache_size_limit = 64
 
-                backend = ThunderCompiler(executors=executors, **jit_options)
+                self.backend = ThunderCompiler(executors=executors)
                 # so we are using the lower level torch._dynamo.optimize function
-                model = torch._dynamo.optimize(backend=backend)(model)
+                model = torch._dynamo.optimize(backend=self.backend)(model)
             else:
+                jit_options = {
+                    "enable_saved_for_backward_recomputation": self.checkpoint_activations,
+                }
                 jit_options["fp8_shard_intermediate_activation"] = self.fp8_shard_intermediate_activation
                 model = thunder.jit(model, executors=executors, **jit_options)
 
@@ -819,9 +817,9 @@ def benchmark_main(return_metrics_as_json=False, json_path="", **kwargs) -> None
 
         print(f"Average iter time: {benchmark.perf_metrics['average_iter_time']:.2f} ms")
         print(f"Memory used: {benchmark.perf_metrics['memory_used_GB']:.02f} GB")
-        print(f"Tokens/s: {benchmark.perf_metrics['tokens_per_sec']:.02f}")
-        print(f"Tokens/s/GPU: {(benchmark.perf_metrics['tokens_per_sec']/world_size):.02f}")
-        print(f"TFLOP/s: {benchmark.perf_metrics['model_flop_per_sec'] / 1e12:.02f}")
+        # print(f"Tokens/s: {benchmark.perf_metrics['tokens_per_sec']:.02f}")
+        # print(f"Tokens/s/GPU: {(benchmark.perf_metrics['tokens_per_sec']/world_size):.02f}")
+        # print(f"TFLOP/s: {benchmark.perf_metrics['model_flop_per_sec'] / 1e12:.02f}")
 
         if benchmark.dump_memory_snapshot:
             file_name = f"{benchmark.model_name}_{benchmark.compile}_{benchmark.distributed_mode}"
@@ -843,16 +841,28 @@ def benchmark_main(return_metrics_as_json=False, json_path="", **kwargs) -> None
                 for jitted in benchmark.thunder_as_torch_compile_backend.gm_to_thunder.values():
                     fwd_traces.append(thunder.last_traces(jitted))
                     bwd_traces.append(thunder.last_backward_traces(jitted))
-            else:
+            elif "dynamo" not in benchmark.compile:
                 fwd_traces = [thunder.last_traces(benchmark.model)]
                 bwd_traces = [thunder.last_backward_traces(benchmark.model)]
 
-            for i, f_traces in enumerate(fwd_traces, start=1):
-                print(f"##########\n#{i}-th ThunderModule\n##########")
-                print(f_traces[-1])
-            for i, b_traces in enumerate(bwd_traces, start=1):
-                print(f"##########\n#{i}-th ThunderModule\n##########")
-                print(b_traces[-1])
+            if "dynamo" in benchmark.compile:
+                for gid, infos in enumerate(benchmark.backend.subgraph_infos):
+                    for subgid, thunder_fn in enumerate(infos.thunder_compiled_fns):
+                        print(f"##########\n#Graph{gid}-ThunderFn{subgid} first forward trace\n##########")
+                        print(thunder.last_traces(thunder_fn)[0])
+                        print(f"##########\n#Graph{gid}-ThunderFn{subgid} last forward trace\n##########")
+                        print(thunder.last_traces(thunder_fn)[-1])
+                        print(f"##########\n#Graph{gid}-ThunderFn{subgid} last backward trace\n##########")
+                        print(thunder.last_backward_traces(thunder_fn)[0])
+                        print(f"##########\n#Graph{gid}-ThunderFn{subgid} last backward trace\n##########")
+                        print(thunder.last_backward_traces(thunder_fn)[-1])
+            else:
+                for i, f_traces in enumerate(fwd_traces, start=1):
+                    print(f"##########\n#{i}-th ThunderModule\n##########")
+                    print(f_traces[-1])
+                for i, b_traces in enumerate(bwd_traces, start=1):
+                    print(f"##########\n#{i}-th ThunderModule\n##########")
+                    print(b_traces[-1])
 
     if global_rank in [0, None]:
         if return_metrics_as_json:
```

</details>

thunder/tests/test_dynamo.py
```diff
@@ -551,7 +552,10 @@ def forward(self, x):
     out.backward(g)
 
     ref_g = torch.ones_like(ref_out)
-    ref_out.backward(ref_g)
+    with warnings.catch_warnings():
+        # FutureWarning: `torch.cpu.amp.autocast(args...)` is deprecated. Please use `torch.amp.autocast('cpu', args...)` instead.
+        warnings.filterwarnings("ignore", category=FutureWarning)
+        ref_out.backward(ref_g)
```
crcrpar, line 558: can't we use [`@pytest.mark.filterwarnings`](https://docs.pytest.org/en/stable/how-to/capture-warnings.html#filterwarnings) instead?

thunder/dynamo/splitter.py
```diff
@@ -1,5 +1,6 @@
 from __future__ import annotations
 from typing import TYPE_CHECKING
+import copy
```
crcrpar, line 3: where is this used? I don't see any use in diff

thunder/tests/test_dynamo.py
```diff
@@ -510,3 +513,88 @@ def f(x, y):
...
+@requiresCUDA
+def test_checkpoint_converter():
+    import torch.utils.checkpoint as checkpoint
+
+    class SimpleModel(nn.Module):
+        def __init__(self):
+            super().__init__()
+            self.layer1 = nn.Linear(10, 20)
+            self.layer2 = nn.Linear(20, 20)
+            self.layer3 = nn.ReLU()
```
crcrpar, line 527:
```suggestion
```

thunder/tests/test_dynamo.py
```diff
@@ -510,3 +513,88 @@ def f(x, y):
...
+
+        def forward(self, x):
+            x = torch.sin(x)
+            x = checkpoint.checkpoint(self.layer1, x)
+            x = checkpoint.checkpoint(self.layer2, x)
+            x = self.layer3(x)
```
crcrpar, line 533:
```suggestion
            x = F.relu(x)
```

thunder/tests/test_dynamo.py
```diff
@@ -1,11 +1,14 @@
 import pytest
+import warnings
 import torch
 import torch.fx
+import torch.nn as nn
```
crcrpar, line 5:
```suggestion
import torch.nn as nn
import torch.nn.functional as F
```

thunder/tests/test_dynamo.py
crcrpar, line 1: My three suggestions are quite up to your taste.

thunder/torch/__init__.py
```diff
@@ -5199,6 +5201,71 @@ def _unwrap_if_dead(tensor):
...
+def _backward_checkpoint(
+    function,
+    args,
+    kwargs,
+    *grad_outputs,
+) -> tuple[None | TensorLike, ...]:
+    from thunder.core.transforms import vjp
+
+    result, grads = vjp(function)(args, grad_outputs, **kwargs)
+    return grads  # result
```
crcrpar, line 5266: not sure but at glance `result` looks like a history of development
```suggestion
    return grads
```

crcrpar (approved):

thunder/dynamo/splitter.py
kiya00: forgot to remove it
```suggestion
```

kiya00:
Hi @IvanYashchuk @kshitij12345 , do you want to take another look, I think it's ready to merge

IvanYashchuk:
> Hi @IvanYashchuk @kshitij12345 , do you want to take another look, I think it's ready to merge

Yes, I want to have another look and I'll do that first thing on Monday or before that.

thunder/dynamo/utils.py
```diff
@@ -306,6 +328,14 @@ def is_node_supported_by_thunder(node: torch.fx.Node) -> tuple[bool, SplitReason
         )
         return False, split_reason
 
+    # If the operation is higher order function for checkpointing, check whether the submodule is supported by Thunder
+    if target is torch.ops.higher_order.tag_activation_checkpoint:
+        m = node.graph.owning_module
+        assert hasattr(m, node.args[0].name)
+        higher_order_module = getattr(m, node.args[0].name)
```
kshitij12345, line 335: I think it would be better to rename this as `checkpointed_fn` or something similar. Wdyt?

thunder/tests/test_dynamo.py
```diff
@@ -510,3 +514,85 @@ def f(x, y):
...
+    ref_model.load_state_dict(model.state_dict())
+
+    backend = ThunderCompiler()
+    jf = torch.compile(backend=backend)(model)
+
+    ref_out = ref_model(x_ref)
+    out = jf(x)
+    torch.testing.assert_close(ref_out, out)
+
+    g = torch.ones_like(out)
```
kshitij12345, line 552: I would recommend using `torch.randn_like` or `torch.rand_like` (if we want more constrained values). Previously while working on some feature, I vaguely recall using `ones_like` in input or gradients which hid errors which occured with more random values.

kshitij12345 (approved):
LGTM, thanks @kiya00 !

thunder/dynamo/splitter.py
```diff
@@ -143,6 +144,8 @@ def is_thunder_supported_partition(node: torch.fx.Node) -> bool:
     for node in split_gm.graph.nodes:
         if is_thunder_supported_partition(node):
             graph_module = getattr(split_gm, node.name)
+            # Replace the torch operators within the function called by activation checkpoint with the corresponding Thunder symbols
```
IvanYashchuk, line 147: The original comment uses "operators" and "symbols" interchangeably. It's better to stick with one term in one sentence for clarity.
```suggestion
            # Replace PyTorch operators within the checkpointed function with the corresponding Thunder operators
```

thunder/dynamo/utils.py
```diff
@@ -5,14 +5,15 @@
...
-import warnings
+import copy
 
 import torch
 
 from thunder.torch.default_torch_ops import torch_auto_registered_ops
 from thunder.torch import _torch_to_thunder_function_map
 from thunder.torch.langctx import torchctx
 from thunder.core.utils import check
+from thunder.core.pytree import tree_map
```
IvanYashchuk, line 16: It seems unused.
```suggestion
```

thunder/dynamo/utils.py
```diff
@@ -306,6 +328,14 @@ def is_node_supported_by_thunder(node: torch.fx.Node) -> tuple[bool, SplitReason
         )
         return False, split_reason
 
+    # If the operation is higher order function for checkpointing, check whether the submodule is supported by Thunder
```
IvanYashchuk, line 331: It's better to explain why this if-statement is needed and not what it does.
```suggestion
    # The checkpointed function must be fully supported by Thunder
```

thunder/dynamo/utils.py
```diff
@@ -418,3 +448,55 @@ def _get_example_inputs_from_placeholder(node) -> tuple[torch.Tensor]:
...
             "The 'example_value' in the placeholder node is expected to be either a Tensor or a Tuple of Tensors."
         )
+
+
+def _checkpoint_function_converter(gm: torch.fx.GraphModule):
+    """
+    Replace the Torch operators in the GraphModule called by activation checkpoint operator with the corresponding Thunder symbols in place
+    Args:
+        gm: The GraphModule of the checkpointed function, which is modified in place
+    """
```
IvanYashchuk, line 458: The docstring mentions inplace modification of the GraphModule but there's a deep copy done in the first line of the function.
Why is the graph deepcopied, but not the GraphModule? Should we skip copying the graph and modify it inplace as well?

thunder/dynamo/utils.py
```diff
@@ -418,3 +448,55 @@ def _get_example_inputs_from_placeholder(node) -> tuple[torch.Tensor]:
...
+    new_graph = copy.deepcopy(gm.graph)
+    for n in new_graph.nodes:
+        # replace the torch operator in "call_function" node
+        if n.op == "call_function":
+            assert isinstance(n.target, Callable)
```
IvanYashchuk, line 463: Can there be other (non-Callable) targets for "call_function" node types?

thunder/dynamo/utils.py
```diff
@@ -418,3 +448,55 @@ def _get_example_inputs_from_placeholder(node) -> tuple[torch.Tensor]:
...
+                thunder_node = new_graph.call_function(
+                    _torch_to_thunder_function_map[n.target], args=n.args, kwargs=n.kwargs
+                )
+            n.replace_all_uses_with(thunder_node)
+            new_graph.erase_node(n)
+        else:
+            if n.op == "call_module":
+                raise RuntimeError(
+                    "Unexpected call_module detected inside a checkpoint. This should have been inlined in dynamo graphs"
+                )
```
IvanYashchuk, line 479: Are all other node types fine to have? Could you please add a test for this error?

thunder/dynamo/utils.py
```diff
@@ -418,3 +448,55 @@ def _get_example_inputs_from_placeholder(node) -> tuple[torch.Tensor]:
         raise TypeError(
             "The 'example_value' in the placeholder node is expected to be either a Tensor or a Tuple of Tensors."
         )
+
+
+def _checkpoint_function_converter(gm: torch.fx.GraphModule):
+    """
+    Replace the Torch operators in the GraphModule called by activation checkpoint operator with the corresponding Thunder symbols in place
+    Args:
+        gm: The GraphModule of the checkpointed function, which is modified in place
```
IvanYashchuk, line 457:
```suggestion
    Replace PyTorch operators in ``gm`` representing a checkpointed function with corresponding Thunder operators. The input ``gm`` is modified inplace.

    Args:
        gm (torch.fx.GraphModule): The GraphModule of the checkpointed function, which is modified inplace.
```

thunder/dynamo/utils.py
```diff
@@ -418,3 +448,55 @@ def _get_example_inputs_from_placeholder(node) -> tuple[torch.Tensor]:
...
+
+    Note:
+        The GraphModule of the checkpointed function is updated inplace
+    """
+    for n in sub_gm.graph.nodes:
+        if n.op == "call_function":
+            if n.target in (torch.ops.higher_order.tag_activation_checkpoint,):
+                name = n.args[0].name
+                assert hasattr(gm, name)
+                function_module = getattr(gm, name)
```
IvanYashchuk, line 501: `getattr` would raise an AttributeError. There's no need to do the assert.
```suggestion
                function_module = getattr(gm, name)
```

IvanYashchuk, line 501: Or use a more descriptive variable name like `checkpointed_function_name = n.args[0].name`
```suggestion
                function_module = getattr(gm, n.args[0].name)
```

thunder/dynamo/utils.py
```diff
@@ -418,3 +448,55 @@ def _get_example_inputs_from_placeholder(node) -> tuple[torch.Tensor]:
...
+    recompile_graph(gm)
+
+
+def checkpoint_converter(gm: torch.fx.GraphModule, sub_gm: torch.fx.GraphModule):
+    """
+    Utility function to convert the GraphModule that uses activation checkpointing into a Thunder-traceable GraphModule.
+
+    Args:
+        gm: The parent GraphModule containing the submodule(sub_gm), as well as the GraphModule of the checkpointed function.
+        sub_gm: the GraphModule containing the checkpoint operator
```
IvanYashchuk, line 491: Doesn't  `sub_gm` store any reference to the GraphModule for the checkpointed function? Is it only available in `gm`?

thunder/torch/__init__.py
```diff
@@ -5262,8 +5262,8 @@ def _backward_checkpoint(
 ) -> tuple[None | TensorLike, ...]:
     from thunder.core.transforms import vjp
 
-    result = vjp(function)(args, grad_outputs, **kwargs)
-    return result
+    result, grads = vjp(function)(args, grad_outputs, **kwargs)
+    return grads
```
IvanYashchuk, line 5266: Why did `test_torch_checkpoint` pass?

thunder/tests/test_dynamo.py
```diff
@@ -535,3 +539,85 @@ def f(x):
...
+        def __init__(self) -> None:
+            super().__init__()
+            self.sub_mod = SubModule()
+
+        def forward(self, x):
+            x = torch.sin(x)
+            x = checkpoint.checkpoint(self.sub_mod, x)
+            return x
+
+    x = torch.randn(5, 10).cuda().requires_grad_()
```
IvanYashchuk, line 608:
```suggestion
    x = torch.randn(5, 10, device="cuda", requires_grad=True)
```

thunder/tests/test_dynamo.py
```diff
@@ -535,3 +539,85 @@ def f(x):
...
+    model = SimpleModel().cuda().train()
```
IvanYashchuk, line 609: `training=True` is the default state of a nn.Module.
```suggestion
    model = SimpleModel().cuda()
```

thunder/tests/test_dynamo.py
```diff
@@ -535,3 +539,85 @@ def f(x):
...
+    backend = ThunderCompiler()
+    jf = torch.compile(backend=backend)(model)
+    out = jf(x)
+
+    subgraph_info = backend.subgraph_infos[0]
+    split_m = subgraph_info.split_graph_module
+    submodule_name = "wrap_body_0"
+    assert hasattr(split_m, submodule_name)
+
+    submodule = getattr(split_m, submodule_name)
```
IvanYashchuk, line 619:
```suggestion
    submodule = getattr(split_m, submodule_name)
```

IvanYashchuk (commented):
Let's resolve https://github.com/Lightning-AI/lightning-thunder/issues/1337 first and then merge this PR so that we are confident it would bring value for activation checkpointing use cases.

thunder/dynamo/utils.py
kiya00: `get_attr`, `placeholder`,`output `,`call_method` are fine to have, according to https://github.com/Lightning-AI/lightning-thunder/blob/287abf79ac7193f5d3dad6f7bdd7d5a42849ae6b/thunder/dynamo/utils.py#L278-L279, the module is inlined unless the flag is disabled.

thunder/dynamo/utils.py
kiya00: I think `sub_gm` has no reference to its parent, only gm has references to all its submodules. yes, `sub_gm` is only available in `gm`

thunder/dynamo/utils.py
kiya00: e.g. : the whole split module is as below:
```
# split_module
GraphModule(
  (wrap_body_0): GraphModule() 
  (thunder_1): ThunderModule( 
    (_model): GraphModule()
  )
)
```
the `gm` is `wrap_body_0`, `split_module` has reference to `wrap_body_0`, if we deepcopy `wrap_body_0` we need to update the reference in `split_module`, since we only need to modify the Graph structure, I think update the `gm.graph` inplace and recompile it should be fine

kiya00:
As mentioned in https://github.com/Lightning-AI/lightning-thunder/pull/1370#issue-2624528316
Here are some results comparing this PR and the [PR1370](https://github.com/Lightning-AI/lightning-thunder/pull/1370) in terms of peak memory usage on single H100 GPU

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/wayan/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/wayan/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">

</head>

<body link="#467886" vlink="#96607D">


H100 1GPU 2layers | this PR (GB) | ThunderFX with checkpoint fallback to inductor(GB)
-- | -- | --
longchat-13b-16k | 12.41 | 15.99
CodeLlama-34b-hf | 25.03 | 30.09
Gemma-2-27b | OOM | OOM
Llama-3-70B | 35.56 | 39.41
Mistral-7B-v0.2 | 21.36 | 28.78
vicuna-7b-v1.5-16k | 12.41 | 15.99

</body>

</html>
cc: @IvanYashchuk

kiya00:
`torchrun --nproc_per_node=8 --nnodes=1 thunder/benchmarks/benchmark_litgpt.py  --model_name Mistral-7B-v0.2  --micro_batch_size 3  --compile thunder+dynamo --checkpoint_activations=True --distributed_mode=fsdp --shard_mode zero3 --max_iters=4 --warmup_iters=1 --bucketing_mode=block` on 8*H100 (container: 20241028)
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/wayan/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/wayan/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">

</head>

<body link="#467886" vlink="#96607D">


  | micro batch size | thunder+dynamo | inductor
-- | -- | -- | --
longchat-13b-16k | 3 | 49.09 | 48.59
CodeLlama-34b-hf | 1 | 50.47 | 50.8
Gemma-2-27b | 1 | OOM | OOM
Llama-3-70B | 1 | OOM | OOM
Mistral-7B-v0.2 | 3 | 63.43 | 64.63
vicuna-7b-v1.5-16k | 5 | 51.67 | 51



</body>

</html>

thunder/dynamo/utils.py
IvanYashchuk: Thank you for the explanation. It's totally fine to replace `.graph` with the new object. Can we modify the `.graph` itself inplace then as well? No need to change anything in this PR but we should probably revisit this in a follow-up.

thunder/dynamo/utils.py
IvanYashchuk: It's me nitpicking here, no need to block merging this pull request because of this.

thunder/dynamo/utils.py
IvanYashchuk: Thank you for checking.

thunder/torch/__init__.py
IvanYashchuk: The test passes because the first element from the return tuple (`result` in this code) was registered as a "gradient" for the "function" argument of the checkpoint and simply discarded.

IvanYashchuk: Updated the test in https://github.com/Lightning-AI/lightning-thunder/pull/1391.

IvanYashchuk (approved):
Awesome work on enabling models with activation checkpointing to run with Thunder (https://github.com/Lightning-AI/lightning-thunder/pull/1261#issuecomment-2449340027)!

IvanYashchuk:

t-vi (approved):
Thank yyou @kiya00 @IvanYashchuk @crcrpar @kshitij12345

Issue #824, opened 2024-07-22
enter/exit_autocast of torch.amp.autocast_mode


tfogal:
## 🚀 Model / language coverage

I'm trying to get a fuller picture of what we need to support NeVA. As such I'm using:

```python
def thunder_backend(gm, args):
  gm.real_recompile()
  # Examine may raise an error
  from thunder.examine import examine
  try:
      thunder.examine.examine(gm, *args)
  except Exception as e:
      print(f"Hit problem with examine:\n{e}")
  # Don't really use Thunder just return the original graph
  return gm

...
#model.model = thunder.jit(model.model)
model.model = torch.compile(backend=thunder_backend)(model.model)
```

And one of the issues that gets reported is e.g.

```
Found 2 distinct operations, of which 0 (0.0%) are supported
Please file an issue requesting the following operators here: https://github.com/Lightning-AI/lightning-thunder/issues/
new
_enter_autocast of torch.amp.autocast_mode
_exit_autocast of torch.amp.autocast_mode
```

### Pitch

This looks like it is going to be important for #343.

### Minimal Repro

tfogal:
Assigning to me until I can fill out a better reproducer

IvanYashchuk:
Thunder doesn't support PyTorch's context managers like autocast, no_grad, enable_grad, etc. inside the compiled function.   With the recent addition of the autocast-specific dispatch at tracing time (https://github.com/Lightning-AI/lightning-thunder/pull/705, https://github.com/Lightning-AI/lightning-thunder/pull/810) supporting this might not take a lot of work, the challenge is not to reorder these enter and exit calls inappropriately.
Another way to approach this problem is to ensure Thunder never sees these calls by adding more graph breaks.

kshitij12345:
I think this could be the minimal repro:

```python
import torch
import thunder

class ThunderJitBackend:
    def __init__(self, **compile_options) -> None:        
        self.thunder_jit_fns = []
        self.dynamo_graphs = []
        self.cnt = 0
        self.compile_options = compile_options

    def compile(self, gm, sample_args):
        self.dynamo_graphs.append(gm)
        gm.real_recompile()
        thunder_jit_fn = thunder.jit(gm, **self.compile_options)
        self.thunder_jit_fns.append(thunder_jit_fn)
        self.cnt += 1
        return thunder_jit_fn

dev = "cuda"

def foo(x):
    with torch.autocast(dev, torch.bfloat16):
        y = x @ x
[...]

PR #1262, opened 2024-10-04
Support torch.autocast in compiled function

kshitij12345:
Related - https://github.com/Lightning-AI/lightning-thunder/issues/824

~~Depends on https://github.com/Lightning-AI/lightning-thunder/issues/1266 to be fixed.~~

This PR introduces tracing through `torch.autocast` context manager and adds corresponding symbols `autocast_enter` and `autocast_exit` for representation in trace. NOTE, we also tag these symbols with `CTX_MANAGER_ENTER_EXIT_OP` so that a future pass can eliminate these symbols.

These symbols don't map to any computation but instead manipulate a global stack (AutocastStack) to push and pop the relevant autocast information which includes autocast device, autocast dtype, if autocast is enabled or not. This information is then queried when relevant symbols with autocast rules are called.

We also introduce a pass which then removes symbols from trace so that they don't affect fusion and reordering.

NOTE - We create a lookaside for `torch.autocast.__enter__` but instead manually map it to the symbol `thunder.torch.autocast_enter` (used to represent autocast enter in the trace)

Eg.
```python
import torch
import thunder

def foo(x):
    x = x + x
    with torch.autocast("cpu", torch.bfloat16):
        x = (x * 2) @ x
        return x

jfoo = thunder.jit(foo)

jfoo(torch.randn(3, 3, requires_grad=True))
print(thunder.last_traces(jfoo)[4])

```

First Trace
```python
def computation(x):
  # x: "cpu f32[3, 3]"

  # /home/kkalambarkar/lightning-thunder/scratchpad/test.py:214:            x = x + x
  t0 = ltorch.add(x, x, alpha=None)  # t0: "cpu f32[3, 3]"
    # t0 = prims.add(x, x)  # t0: "cpu f32[3, 3]"
  ltorch.autocast_enter('cpu', torch.bfloat16, True, True)

  # /home/kkalambarkar/lightning-thunder/scratchpad/test.py:216:                x = (x * 2) @ x
  t2 = ltorch.mul(t0, 2)  # t2: "cpu f32[3, 3]"
    # _ = prims.convert_element_type(2, float)
    # t2 = prims.mul(t0, 2.0)  # t2: "cpu f32[3, 3]"
  t3 = prims.convert_element_type(t2, dtypes.bfloat16)  # t3: "cpu bf16[3, 3]"
  t4 = prims.convert_element_type(t0, dtypes.bfloat16)  # t4: "cpu bf16[3, 3]"
  t5 = ltorch.matmul(t3, t4)  # t5: "cpu bf16[3, 3]"
    # t5 = prims.matmul(t3, t4)  # t5: "cpu bf16[3, 3]"
  ltorch.autocast_exit()
  return t5
```

Trace after context manager removal pass
```python
def computation(x):
  # x: "cpu f32[3, 3]"

  # /home/kkalambarkar/lightning-thunder/scratchpad/test.py:39:             x = x + x
  t0 = ltorch.add(x, x, alpha=1)  # t0: "cpu f32[3, 3]"
    # t0 = prims.add(x, x)  # t0: "cpu f32[3, 3]"

  # /home/kkalambarkar/lightning-thunder/scratchpad/test.py:41:                 x = (x * 2) @ x
  t2 = ltorch.mul(t0, 2)  # t2: "cpu f32[3, 3]"
    # _ = prims.convert_element_type(2, float)
    # t2 = prims.mul(t0, 2.0)  # t2: "cpu f32[3, 3]"
  t3 = prims.convert_element_type(t2, dtypes.bfloat16)  # t3: "cpu bf16[3, 3]"
  t4 = prims.convert_element_type(t0, dtypes.bfloat16)  # t4: "cpu bf16[3, 3]"
  t5 = ltorch.matmul(t3, t4)  # t5: "cpu bf16[3, 3]"
    # t5 = prims.matmul(t3, t4)  # t5: "cpu bf16[3, 3]"
  return {'output': t5, 'flat_args': [x]}
```


kshitij12345:
Have filed an issue for the failures showing up in CI - https://github.com/Lightning-AI/lightning-thunder/issues/1266

thunder/core/jit_ext.py
```diff
@@ -1402,7 +1427,7 @@ def from_binary_subscr(provenance, *, new_output=False):
                 output = Proxy("subscr")  # name? collectify?
             else:
                 output = p
-            if isinstance(idx, (int, str)):
+            if isinstance(idx, (int, str, Proxy)):
```
kshitij12345, line 1430: Patch for https://github.com/Lightning-AI/lightning-thunder/issues/1266 (shouldn't be part of this PR)

kshitij12345:
Opening for review though #1266 needs to be fixed first.

t-vi (commented):
So I love the improvement in our autocast support, thank you @kshitij12345 

The thing I'm wondering whether it may be tricky is whether having these symbols in the trace is tricky for us w.r.t. reordering, fusion etc.
So I wonder whether we should really mark the operations with their autocast state rather than putting symbols in the trace that manipulate global state.
I'm happy to be convinced of this approach though.

WDYT?

kshitij12345:
> If we don't actually need them in the trace, maybe we can keep them in the initial trace and then remove them before functionalization?

Yeah, they are not actually needed and kept them for inspectability. Will update the PR to have them explicitly removed before functionalization.

kshitij12345:
@t-vi have added an explicit pass to remove these symbol. PTAL, thanks :)!

PR #1268, opened 2024-10-07
Add an nvtx decorator for annotating components

tfogal:
## What does this PR do?

Starts to add nvtx around (some) of the larger or at least more variable components of our compilation and execution. These help with identifying performance issues.

Some quick examples of what this does: at a high-level, we quickly see where we're spending compilation time:

image

Drilling into performance of actual execution:
image

On the right one can see how the effects of the `NvtxProfileTransform` (and, in this case, nvFuser's nvtx) nest within the added `computation_fn`.

The intent with this PR is to capture the largest elements of where our time is spent. You may notice in the screenshots that there is still some time unaccounted for--I am happy to go further based on feedback but felt this is in a good enough place to solicit feedback already.

## PR review

## How do people feel about an `nvtx` package dependency?

The nvtx library includes the concept of [a 'domain' argument that can isolate different libraries in a trace](https://github.com/NVIDIA/NVTX?tab=readme-ov-file#isolate-nvtx-annotations-in-a-library-using-a-domain). Unfortunately torch's nvtx.range_push() does not support this.

The [nvtx python package](https://pypi.org/project/nvtx/) supports this. If we used the `nvtx` package, then instead of naming each nvtx region as "thunder foo" we could just use a "thunder" domain. This lets us have a special `thunder` line when looking at traces:

image

image

which can be nice. Of course, this means we have a dependency that we wouldn't otherwise have.

The latest version _tries_ to use nvtx domains, if the package is available, and falls back to torch's implementation if not. But reasonable people may prefer seeing all the nvtx together 🤷 . Please comment.


thunder/__init__.py
```diff
@@ -689,7 +749,9 @@ def get_computation_and_inputs(*args, **kwargs):
                 computation_traces.append(computation_trc)
 
             computation_trc = transform_to_torch_types(computation_trc)
+            _nvtx_push("python_callable main")
```
crcrpar, line 752: "main" reads a bit unclear at glance as we have a namespace named main as well

crcrpar (commented):
I'd find this way nicer if there exists a strict naming rule. Atm I see some names that look like a function name while the others seem to describe what each region does

thunder/__init__.py
```diff
@@ -751,10 +816,13 @@ def fn_(*args, **kwargs) -> Any:
                 *data_for_autograd["flat_args"],
             )
             result = data_for_autograd["output"]
+            _nvtx_pop()
 
         if cache_entry.epilogue_fn:
             result, comp_to_epi = result
+            _nvtx_push("epilogue")
             cache_entry.epilogue_fn(*pro_to_epi, *comp_to_epi)
+            _nvtx_pop()
```
IvanYashchuk, line 825: A Pythonic way to do this is to add a decorator to `epilogue_fn` and it's better to do it where the function is defined instead of modifying the code here.
```py
@nvtx.annotate(domain="thunder")
def epilogue_fn():
    pass
```

thunder/__init__.py
```diff
@@ -735,13 +797,16 @@ def fn_(*args, **kwargs) -> Any:
...
+        _nvtx_push("computation_fn")
         result = cache_entry.computation_fn(*inps)
+        _nvtx_pop()
 
         if cache_entry.backward_fn:
             # Run the compiled forward function
             data_for_autograd, (saved_tensors, saved_other) = result
 
             # Connect produced tensors with PyTorch's autograd graph
+            _nvtx_push("connect autograd")
```
IvanYashchuk, line 809: A Pythonic way to push and pop is using a context manager. Then any exception in the middle wouldn't mess up the annotations.
```py
with nvtx.annotate(message="connect autograd"):
    pass
```

IvanYashchuk (commented):
Before using NVTX domains the execution time should be measured. From https://nvtx.readthedocs.io/en/latest/annotate.html#domains:
> Domains should be used sparingly as they are expensive to create.

From this statement it's not clear whether expensive refers to filesize or CPU overhead.

I don't think Nsight Systems (nsys) is the right tool for everyone to do Python CPU profiling. Yes, these annotations are useful for those who use nsys, but they are completely useless for others and add unnecessary complexity to the code.

It's okay to annotate the hot path for easier mapping of regions to GPU kernels, it should be done using Pythonic constructs like decorators and context managers. It would be awesome to refactor the code of the main function `_fn` to use clear stages calling into other functions so that any other profiling instrumentation could use this information without a practitioner resorting to line profilers or range-based annotations like here.

I agree with Masaki that there should be a cohesive style for these annotations.

thunder/__init__.py
t-vi: Note that epilogue_fn is generated, but last I looked
```
epilogue_fn = nvtx.annotate(domain="thunder")(epilogue_fn)
```
should do if you don't want the context manager.

t-vi (commented):
I must admit I'm not enthused by how we just add whatever.

- Can we please have an abstraction where we just have a single line of `profile.start_segment("foo")` or so in various places (plus one start and end in some decorator) and then that figures out what to do (nvtx or no-op, I don't think layers of automated fallback is good). Preferable, this abstraction would live outside `thundet/__init__.py`. Then at the end have something.
- If we decide that this is how we profile, let's get rid of all times in `pref_counter_ns`. We can't just pile on without removing the approaches that didn't work out.

tfogal:
> Before using NVTX domains the execution time should be measured. From https://nvtx.readthedocs.io/en/latest/annotate.html#domains:
> 
> > Domains should be used sparingly as they are expensive to create.
> 
> From this statement it's not clear whether expensive refers to filesize or CPU overhead.

They are expensive _to create_. They are cheap otherwise.
The nvtx lib initializes on first use and so the first nvtx with a domain is expensive.

> I don't think Nsight Systems (nsys) is the right tool for everyone to do Python CPU profiling. Yes, these annotations are useful for those who use nsys, but they are completely useless for others and add unnecessary complexity to the code.

I haven't found any other CPU profilers to be useful for our situation. Which ones do you see success with?

> It's okay to annotate the hot path for easier mapping of regions to GPU kernels, it should be done using Pythonic constructs like decorators and context managers. It would be awesome to refactor the code of the main function `_fn` to use clear stages calling into other functions so that any other profiling instrumentation could use this information without a practitioner resorting to line profilers or range-based annotations like here.

Refactors are always nice, but I cannot sign up for that here just for profiling, sorry.

> I agree with Masaki that there should be a cohesive style for these annotations.

Okay, could you (or Masaki) please comment either way? I am happy to do whatever here but the guidance at present is "away from" and not "towards" something.

thunder/__init__.py
tfogal: > A Pythonic way to do this is to add a decorator to `epilogue_fn` and it's better to do it where the function is defined instead of modifying the code here.
> 
> ```python
> @nvtx.annotate(domain="thunder")
> def epilogue_fn():
>     pass
> ```

There is no `epilogue_fn` in the code as this is a functor.

thunder/__init__.py
tfogal: Closing this thread as this is tracked in the thread above.

t-vi:
> "If we decide" -- this is really a you decision, Tom

I'm glad we get a new way, but let's remove the old, too.

tfogal:
TomV, Ivan, and I talked offline.

I am going to repurpose this PR to add a decorator that uses either the nvtx library or is a no-op. It won't actually be used at this point. Ivan will restructure thunder.jit to be more amenable to having a decorator attached. Then we'll come back and apply the decorator at key points.

We have looked at nvtx overhead in C, and it's on the order of 300-800 nanoseconds (incl. both the 'start' and 'end' markers) when the profiler is attached, and within the noise of timer resolution when the profiler is not attached. This would imply that we probably don't want to nvtx anything below ~2us. However, the overhead may be worse in Python than it was in C; Ivan is going to measure.

IvanYashchuk:
Below are the timings for the call overhead in Python when the profiler is not attached. We can see that the decorator from `nvtx` is cheap even with the domain specified. Interestingly, the NVTX decorator from PyTorch is 1000x slower than from the `nvtx` Python package.
```py
In [1]: import nvtx

In [2]: def f(): pass

In [3]: nvtxed_f = nvtx.annotate("test")(f)

In [4]: %%timeit
   ...: nvtxed_f()
   ...: 
   ...: 
129 ns ± 0.147 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [5]: import torch

In [6]: torch_nvtxed_f = torch.cuda.nvtx.range("test")(f)

In [7]: %%timeit
   ...: torch_nvtxed_f()
   ...: 
   ...: 
1.26 µs ± 2.71 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [8]: nvtxed_f_with_domain = nvtx.annotate("test", domain="thunder")(f)

In [9]: %%timeit
   ...: nvtxed_f_with_domain()
   ...: 
   ...: 
129 ns ± 0.273 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

thunder/core/profile.py
```diff
@@ -1,11 +1,28 @@
...
+    import nvtx
+
+    _ENABLED = True
+except ImportError:
+    if _ENABLED:
+        msg = "Requested nvtx but the package is not available."
+        msg += "\nUse `pip install -m pip install nvtx`."
+        warnings.warn(msg)
+        _ENABLED = False
+        raise
```
tfogal, line 24: My goal here was that people do not need to learn about our environment variable to get useful information out of thunder, which I would argue is good UX when thunder is just one of hundreds of packages in use.

Interactively we discussed wanting users to see the error if they requested annotations but don't have the lib, hence the re-raise.

thunder/common.py
```diff
@@ -629,6 +630,7 @@ def wait_for_future(f: FutureTensorProxy) -> TensorProxy:
 # TODO Consider making this faster by reusing more data
 # TODO Create a general mechanism for running traces that produces reproducible provenance and the
 #   appropriate error checks
+@thunder.core.profile.profile("transform_for_execution")
```
tfogal, line 633: Happy to remove any/all of these for now if that makes your refactoring easier, @IvanYashchuk. Just wanted a couple places so that the decorator would actually get tested in CI.

tfogal (commented):
I believe the windows failures are unrelated; we're seeing it in other PRs. I started a slack thread on that.

Anyway, as discussed this is now basically just the decorator. Ivan will refactor things in a parallel PR, and then I'm happy to come back and apply the decorator in appropriate places.

A note for one of your requests, @t-vi: mixology is actually using some of the manual timer stuff that is in there. As such, I'd appreciate if we could have a period where both mechanisms exist in the code (as ugly as that is), to give the mixology team time to update their code accordingly. I am happy to take on removing legacy mechanisms after mixology has upgraded, though.

t-vi:
> A note for one of your requests, @t-vi: mixology is actually using some of the manual timer stuff that is in there. As such, I'd appreciate if we could have a period where both mechanisms exist in the code (as ugly as that is), to give the mixology team time to update their code accordingly. I am happy to take on removing legacy mechanisms after mixology has upgraded, though.

Sounds like a plan, who is driving the upgrade to the new method at mixology?

tfogal:
> Sounds like a plan, who is driving the upgrade to the new method at mixology?

Nobody yet; I've given a heads-up this is coming but it's not actionable for them until the new mechanism is in place.

I'll remind them at a Tuesday meeting and maybe we can come up with a loose schedule/timeline. I don't believe it's reasonable for me to push hard on them to schedule this until we have our own schedule for when they could start, sadly.

thunder/common.py
IvanYashchuk: Another way to test the decorator in CI is to add a dedicated test.

thunder/__init__.py
```diff
@@ -368,6 +368,7 @@ def _alias_tensor_of_args_kwargs_dict(*args, **kwargs) -> dict[int, list[int]]:
                 tensor_group_index_to_tensor_indices[tgi].append(idx)
         return tensor_group_index_to_tensor_indices
 
+    @thunder.core.profile.profile("_alias_tensor_of_args_kwargs")
```
IvanYashchuk, line 371: This function doesn't do any profiling. Why do you call it "profile"? Do you intend to add routing to [torch.profiler.profile](https://pytorch.org/docs/stable/profiler.html#torch.profiler.profile)?

tfogal: 🤷 I think of adding nvtx markers as "profiling".

But I don't have any stake in the naming here--what would be a good name, do you think?

tfogal: To your second question:

> Do you intend to add routing to [torch.profiler.profile](https://pytorch.org/docs/stable/profiler.html#torch.profiler.profile)?

I didn't intend to, but you and TomV think that'd be useful I'd be happy to do so.

thunder/common.py
tfogal: Cool, done.

tfogal:
*ping* @t-vi this is waiting on your review and/or comment that [we should extend this to include torch.profile.profile too](https://github.com/Lightning-AI/lightning-thunder/pull/1268#discussion_r1809189327).

> > Sounds like a plan, who is driving the upgrade to the new method at mixology?
> 
> Nobody yet; I've given a heads-up this is coming but it's not actionable for them until the new mechanism is in place.

Martyna and I talked and they actually use a different mechanism for identifying compilation time. So, we can rip out the old timers whenever.

I'd like to avoid ripping out the unused timers in this PR as it will likely produce conflicts for Ivan's #1333, but I am happy to do it in general.

t-vi (approved):
Let's have this.
Thank you @tfogal @IvanYashchuk @crcrpar

thunder/__init__.py
t-vi: Let's name it `annotate_for_profiling`

thunder/core/profile.py
```diff
@@ -27,3 +44,30 @@ def add_markers(msg: str) -> None:
 
         finally:
             torch.cuda.nvtx.range_pop()
+
+
+# The main interface to profiling something. Generally used as a decorator:
+#   @thunder.core.profile.profile("foo")
```
IvanYashchuk, line 50:
```suggestion
#   @thunder.core.profile.annotate_for_profile("foo")
```

thunder/core/profile.py
```diff
@@ -27,3 +44,30 @@ def add_markers(msg: str) -> None:
...
+#   def foo(...): ...
+# but alternatively as a `with` context:
+#   with thunder.core.profile.profile("name for a block of code"):
```
IvanYashchuk, line 53:
```suggestion
#   with thunder.core.profile.annotate_for_profile("name for a block of code"):
```

tfogal: Thank you for this!

PR #1270, opened 2024-10-08
Disallow traceable tensor subclasses

crcrpar:
## What does this PR do?

Conservatively errors out when any of tensor inputs are a traceable tensor subclass and any non-pytorch executor is enabled.
Ideally we should interpret, translate, and disable `__torch_dispatch__` eventually.


t-vi:
> Conservatively errors out when any of tensor inputs are a traceable tensor subclass and any non-pytorch executor is enabled.

I think it would be good to just error out and not depend on the executors here. WDYT?

PR #1280, opened 2024-10-10
stop start import from clang in thunder namespace

crcrpar:
## What does this PR do?

At glance, the star import does not seem to have many uses.


t-vi (approved):
Thank you @crcrpar

PR #1283, opened 2024-10-10
add helper for `SigInfo` creation in `Function.apply` lookaside

crcrpar:
## What does this PR do?

As per title, add a helper, as we repeat the same thrice in the lookaside

cc @borda

### DIFF
--- a/thunder/core/codeutils.py
+++ b/thunder/core/codeutils.py
@@ -339,6 +339,19 @@ class SigInfo:
 
         return f"def {self.name}({arg_str}):"
 
+    @staticmethod
+    def from_name_and_args(name: str, args: Sequence[Any]):
+        si = SigInfo(name)
+        for a in args:
+            if isinstance(a, ProxyInterface):
+                si.args.append((a.name, None))
+            else:
+                from thunder.core.proxies import proxy
+
+                pa = proxy(a)
+                si.args.append((pa.name, None))
+        return si
+
 
 # Creates a SigInfo object from a function and the inputs to it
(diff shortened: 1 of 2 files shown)


IvanYashchuk (approved):
Nice!

t-vi:
This is nice, but wouldn't the real deal be making siginfo optional?
The only thing we don't have is placeholder names for constants, which could just be made up on the fly when doing `.python()`.

crcrpar:
> wouldn't the real deal be making siginfo optional?

If we are to do this, I reckon siginfo should be hidden inside of trace and it should always be created from trace's args, instead of allowing some freedom by making it optional.

t-vi:
> If we are to do this, I reckon siginfo should be hidden inside of trace and it should always be created from trace's args, instead of allowing some freedom by making it optional.

So the siginfo does three things:
- hold names - the proxies in args already have them, the others are actually consts and could just use unused proxies in args with a value equal to the const value,
- hold default args (not used often, but we probably still want this but it could be a dict),
- mark *arg and **kwarg (I'm not sure we support this properly and I don't think it is used, if we want that, it would likely need fixing).

The trouble I have with siginfo that it's most often not used much except for the names and that it can go out of sync, leading to inconsistencies. This is why I'd think using proxies in all arg for the names (might need adjustments, though) and keeping the default arguments elsewhere would let us move to a single source of truth.

t-vi:
That said, this PR strictly improves things, so, yay!

t-vi (approved):
Thank you @crcrpar  @IvanYashchuk.
I'd like to remove siginfo, but clearly this is better than before.

Issue #1276, opened 2024-10-09
Phi3 masked_fill_ in-place support


tfogal:
## 🚀 Model / language coverage

Trying to run the HF model "microsoft/Phi-3-mini-4k-instruct" hits an issue translating an in-place op.

```
Traceback (most recent call last):
  File "/home/tfogal/scratch/tfx-tests/phi3/phi3.py", line 70, in <module>
    outputs = model(input_ids=input_ids, attention_mask=attention_mask, labels=input_ids)
  File "/usr/local/lib/python3.10/dist-packages/torch/nn/modules/module.py", line 1736, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/torch/nn/modules/module.py", line 1747, in _call_impl
    return forward_call(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/torch/_dynamo/eval_frame.py", line 465, in _fn
    return fn(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/torch/nn/modules/module.py", line 1736, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/torch/nn/modules/module.py", line 1747, in _call_impl
    return forward_call(*args, **kwargs)
  File "/home/tfogal/.cache/huggingface/modules/transformers_modules/microsoft/Phi-3-mini-4k-instruct/0a67737cc96d2554230f90338b163bc6380a2a85/modeling_phi3.py", line 1243, in forward
    outputs = self.model(
  File "/usr/local/lib/python3.10/dist-packages/torch/nn/modules/module.py", line 1736, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/torch/nn/modules/module.py", line 1747, in _call_impl
    return forward_call(*args, **kwargs)
  File "/home/tfogal/.cache/huggingface/modules/transformers_modules/microsoft/Phi-3-mini-4k-instruct/0a67737cc96d2554230f90338b163bc6380a2a85/modeling_phi3.py", line 1091, in forward
    attention_mask = _prepare_4d_causal_attention_mask(
  File "/home/tfogal/env/lib/python3.10/site-packages/transformers/modeling_attn_mask_utils.py", line 295, in _prepare_4d_causal_attention_mask
[...]

PR #1292, opened 2024-10-13
specialcasing `Tensor.to`

crcrpar:
## What does this PR do?

A potential fix for #1276.


t-vi (approved):
Thank you @crcrpar 
I wonder if having this logic inside .to somewhere would be better (or maybe check the subsymbols?) but let's have this for now.

PR #1294, opened 2024-10-14
Make graph visualization from a trace

riccardofelluga:
<details>

</details>

## What does this PR do?

In the quest of finding the OOMs I wrote this utility that takes a trace and returns a graph that can be rendered out to visualize the flow of computation.

This requires installing graphviz to work, I haven't added it to the dependencies as it is only needed in this function, however it might be added if you think its better to.

## Usage

Here is an example, from the following code:

```python
import torch
import torch.nn as nn

import thunder
from thunder.examine import make_trace_dot

class MyModel(nn.Module):
    def __init__(self) -> None:
        super().__init__()
        self.layers = nn.Sequential(nn.Linear(10, 10), nn.ReLU(), nn.Linear(10, 10))

    def forward(self, x):
        return self.layers(x)

a = torch.randn((10, 10), requires_grad=True, device="cuda")
m = MyModel().to("cuda")

jm = thunder.jit(m)
jm(a)

bwd_trace = thunder.last_backward_traces(jm)[0]

dot = make_trace_dot(bwd_trace)

dot.render(filename="backward_trace", format="svg")
```

will output something like this:

backward_trace

cc @borda @apaz-cli @carmocca

### DIFF
--- a/thunder/examine/__init__.py
+++ b/thunder/examine/__init__.py
@@ -2,10 +2,10 @@ from typing import Any
 from collections.abc import Callable
 import collections
 import traceback
-import warnings
 
 import thunder
 from thunder.core.trace import TraceCtx
+from thunder.core.transforms import bsym_list_to_dag, Node
 from thunder.core.proxies import TensorProxy
 from thunder.core.symbol import BoundSymbol
 from thunder.torch import _torch_to_thunder_function_map
@@ -14,13 +14,7 @@ from thunder.core.langctxs import resolve_language, LanguageContext, Languages
 import torch
 from warnings import warn
 from itertools import chain
-
-try:
-    import graphviz
-
-    HAS_GRAPHVIZ = True
-except ImportError:
-    HAS_GRAPHVIZ = False
+import importlib
 
 
 # TODO Maybe make collect_into a set?
@@ -282,7 +276,7 @@ def get_nvfuser_repro(trace: TraceCtx, fusion_name: str, /) -> str:
     return get_repro(fusion.last_inputs)
 
 
-def make_trace_dot(trace: TraceCtx) -> graphviz.Digraph:
+def make_trace_dot(trace: TraceCtx):
     """
     Creates a directed graph of the given trace.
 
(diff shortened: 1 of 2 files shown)


thunder/examine/__init__.py
```diff
@@ -14,6 +15,13 @@
 from warnings import warn
 from itertools import chain
 
+try:
```
Borda, line 18: you can use https://github.com/Lightning-AI/utilities/#3-general-python-utilities-lightning_utilitiescore

t-vi (approved):
Who would not love graphs?
Thank you @riccardofelluga @Borda

thunder/examine/__init__.py
```diff
@@ -272,3 +274,59 @@ def get_nvfuser_repro(trace: TraceCtx, fusion_name: str, /) -> str:
...
+    Requires graphviz to be installed, for more information check out -> https://graphviz.readthedocs.io/en/stable/index.html
+
+    Args:
+        trace (TraceCtx): The Thunder trace to be made into a graph.
+
+    Returns:
+        graphviz.Digraph: A graphviz directed graph.
+    """
+    if not importlib.util.find_spec("graphviz"):
+        warn("graphviz is not available. Graph cannot be created.")
```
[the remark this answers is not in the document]
IvanYashchuk: Let's add a more friendly message like here
https://github.com/pytorch/pytorch/blob/7f88bf96f9ff37f3873cc5cb8c962ebcbf509c65/torch/fx/passes/graph_drawer.py#L442-L443

thunder/examine/__init__.py
```diff
@@ -272,3 +274,59 @@ def get_nvfuser_repro(trace: TraceCtx, fusion_name: str, /) -> str:
...
+        return
+
+    import graphviz
```
IvanYashchuk, line 298: PyTorch FX uses `pydot`. What's the difference from `graphviz` and should one be preferred over another?
https://github.com/pytorch/pytorch/blob/7f88bf96f9ff37f3873cc5cb8c962ebcbf509c65/torch/fx/passes/graph_drawer.py#L112

thunder/examine/__init__.py
```diff
@@ -272,3 +274,59 @@ def get_nvfuser_repro(trace: TraceCtx, fusion_name: str, /) -> str:
...
+
+    node_attr = dict(
+        style="filled", shape="box", align="left", fontsize="10", ranksep="0.1", height="0.2", fontname="monospace"
```
IvanYashchuk, line 301: It's been a while since I used FX's drawer and don't remember how it looks. Here's a link to their style and maybe you would like it
https://github.com/pytorch/pytorch/blob/7f88bf96f9ff37f3873cc5cb8c962ebcbf509c65/torch/fx/passes/graph_drawer.py#L150

thunder/examine/__init__.py
IvanYashchuk, line 1: Let's adopt a new habit of creating a new file and putting the implementation there. Inside this `__init__.py` only the import of the new function should be added.

PR #1300, opened 2024-10-15
match number of backward return values to number of forward args

crcrpar:
The test case lacked `None` as the last return value


t-vi (approved):
Thank you @crcrpar

PR #1301, opened 2024-10-15
Use `grad_transform` in `torch.autograd.Function` lookaside

crcrpar:
## What does this PR do?

Fixes https://github.com/Lightning-AI/lightning-thunder/blob/30e4aa1e67005c58219d7f06b46836eedb74b27a/thunder/core/jit_ext.py#L696


t-vi (approved):
Seems good.
I wonder if we lose clarity by not having the backward symbol, but it is neat to go through the grad transform.

t-vi:
I'll merge main after #1329 lands

PR #1305, opened 2024-10-15
const-folding : minor clean-up

kshitij12345:
Remove commented stray print and unused imports


tfogal (approved):
Thank you.

Test failures seem unrelated; maybe we need to increase the timeout?

t-vi (approved):
Thank you @kshitij12345 @tfogal

PR #1315, opened 2024-10-16
Make Thunder tests runnable with non-distributed build of PyTorch: move litgpt imports into tests

IvanYashchuk:
This PR moves litgpt imports into tests to have better visibility of what tests do not work in the PyTorch build without distributed support. Currently, litgpt is unimportable if PyTorch is built without distributed support (https://github.com/Lightning-AI/litgpt/issues/1792).

`get_default_torch_fsdp_executor` was importing a class from the distributed module just for the type annotation prohibiting the use of the whole `thunder/benchmarks/__init__.py` file. I've removed this type of annotation.

cc @borda @apaz-cli

### DIFF
--- a/thunder/benchmarks/__init__.py
+++ b/thunder/benchmarks/__init__.py
@@ -788,9 +788,6 @@ def default_torch_ddp_executor(_) -> Callable:
 
 @dataclass(frozen=True)
 class get_default_torch_fsdp_executor:
-    from torch.distributed.fsdp import ShardingStrategy
-
-    sharding_strategy: ShardingStrategy
     apply_torch_compile: bool
     auto_wrap_policy: Any | None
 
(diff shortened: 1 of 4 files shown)


thunder/benchmarks/__init__.py
```diff
@@ -1221,12 +1217,14 @@ def args(cls) -> tuple[BenchmarkArg, ...]:
 
     def __init__(
         self,
-        config: str | LitGPTConfig,
+        config: str,
```
tfogal, line 1220: nit: since `LitGPTConfig` is still a valid thing to pass, maybe `str | Any`, or even `str | "LitGPTConfig"`?

tfogal (approved):

thunder/benchmarks/__init__.py
IvanYashchuk: Since it's a nit I'll keep it as is. We don't use type checkers anyway.

t-vi (approved):
Thank you @IvanYashchuk

thunder/benchmarks/__init__.py
t-vi: I'm pretty sure the next person will want to have their type checker enabled, but well.

PR #1319, opened 2024-10-16
Avoid rematerializing matmul/linear by default. 

wujingyue:
This was found to be expensive.

### DIFF
--- a/thunder/tests/test_transforms.py
+++ b/thunder/tests/test_transforms.py
@@ -4,6 +4,9 @@ from lightning_utilities.core.imports import package_available
 import pytest
 
 import thunder
+import thunder.examine
+from thunder.core import prims
+from thunder.core.trace import TraceCtx
 from thunder.dev_utils.nvtx_profile_transform import NvtxProfileTransform, nvtx_push, nvtx_pop
 from thunder.tests.framework import requiresCUDA
 from thunder.tests import litgpt_model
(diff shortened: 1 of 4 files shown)


thunder/tests/test_transforms.py
```diff
@@ -437,6 +440,46 @@ def forward(self, x):
...
+        def forward(self, x):
+            return self.layers(x)
+
+    batch_size, sequence_length, embedding_size = 2, 3, 4
+    inp = torch.randn(batch_size, sequence_length, embedding_size, device="cuda", requires_grad=True)
+
+    model = MLP(embedding_size)
+    model.to("cuda")
+
+    jmodel = thunder.jit(model, nv_enable_linear=True, nv_enable_matmul=True)
```
wujingyue, line 465: I'm not sure why nv_enable_linear and nv_enable_matmul need to be on to repro. I believe it's related to [this is_fusion check](https://github.com/Lightning-AI/lightning-thunder/blob/40904c762f6b1c2eb1a0c3a75a7b4b1382108b84/thunder/core/rematerialization.py#L239). @IvanYashchuk, feel free to code-comment an explanation.

crcrpar: Without those are enabled matmul and linear are not fused into nvfuser regions and we have set them off by default since they were added

IvanYashchuk (approved):
Thank you!

t-vi (approved):
Thank you @wujingyue @IvanYashchuk

PR #1321, opened 2024-10-17
remove comment out `BoundSymbol.set_output`

crcrpar:
## What does this PR do?

As per title. I don't see convincing reasons to keep it


t-vi (approved):
Thank you @crcrpar

Issue #1142, opened 2024-09-11
CUDA transform of a NeVA graph


tfogal:
## 🐛 Bug

Inside the `CUDAGraphTransform` `to_arg_descriptor` code, `zip(*map(extract_descriptor, args))` is not getting the values it expects and this leads to a tuple unpacking issue.

```
Traceback (most recent call last):
  File "/home/tfogal/dev/nemo/./examples/multimodal/multimodal_llm/neva/neva_pretrain.py", line 484, in thunder_graph_backend4
    fqn(*args) # run once to force compilation
  File "/usr/local/lib/python3.10/dist-packages/torch/nn/modules/module.py", line 1736, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/torch/nn/modules/module.py", line 1748, in _call_impl
    return forward_call(*args, **kwargs)
  File "/home/tfogal/dev/thunder/thunder/core/module.py", line 81, in forward
    res = self._forward_fn(*args, **kwargs)
  File "/home/tfogal/dev/thunder/thunder/__init__.py", line 787, in fn_
    result = cache_entry.computation_fn(*inps)
  File "/usr/local/lib/python3.10/dist-packages/torch/utils/_contextlib.py", line 116, in decorate_context
    return func(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/torch/amp/autocast_mode.py", line 44, in decorate_autocast
    return func(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/torch/amp/autocast_mode.py", line 44, in decorate_autocast
    return func(*args, **kwargs)
  File "thunder.computation_261", line 8, in computation
    [scale_t, getitem] = CUDAGraph1()
  File "/home/tfogal/dev/thunder/thunder/transforms/cudagraph.py", line 193, in callable
    return self.call_cuda_graph(fn_name, *args)
  File "/home/tfogal/dev/thunder/thunder/transforms/cudagraph.py", line 119, in call_cuda_graph
    cache_key = self.make_cache_key(fn_name, *args)
  File "/home/tfogal/dev/thunder/thunder/transforms/cudagraph.py", line 114, in make_cache_key
    return (fn_name, self.make_static_inputs_mask(fn_name, *args), to_arg_descriptor(*args))
[...]

PR #1324, opened 2024-10-17
fix cudagraphs without tensor inputs

t-vi:
Fixes: #1142

(well fixes the error message hit there. @tfogal do you want this to close the issue or keep it open and update the traceback)


lantiga (approved):
Stamped!

tfogal (approved):

tfogal:
Thanks! We can close the related issue; i'll open a new one with a new traceback (if we still hit this).

PR #1325, opened 2024-10-17
Temporarily disable some inductor tests on windows

tfogal:
These tests require that the `cl` compiler is available and this may take some setup.

## What does this PR do?

Disables some tests since they require some setup on the runners.

### DIFF
--- a/thunder/tests/test_autocast.py
+++ b/thunder/tests/test_autocast.py
@@ -1,4 +1,5 @@
 import itertools
+import platform
 
 import pytest
 import torch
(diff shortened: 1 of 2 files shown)


crcrpar (approved):

IvanYashchuk:
Did Inductor start to work on Windows but Thunder's CI setup is now wrong?

t-vi:
> Did Inductor start to work on Windows but Thunder's CI setup is now wrong?

Yes, probably as "easy" as finding the right way of setting a path to cl (integrating vcvarsall.bat) but apparently the usual way is to go through including third party app in the CI (which is a bit meh to me to have 100s of loc to set a PATH=).

t-vi (approved):
Let's do this for now until we figure out setting up the env.
Thank you @tfogal @crcrpar @IvanYashchuk

PR #1327, opened 2024-10-18
Skip nvFuser's test_sdpa for pre-Ampere cards because flash attention is not available

IvanYashchuk:
This skip is necessary to unblock NV-internal CI on older GPUs.


t-vi (approved):
Seems good, thank you @IvanYashchuk

IvanYashchuk:
CI failures are fixed in https://github.com/Lightning-AI/lightning-thunder/pull/1325.

PR #1329, opened 2024-10-18
bump gpu timeout

t-vi:
Even better would be to have more runners and smaller jobs, but it is what it is.

### DIFF
--- a/.azure/gpu-tests.yml
+++ b/.azure/gpu-tests.yml
@@ -84,7 +84,6 @@ jobs:
             pytest thunder/tests/ \
               -m "not standalone" \
               -v --datefmt="%Y%m%d-%H:%M:%S.%f" \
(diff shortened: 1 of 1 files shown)


.azure/gpu-tests.yml
```diff
@@ -97,7 +96,7 @@ jobs:
           ./codecov --token=$(CODECOV_TOKEN) --commit=$(Build.SourceVersion) \
             --flags=gpu,pytest,regular --name="GPU-coverage" --env=linux,azure
         condition: ne(variables['testing'], 'distributed')
-        timeoutInMinutes: "30"
+        timeoutInMinutes: "35"
```
t-vi, line 99:
```suggestion
        timeoutInMinutes: "40"
```
to not kill it at 99%

Issue #1285, opened 2024-10-10
Support mistralai/Mistral-Nemo-Base-2407


tfogal:
## 🚀 Model / language coverage

Support the [Mistral-Nemo-Base-2407 model](https://huggingface.co/mistralai/Mistral-Nemo-Base-2407) in Thunder.

* #1279 

### Pitch

This is an ask from the NeMo team.

### Minimal Repro

Current version of the model:

```
import torch
from torch.utils.data import DataLoader
from transformers import AutoConfig, AutoModelForCausalLM, AutoTokenizer, get_scheduler
from transformers import get_scheduler
from huggingface_hub import snapshot_download
from pathlib import Path
from datasets import load_dataset

mistral_models_path = Path.home().joinpath('mistral_models', 'Nemo-v0.1')
mistral_models_path.mkdir(parents=True, exist_ok=True)

snapshot_download(
  repo_id="mistralai/Mistral-Nemo-Base-2407",
  allow_patterns=["params.json", "consolidated.safetensors", "tekken.json"],
  local_dir=mistral_models_path
)

model_id = "mistralai/Mistral-Nemo-Base-2407"
tokenizer = AutoTokenizer.from_pretrained(
  model_id,
  torch_dtype=torch.bfloat16,
  ignore_mismatched_sizes=True,
)

config = AutoConfig.from_pretrained(model_id)
config.num_hidden_layers = 2
config.torch_dtype = torch.bfloat16
config.ignore_mismatched_sizes=True
config.max_position_embeddings=1024
model = AutoModelForCausalLM.from_config(config)

# Add a padding token to the tokenizer
if tokenizer.pad_token is None:
    tokenizer.add_special_tokens({'pad_token': '[PAD]'})
    model.resize_token_embeddings(len(tokenizer))

dataset = load_dataset("tiny_shakespeare", split='train',
                       trust_remote_code=True)

# Tokenize the dataset
def tokenize_function(examples):
    return tokenizer(examples["text"], padding="max_length", truncation=True,
                     max_length=2)

tokenized_dataset = dataset.map(tokenize_function, batched=True, remove_columns=["text"])

# Convert the dataset to PyTorch format and specify columns to return as tensors
tokenized_dataset.set_format(type='torch', columns=['input_ids', 'attention_mask'])

# Create PyTorch DataLoader
dataloader = DataLoader(tokenized_dataset, batch_size=1, shuffle=True)

# Define optimizer and learning rate scheduler
[...]

Issue #1240, opened 2024-10-01
VJP for PrimIDs.COPY_WITH_SETITEM is not implemented


kshitij12345:
NOTE : Found while trying to run `hf_Longformer` with torchbench and thunderFX path.

```python
import torch
import thunder

def forward(self, hidden_states_padded_1: "f32[24, 3, 512, 513]", diagonal_attention_scores: "f32[24, 4, 256, 513]"):
        # File: /usr/local/lib/python3.10/dist-packages/transformers/models/longformer/modeling_longformer.py:861 in _sliding_chunks_query_key_matmul, code: diagonal_attention_scores[:, :-1, :, window_overlap:] = diagonal_chunked_attention_scores[
    getitem: "f32[24, 3, 256, 257]" = hidden_states_padded_1[(slice(None, None, None), slice(None, None, None), slice(None, 256, None), slice(None, 257, None))]
    diagonal_attention_scores[(slice(None, None, None), slice(None, -1, None), slice(None, None, None), slice(256, None, None))] = getitem;  setitem = diagonal_attention_scores;  getitem = setitem = None
    return diagonal_attention_scores

hidden_states_padded_1 = torch.randn([24, 3, 512, 513], requires_grad=True)
diagonal_attention_scores = torch.randn([24, 4, 256, 513])

forward(None, hidden_states_padded_1, diagonal_attention_scores)

thunder.jit(forward)(None, hidden_states_padded_1, diagonal_attention_scores)

```

Error (partial stack trace)
```python
File "/opt/pytorch/lightning-thunder/thunder/core/transforms.py", line 2577, in augmented_forward_pass
    result, env = eval_trace(
  File "/opt/pytorch/lightning-thunder/thunder/core/trace_interpreter.py", line 63, in interpret_trace
    prim_func = symbol_mapper(symbol) if symbol_mapper is not None else symbol.sym
  File "/opt/pytorch/lightning-thunder/thunder/core/transforms.py", line 2514, in vjp_symbol_mapper
    raise NotImplementedError(f"VJP for {symbol.sym.id} is not implemented")
NotImplementedError: VJP for PrimIDs.COPY_WITH_SETITEM is not implemented
```

cc @apaz-cli

PR #1340, opened 2024-10-21
Add a test for Mistral-NeMo.

tfogal:
See issue #1285.

## What does this PR do?

Adds a test case for #1285.


thunder/tests/test_mistral_nemo.py
```diff
@@ -0,0 +1,100 @@
...
+    lr_scheduler = transformers.get_scheduler(
+        "linear",
+        optimizer=optimizer,
+        num_warmup_steps=0,
+        num_training_steps=num_epochs * len(dataloader),
+    )
+
+    # Move model to GPU if available
+    # device = torch.device("cuda") if torch.cuda.is_available() else torch.device("cpu")
+    # model.to(device)
```
crcrpar, line 78:
```suggestion
```

thunder/tests/test_mistral_nemo.py
```diff
@@ -0,0 +1,100 @@
...
+        for batch in dataloader:
+            # Move input tensors to device
+            input_ids = batch["input_ids"].to(device)
+            attention_mask = batch["attention_mask"].to(device)
+
+            # Forward pass
+            outputs = mdl(input_ids=input_ids, attention_mask=attention_mask, labels=input_ids)
+            loss = outputs.loss
+            print(loss)
+            total_loss += loss.item()
```
crcrpar, line 92: would there be a plan to use this to make sure something or not?

crcrpar (commented):
We already have https://github.com/Lightning-AI/lightning-thunder/blob/79e59d0c5c5f8aa8ef80eb31f3fe918466d64c1c/thunder/tests/test_networks.py, I thought it's nice to add this to the file instead of the new one

thunder/tests/test_mistral_nemo.py
tfogal: oops, thanks

thunder/tests/test_mistral_nemo.py
tfogal: We are still waiting for accuracy requirements from the NeMo side...
but yes, we will update this when we get those. or rip it out and do validation on the `parameters()`; it depends on how they want to define accuracy.

tfogal:
Latest version moves the code where it belongs, in test_networks.py, as Masaki pointed out.

I also rebuilt my container and that made it clear that there's a couple difficulties with this:
* I needed to turn `trust_remote_code` on to get the tiny shakespeare dataset to work
* One needs to have an active login to huggingface's hub to grab configurations
* This requires some dependencies (`transformers`, `datasets`) that we wouldn't otherwise depend on.

As such I made this `skip`ped by default. I still think it makes sense to store as a thunder test, as that's a logical source of truth for the whole team to work on. But I invite discussion.

tfogal:
@t-vi this is ready for review.

The CI failure was a node timing out running the tests, even though all this does is add a single, skipped test; seems like there's more going on there. Do you want me to add empty commits until this happens to pass or can you override that?

lantiga:
thank you @tfogal 

I don’t think we need tiny shakespeare to be downloaded, we can get away with something simpler or  random (just like with other examples), this way we don’t even have to get the tokenizer

do we actually need credentials for the configs or is it the tokenizer checkpoint that requires it?

IvanYashchuk:
> do we actually need credentials for the configs or is it the tokenizer checkpoint that requires it?

There should be a way to avoid using any credentials since the tokenizer is unnecessary and we don't load any weights for the config.

Usually, we check Thunder's ability to run a network with a sample random input and then invoke backward. A mock training loop is unnecessary here and can be avoided. Here's a patch that would verify that Thunder successfully runs the model without any dataset, optimizers, etc:
```diff
diff --git a/thunder/tests/test_networks.py b/thunder/tests/test_networks.py
index 57a9a759..9bf0cb57 100644
--- a/thunder/tests/test_networks.py
+++ b/thunder/tests/test_networks.py
@@ -362,7 +362,7 @@ def test_quantization():
 
 
 @thunder.tests.framework.requiresCUDA
-@pytest.mark.skip(reason="Dependencies, trust issues")
+# @pytest.mark.skip(reason="Dependencies, trust issues")
 def test_thunderfx_mistral_nemo_small():
     """
     Runs a small version of Mistral-NeMo
@@ -370,17 +370,9 @@ def test_thunderfx_mistral_nemo_small():
     This is largely based on code from Alexandros Koumparoulis.
     """
     import transformers
-    import datasets
 
     model_id = "mistralai/Mistral-Nemo-Base-2407"
 
-    tokenizer = transformers.AutoTokenizer.from_pretrained(
-        model_id,
-        torch_dtype=torch.bfloat16,
-        ignore_mismatched_sizes=True,
-        trust_remote_code=False,
-    )
-
     # Setup a "small" version of NeMo-Mistral that does not require downloading
     # weights. This is not a configuration that is worth benchmarking.
     # This was created by using
@@ -389,7 +381,7 @@ def test_thunderfx_mistral_nemo_small():
     #   transformers.AutoConfig.from_pretrained(model_id)
     # until they lined up.
     config = transformers.models.mistral.configuration_mistral.MistralConfig(
-        num_hidden_layers=2,
+        num_hidden_layers=1,
         torch_dtype=torch.bfloat16,
         max_position_embeddings=1024,
         architectures=["MistralForCausalLM"],
@@ -404,53 +396,18 @@ def test_thunderfx_mistral_nemo_small():
     model = transformers.AutoModelForCausalLM.from_config(config)
     device = torch.device("cuda")
     model.to(device)
-    mdl = torch.compile(model, backend=thunder.dynamo.ThunderCompiler())
+    backend = thunder.dynamo.ThunderCompiler()
+    mdl = torch.compile(model, backend=backend)
     del model
 
-    # Add a padding token to the tokenizer
-    if tokenizer.pad_token is None:
-        tokenizer.add_special_tokens({"pad_token": "[PAD]"})
-        mdl.resize_token_embeddings(len(tokenizer))
-
-    dataset = datasets.load_dataset("tiny_shakespeare", split="train", trust_remote_code=True)
-
-    def tokenize_function(examples):
-        return tokenizer(examples["text"], padding="max_length", truncation=True, max_length=2)
+    batch_size = 1
+    input_ids = torch.randint(0, config.vocab_size, (batch_size, config.max_position_embeddings), device=device)
+    attention_mask = torch.ones_like(input_ids)
 
-    tokenized_dataset = dataset.map(tokenize_function, batched=True, remove_columns=["text"])
-
-    # Convert the dataset to PyTorch format and specify columns to return as tensors
-    tokenized_dataset.set_format(type="torch", columns=["input_ids", "attention_mask"])
-
-    dataloader = torch.utils.data.DataLoader(tokenized_dataset, batch_size=1, shuffle=True)
-
-    # Define optimizer and learning rate scheduler
-    optimizer = torch.optim.AdamW(mdl.parameters(), lr=5e-5)
-    num_epochs = 3
-    lr_scheduler = transformers.get_scheduler(
-        "linear",
-        optimizer=optimizer,
-        num_warmup_steps=0,
-        num_training_steps=num_epochs * len(dataloader),
-    )
+    output = mdl(input_ids=input_ids, attention_mask=attention_mask, labels=input_ids)
+    logits = output.logits
+    grad_logits = torch.randn_like(logits)
+    logits.backward(grad_logits)
 
-    mdl.train()
-    for epoch in range(num_epochs):
-        total_loss = 0
-        for batch in dataloader:
-            # Move input tensors to device
-            input_ids = batch["input_ids"].to(device)
-            attention_mask = batch["attention_mask"].to(device)
-
-            # Forward pass
-            outputs = mdl(input_ids=input_ids, attention_mask=attention_mask, labels=input_ids)
-            loss = outputs.loss
-            total_loss += loss.item()
-
-            # Backward pass
-            optimizer.zero_grad()
-            loss.backward()
-            optimizer.step()
-
-            # Update learning rate
-            lr_scheduler.step()
+    # Check that Thunder has actually compiled the model
+    assert backend.subgraph_infos, "No subgraphs found"
```

tfogal:
> There should be a way to avoid using any credentials

Thanks Luca, Ivan. I've applied Ivan's patch + some minor other changes, and indeed it appears to not download anything new (into `~/.cache/huggingface`, at least), now.

Unfortunately in the interim something to seems to have tickled things so that #1240 is now a blocking issue, so I am leaving the `skip` designation for now :-(

kshitij12345:
#1240 has been fixed just now, so we can probably remove the skip.

thunder/tests/test_networks.py
```diff
@@ -359,3 +359,56 @@ def test_quantization():
...
+        head_dim=128,
+        _name_or_path=model_id,
+    )
+    model = transformers.AutoModelForCausalLM.from_config(config, trust_remote_code=False)
+    device = torch.device("cuda")
+    model.to(device)
+    model.train()
+    th_backend = thunder.dynamo.ThunderCompiler()
+    mdl = torch.compile(model, backend=th_backend)
+    del model
```
kshitij12345, line 402: Curious to know if this is actually needed. As per my understanding, `mdl` will use and hold onto the references for the parameters of `model`, so `del model` shouldn't reclaim a lot of memory except for clearing the `nn.Module` Python object.

Thanks!

tfogal: Indeed, not needed. I just did this as a technique for myself to make sure I didn't use `model` after compiling it; with the `del` were I to screw up and accidentally use the *uncompiled* model, Python would die with an unknown variable error; without it the script would just silently test the wrong thing.

I am happy to delete it for clarity. Since I have to re-upload to unskip, I'll fix it as part of that.

kshitij12345: Got it, thanks for clarifying

tfogal:
hrm my merge of main seemed to have not gone well. meetings now but i will fix after...

tfogal:
Hi, sorry for the weirdness. I couldn't figure it out why the github diff was wild even though `git diff main..tfogal/nemo-test-case` was sane, so I just ended up rebasing it.

It's pretty tiny (thanks to Ivan's patch), so hopefully it's not too painful to review from scratch again.

tfogal:
[CI failure](https://dev.azure.com/Lightning-AI/lightning/_build/results?buildId=218759&view=logs&j=3f274fac-2e11-54ca-487e-194c91f3ae9f&t=244491d3-5bd5-5b27-6d81-66bb4c7264ae&l=551) is real: `Found two different const extents in the same set: { bS39{1}; bS43{1}; bS66{1}; bS70{1 ex 8}; bS50{1}; bS54{1 ex 32} }` nvFuser issue.

However, it works fine with nvFuser `e33316d9480508b49db788a7472f4df52e53af92`. Do we need an nvFuser upgrade for this to work?

t-vi (approved):
The slimmed-down version looks good.
Thannk you @tfogal @IvanYashchuk @lantiga @crcrpar @kshitij12345

thunder/tests/test_networks.py
```diff
@@ -359,3 +359,54 @@ def test_quantization():
...
+    # Setup a "small" version of NeMo-Mistral that does not require downloading
+    # weights. This is not a configuration that is worth benchmarking.
+    # This was created by using
+    #   MistralConfig(num_hidden_layers=1, max_position_embeddings=1024)
+    # and then manually diffing that returned object with:
+    #   transformers.AutoConfig.from_pretrained(model_id)
+    # until they lined up sans the hidden and embeddings changes, above.
+    config = transformers.models.mistral.configuration_mistral.MistralConfig(
+        num_hidden_layers=1,
+        torch_dtype=torch.bfloat16,
```
IvanYashchuk, line 384: This setting is ignored by the model instantiation. It can be checked by inspecting for example `model.model.layers[0].mlp.gate_proj.weight.dtype`.

@riccardofelluga, when you look into what Thunder executes for this and other HF models and what is missing for performance please update this test to use bfloat16 weights.

PR #1345, opened 2024-10-24
[traceable tensor subclasses] Add proxy and transform for a callable with subclasses

crcrpar:
## What does this PR do?
- New `Proxy` for traceable tensor subclasses
- Initial support of traceable tensor subclasses, tested with a simple subclass
- Tensor factory is not supported


mruberry:
Very exciting! Would you also link a (draft?) design doc?

Issue #1314, opened 2024-10-15
ThunderFX: the output memory is not properly released for `nanogpt_model`


kiya00:
*Note*: If you have a model or program that is not supported yet but should be, please use the program coverage template.

### To Reproduce

```
import torch
import thunder
from thunder.dynamo import ThunderCompiler

import thunder.tests.nanogpt_model as nanogpt_model
from thunder.tests.make_tensor import make_tensor
import gc
import weakref
import refcycle

config = nanogpt_model.GPTConfig(dropout=0, block_size=16, n_layer=1, n_head=2, n_embd=128, bias=False)
gpt = nanogpt_model.GPT(config).to(device="cuda", dtype=torch.float32)

backend = ThunderCompiler()
compiled = torch.compile(backend=backend)(gpt)

x = make_tensor((4, 16), dtype=torch.int64, low=0, high=255, device="cuda")
targets = make_tensor((4, 16), dtype=torch.int64, low=0, high=255, device="cuda")

ref=None
# torch.cuda.memory._record_memory_history()
for i in range(4):
    result0, result1 = compiled(x, targets=targets)
    # result0, result1 = gpt(x, targets=targets)
    weakref.finalize(result0, lambda : print(f"output0 is being cleared"))
    if ref is None:
        ref = weakref.ref(result0)
# torch.cuda.memory._dump_snapshot("snapshot_thunderfx_gc.pickle")
print("end process")
print(type(ref()))
# if ref() is not None:
#     graph = refcycle.snapshot()
#     try:
#         cycle = graph.shortest_cycle(ref())
#         print("CYCLE FOUND FROM TENSOR")
#     except ValueError:
#         print("NO CYCLE FROM TENSOR")
#         pass

#     for anc in graph.ancestors(ref()):
#         try:
#             cycle = graph.shortest_cycle(anc)
#             print("CYCLE FOUND FROM ANCESTOR")
#         except ValueError:
#             pass
```
Outputs:
```
end process
<class 'torch.Tensor'>
output0 is being cleared
output0 is being cleared
output0 is being cleared
output0 is being cleared
```
### Expected behavior
expect the output to be released as eager
```
output0 is being cleared
output0 is being cleared
output0 is being cleared
end process
<class 'NoneType'>
output0 is being cleared
```

kiya00:
@kshitij12345 , maybe you have some idea?

kshitij12345:
[...]

PR #1352, opened 2024-10-28
fix leak when view of output is saved for backward in autograd.Function

kshitij12345:
Fixes https://github.com/Lightning-AI/lightning-thunder/issues/1314

PyTorch has a bug where saving a view of the output in `torch.autograd.Function` leads to memory leak. See https://github.com/pytorch/pytorch/issues/94990 and [comment](https://github.com/pytorch/pytorch/issues/94990#issuecomment-1435181804).

We side-step this issue by detaching the saved tensors before saving for backward (as mentioned as an option in [comment](https://github.com/pytorch/pytorch/issues/94990#issuecomment-1435181804)). This can lead to problem with higher order gradients but for now thunder explicitly disables it, so it should be fine.

Have added a small repro test for the same.


IvanYashchuk (approved):
Great find!

lantiga (approved):

Issue #1188, opened 2024-09-23
Recipes and high-level entrypoint


lantiga:
## 🚀 Feature

Thunder recipes and new high-level entrypoint.

This is important
- for new users, or users that just want to take advantage of thunder without getting into the how it works
- for shipping optimizations into reusable components to be utilized across codebases
- for providing model-aware optimizations out of the box

### Motivation

Providing a model to `thunder.jit` requires understanding on:
- what transforms to use and in what order
- what executors to provide and in what order
- what eventual lookasides to register

The above is specific to models and cluster configurations, so it would be good to have a way to package everything up in a reusable class, that applies to certain models or model families and can be shipped alongside the original code.

For instance, one could have a `HFLllama3` recipe, or a more general `HFLlama` recipe that can be applied to all variants of Llama. A recipe could expose options for different configurations as well, like the use of distributed.
Last, one could have a `HFLlama3Hopper` recipe that optimizes the combination of executors for a certain architecture.

In a nutshell, the recipe would orchestrate what is needed to make a thunder run on a model and what gets applied to that model. Code that use recipes would not:
- have to worry about providing the right things in the right order to `thunder.jit`
- incur in significant changes when adopting different recipes

One of the uses for a recipe is when dealing details of implementations and the way we deal with them, e.g.
- use of inplace
- particular lookasides needed to make the model work without failures or graph breaks

The possible introduction of a `ThunderFX` entrypoint makes this even more attractive. The recipe could decide to call into `thunder.jit` or `ThunderFX` according to the recipe implementor decides to go with.

### Pitch

This is how the new entrypoint and a recipe could look like. **NOTE**: the naming is up for grabs, this is just for demonstration purposes.

Here's a skeleton of a base `ThunderRecipe` class and the `thunder.compile` entrypoint

```python
import thunder
from thunder import jit, Transform, Executor


class Lookaside:
    def __init__(self, fn, replace_with):
        self._fn = fn
        self._replace_with = replace_with


class ThunderRecipe:
    def __init__(self):
        pass

    def validate(self, model):
        # this is supposed to raise
        pass

    # def setup_operators(self) -> list[Operator]:
    #     # this is for registering custom kernels on the fly
    #     return None

    def setup_lookasides(self) -> list[Lookaside]:
        return None

    def setup_transforms(self) -> list[Transform]:
        return None

    def setup_executors(self) -> list[Executor]:
        return None

    def setup_config(self):
        return {}

    def setup(self, model):
[...]

PR #1353, opened 2024-10-28
Add recipes and high level entrypoint

lantiga:
Addresses #1188

This is a very simple v0, we'll be adding models and smoothen out the API as we go.

- I've opted for one recipe at a time for now (i.e. no lists of recipes) to keep things simple.
- Also I've included a ThunderFX mode to clarify usage there.
- The single operator shortcut is still TODO
- I need to extend tests to add more coverage before this is merged


t-vi (approved):
Awesome, LGTM and see how it plays out.

PR #1359, opened 2024-10-29
Skip importing Apex if torch distributed is not available

IvanYashchuk:
Fixes https://github.com/NVIDIA/apex/issues/1853.


lantiga (approved):
Thanks @IvanYashchuk !

PR #1360, opened 2024-10-29
Add celu activation function

beverlylytle:
<details>

- [X] Did you read the [contributor guideline](https://github.com/Lightning-AI/pytorch-lightning/blob/main/.github/CONTRIBUTING.md), Pull Request section?
- [X] Did you write any new necessary tests?

</details>

## What does this PR do?

The activation function [celu](https://pytorch.org/docs/stable/generated/torch.nn.CELU.html#torch.nn.CELU) is registered as an op.

## PR review

## Did you have fun?

cc @apaz-cli


thunder/torch/__init__.py
```diff
@@ -1761,6 +1761,18 @@ def real(a):
 # TODO Move these to torch.nn.functional
 
 
+@torchsymbol(torch.celu, torch.nn.functional.celu, id="torch.celu", is_method=True)
+def celu(a: TensorLike, /, alpha: float = 1.0, inplace: bool = False) -> TensorLike:
```
mruberry, line 1765: In the PyTorch reference implementation (https://github.com/pytorch/pytorch/blob/869d1ad0b4d85e0528920949162f1bdcf6536bc6/torch/_refs/nn/functional/__init__.py#L161) `alpha` defaults to `None`:

```
alpha: None | float = None
```

This allows for a special case:

```
if alpha is None:
  rhs = torch.expm1(a)
```

In the future, we might want to add special-cases to division and multiplication so that they would be no-ops, anyway (fyi @kshitij12345, @tfogal).

beverlylytle: Right, this is a point I do not understand.  There is the reference implementation that you link to, but then there are also these two entry points: [here](https://github.com/pytorch/pytorch/blob/main/torch/nn/modules/activation.py#L568) and [here](https://github.com/pytorch/pytorch/blob/a762dc0357d6c78bbb8be942355cad21b704debe/torch/nn/functional.py#L1849).  Both of these provide a default value of 1.0 for `alpha`.  So which signature should thunder ops be mirroring?

IvanYashchuk:
This error is probably due to an unfortunate random input with values too close to 0, which makes the finite differences approximation we use to compare with our autogenerated code inaccurate. It's a rare case since we generate uniformly distributed numbers in the -9 to 9 range. Re-running the CI job might make the test pass.
Another way we approached this problem in the past is to add a `singularity_fn` to the OpInfo, this is a function that returns a distance to the instability point and it's used to push tensor values away from it.

thunder/torch/__init__.py
```diff
@@ -1761,6 +1761,18 @@ def real(a):
...
+    negative_domain_value = expm1(a / alpha) * alpha
+    out = where(a > 0, a, negative_domain_value)
+    if inplace:
+        return prims.copy_(out, a)
+    return out
+
+
+_inplace_to_out_of_place[celu] = celu, 1
```
IvanYashchuk, line 1773: @crcrpar, why is the index of the `inplace` argument needed in the code? What happens if it's specified as a keyword argument and not passed positionally? Are there tests that should fail if the index is wrong?

Hopefully Masaki can explain what's going on here but it seems from the signature the index should be 2.
```suggestion
_inplace_to_out_of_place[celu] = celu, 2
```

IvanYashchuk (approved):
Great first contribution, thank you!

thunder/torch/__init__.py
beverlylytle: I appreciate the efficiency of not doing redundant arithmetic, however I encounter the following error when following the suggestion:
```
input = tensor(3.5244, device='cuda:0', dtype=torch.float64), alpha = None, inplace = False

    def celu(
        input: Tensor,
        alpha: float = 1.0,
        inplace: bool = False,
    ) -> Tensor:  # noqa: D400,D402
        r"""celu(input, alpha=1., inplace=False) -> Tensor

        Applies element-wise,
        :math:`\text{CELU}(x) = \max(0,x) + \min(0, \alpha * (\exp(x/\alpha) - 1))`.

        See :class:`~torch.nn.CELU` for more details.
        """
        if has_torch_function_unary(input):
            return handle_torch_function(
                celu, (input,), input, alpha=alpha, inplace=inplace
            )
        if inplace:
            result = torch.celu_(input, alpha)
        else:
>           result = torch.celu(input, alpha)
E           TypeError: celu(): argument 'alpha' (position 2) must be Number, not NoneType
```

Since PyTorch does not expect `None`, I will leave it as is.

beverlylytle:
I find it less than satisfying that `celu`, which was devised exactly to be a continuously differentiable version of `relu`, should be treated as having a singularity. I'm more inclined to use a custom comparator with an increased tolerance.  I ran the test 200 times with the increased tolerances: no flakes.

t-vi (approved):
Thank you and congrats @beverlylytle 
Thank you @mruberry @IvanYashchuk

PR #1362, opened 2024-10-29
Add an option to print nvFuser repros.

tfogal:
## What does this PR do?

This adds a minor nvidia-specific debug option to print out reproduction information whenever possible.

This could be useful if there is a segfault, or if an nvFuser developer wants to drop all the fusions in a network for importing into their CI.

## PR review

### DIFF
--- a/thunder/executors/nvfuserex_impl.py
+++ b/thunder/executors/nvfuserex_impl.py
@@ -442,6 +442,7 @@ class FusionDefinitionWrapper:
     last_used: None | FusionDefinition = None
     last_inputs: None | Sequence[tuple] = None
     store_inputs: bool = False
+    print_repro: bool = False
 
     def __call__(self, *args):
         fd = self.get_fd(self.to_descriptors(args))
@@ -453,7 +454,7 @@ class FusionDefinitionWrapper:
         # Set device if set in one of the "factory" methods like full, iota, or uniform
         kwargs = {"device": fd._selected_device} if hasattr(fd, "_selected_device") else {}
         with add_markers(self.name):
-            return fd.execute(args, **kwargs)
+            return fd.execute(args, **kwargs, print_repro=self.print_repro)
 
     def __repr__(self):
         return f"FusionDefinitionWrapper({self.name})"
@@ -544,6 +545,13 @@ def create_fusion_definition_wrapper(
     store_inputs: None | bool = get_compile_option(
         "nv_store_fusion_inputs", "Allow nvFuser to store fusion inputs for repro."
     )
+    repro_help = """\
+Tells NVIDIA libraries to print out reproductions for cases. This is largely
+for creating bug reports that can be independent of the full network.
+"""
+    print_repro: None | bool = get_compile_option("nv_print_repro", repro_help)
+    print(f"PRINT repro: {print_repro}")
+    print_repro: bool = print_repro if isinstance(print_repro, bool) else False
 
     tensor_indices = []
     for idx, x in enumerate(sorted_unique_inputs):
@@ -565,6 +573,7 @@ def create_fusion_definition_wrapper(
         get_fd.cache_info,
         get_fd.cache_clear,
         store_inputs=store_inputs,
+        print_repro=print_repro,
     )
     return fdw
 
--- a/thunder/tests/test_nvfuser.py
+++ b/thunder/tests/test_nvfuser.py
@@ -1077,3 +1077,29 @@ def test_sdpa(
     ref_outputs = (ref_attn_out, *(inp.grad for inp in ref_tensor_inputs))
     for nv_out, ref_out in zip(nv_outputs, ref_outputs):
         torch.testing.assert_close(nv_out, ref_out)
+
+@thunder.tests.framework.requiresNVFuser
+def test_nvfuser_repro_print(capsys):
+    def foo(a: torch.Tensor, b: torch.Tensor) -> torch.Tensor:
+        t0 = torch.add(a, b)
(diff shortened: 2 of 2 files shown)


thunder/executors/nvfuserex_impl.py
```diff
@@ -442,6 +442,10 @@ class FusionDefinitionWrapper:
     last_used: None | FusionDefinition = None
     last_inputs: None | Sequence[tuple] = None
     store_inputs: bool = False
+    print_repro: bool = False
+    # Pre-generated so that we don't construct a LooseVersion on the critical
+    # path.
+    repro_version_needed = LooseVersion("0.2.22")
```
jjsjann123, line 448: nitpick: Does this need to be a per instance field, rather than a file scope thing?

jjsjann123 (approved):
LGTM

thunder/executors/nvfuserex_impl.py
tfogal: Good point! Thanks.

riccardofelluga:
How does differ from the util here?

https://github.com/Lightning-AI/lightning-thunder/blob/9c916d9df73f3920b51e5951303a76b25ab2d4d4/thunder/examine/__init__.py#L249

tfogal:
> How does it differ from the util here?
> 
> https://github.com/Lightning-AI/lightning-thunder/blob/9c916d9df73f3920b51e5951303a76b25ab2d4d4/thunder/examine/__init__.py#L249

It's pretty similar, just a more user-accessible way to get at the same functionality really. Inside nvFuser it calls the same code.

The intent is to turn this on when nvFuser is segfaulting and then just grab the last repro printed out for a bug report. To satisfy that case with the existing entry points, one would need to first dump the thunder trace to get all the `nvFusion<X>` names, and then iterate through and print each one.

t-vi:
I have three issues with this
- I don't like random options added to the .jit call it just turns into a giant mess over time.
- It special-cases nvfuser a lot and pretty much duplicates the existing functionality. - We print on errors and we can print after the fact.
- So the remaining use-case is segfaults. In my experience, these are really rare.

tfogal:
Thanks for taking a look!

> * So the remaining use-case is segfaults. In my experience, these are really rare.

Yes, indeed, segfaults are the main reason one would want this.

I'm happy to look at other options but it's not clear (to me) what other ones exist. Or maybe your intent is that we shouldn't do anything, and instead just manually edit thunder to print these repros when we hit such cases?

t-vi:
So lazily search segfault in the NVFuser repo issues gives 8 from this year, not all of which would apply here, maybe 3-4 that affected thunder CI. So my gut feeling is that it is exceedingly rare to hit those.

tfogal:
> So lazily search segfault in the NVFuser repo issues gives 8 from this year, not all of which would apply here, maybe 3-4 that affected thunder CI. So my gut feeling is that it is exceedingly rare to hit those.

Yes, segfaults are rare (thankfully!).

Sorry, it's not clear what behavior or action you're recommending, or what debugging flow you'd like developers to use when they hit a segfault. Could you clarify?

IvanYashchuk:
> I don't like random options added to the .jit call it just turns into a giant mess over time.

It's good to have as few options as possible for a general user to use Thunder. Going to the README of the project we would see "Thunder aims to be usable, understandable, and extensible.".  This option is not random, it aids usability and ease of reporting a problem for debugging. Unfortunately, due to the nature of active development nvFuser segfaults, while rare, are inevitable, and there can be other problems besides segfaults where the help from the nvFuser team would be needed.

> It special-cases nvfuser a lot and pretty much duplicates the existing functionality. - We print on errors and we can print after the fact.

The scope of the change is nvFuser files. nvFuser should be able to introduce any options they think are necessary for a better experience. If there would be any other executor willing to reuse the same option for their reproducer generation they can do so.

@t-vi, what would you recommend doing here?

kevinstephano:
I would not consider this option specific to segfaults.  This option enables you to generally dump nvFuser repros. I would just consider it a productivity improvement as it takes less effort to add a single knob.   In addition, this would be necessary to catch a segfault.

t-vi:
Of course, it is not random to the people implementing it, but it seems relatively random to the user. Part of the problem is the option mechanism that Thunder provides here, which basically makes options very non-discoverable.

To my mind
- options should be discoverable,
- they should be documented,
- we should avoid having many options to the jit cluttering that call
  - if it is specific to an executor or to a transform, maybe setting a flag there with a function / property is better and then
    it can be documented in the executor class documentation.
  - for debug options in particular, I created a DebugOptions class that you can register options with in #1447 . Those options will be automatically added to the doc string and to the documentation of the DebugOptions class, so people can interactively find them or look them up in the docs. We will use these for debug verbosity options in the interpreter and for enabling trace checking etc.
- This adds the third way to get nvfuser repros and to my mind, this is a bit much. I know that nvfuser is very dear to you, but still. If an option like the present one is adopted, will we drop @riccardofelluga 's way of doing it? Printing the repro on error seems to be great value, but why do we need to have two options to print it when no error happens?

t-vi:
@tfogal is there still interest in this? If you could update to use the debug option facility, I'd be glad to merge if you want to.

tfogal:
> @tfogal is there still interest in this? If you could update to use the debug option facility, I'd be glad to merge if you want to.

Yes, thanks for the reminder. I will update it to use the DebugOption mechanism.

t-vi:
Last activity was Jan 10, > 6 months ago. Let's close and reopen when we're prepared to finish this.

PR #1363, opened 2024-10-30
Try removing `_safe_functions` from `jit_ext`

crcrpar:
## What does this PR do?
as per title, it seems that the variable is not used


t-vi (approved):
Thank you @crcrpar

Issue #1232, opened 2024-10-01
AssertionError: t54580_out for rematerialisation for Gemma-2-27b and other models.


mpatel31415:
## 🐛 Bug
When running benchmarking script with `--checkpoint_activations True` we get:

> AssertionError: t54580_out for rematerialisation 

This issue is present for the following models:
'Llama-3-70B', 'Gemma-2-27b', 'longchat-13b-16k', 'Mistral-7B-v0.2', 'vicuna-7b-v1.5-16k', 'Llama-2-13b-hf', 'CodeLlama-34b-hf'

### To Reproduce

Please use:
1 node(s), each with 8 GPUs.
Image "INTERNAL_IMAGE:pjnl-20240930"
Training script:
python /opt/pytorch/lightning-thunder/thunder/benchmarks/benchmark_litgpt.py \
    --model_name Gemma-2-27b \
    --distributed_mode fsdp \
    --shard_mode zero2 \
    --compile thunder \
    --checkpoint_activations True \
    --low_precision_mode none  \
    --micro_batch_size 1

### Expected behavior

We should be able to run the training.

### Environment

"system.device_product_name                      DGXH100
system.gpu_driver_version                    535.129.03
libraries.cuda                               12.6.2.004
libraries.pip.lightning               2.4.0.dev20240728
libraries.pip.lightning-thunder              0.2.0.dev0
libraries.pip.lightning-utilities                0.11.7
libraries.pip.litgpt                             0.4.11
libraries.pip.nvfuser                 0.2.13+git2cee59d
libraries.pip.pytorch-lightning                   2.4.0
libraries.pip.torch                  2.6.0a0+gitc4ae451
libraries.pip.torchmetrics                        1.4.2
libraries.pip.torchvision              0.19.0a0+d23a6e1"

riccardofelluga:
As an update to this issue, after the move to the networkx library for finding a minimal graph cut in rematerialization the error  looks like the following:

``` shell
  File "/opt/pytorch/lightning-thunder/thunder/__init__.py", line 649, in get_computation_and_inputs
    computation_trc, backward_trc = split_forward_backward(computation_trc, cd, cs, *inps)
  File "/opt/pytorch/lightning-thunder/thunder/executors/torch_autograd.py", line 231, in split_forward_backward
    bw_extrace = transform_for_execution(
  File "/opt/pytorch/lightning-thunder/thunder/executors/passes.py", line 160, in transform_for_execution
[...]

PR #1367, opened 2024-10-30
Rename rematerialization of saved for backward symbols

riccardofelluga:
This is part of #1232. PR renames the outputs of recomputed symbols so that they do not overlap with names used in the forward trace. Fusion rematerialization requires names used in producer and consumer fusions to be unique.


thunder/core/transforms.py
```diff
@@ -3189,10 +3196,11 @@ def recompute_saved_for_backward(fwd_trace: Trace, bwd_trace: Trace) -> tuple[Tr
...
             new_bwd_trace.bound_symbols.append(new_unpack)
         elif idx == 6:
-            new_bwd_trace.bound_symbols.extend(producers)
-            new_bwd_trace.bound_symbols.append(bsym)
+            for p in producers:
+                new_bwd_trace.bound_symbols.append(p.from_bsym_swap_proxies(swap_map=swap_map))
+            new_bwd_trace.bound_symbols.append(bsym.from_bsym_swap_proxies(swap_map=swap_map))
         else:
-            new_bwd_trace.bound_symbols.append(bsym)
+            new_bwd_trace.bound_symbols.append(bsym.from_bsym_swap_proxies(swap_map=swap_map))
```
riccardofelluga, line 3203: Hmm apparently torch compile backend breaks with the rename:
```shell
 File "/opt/pytorch/lightning-thunder/thunder/core/langctxs.py", line 136, in _fn
    result = fn(*args, **kwargs)
  File "/opt/pytorch/lightning-thunder/thunder/__init__.py", line 233, in cache_info_wrapper
    res = fn(*args, **kwargs)
  File "/opt/pytorch/lightning-thunder/thunder/__init__.py", line 657, in get_computation_and_inputs
    computation_trc, backward_trc = split_forward_backward(computation_trc, cd, cs, *inps)
  File "/opt/pytorch/lightning-thunder/thunder/executors/torch_autograd.py", line 250, in split_forward_backward
    bw_extrace = transform_for_execution(
  File "/opt/pytorch/lightning-thunder/thunder/executors/passes.py", line 160, in transform_for_execution
    extrace = ex.fusion_pass(extrace)
  File "/opt/pytorch/lightning-thunder/thunder/executors/torch_compile.py", line 194, in fusion_pass
    fusedtrace = dce(fusedtrace)
  File "/opt/pytorch/lightning-thunder/thunder/core/transform_common.py", line 121, in dce
    flat_trace_outputs, _ = tree_flatten(trace.output)
  File "/opt/pytorch/lightning-thunder/thunder/core/trace.py", line 240, in output
    assert False, "Trace has no return, and so no output"
AssertionError: Trace has no return, and so no output
```

Even tho the trace does pass the asserts earlier and therefore it contains a return before going to the fusion pass. Let's fix that before merging this

IvanYashchuk (changes requested):
Other reviewers, please don't merge this PR without my review.

thunder/core/transforms.py
```diff
@@ -3148,6 +3148,9 @@ def recompute_saved_for_backward(fwd_trace: Trace, bwd_trace: Trace) -> tuple[Tr
 
     producers = find_producer_symbols(fwd_trace, tuple(unvariableify(i) for i in rematerializable), fwd_trace.args)
 
+    trace_tok = set_tracectx(bwd_trace)
```
IvanYashchuk, line 3151: Please set and reset traces only with "try: finally:" blocks. If there's any error between the calls, the trace will not be reset.

thunder/core/transforms.py
```diff
@@ -3156,8 +3159,12 @@ def recompute_saved_for_backward(fwd_trace: Trace, bwd_trace: Trace) -> tuple[Tr
             if prod_arg in fwd_trace_args:
                 required_fw_args.add(prod_arg)
         for prod_out in prod.flat_outs:
+            if isinstance(prod_out, TensorProxy):
+                swap_map[variableify(prod_out)] = prod_out.replace_name(f"remat_for_{prod_out.name}")
```
IvanYashchuk, line 3163: What does "remat_for" indicate in a full sentence? I think using "of" is a more suitable preposition here.

Could you please update the `replace_name` method to add this name to the current trace:
https://github.com/Lightning-AI/lightning-thunder/blob/1de5dd81d80a6c6036eeb9f77aa1116cb429fbc3/thunder/core/proxies.py#L178

thunder/core/transforms.py
IvanYashchuk: Why do you set the input `bwd_trace` as the active trace? There are no Thunder operations calls between set and reset, and the input trace shouldn't be modified.

thunder/core/transforms.py
IvanYashchuk: Could you please add a test first before attempting to fix the issue? CI hasn't detected anything.

thunder/core/transforms.py
t-vi: would we not use `with tracectx(bwd_trace)`?

riccardofelluga: For this I've taken inspiration from the code in the torch_autograd executor, in particular these lines explain why the need to set the trace context: https://github.com/Lightning-AI/lightning-thunder/blob/3390c922cfbe1b70c42118dfa8aa71adb3bec692/thunder/executors/torch_autograd.py#L33-L40 

@IvanYashchuk Would an acceptable workaround be to create a new empty trace and use it as ctx?

thunder/core/transforms.py
riccardofelluga: What do you mean by that? Are you suggesting to change the `replace_name` method by adding a default value "remat_of_" ?

thunder/core/transforms.py
IvanYashchuk: Yes, or allow creating Proxies with any name without active tracectx. Maybe all is needed is to return True if trc is None in this function https://github.com/Lightning-AI/lightning-thunder/blob/3390c922cfbe1b70c42118dfa8aa71adb3bec692/thunder/core/proxies.py#L75-L82

thunder/core/transforms.py
riccardofelluga: It doesn't seem to be a problem anymore for two main reasons:
- saved for backward remat is not in the code path anymore when using torch compile for the example that was showing up the issue.
- Can't reproduce the problem anymore even when using ThunderFX.

At this point  I'd rely on the test and mark this as solved for now.

t-vi:
@riccardofelluga I think this is  superseded by doing the renaming in the rematerialization itself. 
Please reopen if I am missing something (entierly possible).

Issue #1233, opened 2024-10-01
OOM for training on 4 gpus for falcon-40b' and  'vicuna-33b-v1.3'


mpatel31415:
## 🐛 Bug

This might be related to [old OOM issue](https://github.com/Lightning-AI/lightning-thunder/issues/474), but the models and # nodes is different, so I decided to create another issue. 
We get OOM error, but torch.compile can run the models successfully. 

### To Reproduce

Please use:
4 node(s), each with 8 GPUs.
Image "INTERNAL_IMAGE:pjnl-20240930"
Training script:
python /opt/pytorch/lightning-thunder/thunder/benchmarks/benchmark_litgpt.py \
    --model_name falcon-40b \
    --distributed_mode fsdp \
    --shard_mode zero3 \
    --compile thunder \
    --checkpoint_activations False \
    --low_precision_mode none  \
    --micro_batch_size 1

### Expected behavior

We should be able to run the benchmarking script.

### Environment

system.device_product_name                      DGXH100
system.gpu_driver_version                    535.129.03
libraries.cuda                               12.6.2.004
libraries.pip.lightning               2.4.0.dev20240728
libraries.pip.lightning-thunder              0.2.0.dev0
libraries.pip.lightning-utilities                0.11.7
libraries.pip.litgpt                             0.4.11
libraries.pip.nvfuser                 0.2.13+git2cee59d
libraries.pip.pytorch-lightning                   2.4.0
libraries.pip.torch                  2.6.0a0+gitc4ae451
libraries.pip.torchmetrics                        1.4.2
libraries.pip.torchvision              0.19.0a0+d23a6e1

IvanYashchuk:
falcon-40b seems to use `parallel_residual=True` so I expect this model config to work once https://github.com/Lightning-AI/lightning-thunder/issues/1175 (and https://github.com/Lightning-AI/lightning-thunder/issues/474) is resolved.

mpatel31415:
We still see OOM for these models. I'm not sure if this is relevant, bu right now OOM is preceded by

> [ERROR    | nvfuser            ]: An error occurred while executing nvFuser FusionDefinition 8.

IvanYashchuk:
Yes, it's probably the same OOM error just happening while nvFuser tries to allocate more memory than available, slightly different place in the code.

Issue #1175, opened 2024-09-20
Thunder seems to use way more memory for longer sequences than others


crcrpar:
## 🐛 Bug

When input sequences get longer, Thunder seems to tend to use more memory than eager and torch.compile.

Let's take litgpt's `stablecode-completion-alpha-3b` as an example whose sequence length (`Config.block_size`) is 16384.

With the following table, Thunder's memory consumption can be more prone to sequence length

sequence length | Thunder | Torch Compile | Diff
-- | -- | -- | --
16384 | 77.02 | 71.73 | 5.29
12288 | 61.9 | 57.94 | 3.96
8192 | 46.85 | 44.2 | 2.65
4096 | 31.78 | 30.46 | 1.32
2048 | 24.28 | 23.63 | 0.65

### To Reproduce

Apply a diff like this and run commands like 

```
python thunder/benchmarks/benchmark_litgpt.py --model_name stablecode-completion-alpha-3b --warmup_iters 0 --max_iters 3 --compile eager --dump_memory_snapshot false --block_size 2048
```

```diff
@@ -227,6 +269,7 @@ class Benchmark_litGPT:
         fsdp_bucket_params: float | None = None,
         checkpoint_activations: bool = False,
         n_layers: int | None = None,
+        block_size: int | None = None,
         profiler_start: int = 15,
         profiler_stop: int = 15,
         skip_data_sync: bool = False,
@@ -360,6 +403,8 @@ class Benchmark_litGPT:

         if n_layers is not None:
             self.config.n_layer = n_layers
+        if block_size is not None:
+            self.config.block_size = block_size

         # Initialize the model
         t0 = time.perf_counter()
```

### Environment

pjnl-20240919

nvMelissa:
Add to Q4 planning

crcrpar:
The cause seems to be basically `parallel_residual=True` as in https://github.com/Lightning-AI/lightning-thunder/issues/246#issuecomment-2302121789

crcrpar:
Script to run `litgpt.model.Block` with the config of "stablecode-completion-alpha-3b" whose `parallel_residual` by default is `True`.
```python
import argparse
import gc

import torch
from litgpt import Config, GPT
from litgpt.model import Block, CausalSelfAttention, GptNeoxMLP, LLaMAMLP

import thunder


def init_model(config: Config, compiler: str, device: torch.device, dtype: torch.dtype) -> GPT:
[...]

PR #1371, opened 2024-10-30
Allow linear and matmul to be consumed by nvFuser by default

IvanYashchuk:
This change lowers peak memory usage of LitGPT implementations that use `mlp_class_name="GptNeoxMLP"` configuration (https://github.com/Lightning-AI/lightning-thunder/issues/1175, https://github.com/Lightning-AI/lightning-thunder/issues/1233, https://github.com/Lightning-AI/lightning-thunder/issues/246).

Updates from Jan 27th:
* `enable_linear=True` works as expected only together with allowing "shape only" fusions which were disabled in https://github.com/Lightning-AI/lightning-thunder/pull/1559, without reverting #1559 there's no change in peak used memory.
* `enable_linear=True` together with `enable_matmul=True` give substantial memory savings both for GptNeoxMLP- and LLaMAMLP-based models.

|           config (mlp_class_name)           |     Current main    | enable_linear + shape-only fusions | enable_linear + enable_matmul |
|:-------------------------------------------:|:---------------------:|:----------------------------------:|-------------------------------|
| stablecode-completion-alpha-3b (GptNeoxMLP) | 894.26 ms \| 76.99 GB | 894.84 ms \| 74.30 GB (1.036x)              | 919.44 ms \| 58.10 GB (1.325x)        |
| Llama-2-7b-hf (LLaMAMLP)                    | 342.67 ms \| 64.18 GB | 349.15 ms \| 63.11 GB (1.016x)              | 350.66 ms \| 57.55 GB (1.115x)        |

Old results from Oct 30th: 
| config (mlp_class_name)                     | Current main                | This PR               |
|---------------------------------------------|-----------------------|-----------------------|
| stablecode-completion-alpha-3b (GptNeoxMLP) | 889.68 ms \| 77.02 GB | 892.37 ms \| 74.33 GB |
| Llama-2-7b-hf (LLaMAMLP)                    | 336.06 ms \| 64.22 GB | 340.86 ms \| 64.18 GB |

Better memory usage comes from simplifying the setup for Thunder's fusion rematerialization. With this change, there are fewer "producer" fusions.

Commands for reproducing the results:
```
python thunder/benchmarks/benchmark_litgpt.py --model_name=stablecode-completion-alpha-3b --compile=dynamo_thunder
python thunder/benchmarks/benchmark_litgpt.py --model_name=Llama-2-7b-hf --compile=dynamo_thunder
```

cc @Priya2698, @wujingyue


wujingyue:
There are multiple concerning CI failures: 
1. `FAILED thunder/tests/test_nvfuser.py::test_cse_rematerialization_nvfuser_cuda_None - assert 3 == 11` is likely due to a behavior change in rematerialization that should be reflected as well in the test. 
2. `FAILED thunder/tests/test_jit_general.py::test_litgpt_variants_kvcache[cuda-llama1-like] - RuntimeError: !detect_exception_in_thread_pool.load() INTERNAL ASSERT FAILED at "/workspace/Fuser/csrc/kernel_cache.cpp":1234, please report a bug with repro script to NVFuser at https://github.com/NVIDIA/Fuser/issues. Detected exception while compiling fusion segments in parallel. Error messages from all threads are printed below.` sounds like a bug in nvFuser. 

```
Error from segmentation group 4: producer->getMemoryType() == MemoryType::Global INTERNAL ASSERT FAILED at "/workspace/Fuser/csrc/device_lower/analysis/sync_information.cpp":699, please report a bug with repro script to NVFuser at https://github.com/NVIDIA/Fuser/issues. Inconsistent parallelization found between TV30 (T30_l[ iblockIdx.x284{( ceilDiv(( ceilDiv(( ceilDiv(( 1 * 16 ), 128) ), 1) ), 1) )}, iUS285{1}, iS283{1}, ithreadIdx.x281{128} ]) and TV25(T25_l[ iblockIdx.x232{( ceilDiv(( ceilDiv(( ceilDiv(( 1 * ( 1 * ( 1 * 16 ) ) ), 128) ), 1) ), 1) )}, iUS233{1}, iS231{1}, ithreadIdx.x229{128} ] ca_pos( 4 )). Producer is required to be in Global Memory based on parallelization strategy. RAW flags: (blockIdx.x threadIdx.x)
```

thunder/executors/nvfuserex_impl.py
```diff
@@ -2241,6 +2241,7 @@ def map_inside_replacement(x: Any) -> None:
 
 def _linear_check(a: TensorProxy, b: TensorProxy, bias: TensorProxy | None) -> bool:
     enable_linear: None | bool = get_compile_option("nv_enable_linear", "Enable nvFuser linear.")
+    enable_linear = enable_linear if enable_linear is not None else True
```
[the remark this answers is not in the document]
IvanYashchuk: As you wish 🙂

Priya2698:
We did not run Thunder benchmarks using nvfuser linear. Should we run other benchmarks as well before enabling it by default?

Additionally, @wujingyue needed to remove support for 1D weights to facilitate DID-aware execution. We might have to add an additional check on Thunder side or use `unsqueeze-squeeze` operators to support it. I believe we currently do not have cases exercising 1D weights in Thunder so this should not break anything right now.

crcrpar (approved):
orthogonal to this PR but related to nvfuser knobs: are there `enable_foobar`s other than linear and matmul?

IvanYashchuk:
> are there `enable_foobar`s other than linear and matmul?

Yes, there's also `nv_enable_sdpa`. I didn't enable matmul in this PR because it leads to excessive memory usage in my tests. I haven't tried `nv_enable_sdpa` I thin we want to continue using the one from cuDNN for performance.

thunder/executors/nvfuserex_impl.py
```diff
@@ -2241,6 +2241,8 @@ def map_inside_replacement(x: Any) -> None:
 
 def _linear_check(a: TensorProxy, b: TensorProxy, bias: TensorProxy | None) -> bool:
     enable_linear: None | bool = get_compile_option("nv_enable_linear", "Enable nvFuser linear.")
+    if enable_linear is None:
+        enable_linear = True
     if not enable_linear:
         return False
```
riccardofelluga, line 2247: Idk, this doesn't look that nice to me :(

IvanYashchuk: It was decided here https://github.com/Lightning-AI/lightning-thunder/pull/1371#discussion_r1823075638. Only one style change per PR per line.

crcrpar (approved):

wujingyue:
CI with torch-nightly is now passing with https://github.com/NVIDIA/Fuser/issues/3369 fixed. CI with older versions of torch (and therefore older versions of nvFuser) still fail, because we can fix a past version. @IvanYashchuk, can you up nvFuser's version and enable linear only for that?

IvanYashchuk:
Yes, I will do that.

wujingyue (approved):

wujingyue:
I'm unsure about the lightning-thunder (ipynb) test. Is that an infra failure or indeed a regression?

IvanYashchuk:
For some reason, there was a timeout in the cell execution. Let's try rerunning the failed job.

IvanYashchuk:
@kiya00, could you please check what's going on with the Notebooks CI job?

t-vi (dismissed):
Thank you for this PR, but if this means needing to add correctness exceptions to half a dozen models, and the savings on Llama are marginal, we should not do this by default and instead use recipes to enable it where desired.

mruberry:
> Thank you for this PR, but if this means needing to add correctness exceptions to half a dozen models, and the savings on Llama are marginal, we should not do this by default and instead use recipes to enable it where desired.

Which correctness exceptions are you referring to, @t-vi? I agree we can't enable something by default if it causes timeouts in some models, but that's just a speed motivation. 

@kiya00 Is there a bug for the slowness issue? @kevinstephano, do you think it's worth investigating?

@kshitij12345, @IvanYashchuk, @jjsjann123, should we consider requiring nv fusions have enough nodes to be valuable (2+?) and not so many nodes they might be slow (< 30?). Maybe we could develop better heuristics in the future

kevinstephano:
> @kiya00 Is there a bug for the slowness issue? @kevinstephano, do you think it's worth investigating?
While there is a real issue, between 20 to 30 matmuls in a fusion that leads to an explosion in segmentation time, I am not sure how often we are going to see so many matmuls in one fusion.  It would be interesting to see if this comes up in actual models.

t-vi:
I think it would be good to re-gather the memory impact (baseline and with PR) before merging.
We should see much better memory footprint in the baseline, potentially rendering this PR moot.

IvanYashchuk:
> I think it would be good to re-gather the memory impact (baseline and with PR) before merging. We should see much better memory footprint in the baseline, potentially rendering this PR moot.

Sure, I will do that.

IvanYashchuk:
> I think it would be good to re-gather the memory impact (baseline and with PR) before merging. We should see much better memory footprint in the baseline, potentially rendering this PR moot.

The baseline numbers didn't change much from October (77.02 GB -> 76.99 GB, 64.22 GB -> 64.18 GB). 

`enable_linear=True` doesn't give any memory or performance changes by itself, it requires reverting changes from https://github.com/Lightning-AI/lightning-thunder/pull/1559, but the good news is that using `enable_linear=True` and `enable_matmul=True` together give great results now (it wasn't the case in Oct).

Before merging this change, I want to check the impact on all models we track.

t-vi (commented):
Memory savings sound great! Thank you @IvanYashchuk

IvanYashchuk:
There is no benefit in throughput (memory savings are not enough to increase the batch size for better throughput) for a larger set of models so not pursuing further as the default. In addition, we can't rely on sending linear and matmul to nvFuser when using the TransformerEngine executor.

Issue #1372, opened 2024-10-30
Upgrade nvFuser in CI


tfogal:
## 🚀 Feature

Please upgrade nvFuser to 0.2.22 or later.

### Motivation

We are [hitting bugs](https://github.com/Lightning-AI/lightning-thunder/pull/1340/checks?check_run_id=32239953960) that are fixed in more recent versions of nvFuser, and this is blocking #1340 at least.

### Additional context

The command you'd want to use to do this will depend on which version of torch and CUDA you are using.

* CUDA 12.4, torch 2.4: `python3 -m pip install --extra-index-url https://pypi.nvidia.com --pre nvfuser-cu124-torch24`
* CUDA 12.1, torch 2.3: `python3 -m pip install --extra-index-url https://pypi.nvidia.com --pre nvfuser-cu121-torch23`
* CUDA 12.4 nightly torch: `python3 -m pip install --extra-index-url https://pypi.nvidia.com --pre nvfuser-cu124`

t-vi:
I'll take care of it.

t-vi:
Note that the NVIDIA PyPI does not seem to have recent packages built against stable PyTorch, so there is little we can do without updates to the packages (torch25 might be interesting, too)...

e.g. https://pypi.nvidia.com/nvfuser-cu121-torch24/

xwang233:
Sorry about the delay there. We decided to push nvfuser build against torch stable to pypi.org _twice a month_ (1st and 15th).

Nvfuser build against torch nightly will still be pushed _nightly_ to pypi.nvidia.com every day.

I'll trigger a build for nvfuser build against torch stable 2.5 now to unblock CI upgrades.

xwang233:
Also, the nightly and stable builds now supports (python 3.10, and **3.12**) x (cuda 11.8, 12.1, and 12.4)

jjsjann123:
> Nvfuser build against torch nightly will still be pushed _nightly_ to pypi.nvidia.com every day.

do you mind updating the installation instruction in TomF's comment at the top? (for build against stable torch from pypi.org)

t-vi:
We currently take nvfuser from PyPI, but I don't seem to see recent updates, e.g in https://pypi.org/project/nvfuser-cu121-torch24/#history

xwang233:
> We currently take nvfuser from PyPI, but I don't seem to see recent updates, e.g in https://pypi.org/project/nvfuser-cu121-torch24/#history

[...]

PR #1373, opened 2024-10-30
bump pt to 2.5.1

t-vi:
Fixes: #1372

based on the discussion in #1372

- pypi.python.org is actually the right repo for nvfuser built for stable PT,
- as of now, we have -torch25 packages there and the plan is to update these regularly (and it's not so clear with -toirch24).

### DIFF
--- a/.azure/docker-build.yml
+++ b/.azure/docker-build.yml
@@ -40,8 +40,8 @@ jobs:
       #maxParallel: "3"
       matrix:
         # CUDA 12.1
-        "cuda 12.1 | torch 2.4.0 | cudnn FE v1.5.2":
-          { CUDA_VERSION: "12.1.1", TORCH_VERSION: "2.4.0", TRITON_VERSION: "3.0.0", CUDNN_FRONTEND_VERSION: "1.5.2" }
+        "cuda 12.1 | torch 2.5.1 | cudnn FE v1.5.2":
+          { CUDA_VERSION: "12.1.1", TORCH_VERSION: "2.5.1", TRITON_VERSION: "3.0.0", CUDNN_FRONTEND_VERSION: "1.5.2" }
         "cuda 12.1 | torch 2.5 /nightly | cudnn FE v1.5.2":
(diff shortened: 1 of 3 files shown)


dockers/ubuntu-cuda/Dockerfile
```diff
@@ -160,7 +160,7 @@ RUN \
         # installing pytorch from wheels \
         CUDA_VERSION_MM=${CUDA_VERSION%.*} && \
         TORCH_VERSION_MM=${TORCH_VERSION%.*} && \
-        pip install -U "nvfuser-cu${CUDA_VERSION_MM/./}-torch${TORCH_VERSION_MM/./}" ; \
+        pip install -U --extra-index-url https://pypi.nvidia.com --pre "nvfuser-cu${CUDA_VERSION_MM/./}-torch${TORCH_VERSION_MM/./}" ; \
```
Borda, line 163:
```suggestion
        pip install -U --extra-index-url https://pypi.nvidia.com "nvfuser-cu${CUDA_VERSION_MM/./}-torch${TORCH_VERSION_MM/./}" ; \
```
Is the Pre needed? Does this mean you are using unstable versions?

.azure/docker-build.yml
```diff
@@ -110,6 +110,6 @@ jobs:
           echo $(imageRepository):$(imageTag)
           echo $(DOCKERHUB_PAT) | docker login --username $(DOCKERHUB_USER) --password-stdin
           docker push $(imageRepository):$(imageTag)-dev
-        condition: ne(variables['Build.Reason'], 'PullRequest')
+        #condition: ne(variables['Build.Reason'], 'PullRequest')
```
Borda, line 113:
```suggestion
        condition: ne(variables['Build.Reason'], 'PullRequest')
```

Borda (approved):
`nvfuser_cu121_torch24    0.2.9.dev20240808`

dockers/ubuntu-cuda/Dockerfile
t-vi: nvfuser only has pre-releases.

PR #1374, opened 2024-10-30
Print info on saved for backward tensors for LitGPT benchmarking script

IvanYashchuk:
This PR adds to the benchmark summary information on the number of tensors that are saved for backward and how much memory they consume.
It is useful for comparing Thunder and the default torch.compile.


riccardofelluga (approved):
LGTM

t-vi (approved):
Thank you @IvanYashchuk @riccardofelluga

PR #1375, opened 2024-10-30
LitGPT benchmarking: Use block FSDP bucketing for Eager, Inductor, ThunderFX by default

IvanYashchuk:
Using [`transformer_auto_wrap_policy`](https://github.com/pytorch/pytorch/blob/ef380f7b8e6419909ea3c20e2d6506db45d76463/torch/distributed/fsdp/wrap.py#L307) is recommended for performant FSDP in PyTorch. The previous default value wasn't choosing this type of applying FSDP (and consequently bucketing parameters together based on the FSDP group) and it was easy to miss when doing performance comparisons.

cc @crcrpar


crcrpar (approved):

t-vi (approved):
Thank you @IvanYashchuk @crcrpar

PR #1389, opened 2024-11-01
[visualization] update make_trace_dot

kshitij12345:
Builds on great feature introduced by @riccardofelluga 

Changes 
* We add a seperate node for input/output to a Symbol (with option to add more metadata)
* Symbol node prints the python repr to show more information (currently we print the symbol name)
* Update color scheme - `Roots nodes are colored "green", intermediates are colored "lightblue" and leaves are colored "orange"`

```python
import thunder
from thunder.examine import make_trace_dot
import torch

def foo(x, y):
    return x + y / (x * y)

jfoo = thunder.jit(foo)

jfoo(torch.randn(3, requires_grad=True), torch.randn(3,))

t = thunder.last_traces(jfoo)[-3]
dot = make_trace_dot(t)
dot.render(format='png')

```

Output

Forward Graph
image

Backward Graph
image

### DIFF
--- a/thunder/examine/__init__.py
+++ b/thunder/examine/__init__.py
@@ -365,7 +365,7 @@ def make_trace_dot(trace: TraceCtx, show_metadata=False):
                 dot.node(arg_id, _repr_tensor_proxy(arg, show_metadata))
                 dot.edge(arg_id, str(node_id))
 
-        # Add node for outputs and connect outputs
+        # Connect outputs
         for out in node.bsym.flat_outs:
             if isinstance(out, TensorProxy):
                 out_id = out.name
--- a/thunder/examine/__init__.py
+++ b/thunder/examine/__init__.py
@@ -276,7 +276,28 @@ def get_nvfuser_repro(trace: TraceCtx, fusion_name: str, /) -> str:
     return get_repro(fusion.last_inputs)
 
 
-def make_trace_dot(trace: TraceCtx):
+# Copied from `pytorchviz` which has MIT License (Thank you!)
+# See https://github.com/szagoruyko/pytorchviz/blob/0adcd83af8aa7ab36d6afd139cabbd9df598edb7/torchviz/dot.py#L180
+def resize_graph(dot, size_per_element=0.15, min_size=12):
+    """Resize the graph according to how much content it contains.
+
+    Modify the graph in place.
+    """
+    # Get the approximate number of nodes and edges
+    num_rows = len(dot.body)
+    content_size = num_rows * size_per_element
+    size = max(min_size, content_size)
+    size_str = str(size) + "," + str(size)
+    dot.graph_attr.update(size=size_str)
+
+
+def _repr_tensor_proxy(t_proxy, show_metadata=False):
+    assert isinstance(t_proxy, TensorProxy)
+    extra_meta = f"\n shape:{t_proxy.shape} \n dtype:{t_proxy.dtype}" if show_metadata else ""
+    return f"name:{t_proxy.name}" + extra_meta
+
+
+def make_trace_dot(trace: TraceCtx, show_proxy_metadata=False):
     """
     Creates a directed graph of the given trace.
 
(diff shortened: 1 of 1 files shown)


thunder/examine/__init__.py
```diff
@@ -308,25 +332,51 @@ def make_trace_dot(trace: TraceCtx):
...
     while stack:
-        node: Node = stack.pop()
+        node: Node = stack.pop(0)
         node_id = id(node)
         visited.add(node_id)
-        dot.node(str(node_id), node.bsym.sym.name, fillcolor="orange" if node_id in leaves_id else "white")
-
+        color = _get_color(node_id)
+        dot.node(str(node_id), node.bsym.python(indent=0, print_depth=1)[0], fillcolor=color)
+
```
riccardofelluga, line 360: I think here there is the need to do some sanitizing. When trying to make a graph of a model's bwd trace I get `[...]/graphviz/quoting.py:86: expect syntax error scanning invalid quoted string: 'cos, idx, sin, t_lm_head_weight, t_transformer_h_0_attn_attn_weight, \\'`

thunder/examine/__init__.py
```diff
@@ -308,25 +332,51 @@ def make_trace_dot(trace: TraceCtx):
...
+        # Add node for args and connect args
+        for arg in node.bsym.flat_args:
+            if isinstance(arg, TensorProxy):
+                arg_id = arg.name
+                dot.node(arg_id, _repr_tensor_proxy(arg, show_metadata))
```
riccardofelluga, line 365: My concern here is that with big graphs adding the node increases the time that it takes to render out the graph because it will try to layout more nodes that it is now. It's ok like this but I would want to try it first

riccardofelluga (changes requested):
Minor things but overall it's a great addition!

thunder/examine/__init__.py
kshitij12345: Can you share the snippet where you see this, for a small snippet it seems to work fine. Thanks!

riccardofelluga: here is the diff where I introduced the `make_trace_dot` in the `benchmark_litgpt.py` file

```diff
diff --git a/thunder/benchmarks/benchmark_litgpt.py b/thunder/benchmarks/benchmark_litgpt.py
index 2aa78cbe..8ab1e86d 100644
--- a/thunder/benchmarks/benchmark_litgpt.py
+++ b/thunder/benchmarks/benchmark_litgpt.py
@@ -747,6 +747,11 @@ class Benchmark_litGPT:
             self.perf_metrics["saved_for_backward_tensor_size_mib"] = saved_tensors_size_in_mib
             self.perf_metrics["saved_for_backward_number_of_tensors"] = saved_tensors_len
 
+        from thunder.examine import make_trace_dot
+        bwd_trace = thunder.last_backward_traces(self.model)[-1]
+        dot = make_trace_dot(bwd_trace)
+        dot.render(format='pdf', filename='bwd_trace_graph')
+
     def add_perf_metrics(self):
         if self.throughput:
             metrics = self.throughput.compute()
```

and called the script with:

```shell
python thunder/benchmarks/benchmark_litgpt.py --model_name vicuna-7b-v1.5-16k --compile thunder  --max_iters 3 --warmup_iters 1 --n_layers 2
```

thunder/examine/__init__.py
IvanYashchuk: Creating a graph is a tool to aid understandability or debugging. It's okay if the graph generation is slow.

thunder/examine/__init__.py
kshitij12345: Have updated the patch to handle this.

thunder/examine/__init__.py
kshitij12345: I found a few options to improve the graph rendering times thought at the cost of more clumsy placement for nodes.

```python
dot = make_trace_dot(t)

# Update `nlimit` for less iterations of layout algorithm.
dot.graph_attr["nslimit"] = "5"
dot.graph_attr["nslimit1"] = "5"

# Or maybe use a different engine.
dot.render(format='png')
```

I don't know exactly how much they affect the times (and what are a good value) but we can potentially document it. Wdyt? Or maybe think about it in a follow-up.

Ref:
https://graphviz.org/docs/attrs/nslimit/
https://stackoverflow.com/questions/64130276/graphviz-is-too-slow

riccardofelluga: Yeah I also experimented with those options at the start and they result in messier graphs. But I think using `nslimit` in worst case scenarios it's a good compromise! 

@IvanYashchuk we have the same idea here, tho I have some more context from playing around with graphviz and with the default renderer. I've seen it take 30 minutes to render out complex graphs so I just wanted to share the awareness on this and of course if it takes 2 minutes it's fine. So it's worth testing in my opinion

riccardofelluga: I've quickly measured and it compares like this for the example of the bwd trace: 142,56s vs 58,62s of the previous version in main. Maybe we should have a "fast mode" that sets the nslimit and then the speedup is significant (2,79 seconds on same example!).

riccardofelluga (approved):
It's a good idea to leave the rendering to the user at they will have the dot object anyway. Stamping! 🎉

t-vi (approved):
Great, thank you @kshitij12345 @riccardofelluga

PR #1391, opened 2024-11-01
Fix thunder.torch.checkpoint to support multiple arguments

IvanYashchuk:
In https://github.com/Lightning-AI/lightning-thunder/pull/1261 it was discovered that there is a bug in the `thunder.torch.checkpoint` implementation leading to a problem when running multiple arg case. Now it's fixed.


t-vi (approved):
Thank you @IvanYashchuk

PR #1393, opened 2024-11-02
[Tensor Subclasses] Support func calling only `Subclass(...)`

crcrpar:
## What does this PR do?

The scope of this PR is Tensor Subclasses that call `torch.Tensor._make_wrapper_subclass` in their dunder new (and define `__torch_dispatch__`, `__tensor_flatten__`, and `__tensor_unflatten__`).

This PR adds a new proxy for such tensor subclasses and implements a lookaside for `_make_wrapper_subclass` that returns an instance of the new proxy.

`MySubclass(...)` calls `MySubclass.__new__(cls, ...)` before calling `__init__(...)` on the return value of the dunder new.
Since this PR has the lookaside and it returns an instance of a proxy, not `MySubclass`, the `__init__` of the new proxy is called.
This is the reason the new proxy has if-else branches inside of its dunder new.

Caveat: This assumes that `Subclass.__new__` does not have kwargs, only positional args.

- ~[ ] The proxy should express `__tensor_flatten__` and `__tensor_unflatten__`~
    - #1345 might be reusable in this regard
    - will be resolved in #1394


thunder/core/jit_ext.py
```diff
@@ -743,6 +744,42 @@ def grad_transform(*args, **kwargs):
...
+    dispatch_device: bool = False,
+    dispatch_layout: bool = False,
+    _extra_dispatch_keys: torch.DispatchKeySet | None = None,
+    storage_size: int | None = None,
+):
+    ucls = unwrap(cls)
+    usize = unwrap(size)
+    udtype = unwrap(dtype)
+    udevice = unwrap(device)
+    urequires_grad = unwrap(requires_grad)
```
crcrpar, line 769: should be clear about ignored args

thunder/core/prims.py
```diff
@@ -4048,3 +4053,27 @@ def sink_meta(*args, **kwargs):
...
+    s = SubclassTensorProxy(
+        name,
+        subclass_type=cls,
+        shape=shape,
+        device=device,
+        dtype=dtype,
+        requires_grad=requires_grad,
+        tensors=tensors,
+        non_tensors=non_tensors,
+        history=[t.history for t in tensors],
```
crcrpar, line 4070: `history` should be able to get better

thunder/core/proxies.py
```diff
@@ -1880,6 +1880,111 @@ def real(self):
         return method(self)
 
 
+class SubclassTensorProxy(TensorProxy):
```
crcrpar, line 1883: this should be able to express `__tensor_flatten__(self) -> tuple[list[str], dict[str, Any]]` (and `__tensor_unflatten__(inner_tensors: dict[str, Tensor], metadata: dict[str, Any], outer_size, outer_stride) -> MySubclass`).
For it to happen, somewhere I have to give instances of this class the attribute names of tensors and non-tensor values.

thunder/tests/test_tensor_subclass.py
```diff
@@ -0,0 +1,182 @@
...
+class ScaleTensorSubclass(torch.Tensor):
+    _x: torch.Tensor
+    _scale: torch.Tensor
+    __slots__ = ["_x", "_scale"]
+
+    def __new__(cls, x: torch.Tensor, scale: torch.Tensor):
+        assert scale.numel() == 1, f"Invalid `scale`: {scale}"
+        dtype = x.dtype
+        device = x.device
+        self = torch.Tensor._make_wrapper_subclass(
```
IvanYashchuk, line 53: Do we need to use `make_wrapper_subclass` when `requires_grad=False`?

Here the behavior is different depending on the requires_grad value:
https://github.com/albanD/subclass_zoo/blob/ec47458346c2a1cfcd5e676926a4bbc6709ff62e/base_tensor.py#L12-L15

thunder/core/jit_ext.py
```diff
@@ -743,6 +744,42 @@ def grad_transform(*args, **kwargs):
     return forward_result
 
 
+@register_general_jit_lookaside(torch.Tensor._make_wrapper_subclass)
```
IvanYashchuk, line 747: Does it need to be a jit lookaside? Can the implementation be moved to `thunder/torch/__init__.py` if `@torchsymbol` is used?

thunder/__init__.py
```diff
@@ -369,7 +369,7 @@ def _alias_tensor_of_args_kwargs_dict(*args, **kwargs) -> dict[int, list[int]]:
         data_ptr_to_tensor_group_index = {}
         tensor_group_index_to_tensor_indices = defaultdict(list)
         for idx, t in enumerate(flat_args):
-            if pytorch.is_tensor(t) and t.layout == pytorch.strided:
+            if type(t) in {pytorch.Tensor, pytorch.nn.Parameter} and t.layout == pytorch.strided:
```
IvanYashchuk, line 372: Could you please submit this fix with a test in a separate pull request?

thunder/tests/test_tensor_subclass.py
crcrpar: The link uses `_make_subclass` not `_make_wrapper_subclass` and the last update is 2 years ago, so it doesn't sound convincing to me

thunder/core/prims.py
crcrpar: dunno, something disallows history being empty/none

thunder/core/jit_ext.py
crcrpar: I rather want to hide this method as possible. so making it a lookaside feels more right than a torchsymbol

thunder/__init__.py
crcrpar: That pull request would need a subclass then I'm not quite convinced by the option

thunder/core/proxies.py
mruberry: Would you add a comment describing what this class is intended for?

thunder/core/proxies.py
```diff
@@ -1880,6 +1880,125 @@ def real(self):
...
+    _subclass_type: torch._C._TensorMeta
+
+    def __init__(self, *args, **kwargs):
+        from thunder.core.pytree import tree_flatten
+
+        kwarg_tensors = kwargs.pop("tensors", [])
+        kwarg_non_tensors = kwargs.pop("non_tensors", [])
+        subclass_type = kwargs.pop("subclass_type", None)
+
+        # If tensors (and non_tensors) are not empty, then it should be the path of `_make_wrapper_subclass`
```
mruberry, line 1895: Would you elaborate on this comment?

thunder/core/proxies.py
```diff
@@ -1880,6 +1880,125 @@ def real(self):
...
+        # where `self` should already have gotten its name.
+        flat_args, spec = tree_flatten((args, kwargs))
+        tensors = list(filter(lambda t: isinstance(t, TensorProxy), flat_args))
+        non_tensors = list(filter(lambda t: not isinstance(t, TensorProxy), flat_args))
+        has_name_before_init = hasattr(self, "_name")
+
+        is_dunder_init_following_make_wrapper_subclass: bool = False
```
mruberry, line 1902: What's going on here?

thunder/core/proxies.py
```diff
@@ -1880,6 +1880,125 @@ def real(self):
...
 
+class SubclassTensorProxy(TensorProxy):
+    _tensors: list[TensorProxy]
+    _non_tensors: list[Any]
+    _subclass_type: torch._C._TensorMeta
+
+    def __init__(self, *args, **kwargs):
+        from thunder.core.pytree import tree_flatten
+
+        kwarg_tensors = kwargs.pop("tensors", [])
```
mruberry, line 1891:
```
__init__(self, *args, tensor=[], non_tensors=[], subclass_type=None, **kwargs)
```
?

thunder/core/proxies.py
```diff
@@ -1880,6 +1880,125 @@ def real(self):
...
+        kwarg_non_tensors = kwargs.pop("non_tensors", [])
+        subclass_type = kwargs.pop("subclass_type", None)
+
+        # If tensors (and non_tensors) are not empty, then it should be the path of `_make_wrapper_subclass`
+        # where `self` should already have gotten its name.
+        flat_args, spec = tree_flatten((args, kwargs))
+        tensors = list(filter(lambda t: isinstance(t, TensorProxy), flat_args))
+        non_tensors = list(filter(lambda t: not isinstance(t, TensorProxy), flat_args))
+        has_name_before_init = hasattr(self, "_name")
```
mruberry, line 1900: How can this happen?

thunder/core/proxies.py
```diff
@@ -1880,6 +1880,125 @@ def real(self):
...
+            self._non_tensors = kwarg_non_tensors
+            self._subclass_type = subclass_type
+        else:
+            # TODO(crcrpar): Think about materializing `self` so that we can
+            # call `__tensor_init__` to know each attribute names.
+            from thunder.core import prims
+
+            self._tensors = tensors
+            self._non_tensors = non_tensors
+            bsym = prims.tensor_subclass_ctor.bind(
```
mruberry, line 1929: This is interesting. Why does this class either create a new proxy or change the actual trace?

thunder/core/proxies.py
```diff
@@ -1880,6 +1880,125 @@ def real(self):
...
+            # but not, otherwise. As [the lookasdie of `torch.autograd.Function`](
+            # https://github.com/Lightning-AI/lightning-thunder/blob/3d42c10/thunder/core/jit_ext.py#L655)
+            # puts the temporary scope to the current trace.
+            current_trace = get_tracectx()
+            if id(current_trace.bound_symbols) == id(cur_tail_scope := current_trace.scopes[-1]):
+                current_trace.add_bound_symbol(bsym)
+            else:
+                cur_tail_scope.append(bsym)
+
+    def replace(self, **changes):
```
mruberry, line 1954: Why is a replace function needed?

thunder/core/proxies.py
```diff
@@ -1880,6 +1880,125 @@ def real(self):
...
+        r"""Return a copy of the SubclassTensorProxy object with new values for the specified fields as given to the constructor as arguments.
+        Valid keyword arguments are ``name``, ``history``, ``shape``, ``dtype``, ``device``, ``requires_grad``, ``distparallel_type``,  ``thunder_fsdp_padding_size``.
+        ``like`` is also a valid keyword and will take metadata from the tensor proxy argument
+        in preference to the old values but overridable by keyword arguments.
+        Note that the copy will use the current (environment) tracectx."""
+
+        like = changes.get("like")
```
mruberry, line 1961: If `changes` doesn't have `like` as a key won't this throw a key error?

thunder/core/proxies.py
```diff
@@ -1880,6 +1880,125 @@ def real(self):
...
+            true_dtype,
+            numel,
+            ndim,
+            requires_grad,
+            grad,
+            distparallel_type,
+            thunder_fsdp_padding_size,
+        ) = _infer_tensor_properties(
+            like,
+            changes.get("shape", self._shape if like is None else None),
```
mruberry, line 1975: Why not 

```
*[myDictionary.get(key, None) for key in ('shape', 'device', ...)]
```

?

thunder/__init__.py
mruberry: How do subclasses appear here? Do all subclasses have the actual torch.Tensor type?

thunder/tests/test_tensor_subclass.py
```diff
@@ -0,0 +1,182 @@
...
+        if not isinstance(out, torch.Tensor):
+            return out
+        else:
+            return ScaleTensorSubclass(out, scales[0])
+
+
+@instantiate(
+    dtypes=(thunder.core.dtypes.float32,),
+)
+def test_func_of_subclass_ctor_wrapper(executor, device, _):
```
mruberry, line 124: Do these tests just check that the tensor subclass can be constructed?

thunder/core/proxies.py
crcrpar: because I don't want `.replace(...)` call to replace subclass tensor proxy with tensor proxy

thunder/core/proxies.py
crcrpar: If key is not found then the default value, in this case, `None` is returned. https://docs.python.org/3/library/stdtypes.html#dict.get

thunder/core/proxies.py
crcrpar: I copied these lines from `TensorProxy.replace` and I'm lazy enough not to do that

thunder/tests/test_tensor_subclass.py
crcrpar: Yes, because I was struggling to support previously, in #1345. The others are in #1394

thunder/core/proxies.py
crcrpar: If thunder sees a tensor wrapper subclass of `MySubclass(...)` that has its own dunder new calling `_make_wrapper_subclass` in a function thunder's tracing, the lookaside of https://github.com/Lightning-AI/lightning-thunder/pull/1393/files#diff-3d1ea50ad3b0e3ad6fc369f91a7e42011d1d33d770ce25f800637c99de85f4b5R762 creates an instance of SubclassTensorProxy, then the dunder init of that instance is called, not the dunder init of `MySubclass` instance.

thunder/core/proxies.py
crcrpar: ditto

thunder/core/proxies.py
crcrpar: this path is happening while tracing and the call of `Proxy.__init__` is not recorded by default in a trace so it's necessary to deliberately register a boundsymbol to the currently active scope

mruberry: OK; but why does proxy creation get recorded into the trace?

thunder/core/proxies.py
mruberry: But doesn't the current code mean that if `like` is specified then the shape of the like tensor is overriden with the shape of the current tensor? Is that what's intended?

thunder/core/proxies.py
mruberry: I guess I'm still confused about how and why the proxy class is entangled with the actual subclass

thunder/core/proxies.py
crcrpar: if a function creates a subclass inside it then shouldn't a trace of it have a BoundSymbol to represent it?

mruberry: I think it's OK to put tensor subclass creation into a trace (although I'm curious if we can dce the creation if the subclass is just flattened and used once later), but I'm not sure why the operator that creates the actual tensor subclass is also a constructor for the proxy. The existing tensor proxies don't entangle their tensor factory methods with the creation of the proxy, for example

crcrpar: The short answer would be that it's discouraged to use `torch.Tensor(...)` but we do `MySubclass(...)` and `MySubclass.__new__` is visible to Thunder.

`MySubclass(...)` (in this PR, it's a subclass that calls `_make_wrapper_subclass` in its `__new__`) calls `__new__`. thunder's tracing sees this dunder new. `_make_wrapper_subclass` used in this method requires a new lookaside or a new torchsymbol (Both would be supposed to return a proxified instance of `MySubclass`). Dunder init of `MySubclass` is called on the return value of `MySubclass.__new__`.
But here, as we override the behavior of `MySubclass.__new__`, the init of the proxified `MySubclass` instance is called.

crcrpar: In the end, a trace might end up free from bsyms representing `MySubclass(...)`, currently this is necessary to just get an initial trace.

mruberry: I still don't really get this. I understand that Thunder sees the call to `__new__`, I'm just confused why that call to `__new__` doesn't produce a bound symbol like other lookasides, and then the meta function for that bound symbol produces a SubclassTensorProxy?

I wouldn't say that we "discourage" the use of `torch.Tensor` (although I don't think Thunder supports it, so I guess that's discouraging) any more than PyTorch does. Thunder does support `torch.tensor`, however, and that does directly call the TensorProxy constructor

crcrpar: As for `torch.Tensor(...)`, I should've put the link to this page: https://pytorch.org/docs/stable/tensors.html#torch.Tensor.__init__.
> This constructor is deprecated, we recommend using [torch.tensor()](https://pytorch.org/docs/stable/generated/torch.tensor.html#torch.tensor) instead. What this constructor does depends on the type of data.

> why that call to __new__ doesn't produce a bound symbol like other lookasides, and then the meta function for that bound symbol produces a SubclassTensorProxy?

Would this by chance mean that a lookaside for `__new__` should work better than a lookaside for `torch.Tensor._make_wrapper_subclass`?
From my perspective, either way, as we see `__new__`, then `__init__` is to be called so I guess it'd be necessary to do some meticulous handling of `__init__` of tensor subclasses because either way we won't dodge the call of the dunder new.

mruberry: OK; I think I'm understanding better what the issue is. Let me try to summarize it. 

When someone creates a tensor subclass, it looks like this:

```
x = ComplexTensor(torch.tensor([[1]]), torch.tensor([[2]]))
```

(taken from https://github.com/albanD/subclass_zoo/blob/main/complex_tensor.py)

And that makes sense. The class's constructor is called, because of course it is.

Internally, Python looks at the class constructor and then constructs an object of the requested type (this process is a little complicated), then initializes it (optionally) using dunder init.

How we understand the object created by the call to `ComplexTensor` is determined by the subclass's implementation of dunder torch_dispatch, dunder tensor flatten, and dunder tensor unflatten. Does the actual call to _make_wrapper_subclass tell us anything about the subclass, or do we get that information by querying the object returned from the constructor?

I guess my question, then, is if we can also just call `ComplexTensor` like a practitioner would? Then the meta for `ComplexTensor` could return a `SubclassTensorProxy` or even a `ComplexTensorSubclassProxy`.

I think that would provide a clear separation between the proxy and the actual call.

crcrpar: > Does the actual call to `_make_wrapper_subclass` tell us anything about the subclass

`torch.Tensor._make_wrapper_subclass` takes `cls`. The lookaside could let us tell things other than actual tensors and other values to be registered to the output of the method.

> I guess my question, then, is if we can also just call `ComplexTensor` like a practitioner would? 

I don't get where this call of `ComplexTensor` would be. Could you elaborate on it?

> Then the meta for `ComplexTensor` could return a `SubclassTensorProxy` or even a `ComplexTensorSubclassProxy`.

How and when would we register `ComplexTensorSubclassProxy`? Would this indicate that Thunder is going to detect a subclass and define a proxy for it dynamically?

mruberry: Yes! What if we detected that a subclass constructor was called and dynamically created a symbol for the constructor and the corresponding proxy?

The call to `ComplexTensor` could be in the resulting Thunder program, like other bound symbols. Or maybe we want to represent `ComplexTensor` construction as an unflatten call, like `ComplexTensor.__tensor_unflatten__`? Both seem reasonable to me

crcrpar: If I do that dynamically, I want that subclass proxy to correctly capture the result type of ops with that subclass. I however don't see a clear path to infer the result type of ops with subclasses, other than we do what AOT Autograd does with torch.fx tracing. Thus I don't think dynamic registration is that different from what I have in this PR.

Representing subclass construction as unflatten call would then call `__init__` of that subclass, so I'm not following the point.

mruberry: The goal is to divorce the proxy from the actual runtime object to simplify the code and its concepts.

How else would we infer the decomposition of operations called on a tensor subclass, and if the result is also a member of the subclass? 

I understand that __init__ gets called on the object at runtime, of course. The goal is not to someone circumvent the proper construction of the object at runtime (unless it can be elided for performance).

t-vi: Maybe I'm not understanding this well, but I'm a bit worried that we don't have well-defined semantics here and that the traces are not representing what's up.

We should absolutely know when we want to construct and use the subclass object (if we return it or something wants the subclass object) and when we don't (all other cases) and should also represent what will be the compute in the trace.

We also want to minimize admin overhead during the compute, which we are not doing a great job today (having looked at wall clock vs. GPU self time for Llama 1b today), so adding the overhead of dealing with subclasses at compute time should likely be minimized.

t-vi (approved):
On my part this is good to merge (thanks @crcrpar and @IvanYashchuk for the sync with @lantiga and me last week). We can revise the bits from this PR when the need arises while we build on it.

@mruberry do you think we should go ahead?

crcrpar:
I'd like to move to #1394. #1394 is based on this one so I think changing the target of #1394 to main would be better than merge 1393 then 1394

t-vi:
Works for me, too.

thunder/core/proxies.py
crcrpar: excuse me for my idling on this. I guess I understand your points a bit better. The separation does sound nice. For it, I think one way is to do the interpretation of `__torch_dispatch__` while acquiring the initial trace, not what #1394 is doing, i.e., get the initial trace with opaque tensor subclass proxies before unrolling their `__torch_dispatch__`. With this, some traces like ones we'd get from torchao float8 programs would be completely free from tensor subclass proxies. If any of arguments or return values of a function is a tensor subclass, we might end up having one or more proxy classes for tensor subclasses observed in it and we might need to implement a dynamic registration of them while interpretation, which I almost have no idea at the moment.

mruberry: Happy to talk through this more over VC if it would be helpful, too!

crcrpar: Tensor subclass proxy objects would have `type(tensor_subclass_object)` as their attribute but not actual instances.

I'd be rather hesitant to complicate the current source code of interpreter by introducing mechanism which unrolls tensor subclasses, defines a proxy class for the observed tensor subclass, and registers that proxy class to the namespace.
Currently, the implementation itself isn't the best for sure, but the unrolling of tensor subclasses happens after the initial computation is acquired (This initial trace would have some wrong expressions, e.g., when a tensor subclass' `__torch_dispatch__` returns one or multiple objects of that tensor subclass), thus I think the change of this PR (more precisely, #1394 and #1415) would have less interactions.

thunder/core/proxies.py
crcrpar: I think `changes > like > self` is the priority order.

PR #1411, opened 2024-11-08
Check `_interpret_call` output in the lookaside of `torch.autograd.Function`.

crcrpar:
## What does this PR do?

return it if it's exception


t-vi (approved):
Should we have a test for this, too?

PR #1421, opened 2024-11-11
remove `from baseutils import *`

crcrpar:
## What does this PR do?

Stop start import in codeutils.

### DIFF
--- a/thunder/core/codeutils.py
+++ b/thunder/core/codeutils.py
@@ -11,6 +11,7 @@ from functools import partial
 import dis
 import linecache
 import dataclasses
+import sys
 
 import torch
 
--- a/thunder/core/codeutils.py
+++ b/thunder/core/codeutils.py
@@ -260,7 +260,7 @@ def prettyprint(
         return m(f"{baseutils.print_type(x, with_quotes=False)}")
 
     # Handles objects that this doesn't know how to serialize as a string
-    return m(f"(object of type {print_type(type(x), with_quotes=False)})")
+    return m(f"(object of type {baseutils.print_type(type(x), with_quotes=False)})")
 
 
 # Use dis.Positions in 3.11+ and make it up in <3.11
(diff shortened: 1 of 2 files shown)


thunder/core/codeutils.py
```diff
@@ -19,7 +20,6 @@
 import thunder.core.dtypes as dtypes
 import thunder.core.devices as devices
 from thunder.core.pytree import tree_flatten, tree_unflatten
-from thunder.core.baseutils import *
```
IvanYashchuk, line 22: Great to get rid of the star import!

IvanYashchuk (approved):

thunder/core/codeutils.py
mruberry, line 22: It's OK to replace uses of `baseutils` with explicit `baseutils.foo` calls in this file, but conceptually `baseutils`, `codeutils` and `utils` offer a progressive expansion of utility functions. Most files can just import `utils` (and doing so gives them all the functions in `codeutils` and `baseutils`), but some files can only import `codeutils` or `baseutils`. `baseutils` depends only on Python and PyTorch. `codeutils` depends on thunder's `dtypes`, `devices` and `pytree` submodules. `utils` also depends on `prims`, `trace` and `proxies`. 

For files that cannot import `utils` but can import `codeutils`, they probably want `baseutils`, too, just like how importing `utils` provides access to the functions in `baseutils` and `codeutils`. 

There are some files, like `trace.py`, that import both `codeutils` and `baseutils`, but they could just import `codeutils`, instead.

Borda (approved):

thunder/core/codeutils.py
t-vi: Let's not aim to minimize the number of import lines but maximize that there is one standard way to access a given function and only deviate from it when there is a need to. I don't mind if we use baseutils.foo or utils.foo by default but let's not do a thing codeutils.foo if you import codeutils and otherwise baseutils.foo .

t-vi (approved):
Thank you @crcrpar @IvanYashchuk @mruberry

Issue #1420, opened 2024-11-11
ThunderFX splitter should send no_grad regions to Thunder


IvanYashchuk:
## 🐛 Bug

Splitter for Thunder as a Dynamo backend should send regions of code under a "no_grad" context manager to Thunder. Currently, it chooses to send these computations to Inductor.

```py
from thunder.dynamo import ThunderCompiler
import thunder
import torch

def f(x):
    with torch.no_grad():
        return x * x

jit_f = thunder.jit(f)
backend = ThunderCompiler()
compile_f = torch.compile(backend=backend)(f)

x = torch.randn(3, 3, requires_grad=True)

out = jit_f(x) # Works with thunder.jit with a warning
out_1 = compile_f(x) # Works with torch.compile but sends the computation to the Inductor instead of Thunder

print(backend.subgraph_infos[0].split_graph_module.print_readable())
```
prints:
```py
class GraphModule(torch.nn.Module):
    def forward(self, l_x_: "f32[3, 3]"):
        # No stacktrace found for following nodes
        inductor_1 = self.inductor_1(l_x_);  l_x_ = None
        return (inductor_1,)
        
    class inductor_1(torch.nn.Module):
        def forward(self, l_x_: "f32[3, 3]"):
            # No stacktrace found for following nodes
            _set_grad_enabled = torch._C._set_grad_enabled(False);  _set_grad_enabled = None
            
             # File: <ipython-input-3-f5a278fc59c9>:7 in f, code: return x * x
            mul: "f32[3, 3]" = l_x_ * l_x_;  l_x_ = None
            
            # No stacktrace found for following nodes
            _set_grad_enabled_1 = torch._C._set_grad_enabled(True);  _set_grad_enabled_1 = None
            return mul
```

HF's Qwen 2 model added in https://github.com/Lightning-AI/lightning-thunder/pull/1406 creates a small Inductor region because `Qwen2RotaryEmbedding.forward` [is decorated with a `torch.no_grad`](https://github.com/huggingface/transformers/blob/3ea3ab62d80d91f9bdd16bd3cacd8133fb0d4566/src/transformers/models/qwen2/modeling_qwen2.py#L151).

IvanYashchuk:
Sending no_grad to Thunder was disabled in https://github.com/Lightning-AI/lightning-thunder/pull/1282 for a good reason (https://github.com/Lightning-AI/lightning-thunder/issues/1219). Maybe it's time to properly support the `no_grad` context in Thunder?

PR #1423, opened 2024-11-11
support no_grad in thunder.jit

kshitij12345:
Related #1420 

In this PR, we support `no_grad` (and `torch._C._set_grad_enabled`) in user function passed to `thunder.jit`.

Approach:

There are 3 steps.
1. We add `torch._C._set_grad_enabled`  as a marker symbol which updates state `is_grad_enabled` in CompileData. These symbols are removed in `remove_context_manager_prims_from_trace` pass.
2. When a Symbol is called during tracing, it queries `is_grad_enabled` on CompileData and accordingly tags it's output Proxy.
3. We update VJP transform to treat Bsyms whose output contains `DETACHED_AUTOGRAD_GRAPH` as constants. (We already have logic to treat Symbols as constant for VJP, we just add a new case here).

<details>

<summary> Example and Forward-Backward Traces </summary>

```python
import torch
import thunder

def fn(x):
    with torch.no_grad():
        y = x * torch.ones(3,)
    return x + y

x = torch.ones(3, requires_grad=True)
jfn = thunder.jit(fn)

o = jfn(x)
o.sum().backward()
print(x.grad)

# Check against PyTorch eager
# x = torch.ones(3, requires_grad=True)
# o = fn(x)
# o.sum().backward()
# print(x.grad)
```

First Trace
```python
def computation(x):
  # x: "cpu f32[3]"

  # /home/kkalambarkar/git/pytorch/torch/autograd/grad_mode.py:187:             torch._C._set_grad_enabled(mode)
  ltorch._set_grad_enabled_with_warning(False)

  # /home/kkalambarkar/lightning-thunder/scratchpad/test_no_grad.py:6:          y = x * torch.ones(3,)
  t0 = ltorch.ones(3, device=None, dtype=None)  # t0: "cpu f32[3]"
    # t0 = ltorch.full((3,), 1, device=None, dtype=torch.float32)  # t0: "cpu f32[3]"
      # t0 = prims.full((3,), 1, device=devices.Device("cpu"), dtype=dtypes.float32)  # t0: "cpu f32[3]"
  t1 = ltorch.mul(x, t0)  # t1: "cpu f32[3]"
    # t1 = prims.mul(x, t0)  # t1: "cpu f32[3]"

  # /home/kkalambarkar/git/pytorch/torch/autograd/grad_mode.py:187:             torch._C._set_grad_enabled(mode)
  ltorch._set_grad_enabled_with_warning(True)

  # /home/kkalambarkar/lightning-thunder/scratchpad/test_no_grad.py:7:      return x + y
  t2 = ltorch.add(x, t1, alpha=1)  # t2: "cpu f32[3]"
    # t2 = prims.add(x, t1)  # t2: "cpu f32[3]"
  return t2
```

Execution Trace
```python
def computation(x):
  # x: "cpu f32[3]"
  t0 = torch.ones(3, device=None, dtype=None)  # t0: "cpu f32[3]"
    # t0 = ltorch.ones(3, device=None, dtype=None)  # t0: "cpu f32[3]"
      # t0 = ltorch.full((3,), 1, device=None, dtype=torch.float32)  # t0: "cpu f32[3]"
        # t0 = prims.full((3,), 1, device=devices.Device("cpu"), dtype=dtypes.float32)  # t0: "cpu f32[3]"
  t1 = torch.mul(x, t0)  # t1: "cpu f32[3]"
    # t1 = ltorch.mul(x, t0)  # t1: "cpu f32[3]"
      # t1 = prims.mul(x, t0)  # t1: "cpu f32[3]"
  del t0
  t2 = torch.add(x, t1, alpha=1)  # t2: "cpu f32[3]"
    # t2 = ltorch.add(x, t1, alpha=1)  # t2: "cpu f32[3]"
      # t2 = prims.add(x, t1)  # t2: "cpu f32[3]"
  del t1
  return {'output': t2, 'flat_args': [x], 'flat_output': (t2,)}, ((), ())
```

Backward Execution Trace
```python
  # cotangents: "Collection"
  t6, = cotangents
  del cotangents
  return (t6,)
```

</details>


kshitij12345:
@IvanYashchuk do you think this approach looks good or do you have anything else in mind?

thunder/core/transform_common.py
```diff
@@ -492,3 +492,52 @@ def is_context_manager_prim(bsym):
...
+
+            def create_detached_output(t):
+                if isinstance(t, TensorProxy):
+                    # NOTE - We need `tracectx` as creating/replacing name for proxy
+                    # tries a look-up in current trace.
+                    with tracectx(new_trace):
+                        # Remove the name so that we can re-use it.
+                        # Otherwise, we get a proxy with new name.
+                        new_trace.names.remove(t.name)
+                        return t.replace(requires_grad=False, tags=(ProxyTag.DETACHED_AUTOGRAD_GRAPH,))
```
IvanYashchuk, line 531: `requires_grad` makes a difference only for input proxies.
```suggestion
                        return t.replace(tags=(ProxyTag.DETACHED_AUTOGRAD_GRAPH,))
```

kshitij12345: Right, forgot to remove this. Initially I had planned to set `requires_grad=False` to treat these as constant but then I recalled that vjp and grad transforms should work even if `requires_grad=False` for inputs.

Thank you!

```python
import torch
import thunder

def fn(x):
    return x * 2

jfn = thunder.jit(fn)
g_jfn = thunder.core.transforms.grad(jfn)
g = g_jfn(torch.randn(3,))
print(g)

```

thunder/core/symbol.py
```diff
@@ -320,6 +321,18 @@ def __call__(self, *args, **kwargs):
             result = self.meta(*args, **kwargs)
             trace.pop_scope()
 
+        cd = get_compile_data()
+        if cd is not None and not cd.is_grad_enabled:
```
kshitij12345, line 325: There are paths in test which compile_data could be False - example while using `thunder.trace`

t-vi: Do we still have cases where thunder.trace is used without a compile data coming from elsewhere?

kshitij12345: These are the hits for `thunder.trace` with no compile-data (mainly in tests and some in example notebooks)

https://github.com/search?q=repo%3ALightning-AI%2Flightning-thunder%20thunder.trace()&type=code

thunder/core/symbol.py
```diff
@@ -320,6 +321,20 @@ def __call__(self, *args, **kwargs):
...
+            # If grad is disabled using `torch.no_grad` or `torch._C._set_grad_enabled(False)`,
+            # tag the results with `DETACHED_AUTOGRAD_GRAPH` which makes this Symbol a constant for
+            # vjp transform (applied later).
+            def tag_tensorproxy_output_as_detached(proxy):
+                if isinstance(proxy, TensorProxy):
+                    # We need to remove name from trace, otherwise replace will return a proxy with new name.
+                    trace.names.remove(proxy.name)
+                    return proxy.replace(tags=(ProxyTag.DETACHED_AUTOGRAD_GRAPH,))
```
kshitij12345, line 333: Is there a better way than this to enforce the same name is used for the output of `replace`

IvanYashchuk: Seems like there's no better way now. Proxies' init method calls 
https://github.com/Lightning-AI/lightning-thunder/blob/d60f85cf009793894e9329eca1ee543037237057/thunder/core/proxies.py#L149
which does not force provided name usage if it's already used (included in the `trace.names` set).

thunder/core/symbol.py
t-vi: Right, to my mind, we should try to drop them. Thanks for bringing this up.

thunder/core/symbol.py
t-vi: Unfortunately, currently not. :/

thunder/__init__.py
```diff
@@ -442,6 +442,11 @@ def get_computation_and_inputs(*args, **kwargs):
         # which seems to break the consistency of cache_info, leading to a failure in cache_info check.
         cache_info["alias_tensor_indices"] = _alias_tensor_of_args_kwargs(*args, **kwargs)
 
+        # Store the `is_grad_enabled` state of PyTorch. This is used by vjp transform
+        # to treat certain Symbols as constant.
+        cache_info["is_grad_enabled"] = pytorch.is_grad_enabled()
```
IvanYashchuk, line 447: Is this cache_info entry needed?

IvanYashchuk (approved):

thunder/__init__.py
t-vi: Well if you call with grad enabled and then without, you would want to have a cache miss?

t-vi: If we think we don't need it, let's remove it in a follow-up.

kshitij12345: It is for the following cases

Example 1
```python
jfn = thunder.jit(fn)
with torch.no_grad():
    jfn(x)   # This will be compiled with no_grad

jfn(x)   # We want this to be recompiled.
```

Example 2
```python
jfn = thunder.jit(fn)

jfn(x)

with torch.no_grad():
    jfn(x)   # We want this to be recompiled
```

IvanYashchuk: > Well if you call with grad enabled and then without, you would want to have a cache miss?

Of course. 

How does this work if the content of `cache_info["is_grad_enabled"]` is not checked anywhere in this pull request?

kshitij12345: In `jit_ext.py`, we read the values from cache_info and add corresponding checks in prologue.

https://github.com/Lightning-AI/lightning-thunder/blob/11a32a4b9575acc6de670450e1c0d66cab541305/thunder/core/jit_ext.py#L1580-L1598

IvanYashchuk: Thank you!

PR #1417, opened 2024-11-08
add elu op

beverlylytle:
<details>
  <summary><b>Before submitting</b></summary>

- [ ] Was this discussed/approved via a Github issue? (no need for typos and docs improvements)
- [X] Did you read the [contributor guideline](https://github.com/Lightning-AI/pytorch-lightning/blob/main/.github/CONTRIBUTING.md), Pull Request section?
- [ ] Did you make sure to update the docs?
- [X] Did you write any new necessary tests?

</details>

## What does this PR do?

Adds the elu operator.

## PR review

Anyone in the community is free to review the PR once the tests have passed.
If we didn't discuss your PR in Github issues there's a high chance it will not be merged.

## Did you have fun?

Make sure you had fun coding 🙃

PR #1427, opened 2024-11-12
Use the prologue in check_vjp

beverlylytle:
In adding support for the elu op (https://github.com/Lightning-AI/lightning-thunder/pull/1417), it was discovered that 'alpha' parameters are not being included in the calculations of jacobian vector products as the prologue is being skipped in this test.  Here the prologue is added and the celu op, which also suffers from this inaccuracy, is adapted.

### DIFF
--- a/thunder/tests/opinfos.py
+++ b/thunder/tests/opinfos.py
@@ -1648,12 +1648,7 @@ celu_opinfo = OpInfo(
     dtypes=(datatypes.floating,),
     sample_input_generator=celu_sample_generator,
     torch_reference=_elementwise_unary_torch(torch.celu),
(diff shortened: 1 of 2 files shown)


t-vi (approved):
Thank you @beverlylytle @IvanYashchuk

t-vi:
@beverlylytle @IvanYashchuk Merging this added 10 minutes (27->37) to the GPU CI runtime. I wonder if we should use it only for things with keyword args.

IvanYashchuk:
Is there a thunder.jit caching option that considers tensor inputs static and keyword arguments dynamic if they are present?

PR #1432, opened 2024-11-13
Add `nv_enable_options` and `nv_disable_options`

Priya2698:
`nvFuser version >= 0.2.23` will support setting `NVFUSER_ENABLE/NVFUSER_DISABLE` through the python frontend using `_enable_options` and `_disable_options` flags. This is to enable matmul/linear codegen for nvfuser through Thunder using flags `nv_enable_options=["fuse_matmul"], nv_disable_options=["matmul_expr_eval"]`
Issue: https://github.com/NVIDIA/Fuser/issues/3022

nvFuser PR: https://github.com/NVIDIA/Fuser/pull/3270

### DIFF
--- a/thunder/executors/nvfuserex_impl.py
+++ b/thunder/executors/nvfuserex_impl.py
@@ -460,7 +460,7 @@ class FusionDefinitionWrapper:
 
         elif self.enable_options or self.disable_options:
             warnings.warn(
-                f"nvFuser _enable_options/_disable_options requires version 0.2.23 and above, using version {nvfuser_version()}. These options will be ignored."
+                f"nv_enable_options/nv_disable_options require nvFuser version 0.2.23 and above, found version {nvfuser_version()}. These options will be ignored."
             )
 
         with annotate_for_profile(self.name):
--- a/thunder/executors/nvfuserex_impl.py
+++ b/thunder/executors/nvfuserex_impl.py
@@ -437,7 +437,10 @@ class FusionDefinitionWrapper:
     cache_clear: None | Callable = None
     last_used: None | FusionDefinition = None
     last_inputs: None | Sequence[tuple] = None
-    store_inputs: bool = False
+    store_inputs: bool = False,
+    profile: bool = False,
+    _enable_options: list = [],
+    _disable_options: list = [],
 
     def __call__(self, *args):
         fd = self.get_fd(self.to_descriptors(args))
@@ -446,8 +449,14 @@ class FusionDefinitionWrapper:
         if self.store_inputs:
             self.last_inputs = args
 
+        kwargs = {}
         # Set device if set in one of the "factory" methods like full, iota, or uniform
-        kwargs = {"device": fd._selected_device} if hasattr(fd, "_selected_device") else {}
+        if hasattr(fd, "_selected_device"):
+            kwargs["device"] = fd._selected_device
+        if not self._enable_options:
+            kwargs["_enable_options"] = self._enable_options
+        if not self._disable_options:
+            kwargs["_disable_options"] = self._disable_options
         with annotate_for_profile(self.name):
             return fd.execute(args, **kwargs)
 
@@ -540,6 +549,12 @@ def create_fusion_definition_wrapper(
     store_inputs: None | bool = get_compile_option(
         "nv_store_fusion_inputs", "Allow nvFuser to store fusion inputs for repro."
     )
+    _enable_options: None | list = get_compile_option(
+        "nv_enable_options", "List of NVFUSER_ENABLE options to set."
+    )
+    _disable_options: None | list = get_compile_option(
+        "nv_disable_options", "List of NVFUSER_DISABLE options to set."
+    )
 
     tensor_indices = []
     for idx, x in enumerate(sorted_unique_inputs):
@@ -553,7 +568,7 @@ def create_fusion_definition_wrapper(
     def get_fd(input_descriptors) -> FusionDefinition:
         # A closure over local trace and region
         return create_fd(bsyms, input_descriptors, sorted_unique_inputs, sorted_unique_outputs)
-
+    # breakpoint()
     fdw = FusionDefinitionWrapper(
         get_fd,
         partial(to_descriptors, sorted_unique_inputs),
@@ -561,6 +576,8 @@ def create_fusion_definition_wrapper(
         get_fd.cache_info,
         get_fd.cache_clear,
         store_inputs=store_inputs,
+        _enable_options = _enable_options if _enable_options is not None else [],
+        _disable_options = _disable_options if _disable_options is not None else [],
     )
     return fdw
 
--- a/thunder/executors/nvfuserex_impl.py
+++ b/thunder/executors/nvfuserex_impl.py
@@ -454,6 +454,7 @@ class FusionDefinitionWrapper:
             kwargs["device"] = fd._selected_device
 
         if nvfuser_version() >= LooseVersion("0.2.23"):
+            # nvFuser expects empty list instead of None values.
             kwargs["_enable_options"] = self._enable_options if self._enable_options is not None else []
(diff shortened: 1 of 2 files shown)


thunder/executors/nvfuserex_impl.py
```diff
@@ -441,7 +441,9 @@ class FusionDefinitionWrapper:
     cache_clear: None | Callable = None
     last_used: None | FusionDefinition = None
     last_inputs: None | Sequence[tuple] = None
-    store_inputs: bool = False
+    store_inputs: bool = (False,)
+    _enable_options: list = ([],)
+    _disable_options: list = ([],)
```
crcrpar, line 446: Could you improve either the type hint or the default values?
Because the type annotations seem to contradict with the default values. The default ones are `tuple[bool]`, `tuple[list[Any]]`, and `tuple[list[Any]]`, respectively.

mruberry:
Maybe @jjsjann123 or @kevinstephano would like to review this?

thunder/executors/nvfuserex_impl.py
```diff
@@ -544,6 +559,12 @@ def create_fusion_definition_wrapper(
     store_inputs: None | bool = get_compile_option(
         "nv_store_fusion_inputs", "Allow nvFuser to store fusion inputs for repro."
     )
+    _enable_options: None | list[str] = get_compile_option(
```
jjsjann123, line 562: nitpick: I think just `enable_options` should be fine. same thing for `disable_options`

thunder/tests/test_nvfuser.py
```diff
@@ -1077,3 +1077,42 @@ def sdpa_fn(q, k, v, dropout_p, is_causal, scale):
...
+        nv_enable_matmul=True,
+        nv_enable_options=["fuse_matmul"],
+        nv_disable_options=["matmul_expr_eval", "kernel_reuse"],
+    )
+    # The above combination of options enables matmul codegen and disables expr evaluation for matmul.
+    # Since matmul scheduler does not support float32 inputs, the execution should raise an error.
+    # By default, without using these options, the given fusion will run through expr eval scheduler correctly.
+
+    with pytest.raises(RuntimeError, match="Can not find a scheduler to schedule fusion segment"):
+        out = compiled_func(*inps)
```
jjsjann123, line 1118: nitpick: this is a strange test, since we are verifying the limited support of nvfuser.

i.e. There's going to be a test regression as we improve matmul scheduler..

jjsjann123 (approved):
nitpick on the test. Code change looks good to me otherwise.

thunder/executors/nvfuserex_impl.py
Priya2698: Sure. For values, this naming is fine too. We do use `_enable_options` for teh argument names in `fd.execute`

thunder/tests/test_nvfuser.py
jjsjann123: :face_with_head_bandage: I see your struggle there.

Let's just put a note here acknowledge that this is a *temporary* test and if the exception is gone at a later point, we should remove this test, since it's no longer an observable test to verify the functionality of the flags.

Priya2698: Done!

thunder/executors/nvfuserex_impl.py
Priya2698: Done!

thunder/executors/nvfuserex_impl.py
```diff
@@ -446,8 +448,21 @@ def __call__(self, *args):
...
+            kwargs["device"] = fd._selected_device
+
+        if nvfuser_version() >= LooseVersion("0.2.23"):
+            # nvFuser expects empty list instead of None values.
+            kwargs["_enable_options"] = self.enable_options if self.enable_options is not None else []
+            kwargs["_disable_options"] = self.disable_options if self.disable_options is not None else []
+
+        elif self.enable_options or self.disable_options:
+            warnings.warn(
+                f"nvFuser _enable_options/_disable_options requires version 0.2.23 and above, using version {nvfuser_version()}. These options will be ignored."
```
jjsjann123, line 463: nitpick: we missed the prefix underscore in the warning message.

Priya2698: I was keeping them the same as nvFuser argument names, but on second thought, `nv_enable_options` might be better on Thunder side.

t-vi:
What would be sane defaults here?
How will user know they need to deviate?
So here we are adding two options that users need to know about and then they need to know the magic strings to be filled in...

Priya2698:
> What would be sane defaults here? How will user know they need to deviate? 

By default, no options need to be used.

The primary objective of these options is nvfuser development. 
We can enable `matmul/linear` for nvfuser using `nv_enable_matmul/linear` flags. Within nvfuser, we can further control the scheduler used for these operators (i.e. whether we use ATen or nvfuser codegen) through `NVFUSER_ENABLE/DISABLE` environment flags.

This PR coupled with https://github.com/NVIDIA/Fuser/pull/3270 allows us to enable nvfuser codegen for matmul/linear through Thunder using these options. This is to primarily support direct experimentation from Thunder.

mruberry:
> What would be sane defaults here? How will user know they need to deviate? So here we are adding two options that users need to know about and then they need to know the magic strings to be filled in...

These are good questions. I think it's OK for now because these are just developer-facing options to continue experimentation. Maybe we should review how options are presented more generally, to clarify what is for practitioners vs. internal developers? Maybe developer options should be better documented somewhere, too?

Priya2698:
> > What would be sane defaults here? How will user know they need to deviate? So here we are adding two options that users need to know about and then they need to know the magic strings to be filled in...
> 
> These are good questions. I think it's OK for now because these are just developer-facing options to continue experimentation. Maybe we should review how options are presented more generally, to clarify what is for practitioners vs. internal developers? 

We could rename the above options to `_nv_enable/disable_options` to hint these are not user-facing options?

>Maybe developer options should be better documented somewhere, too?

I can add a wiki to the nvfuser page listing our options.

t-vi:
I wonder if it would be useful to think about making this an option for NVFuser now.
The global options are not good UX and we should not be adding features that need an entire wiki.

mruberry:
> I wonder if it would be useful to think about making this an option for NVFuser now. The global options are not good UX and we should not be adding features that need an entire wiki.

I think what we need to follow-up on is that this isn't UX, it's DevX, and maybe we need to do something to make that clearer going forward.

lantiga:
Fair that it's DevX, but it will easily become UX if we don't pay attention. Like I can totally see these options percolate into reference models, benchmarks, etc just because they are there. So, we can merge, but then we will need to call out when these are actually used in the wild because it could spiral out of control.

mruberry (approved):
Let's follow-up with how we present developer options. Maybe a technical discussion?

Priya2698:
@t-vi @mruberry: We discussed the feedback on this PR briefly in the Thunder meeting before the holidays.
Does changing the above `nv_enable/disable_options` (that can consume a list of nvfuser supported options) to boolean flags `nv_enable_fuse_matmul` and `nv_disable_aten_matmul` (naming can be discussed later) sound better? These are the only two options that we intend to use from within Thunder.

mruberry:
> @t-vi @mruberry: We discussed the feedback on this PR briefly in the Thunder meeting before the holidays. Does changing the above `nv_enable/disable_options` (that can consume a list of nvfuser supported options) to boolean flags `nv_enable_fuse_matmul` and `nv_disable_aten_matmul` (naming can be discussed later) align with your feedback? These are the only two options that we intend to use from within Thunder.

That's a cool idea, @Priya2698! I think it's OK to leave things as-is for now, but in the future we may want to revisit how we present developer-facing options vs. practitioner-facing options more generally.

Issue #1278, opened 2024-10-09
Support Phi3-mini-128k


tfogal:
## 🚀 Model / language coverage

Support the https://huggingface.co/microsoft/Phi-3-mini-128k-instruct model. This is a tracking issue.

Dynamo is splitting this into 13 subgraphs. The good news is that `examine` doesn't find much for us:

* _exit_autocast of torch.amp.autocast_mode
* _enter_autocast of torch.amp.autocast_mode
* and examine notes that `TensorBase.bool` is automatically registered, which is probably fine.

Issues:
* #1276 

### Pitch

This is an ask from our NeMo colleagues.

### Minimal Repro

```python
import torch
from torch.utils.data import DataLoader
from transformers import AutoModelForCausalLM, AutoTokenizer, get_scheduler
from datasets import load_dataset
import thunder
import thunder.dynamo

# Define model and tokenizer
model_name = "microsoft/Phi-3-mini-128k-instruct"  # Replace with the model you want to fine-tune
model = AutoModelForCausalLM.from_pretrained(
  model_name,
  trust_remote_code=True,
  torch_dtype='auto'
)
model = torch.compile(model, dynamic=False,
                      backend=thunder.dynamo.ThunderCompiler())
tokenizer = AutoTokenizer.from_pretrained(model_name)

# Add a padding token to the tokenizer
if tokenizer.pad_token is None:
    tokenizer.add_special_tokens({'pad_token': '[PAD]'})
    model.resize_token_embeddings(len(tokenizer))  # Resize the model's embeddings to accommodate new tokens

# Load a smaller dataset
dataset = load_dataset("tiny_shakespeare", split='train')

# Tokenize the dataset
def tokenize_function(examples):
    return tokenizer(examples["text"], padding="max_length", truncation=True,
                     max_length=128)

tokenized_dataset = dataset.map(tokenize_function, batched=True, remove_columns=["text"])

# Convert the dataset to PyTorch format and specify columns to return as tensors
tokenized_dataset.set_format(type='torch', columns=['input_ids', 'attention_mask'])

# Create PyTorch DataLoader
dataloader = DataLoader(tokenized_dataset, batch_size=1, shuffle=False)

# Define optimizer and learning rate scheduler
optimizer = torch.optim.AdamW(model.parameters(), lr=5e-5)
num_epochs = 3
lr_scheduler = get_scheduler(
    "linear",
    optimizer=optimizer,
    num_warmup_steps=0,
[...]

PR #1439, opened 2024-11-14
Test HF's implementation of Phi 3 model

IvanYashchuk:
This PR adds a test to verify that the computed cross-entropy loss matches between HF's implementation and application of Thunder on this model.

Ref. https://github.com/Lightning-AI/lightning-thunder/issues/1278.


riccardofelluga (approved):
LGTM, it's better to have it parametrized then to have two different tests with almost the same code. HF transformers library is nice and flexible

t-vi (approved):
Thank you @IvanYashchuk @riccardofelluga

PR #1440, opened 2024-11-14
Add HF Phi3 to the test networks

riccardofelluga:
## What does this PR do?

This PR adds the HF Phi3-mini-128k model to the network test suite. 

It is part of #1278


thunder/tests/test_networks.py
```diff
@@ -445,6 +445,69 @@ def test_hf_qwen2():
     torch.testing.assert_close(grads_ref, grads_compiled, rtol=1e-2, atol=1e-2)
 
 
+PHI3_MINI_128K_CFG = {
```
IvanYashchuk, line 448: There's no need to copy the config here because it's not behind registration wall and downloadable with AutoConfig.

riccardofelluga:
Duplicate of #1439.

PR #1447, opened 2024-11-18
add extensible DebugOptions class

t-vi:
This intends to make debug options more discoverable:

- dynamic registration,
- automatic docstring support,
- also included in documentation.


lantiga (approved):
Looks great.

At first I thought we could provide a singleton that one could modify for better discoverability, but ultimately having objects with combinations of options I can store and swap during debugging is going to be a lot more effective.

PR #1450, opened 2024-11-18
support dynamic getitem in dictionary for symbolic values

jjsjann123:
The old logic assumes getitem in dict is always a constant.

With a script like this:
```
def foo(a, v):
    return a[v].relu()

a = {2: torch.randn(2, 2, device="cuda"), 8: torch.randn(3, 3, device="cuda")}
out = jfoo(a, 2)
```

Running into exceptions here:

```
NotImplementedError: Exception occured unpacking object from ProvenanceRecord(
  i1 = PseudoInst.INPUT_ARGS()
  i2 = PseudoInst.BINARY_SUBSCR(i1, 0)
  i3 = PseudoInst.BINARY_SUBSCR(i2, PseudoInst.CONSTANT([IntegerProxy name=i0, value=2, static=CONSTRAINT.CONSTRAINABLE]))
)
```

Since we now have `PseudoInst.CONSTANT([IntegerProxy name=i0, value=2, static=CONSTRAINT.CONSTRAINABLE])`, which tries to unwrap an IntegerProxy as constant.

The updated logic would preserve the wrapped value if it's holding a number proxy. So the above trace becomes something like this:
```
ProvenanceRecord(
  i1 = PseudoInst.INPUT_ARGS()
  i2 = PseudoInst.BINARY_SUBSCR(i1, 0)
  i3 = PseudoInst.BINARY_SUBSCR(i1, 1)
  i4 = PseudoInst.BINARY_SUBSCR(i2, i3)
)
```


t-vi:
The failure was the one mentioned in the CI weather in the `#thunder` slack.

PR #1452, opened 2024-11-19
fix distributed tests with pt main

t-vi:
PyTorch modified the test class to destroy the process group by default, but we want to do this ourselves.

### DIFF
--- a/thunder/tests/distributed/helper.py
+++ b/thunder/tests/distributed/helper.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 from functools import partial
 from functools import wraps
 from typing import ClassVar, TYPE_CHECKING
+import inspect
 import math
 import os
 import sys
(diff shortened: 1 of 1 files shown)


lantiga (approved):

thunder/tests/distributed/helper.py
```diff
@@ -129,10 +130,14 @@ def _run(cls, rank, test_name, file_name, pipe, *, fake_pg=False):
         local_rank = self.rank % torch.cuda.device_count()
         torch.cuda.set_device(local_rank)
         os.environ["LOCAL_RANK"] = str(local_rank)
+        if "destroy_process_group" in inspect.signature(self.run_test).parameters:
+            run_test_kwargs = {"destroy_process_group": False}
+        else:
+            run_test_kwargs = {}
```
IvanYashchuk, line 136: Can you explain how does this work?

I see failures of the TP tests with this PR. Another way to resolve the problem would be to redefine `destroy_pg_upon_exit` to return False instead of True.
https://github.com/pytorch/pytorch/blob/c418a9ac7501184d4a8cb4443f49e727d43f6be0/torch/testing/_internal/common_distributed.py#L580-L587

t-vi: Yeah, well, last week the way things worked were different.

PR #1453, opened 2024-11-19
fix: using te and fsdp leads to multiple device found error

kshitij12345:
Fixes the error below:

As reported by Mixology team,

`
NVFUSER_DISABLE=multidevice torchrun --standalone --max-restarts=0 --no-python --nproc-per-node=8 python /opt/pytorch/lightning-thunder/thunder/benchmarks/benchmark_litgpt.py --model_name Llama-3-8B --distributed_mode fsdp --shard_mode zero2 --compile inductor --checkpoint_activations False --low_precision_mode fp8-delayed-te --micro_batch_size 1 --bucketing_mode block
`

we see the error
```python
RuntimeError: FSDP only supports single device modules but got params on {device(type='cuda', index=2), device(type='meta')}
```

### DIFF
--- a/thunder/benchmarks/benchmark_litgpt.py
+++ b/thunder/benchmarks/benchmark_litgpt.py
@@ -114,11 +114,13 @@ def swap_linear_layers_for_te(model: torch.nn.Module, device: Any, swap_layernor
 
             if isinstance(m, torch.nn.Linear):
                 has_bias = m.bias is not None
-                new_linear = te.Linear(m.in_features, m.out_features, bias=has_bias, device=device)
+                # Pass device as str (as there is a bug in TransformerEngine's handling of torch.device)
+                new_linear = te.Linear(m.in_features, m.out_features, bias=has_bias, device=str(device))
                 setattr(module, n, new_linear)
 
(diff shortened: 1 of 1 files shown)


thunder/benchmarks/benchmark_litgpt.py
```diff
@@ -114,11 +114,13 @@ def _resursively_swap_linear_layers_for_te(module: torch.nn.Module) -> None:
 
             if isinstance(m, torch.nn.Linear):
                 has_bias = m.bias is not None
-                new_linear = te.Linear(m.in_features, m.out_features, bias=has_bias, device=device)
+                # Pass device as str (as there is a bug in TransformerEngine's handling of torch.device)
+                new_linear = te.Linear(m.in_features, m.out_features, bias=has_bias, device=str(device))
```
kshitij12345, line 118:
```python
import transformer_engine.pytorch as tep
import torch

linear = tep.Linear(16, 16, device="meta")

print(linear.weight.device)  # meta

linear = tep.Linear(16, 16, device=torch.device("meta"))

print(linear.weight.device)  # cuda:0

```

PR #1451, opened 2024-11-18
remove check_storage_aliases

jjsjann123:
The motivation is to remove the use of `numel` in code logic, since number of elements doesn't play well with dynamic shape yet.

I don't know what `check_storage_aliases` is protecting us against. So I removed the whole thing and looks like we don't have any test validating the logic neither.
This is a follow up on conversation here: https://github.com/Lightning-AI/lightning-thunder/pull/877/files#r1844732984
cc'ing @crcrpar for visibility.

PR #1455, opened 2024-11-20
needs to split with `-`

crcrpar:
## What does this PR do?

Fixes the lack of the split of string representing alias tensor indices.


kshitij12345 (commented):
I think we should add a test with multiple groups of aliased tensors (which I think would fail on current main).

crcrpar:
close this one following the removal -- ref: https://github.com/Lightning-AI/lightning-thunder/pull/1451

PR #1458, opened 2024-11-20
Raise when in place operations occur on leafs requiring grad

### LINKED ISSUE #1284: [inplace] Silently incorrect gradient when leaf variable is used in an inplace operation
```python
import torch
import thunder

def forward(x):
    return x.mul_(2)

x = torch.ones(3, requires_grad=True)
v = torch.ones_like(x, requires_grad=False)

# PyTorch Eager errors out.
# forward(x) # RuntimeError: a leaf Variable that requires grad is being used in an in-place operation.

jforward = thunder.jit(forward)
jforward(x).backward(v)
print(x.grad)  # tensor([1., 1., 1.])
```

beverlylytle:
<details>

</details>

## What does this PR do?

Fixes #1284

## PR review

## Did you have fun?


thunder/tests/test_inplace_functionalization.py
```diff
@@ -551,6 +547,8 @@ def single_tensor_adam(
 
     jitted = executor.make_callable(single_tensor_adam)
     params, grads, exp_avgs, exp_avg_sqs = tensors
+    cd = thunder.compile_data(jitted)
+    cd.compile_options["torch_compile_fullgraph"] = False
```
kshitij12345, line 551: I am curious as to why this is required. I think a comment will be helpful.

Also, I think `jitted = executor.make_callable(single_tensor_adam, torch_compile_fullgraph=False)` would work. So, we won't need to grab the compile data and update it.

thunder/tests/test_inplace_functionalization.py
```diff
@@ -476,31 +476,27 @@ def f(xs, ys, z):
     dtypes=NOTHING,
 )
 def test_inplace_to_tensors_with_grad(executor, device, _):
-    @torch.no_grad
```
kshitij12345, line 479: I am not sure what exactly do we test for here and also why did we get rid of `add_grad` case.

kshitij12345 (commented):
The fix looks good. We should add a small test to verify that this error raised when expected. Thanks @beverlylytle

thunder/tests/test_inplace_functionalization.py
beverlylytle: Done.

thunder/tests/test_inplace_functionalization.py
beverlylytle: The test came with the addition of the `grad` member variable for `TensorProxy`.  The `add_grad` case is removed because I was having trouble making the test make sense for even pytorch operations and I don't see what it adds on top of the `add_y` case.

kshitij12345: Actually I think this test should still work as is.

PyTorch allows update on leaf tensors under `no_grad`.
```python
import torch

@torch.no_grad  # Commenting will lead to error
def add_y(x, y):
    x.add_(y, alpha=0.1)

x = torch.randn(3, 3, requires_grad=True)
y = torch.randn(3, 3, requires_grad=True)

add_y(x, y)
```

Also, I think `add_grad` is important as it was added in the PR that added support for `grad` attribute on TensorProxy: https://github.com/Lightning-AI/lightning-thunder/pull/1070, so it seems to be testing that as well. (We should probably add comment that this is happening).

thunder/tests/test_inplace_copy.py
```diff
@@ -178,3 +178,11 @@ def func(T0):
...
     for o, o_ref in zip(o_thunder, o_eager):
         assert_close(o, o_ref)
+
+
+@instantiate(dtypes=datatypes.float_math_dtypes)
+def test_inplace_copy_of_leaf_requiring_grad_fails(executor, device, dtype):
+    tdtype = ttorch.to_torch_dtype(dtype)
+    a = make_tensor((4, 4), device=device, dtype=tdtype, requires_grad=True)
+    with pytest.raises(RuntimeError):
+        a.copy_(a)
```
kshitij12345, line 188: I think this is testing that PyTorch raises error when calling inplace on leaf tensor.

It should probably be something like

```python

def fn(a):
    a.copy_(a)

jfn = thunder.jit(fn)

with pytest.raises(RuntimeError, match="a leaf Variable"):
    jfn(a)

```

thunder/executors/torchex.py
```diff
@@ -2190,6 +2182,9 @@ def is_float_type(self, input):
 
 
 def _copy__impl(copy_from, copy_to):
+    cd = get_compile_data()
+    if cd is not None and cd.is_grad_enabled and copy_to.is_leaf and copy_to.requires_grad:
+        raise RuntimeError("a leaf Variable that requires grad is being used in an in-place operation.")
```
kshitij12345, line 2187: I am wondering if Symbol `copy_` in `thunder/torch/__init__.py` is more appropriate location for the check. 

https://github.com/Lightning-AI/lightning-thunder/blob/60f3ee1ec536ee8d6fdef503af54525e0a3978a4/thunder/torch/__init__.py#L1961-L1963

thunder/tests/test_inplace_functionalization.py
```diff
@@ -549,7 +545,8 @@ def single_tensor_adam(
     ref_state_steps = [torch.tensor(1, device=device) for _ in range(2)]
     single_tensor_adam(*ref_tensors, state_steps=ref_state_steps)
 
-    jitted = executor.make_callable(single_tensor_adam)
+    # torch.compile does not support accessing the ContextVariable compile data used in _copy__impl_
+    jitted = executor.make_callable(single_tensor_adam, torch_compile_fullgraph=False)
```
kshitij12345, line 549: Interesting that `torch.compile` creates a graph break when calling `get` on ContextVariable.

```python
import torch
from contextvars import ContextVar

_compile_data = ContextVar("compile_data", default=(None, None))

def fn(x):
    _compile_data.get()
    return x + 1

torch.compile(fn, fullgraph=False)(torch.randn(3, 3))  # Works with GraphBreak at _compile_data.get()
torch.compile(fn, fullgraph=True)(torch.randn(3, 3))  # Fails
```

thunder/executors/torchex.py
beverlylytle: `a` and `b` are proxies and it it not clear to me if a proxy knows that it is a leaf.

thunder/tests/test_inplace_copy.py
beverlylytle: True, whoops.

thunder/executors/torchex.py
IvanYashchuk: They do not. It's only a PyTorch concept that's available at runtime inside `_copy__impl`.

thunder/tests/test_inplace_functionalization.py
IvanYashchuk: What does Thunder's Interpreter do? It probably fails

kshitij12345: thunder just burns the value in trace (if used) without having a corresponding check in prologue. (Will file an issue for the same).

Eg.
```python
import torch
import thunder
from contextvars import ContextVar

_compile_data = ContextVar("compile_data", default=1)

def fn(x):
    v = _compile_data.get()
    return x + v

jfn = thunder.jit(fn)
o = jfn(torch.ones(3,))
print(o)  # tensor([2., 2., 2.])

_compile_data.set((2,))
o = jfn(torch.ones(3,))
print(o)  # tensor([2., 2., 2.])

print(thunder.last_prologue_traces(jfn)[-1])
# @torch.no_grad()
# @no_autocast
# def prologue(*args, **kwargs):
#   # args: "Any"
#   check_len(args, 1)
#     # prims.check_len(args, 1)
#   # kwargs: "Any"
#   check_len(kwargs, 0)
#     # prims.check_len(kwargs, 0)
#   x: "cpu f32[3]" = args[0]
#   check_tensor_metadata(x, (3,), 'cpu', torch.float32, False)
#     # prims.check_tensor_shape_and_metadata(x, (3,), 'cpu', torch.float32, False)
#   cache_info: "Any" = thunder._get_cache_info()
#   cache_info_default_dtype: "<class 'torch.dtype'>" = cache_info['default_dtype']
#   check_literal_like(cache_info_default_dtype, torch.float32)
#     # prims.check_literal_like(cache_info_default_dtype, torch.float32)
#   cache_info_default_device: "<class 'torch.device'>" = cache_info['default_device']
#   check_literal_like(cache_info_default_device, torch.device("cpu"))
#     # prims.check_literal_like(cache_info_default_device, torch.device("cpu"))
#   cache_info_is_autocast_enabled: "bool False" = cache_info['is_autocast_enabled']
#   check_number_type_and_value(cache_info_is_autocast_enabled, False)
#     # prims.check_number_type_and_value(cache_info_is_autocast_enabled, False)
#   cache_info_no_grad_sync: "bool False" = cache_info['no_grad_sync']
#   check_number_type_and_value(cache_info_no_grad_sync, False)
#     # prims.check_number_type_and_value(cache_info_no_grad_sync, False)
#   cache_info_alias_tensor_indices: "str" = cache_info['alias_tensor_indices']
#   check_string_value(cache_info_alias_tensor_indices, '')
#     # prims.check_string_value(cache_info_alias_tensor_indices, '')
#   cache_info_is_grad_enabled: "bool True" = cache_info['is_grad_enabled']
#   check_number_type_and_value(cache_info_is_grad_enabled, True)
#     # prims.check_number_type_and_value(cache_info_is_grad_enabled, True)
#   return ((x,), ())

print(thunder.last_traces(jfn)[-1])
# @torch.no_grad()
# @no_autocast
# def computation(x):
#   # x: "cpu f32[3]"
#   t0 = torch.add(x, 1, alpha=1)  # t0: "cpu f32[3]"
#     # t0 = ltorch.add(x, 1, alpha=1)  # t0: "cpu f32[3]"
#       # _ = prims.convert_element_type(1, float)
#       # t0 = prims.add(x, 1.0)  # t0: "cpu f32[3]"
#   return t0
```

kshitij12345: Issue filed at https://github.com/Lightning-AI/lightning-thunder/issues/1464

thunder/tests/test_inplace_functionalization.py
beverlylytle: OK, test restored, and, I dare say, even improved.

thunder/executors/torchex.py
beverlylytle: Indeed, compile_data was not available, but now it should be with the added context manager in thunder/__init__.py

kshitij12345: I think this is still incorrect because as discussed in https://github.com/Lightning-AI/lightning-thunder/issues/1486, the value of compile_data.is_grad_enabled here would be that of last updated state which can lead to incorrectness when used outside of tracing context.

We can see the discrepancy here.
```python
import torch
import thunder

x = torch.randn(3, 3, requires_grad=True)

@torch.no_grad
def fn(x):
  return x.add_(1)

fn(x)  # This works

thunder.jit(fn)(x)  # This raises error
```

So, whether the `copy` is in no_grad region needs to be captured during the tracing time.

beverlylytle: Right, this is why I created the other issue.  This PR fixes the leaf/grad issue when there is no annotation.  When there is an annotation, another approach is required.  This other approach may or may not involve using compile data in `_copy__impl`.

As far as I understand, compile data is the medium for passing around data such as whether grad is enabled.  But as the other issue points out, compile data reflects the end state of a function call and not the "live" state, at least at the time it reaches `_copy__impl`.  So I'm left with the questions "are there other mechanisms for passing around whether grad is enabled?" "where else in the execution is it simultaneously knowable that a (1) leaf tensor (2) requiring grad is being (3) copied when (4) grad is enabled?" "is it feasible/desirable to make the compile data more dynamic?"  "is there a way to context-manage the tensors so that their `requires_grad` flags are set to `False` when the interpreter sees `torch._C._set_grad_enabled(False)`, and then later restored, thereby obviating the need for the compile data for this check?"  Do you have suggestions for a fix that addresses both issues? Or can we close out this issue and move the discussion to the more involved issue?

kshitij12345: So to tackle - leaf tensor requiring grad being copied into when grad is enabled, I think similar to a previous commit, 
we can update `prims.copy` to take a argument `is_grad_enabled`. With this, `ltorch.copy` will query `cd.is_grad_enabled` and call `prims.copy` by also passing this argument.

https://github.com/Lightning-AI/lightning-thunder/blob/9de543417260bd32740e2d7bd0763a5871fdee5c/thunder/torch/__init__.py#L1984-L1986

With these changes, the `copy_impl`'s signature will also change to accept `is_grad_enabled` and it will be called at runtime with a tensor which we can query if it is a leaf and also whether grad was enabled or not when calling that particular copy. Wdyt @beverlylytle?

Though, I am curious if there is another approach to this - cc: @IvanYashchuk

beverlylytle: Let's see what the CI thinks.

thunder/executors/nvfuserex_impl.py
```diff
@@ -2085,6 +2087,7 @@ def copy_(
     *,
     fd: FusionDefinition,
     lc_to_nv_map: dict,
+    grad_enabled: bool,
```
kshitij12345, line 2090: What is the behaviour for `nvfuser`? I think that we ignore this argument. Should we raise a warning instead?

thunder/tests/test_inplace_functionalization.py
kshitij12345: With the new implementation, we don't use `CompileData` in `copy_impl`, does this test pass now with the default value i.e.  torch_compile_fullgraph=True.

thunder/torch/__init__.py
```diff
@@ -1983,7 +1983,8 @@ def copysign_(a, b, /):
 
 @torchsymbol(torch.Tensor.copy_, is_method=True)  # , tags=(prims.OpTags.IN_PLACE,))
 def copy_(a, b, /):
-    return prims.copy_(b, a)
+    cd = get_compile_data()
+    return prims.copy_(b, a, grad_enabled=cd.is_grad_enabled if cd is not None else False)
```
kshitij12345, line 1987: `if cd is None` (probably happens for `thunder.trace` with default arguments), should we assume that we are running with `grad_enabled`? I think that it is likely case. Wdyt?

cc: @IvanYashchuk

kshitij12345 (commented):
Overall looks good to me, I just have a couple of questions. Thank you @beverlylytle

thunder/torch/__init__.py
beverlylytle: I'm don't have the background to have an opinion on this.  I defer to you @kshitij12345 and @IvanYashchuk.

thunder/tests/test_inplace_functionalization.py
beverlylytle: yes.

thunder/executors/nvfuserex_impl.py
beverlylytle: Yes, the argument is ignored, and nvfuser does not fail.

thunder/torch/__init__.py
IvanYashchuk: `cd` is our controlled way of specifying and querying the state of PyTorch. If it's None, I don't think we should do anything special. It's the responsibility of the outside system to set up a correct `cd` object. `is_grad_enabled` is a sensible default because if nothing else is specified, we should assume we are executing a program as given in the "inference" mode with no additional side transformations.

thunder/core/prims.py
```diff
@@ -4030,6 +4030,8 @@ def embedding_backward_meta(grad, indices, num_weights, padding_idx, scale_grad_
 def copy__meta(
     copy_from: TensorProxy,
     copy_to: TensorProxy,
+    *,
+    grad_enabled: bool = False,
```
IvanYashchuk, line 4034: To be as simple as possible prims should not have default values.

The current choice works well with all existing code but the introduced behavior of raising warnings for nvFuser-executed copies is not working:
 ```py
import torch
import thunder

test = torch.nn.BatchNorm3d(2, track_running_stats=True).to("cuda")
x = torch.ones((1, 2, 3, 4, 12), device="cuda")
jtest = thunder.jit(test)

out = jtest(x) # No warning
print(thunder.last_traces(jtest)[-1]) # We can see in the trace that grad_enabled is set to False
```

File: thunder/executors/nvfuserex_impl.py [hunk 9]
```diff
@@ -2093,7 +2093,11 @@ def var_mean(
 def _copy__check(
     copy_from: TensorProxy,
     copy_to: TensorProxy,
+    *,
+    grad_enabled: bool,
 ) -> bool:
+    if grad_enabled:
+        warnings.warn("An inplace operation requiring grad is occurring on a variable which may be a leaf.")
```
IvanYashchuk, line 2100: We can easily query at the construction time of input TensorProxy whether the corresponding PyTorch tensors have `is_leaf` set to True similar to how we do this for `requires_grad`:
https://github.com/Lightning-AI/lightning-thunder/blob/eff35942f52156345629a04182564efe9e209782/thunder/core/proxies.py#L1993

With this information on the proxy, we can raise a real error.

@beverlylytle, could you please create an issue about this to discuss whether it's a good idea to do so?

thunder/executors/torchex.py
IvanYashchuk: I agree with modifying `thunder.torch.copy` to query `cd.is_grad_enabled` and passing that to `prims.copy`.

thunder/core/prims.py
IvanYashchuk: It would be better not to use a default value here and replace all occurrences of `prims.copy_` with `thunder.torch.copy_`.

thunder/tests/test_inplace_functionalization.py
```diff
@@ -549,6 +549,7 @@ def single_tensor_adam(
     ref_state_steps = [torch.tensor(1, device=device) for _ in range(2)]
     single_tensor_adam(*ref_tensors, state_steps=ref_state_steps)
 
+    # torch.compile does not support accessing the ContextVariable compile data used in _copy__impl_
```
IvanYashchuk, line 552: This comment is outdated as `_copy__impl_` doesn't access the compile data, could you please remove it?

thunder/tests/test_inplace_functionalization.py
```diff
@@ -477,15 +477,15 @@ def f(xs, ys, z):
...
 
     @torch.no_grad
-    def add_grad(x, y):
-        x.add_(x.grad, alpha=0.1)
+    def add_y(x, y):
+        return x.add_(y, alpha=0.1)
 
-    for f in (add_y, add_grad):
-        jitted_f = executor.make_callable(f)
+    for fn in (add_grad, add_y):
```
IvanYashchuk, line 487: The order of functions is changed, does it affect anything? If not let's revert the changes to this file to keep the diff smaller.

thunder/executors/nvfuserex_impl.py
```diff
@@ -2093,7 +2093,11 @@ def var_mean(
 def _copy__check(
     copy_from: TensorProxy,
     copy_to: TensorProxy,
+    *,
+    grad_enabled: bool,
 ) -> bool:
+    if grad_enabled:
```
IvanYashchuk, line 2099: Effectively it would mean raising a warning for every invocation of batch norm training. We should raise warnings if they are useful to users, here it would be just noisy. Let's not do that for now, but there's a path of raising a real error if we can query `is_leaf` on input tensor proxies.

thunder/tests/test_inplace_functionalization.py
beverlylytle: No the order does not matter, except for readability.  When I see something in not alphabetical order, I assume that means that the order ~ does ~ matter.  But ok, smaller diff...

thunder/examine/__init__.py
```diff
@@ -238,6 +238,8 @@ def get_nvfuser_fusion_definition(trace: TraceCtx, name: str, warn_if_fusion_una
         return None
 
     for bsym in trace.bound_symbols:
+        if bsym.sym.is_fusion:
+            print(bsym.sym.name)
```
crcrpar, line 242:
```suggestion
```

beverlylytle: thanks, removed.

thunder/torch/__init__.py
```diff
@@ -222,6 +222,16 @@ def register_function(torchfn, thunderfn_impl):
...
+def _copy_(a, b, /):
+    cd = get_compile_data()
+    return prims.copy_(b, a, grad_enabled=cd.is_grad_enabled if cd is not None else False)
+
+
+@torchsymbol(torch.Tensor.copy_, is_method=True)  # , tags=(prims.OpTags.IN_PLACE,))
+def copy_(a, b, /):
+    return _copy_(a, b)
+
+
```
beverlylytle, line 234: Consider the following snippet:
```
import thunder
import torch


x = torch.rand((2,3), dtype=torch.float32, device='cuda')

def f(x):
    return x.to(torch.float64).sin_()
```

Here a new tensor of type float64 is being created and an inplace operation is occurring on it.  Explicitly:
```
def f(x):
    y = x.to(torch.float64)
    return y.sin_()
```
One might expect that the following would be a less efficient, but still more or less equivalent version of the above:
```
def g(x):
    y = x.to(torch.float64)
    z = y.sin()
    return y.copy_(z)
```
However, for 
```
jf = thunder.jit(f); jg = thunder.jit(g)
```
`jf(x)` executes successfully while `jg(x)` results in 
```
An error occurred while defining nvFuser FusionDefinition None.
If you believe this is a bug or need assistance, please file an issue at https://github.com/NVIDIA/Fuser/issues/new
Here's a script to reproduce the error:
```python
# CUDA devices:
#  0: NVIDIA RTX 6000 Ada Generation
# torch version: 2.5.1+cu124
# cuda version: 12.4
# nvfuser version: 0.2.23+gitd53be45
import torch

def nvfuser_incomplete_fusion(fd : FusionDefinition) -> None :
    T0 = fd.define_tensor(shape=[2, 3], contiguity=[True, True], dtype=DataType.Float, is_cpu=False, stride_order=[1, 0])
    T1 = fd.ops.cast(T0, dtype=DataType.Double)
    T2 = fd.ops.sin(T1)
    T3 = fd.ops.set(T2)
    fd.add_output(T3, T1)
    fd.add_output(T1)

    nvfuser_fusion_idNone(fd)
```<adding extra characters for md>
  File "/home/blytle/miniforge3/envs/thdrs/lib/python3.10/site-packages/nvfuser/__init__.py", line 105, in __exit__
    self._finalize_definition()
RuntimeError:  INTERNAL ASSERT FAILED at "/workspace/Fuser/csrc/python_frontend/fusion_state.cpp":141, please report a bug with repro script to NVFuser at https://github.com/NVIDIA/Fuser/issues. 
Detected exception while building Fusion Ir. The failing RecordFunctor is: fd.add_output(T3, T1)
NvFuser error message is:  INTERNAL ASSERT FAILED at "/workspace/Fuser/csrc/fusion.cpp":784, please report a bug with repro script to NVFuser at https://github.com/NVIDIA/Fuser/issues. alias source can only be a fusion input
Exception raised from aliasOutputToInput at /workspace/Fuser/csrc/fusion.cpp:784 (most recent call first):
….
```
nvFuser does not like inplace operations performed on tensors which are not input to the fusion definition.  Happily, they are usually functionalized away in during tracing.  The functionalization pass makes assumptions about what an inplace op looks like.  Using a version of `copy_` annotated by `torchsymbol` within the other  inplace ops, like `sin_` , breaks those assumptions and leads to many tests failing with errors as above.  Hence, the split between `_copy_` and `copy_`.  

As a side note, I was also surprised to discover that 
```
def f(x):
    x.sin_()
    return x

def g(x):
    z = torch.sin(x)
    x.copy_(z)
    return x


x = torch.rand((2, 2), device='cuda', dtype=torch.float64)

jf = thunder.jit(f)
jf(x)                  # this is fine
jg = thunder.jit(g)
jg(x)                  # fails with an AssertionError on "assert return_bsym.sym.id == prims.PrimIDs.RETURN"
```

thunder/executors/nvfuserex_impl.py
beverlylytle: done. #1577 1577

thunder/core/prims.py
beverlylytle: Removed.

thunder/torch/__init__.py
```diff
@@ -2241,7 +2246,7 @@ def true_divide(a: NumberLike | TensorLike, b: NumberLike | TensorLike, /) -> Nu
 
 @torchsymbol(torch.Tensor.true_divide_, is_method=True, tags=(prims.OpTags.IN_PLACE,))
 def true_divide_(a: TensorLike, b: NumberLike | TensorLike, /) -> TensorLike:
-    return prims.copy_(true_divide(a, b))
```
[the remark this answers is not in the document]
t-vi: That probably means it's not tested yet. :)

t-vi:
@IvanYashchuk what's your view on this? I'd be happy to merge as is and iterate in a next PR if needed, but I'll happily go with what you prefer.

thunder/torch/__init__.py
IvanYashchuk: Interesting discoveries! 
> nvFuser does not like inplace operations performed on tensors which are not input to the fusion definition

That's a good reason to build narrow-in-scope functionalization in Thunder. When no view operations are involved it's easy to do. When views are involved we need to finalize implementing the approach described in https://docs.google.com/document/d/1TPrPuTbfz7uHjlYSWLhLrLWGZdI2_d9jf-RBPoU6q1I/edit?tab=t.0

In this example, the function `g` could be transformed by Thunder into something that would result in 
```py
def nvfuser_incomplete_fusion(fd : FusionDefinition) -> None :
    T0 = fd.define_tensor(shape=[2, 3], contiguity=[True, True], dtype=DataType.Float, is_cpu=False, stride_order=[1, 0])
    T1 = fd.ops.cast(T0, dtype=DataType.Double)
    T2 = fd.ops.sin(T1)
    T3 = fd.ops.set(T2)
    fd.add_output(fd.ops.set(T3)) # was fd.add_output(T3, T1)
```
because the intermediate `y` is created inside the fusion and used once, we can think of this copy_ call as a variable assignment.

IvanYashchuk (approved):
I apologize for the delay. Let's merge this. Looks great!
Thank you for going through multiple rounds of review and testing different approaches!

In the future, we will be able to query more per-tensor information that's available in PyTorch and the first step is in this draft https://github.com/Lightning-AI/lightning-thunder/pull/1599.

t-vi (approved):
Thank you @beverlylytle @crcrpar @kshitij12345 @IvanYashchuk

t-vi:
This needed to be reverted because it broke main. Please rebase and open a new PR to re-land. Thank you.

Issue #1299, opened 2024-10-14
default transforms for thunder.jit


kshitij12345:
Similar to default executors, we should have a good default for transforms to `thunder.jit`. This issue is to discuss how an API for the same would look like.

One option is that it could look similar to `get_default_executors() -> List[Executor]`. It makes more sense for executors as the higher priority executor will pick-up an operator (and lower priority would be ignored). In case of transforms, the transforms will applied in order and order of transform application maybe important. However, note that this is how `thunder.jit` takes in transforms currently (see [docs](https://lightning-thunder.readthedocs.io/en/latest/reference/generated/thunder.jit.html#thunder.jit)), and applies them in the order they appear.

Context - We may want to make the ConstantFolding transform a default and this issue is to facilitate the discussion around the same. See https://github.com/Lightning-AI/lightning-thunder/issues/1250 and https://github.com/Lightning-AI/lightning-thunder/pull/1273

tfogal:
@kshitij12345 is the constant folding as currently committed applied by default on at least the ThunderFX path, today? Or would we need this or manual intervention to apply that pass at all?

kshitij12345:
> is the constant folding as currently committed applied by default on at least the ThunderFX path, today?

It is not on by default even on ThunderFX path. For now, we can manually specify the pass with

```python
from thunder.transforms.constant_folding import ConstantFolding
backend = ThunderCompiler(transforms=[ConstantFolding(),])
```

tfogal:
ack, appreciate the fast response! but also, take a break, it's late on a Friday ;-)

PR #1468, opened 2024-11-25
Include ConstantFolding in the default list of transforms used in thunder.jit

IvanYashchuk:
According to the discussion in https://github.com/Lightning-AI/lightning-thunder/issues/1299 this transform is not used by default and requires users to opt-in.
This PR makes this transform to be used by default.


kshitij12345:
Looks like we will need to update `ConstantFolding` to fix the failing tests.

t-vi:
I think this would need to have more discussion, both for whether we want to do this in general, and for the implementation.

Issue #1424, opened 2024-11-12
ThunderFX fails with FP8 and Activation Checkpointing


mpatel31415:
## 🐛 Bug

When training models: 'vicuna-7b-v1.5-16k', 'longchat-13b-16k', 'Mistral-7B-v0.2', 'falcon-180B', 'Llama-3-70B', 'CodeLlama-34b-hf' with FSDP and FP8 we get KeyError: 'scaling_fwd'. This might be also issue with Transformer Engine,, so I'm happy to move this issue to TE if needed. 

Full traceback: 

> [rank7]: Traceback (most recent call last):
7: [rank7]:   File "/workspace/lightning-thunder/thunder/benchmarks/benchmark_litgpt.py", line 974, in <module>
7: [rank7]:     CLI(benchmark_main)
7: [rank7]:   File "/usr/local/lib/python3.12/dist-packages/jsonargparse/_cli.py", line 96, in CLI
7: [rank7]:     return _run_component(components, init)
7: [rank7]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
7: [rank7]:   File "/usr/local/lib/python3.12/dist-packages/jsonargparse/_cli.py", line 204, in _run_component
7: [rank7]:     return component(**cfg)
7: [rank7]:            ^^^^^^^^^^^^^^^^
7: [rank7]:   File "/workspace/lightning-thunder/thunder/benchmarks/benchmark_litgpt.py", line 871, in benchmark_main
7: [rank7]:     benchmark.train()
7: [rank7]:   File "/workspace/lightning-thunder/thunder/benchmarks/benchmark_litgpt.py", line 765, in train
7: [rank7]:     loss.backward()
7: [rank7]:   File "/usr/local/lib/python3.12/dist-packages/torch/_tensor.py", line 624, in backward
7: [rank7]:     torch.autograd.backward(
7: [rank7]:   File "/usr/local/lib/python3.12/dist-packages/torch/autograd/__init__.py", line 347, in backward
7: [rank7]:     _engine_run_backward(
7: [rank7]:   File "/usr/local/lib/python3.12/dist-packages/torch/autograd/graph.py", line 825, in _engine_run_backward
7: [rank7]:     return Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
[...]

PR #1473, opened 2024-11-26
Fix TransformerEngine and Activation checkpointing interaction for thunderFX

kshitij12345:
Fixes https://github.com/Lightning-AI/lightning-thunder/issues/1424

TODO - Add test and run sanity check


kshitij12345:
Fix is incorrect and this should be fixed with the new TE executor with Operation based API.

Issue #1479, opened 2024-11-26
NotImplementedError: VJP for PrimIDs.ITEM is not implemented for ThunderFX and litgGPT models.


mpatel31415:
## 🐛 Bug

For Llama-3-8B, Mistral-7B-v0.1 and Phi-3-mini-4k-instruct we get the following error:

> 6: [rank6]: Traceback (most recent call last):
6: [rank6]:   File "/workspace/lightning-thunder/thunder/benchmarks/benchmark_litgpt.py", line 980, in <module>
6: [rank6]:     CLI(benchmark_main)
6: [rank6]:   File "/usr/local/lib/python3.12/dist-packages/jsonargparse/_cli.py", line 96, in CLI
6: [rank6]:     return _run_component(components, init)
6: [rank6]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
6: [rank6]:   File "/usr/local/lib/python3.12/dist-packages/jsonargparse/_cli.py", line 204, in _run_component
6: [rank6]:     return component(**cfg)
6: [rank6]:            ^^^^^^^^^^^^^^^^
6: [rank6]:   File "/workspace/lightning-thunder/thunder/benchmarks/benchmark_litgpt.py", line 877, in benchmark_main
6: [rank6]:     benchmark.train()
6: [rank6]:   File "/workspace/lightning-thunder/thunder/benchmarks/benchmark_litgpt.py", line 748, in train
6: [rank6]:     logits = self.model(input_ids)
6: [rank6]:              ^^^^^^^^^^^^^^^^^^^^^
6: [rank6]:   File "/usr/local/lib/python3.12/dist-packages/torch/nn/modules/module.py", line 1739, in _wrapped_call_impl
6: [rank6]:     return self._call_impl(*args, **kwargs)
6: [rank6]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
6: [rank6]:   File "/usr/local/lib/python3.12/dist-packages/torch/nn/modules/module.py", line 1750, in _call_impl
6: [rank6]:     return forward_call(*args, **kwargs)
6: [rank6]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
6: [rank6]:   File "/usr/local/lib/python3.12/dist-packages/torch/_dynamo/eval_frame.py", line 573, in _fn
6: [rank6]:     return fn(*args, **kwargs)
[...]

PR #1481, opened 2024-11-27
Register grad for torch.Tensor.item

IvanYashchuk:
The rule is simple because PyTorch doesn't support propagating grads through item calls as it returns a disconnected from PyTorch Autograd graph Python scalar.

Fixes https://github.com/Lightning-AI/lightning-thunder/issues/1479.


t-vi (approved):
Thank you @IvanYashchuk 

Do you also want to enable the tests that were failing on the item vjp?

IvanYashchuk:
> Do you also want to enable the tests that were failing on the item vjp?

No, not in this pull request.

PR #1483, opened 2024-11-27
remove redundant `if backward_trc is None`

crcrpar:
## What does this PR do?

As per title, `if backward_trc is None` check does not look necessary.


t-vi (approved):
Thank you @crcrpar

PR #1485, opened 2024-11-27
[lookaside][torch.autograd.Function lookaside] use `shallow_copy` iff forward is empty

crcrpar:
## What does this PR do?

As per #1221 description, we don't strongly need `shallow_copy` if a trace representing a custom `torch.autograd.Function.forward` has multiple bsyms.


mruberry:
@t-vi would you like to review this?

PR #1415, opened 2024-11-08
[torchao float8tensor] 

crcrpar:
<details>
  <summary><b>Before submitting</b></summary>

- [ ] Was this discussed/approved via a Github issue? (no need for typos and docs improvements)
- [ ] Did you read the [contributor guideline](https://github.com/Lightning-AI/pytorch-lightning/blob/main/.github/CONTRIBUTING.md), Pull Request section?
- [ ] Did you make sure to update the docs?
- [ ] Did you write any new necessary tests?

</details>

## What does this PR do?

Fixes # (issue).

## PR review

Anyone in the community is free to review the PR once the tests have passed.
If we didn't discuss your PR in Github issues there's a high chance it will not be merged.

## Did you have fun?

Make sure you had fun coding 🙃

Issue #1487, opened 2024-11-27
`torch.transpose` seems to be mapped to different ops depending on `requires_grad`


crcrpar:
*Note*: If you have a model or program that is not supported yet but should be, please use the program coverage template.

## 🐛 Bug

I admit I'm not sure if this is a bug of an expected behavior, but `torch.transpose` is mapped to `torch.transpose` if `requires_grad=False`, `torch.permute` otherwise.

#### Code sample

```python
import torch
import thunder


@thunder.jit
def f(x):
    return x.transpose(0, 1)


for requires_grad in (False, True):
    print("#" * 120)
    x = torch.rand((4, 2), requires_grad=requires_grad)
    f(x)

    print(f"### {requires_grad = }")
    print(f"### thunder.last_traces(f)[-1]\n{thunder.last_traces(f)[-1]}")
```

#### output

```
########################################################################################################################
### requires_grad = False
### thunder.last_traces(f)[-1]
# Constructed by Unwrap the actual return value
import torch
from thunder.executors.torchex import no_autocast

@torch.no_grad()
@no_autocast
def computation(x):
  # x: "cpu f32[4, 2]"
  t0 = torch.transpose(x, 0, 1)  # t0: "cpu f32[2, 4]"
    # t0 = ltorch.transpose(x, 0, 1)  # t0: "cpu f32[2, 4]"
      # t0 = prims.transpose(x, (1, 0))  # t0: "cpu f32[2, 4]"
  return t0
########################################################################################################################
### requires_grad = True
### thunder.last_traces(f)[-1]
# Constructed by Delete Last Used (took 0 milliseconds)
import torch
from thunder.executors.torchex import no_autocast

@torch.no_grad()
@no_autocast
def computation(x):
  # x: "cpu f32[4, 2]"
  t0 = torch.permute(x, (1, 0))  # t0: "cpu f32[2, 4]"
    # t0 = ltorch.permute(x, (1, 0))  # t0: "cpu f32[2, 4]"
      # t0 = prims.transpose(x, (1, 0))  # t0: "cpu f32[2, 4]"
  return {'output': t0, 'flat_args': [x], 'flat_output': (t0,)}, ((), ())
```

### Environment

- PyTorch Version (e.g., 1.0):
- OS (e.g., Linux):
- How you installed PyTorch (`conda`, `pip`, source):
- Build command you used (if compiling from source):
- Python version:
- CUDA/cuDNN version:
- GPU models and configuration:
[...]

PR #1489, opened 2024-11-28
Unrolling tensor subclasses in fwd/bwd split

crcrpar:
## What does this PR do?

In #1415 and #1394, tensor subclasses and their `__torch_dispatch__` are unrolled before forward-backward split.
It turned out that we want to postpone it in the split as the unrolling seems to be harmful to backward generation.

TODO
- [x] support no autograd cases

note: https://github.com/pytorch/ao/pull/1339 is used

### DIFF
--- a/thunder/core/jit_ext.py
+++ b/thunder/core/jit_ext.py
@@ -637,7 +637,7 @@ def _convert_pytorchfunc_to_thundertrace(
     trace = TraceCtx()
     trace.bound_symbols.extend(active_jit_ctx.computation_trace.pop_scope())
     func_result = unwrap(wrapped_func_result)
-    if shallow_copy_output:
+    if shallow_copy_output and not trace.bound_symbols:
         from thunder.core.baseutils import sequencify
 
         out_to_shallow_copy: dict[Variable, TensorProxy] = {}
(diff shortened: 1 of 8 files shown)


thunder/core/jit_ext.py
```diff
@@ -637,7 +637,7 @@ def _convert_pytorchfunc_to_thundertrace(
     trace = TraceCtx()
     trace.bound_symbols.extend(active_jit_ctx.computation_trace.pop_scope())
     func_result = unwrap(wrapped_func_result)
-    if shallow_copy_output:
+    if shallow_copy_output and not trace.bound_symbols:
```
crcrpar, line 640: copy from #1485

thunder/torch/__init__.py
```diff
@@ -1312,6 +1314,17 @@ def transpose(a: TensorLike, /, dim0: int, dim1: int) -> TensorLike:
...
 
+def _transpose_grad(a: TensorLike, /, dim0: int, dim1: int) -> TensorLike:
+    fwd = transpose(a, dim0, dim1)
+    g = get_grad(fwd)
+    a_grad = transpose(g, dim0, dim1)
+    put_grad(a, a_grad)
+    return fwd
+
+
+register_grad(transpose, _transpose_grad)
```
crcrpar, line 1325: rel: #1487

needed to avoid `prims.permute`

Issue #1131, opened 2024-09-09
look at transform_for_execution for generation torch compile callables


t-vi:
Currently, the torch compile executors have their own way of creating callables operator by operator.
It would be more principled to use transform for execution to do so, however when #1041 tried, it needed to be reverted due to correctness issues. It would be good to find and fix the cause.

PR #1500, opened 2024-12-02
use transform for execution to get torch_compile executable

t-vi:
Transform for execution used to produce duplicate names, this changes the implementation to be close to interpret_trace (with the plan of merging the two eventually).

Unfortunately, there seems to be some bad interaction, the newly skipped tests have been leading to segfaults, this needs investigation. 

Fixes: #1131
Also:
Fixes: #1501

### DIFF
--- a/thunder/core/trace_interpreter.py
+++ b/thunder/core/trace_interpreter.py
@@ -207,6 +207,22 @@ def interpret_trace_to_trace(trace, *args, symbol_mapper=None, with_env=False, *
 
 
 class TraceSubstitutionProcessor:
+    """This processes a trace in an interpretation-style way by looping over the bound symbols.
+    This processing aims to preserve as much information on the proxies as possible.
+
+    Args:
+        trace: trace to process
+        *args: arguments to process the trace with
+        **kwargs: keyword arguments to process the trace with
+
+    The user is expected to subclass the trace and implement process_bsym with the help of add_unprocessed_bsyms (useful eg for using subsymbols to compute a symbol), add_processed_bsyms, and add_bsyms_from_function.
+
+    Calling the instantiated object initiates the processing and returns
+    the new trace and a mapping of the outputs.
+
+    See the OpExProcessor in thunder.executors.passes._transform_for_operator_executor_execution for an example of subclassing.
+    """
+
     NULL = object()
 
     def __init__(self, trace, *args, **kwargs):
@@ -267,7 +283,7 @@ class TraceSubstitutionProcessor:
     def add_unprocessed_bsyms(self, bsyms):
         self.unprocessed_bsyms[:0] = bsyms
 
-    def bsyms_from_function(self, fn, /, *args, **kwargs):
+    def add_bsyms_from_function(self, fn, /, *args, **kwargs):
         self.new_trace.push_scope([])
         result = fn(*args, **kwargs)
         self.new_bsyms += self.new_trace.pop_scope()
@@ -275,8 +291,6 @@ class TraceSubstitutionProcessor:
         return result
 
     def add_processed_bsyms(self, bsyms):
-
-        ### replacements of inputs!
         self.new_bsyms += bsyms
 
     def set_result(self, result):
@@ -298,8 +312,6 @@ class TraceSubstitutionProcessor:
             safe_map_flat(self.write, list(self.trace.kwargs.values()), list(kwargs.values()))
 
     def __call__(self):
-        # if not self.have_processed_args and self.trace.args is not None:
-        #    self.process_args(*self.args, **self.kwargs)
         with tracectx(self.new_trace):
             self.unprocessed_bsyms = self.trace.bound_symbols[:]
 
--- a/thunder/core/trace_interpreter.py
+++ b/thunder/core/trace_interpreter.py
@@ -203,3 +203,161 @@ def interpret_trace_to_trace(trace, *args, symbol_mapper=None, with_env=False, *
         return new_trace, tree_map(read, trace.output), env
 
     return new_trace, tree_map(read, trace.output)
+
+
+class TraceSubstitutionProcessor:
+    NULL = object()
+
+    def __init__(self, trace, *args, **kwargs):
+        self.env = {}
+        self.trace = trace
+        self.new_trace = from_trace(self.trace)
+        self.have_processed_args = False
+        print(self.trace)
+
+    def read(self, x: VariableInterface | Any) -> Any:
+        if isinstance(x, VariableInterface):
+            return self.env[x.name]
+        else:
+            return x
+
+    def write(self, v: VariableInterface | Any, val: Any) -> None:
+        if not isinstance(v, VariableInterface):
+            return
+        # Duplicates are allowed and overwritten
+        if v.name in self.env:
+            raise ValueError(f"Variable {v.name} is being overwritten this is not allowed")
+        self.env[v.name] = val
+
+    def add_to_swap_map(self, old, new):
+        if old is new:
+            return
+        if isinstance(old, ProxyInterface):
+            if isinstance(new, ProxyInterface) and variableify(new) in self.env:
+                # the new isn't new, but something returned the input
+                # this means we need to map the old to the new
+                old, new = new, old
+            elif isinstance(old, TensorProxyInterface):
+                # should we have a fix shapes pass? the sharding
+                # (FSDP, tensor parallel) transforms do "break" shape metadata
+                self.new_trace.names.remove(old.name)  # taken by the .replace proxy
+                if isinstance(new, VJPDual):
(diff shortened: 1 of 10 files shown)


thunder/core/trace_interpreter.py
```diff
@@ -203,3 +204,167 @@ def do_swap(v):
...
+
+    def add_to_swap_map(self, old, new):
+        if old is new:
+            return
+        if isinstance(old, ProxyInterface):
+            if isinstance(new, ProxyInterface) and variableify(new) in self.env:
+                # the new isn't new, but something returned the input
+                # this means we need to map the old to the new
+                old, new = new, old
+            elif isinstance(old, TensorProxyInterface):
```
mruberry, line 242: Maybe in the future we won't need to have logic specific to VJPDuals, for example, here

thunder/core/transform_common.py
```diff
@@ -111,6 +111,24 @@ def check(inp, log_str):
             check(copy_to_out, "output")
 
 
+def remove_duplicate_number_proxies(bsyms):
```
mruberry, line 114: Comment or docstring here explaining the function would be nice

thunder/core/transform_common.py
```diff
@@ -174,7 +192,9 @@ def _helper(x):
                 needed_proxies.add(variableify(x))
 
     dcetrace = from_trace(trace)
-    dcetrace.bound_symbols = list(reversed(dced))
+    dced_bound_symbols = list(reversed(dced))
+    dced_bound_symbols = dced_bound_symbols
```
mruberry, line 196: What's up with this line?

thunder/core/transform_common.py
```diff
@@ -174,7 +192,9 @@ def _helper(x):
...
+    dcetrace.bound_symbols = remove_duplicate_number_proxies(dced_bound_symbols)
```
mruberry, line 197: This is so interesting. Would you add a comment here, or maybe to the description of the dce pass itself, about how duplicate number proxies might happen? What does it look like to have them and then to remove them?

fyi @jjsjann123

thunder/core/trace_interpreter.py
```diff
@@ -203,3 +204,167 @@ def do_swap(v):
         return new_trace, tree_map(read, trace.output), env
 
     return new_trace, tree_map(read, trace.output)
+
+
+class TraceSubstitutionProcessor:
```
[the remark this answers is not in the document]
mruberry: (Maybe with a description of how it should be subclasses, like how the `OpExProcessor` subclasses it)

thunder/executors/passes.py
```diff
@@ -61,68 +62,54 @@ def preserve_bsym(bsym: BoundSymbol) -> Any:
...
-    bound_symbols: list[BoundSymbol] = []
-    for bsym in extrace.bound_symbols:
-        nbsym: BoundSymbol = bsym.from_bsym_swap_proxies(swapmap)
-        bound_symbols.append(nbsym)
-
-    extrace.bound_symbols = bound_symbols
+
+    class OpExProcessor(TraceSubstitutionProcessor):
+        def process_bsym(self, bsym):
+            if bsym.sym.python_impl is not None:
```
mruberry, line 68: Comment for this short-circuit would be cool

thunder/executors/passes.py
```diff
@@ -61,68 +62,54 @@ def preserve_bsym(bsym: BoundSymbol) -> Any:
...
+                        return
+                    else:
+                        raise AssertionError("Unknown executor")
+
+            if bsym.sym.executor is not None:
+                self.add_processed_bsyms([bsym])
+                self.set_result(bsym.output)
+                return
+
+            # No executor found, need to descend
```
mruberry, line 107: Nice comment

thunder/executors/torch_autograd.py
```diff
@@ -240,6 +240,23 @@ def split_forward_backward(computation_trc: TraceCtx, compile_data, compile_stat
         skip_output=False,
         skip_subsymbols=False,
     )
+
+    # remove duplicates
+    # The NVFuser (and possibly others) fusion pass applied on the forward during has a
```
mruberry, line 245: This is super interesting!

fyi @jjsjann123 @kshitij12345 @IvanYashchuk

thunder/executors/torch_compile.py
```diff
@@ -63,36 +68,34 @@ def make_compiled(
...
+        for a in sorted_unique_inputs:
+            prims.unpack_trivial(a, name=a.name)
+
+    region_trace.bound_symbols += list(bsyms)
     region_trace.bound_symbols.append(prims.python_return.bind(sorted_unique_outputs, output=None))
+    # for a in region_trace.args:
+    #    region_trace.add_name(a.name)
+    for bsym in region_trace.bound_symbols:
+        for o in bsym.flat_outs:
+            if o is not None:  # TODO: investigate
```
mruberry, line 89: Would you elaborate on this TODO -- is the question about what's happening with a bsym with no outputs?

thunder/torch/__init__.py
```diff
@@ -5754,6 +5754,7 @@ def register_default_torch_op(torchfn: Callable, torch_module):
         name=torchfn_name,
         meta=_fn,
         id=f"{torch_module.__name__}.{torchfn_name}",
+        is_prim=True,
```
mruberry, line 5757: Interesting! Why do we want to change this? Won't this default torch operations to be primitives? What's the effect of that?

thunder/core/trace_interpreter.py
t-vi: The irony is that `VJPDuals` neither have VJP-specific logic nor is there any duality involved, so this should probably just be called `AugmentedValue`, I will eventually send a PR to rename it.
As the next thing is to consolidate `interpret_trace_to_trace` into this, let's keep it for now.

thunder/core/transform_common.py
t-vi: Ahahah. I prepared the next line and then forgot the preparations.

thunder/core/transform_common.py
t-vi: Well, so basically remove the line and run the tests.

thunder/executors/passes.py
t-vi: Yeah, I copy-pasted this from your version of the visitor logic. What would be the comment?

thunder/torch/__init__.py
t-vi: Note that these are the auto-registered operations that don't have a decomposition into subsymbols. Arguably, this has been a bug before, as dce on a symbol with auto-registered operations as subsymbols would remove them.

thunder/executors/torch_compile.py
t-vi: I think the todo can be dropped. I was surprised that we get None outputs here, but it is OK.

thunder/executors/passes.py
```diff
@@ -61,68 +62,54 @@ def preserve_bsym(bsym: BoundSymbol) -> Any:
...
-    # Restores original variables
-    bound_symbols: list[BoundSymbol] = []
-    for bsym in extrace.bound_symbols:
-        nbsym: BoundSymbol = bsym.from_bsym_swap_proxies(swapmap)
-        bound_symbols.append(nbsym)
-
-    extrace.bound_symbols = bound_symbols
+
+    class OpExProcessor(TraceSubstitutionProcessor):
+        def process_bsym(self, bsym):
```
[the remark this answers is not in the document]
t-vi: Added a comment.

thunder/core/trace_interpreter.py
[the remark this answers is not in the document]
t-vi: Added a docstring with a reference to the OpExProcessor.

thunder/core/trace_interpreter.py
```diff
@@ -203,3 +204,179 @@ def do_swap(v):
...
+        if isinstance(v, VJPDual):
+            v.primal = tree_map(self.do_swap, v.primal)
+            v.residuals = tree_map(self.do_swap, v.residuals)
+            return v
+        if not isinstance(v, ProxyInterface):
+            return v
+        return self.swap_map.get(variableify(v), v)
+
+    def add_unprocessed_bsyms(self, bsyms):
+        self.unprocessed_bsyms[:0] = bsyms
```
lantiga, line 284: we could probably add `self.unprocessed_bsyms` and `self.new_bsyms` to `__init__` for readability - not mandatory

lantiga (approved):
Looks great, also great docstring addition! Let's merge as is

thunder/executors/torch_autograd.py
jjsjann123: Out of curiosity, do duplicates cause any functional issue?

PR #1505, opened 2024-12-02
Add hardshrink

beverlylytle:
<details>

</details>

## What does this PR do?

Add the hardshrink op.

## PR review

## Did you have fun?


thunder/tests/opinfos.py
```diff
@@ -69,8 +69,8 @@ def push_away_from_singularities(x, singularity_fn, eps):
     `eps` away from them. The `singularity_fn`  returns the (signed)
     distance from `x` to the nearest singularity."""
     x_dist = singularity_fn(x)
-    x_ = torch.where((x_dist > 0) & (x_dist < eps), x + eps, x)
-    return torch.where((x_dist < 0) & (x_dist > -eps), x - eps, x_)
+    x_ = torch.where((x_dist >= 0) & (x_dist < eps), x + eps, x)
+    return torch.where((x_dist <= 0) & (x_dist > -eps), x_ - eps, x_)
```
mruberry, line 73: Nice update

thunder/tests/opinfos.py
```diff
@@ -1730,6 +1730,35 @@ def gen(op, device, dtype, requires_grad):
 elementwise_unary_ops.append(relu6_opinfo)
 
 
+def hardshrink_singularity_at(lambd):
```
mruberry, line 1733: A comment here would be helpful

thunder/tests/opinfos.py
```diff
@@ -1730,6 +1730,35 @@ def gen(op, device, dtype, requires_grad):
...
+    return lambda a: torch.where(a >= 0, a - lambd, a + lambd)
+
+
+hardshrink_opinfo = OpInfo(
```
mruberry, line 1737: Why separate the OpInfos? A comment would be helpful for that unusual pattern, too

thunder/torch/__init__.py
```diff
@@ -1849,6 +1849,15 @@ def relu6(a: TensorProxy, /, inplace: bool = False) -> TensorLike:
 _inplace_to_out_of_place[relu6] = relu6, 1
 
 
+@torchsymbol(torch.nn.functional.hardshrink, is_method=False)
+def hardshrink(a: TensorProxy, /, lambd: float = 0.5) -> TensorLike:
+    out = where(a <= lambd, where(a < -lambd, a, 0), a)
```
mruberry, line 1854: Hardshrink is one of those functions that encourages code golfing (or maybe "kernel golfing").

The current implementation calls an elementwise binary kernel for `<=` and `<`, then calls two elementwise ternary kernels for the `where` operations. It requires the tensor `a`, the result of `a <= lambd`, and the result of `where(a < -lambd)` to be in memory at the same time, for a total input memory use of `mem(a) * 3`. 

Here are some alternatives to consider:

```
torch.where((a > lambd) | (a < -lambd), a, 0)
```

This would be three elementwise binary kernels but only one elementwise ternary kernel. The input memory use is only `mem(a) * 2`, however.

An alternative could be:

```
torch.where(abs(a) > abs(lambd), a, 0)
```

This is an elementwise unary kernel, an elementwise binary kernel, and an elementwise ternary kernel. The input memory use is again `mem(a) * 2`. 

This last implementation is similar to the PyTorch reference implementation:

https://github.com/pytorch/pytorch/blob/9012e7a62f756f7cf42e73d7dd375f8c468a0c66/torch/_refs/nn/functional/__init__.py#L493

Although the PyTorch reference implementation flips the comparison operator. Is that an interesting variation to consider?

There are a couple caveats with this last formulation of hardshrink. First, it extends the function to the complex numbers, so an explicit check that only real-valued numbers are given to the function might be a good idea (unless we're sure that we want to support complex hardshrink this way). Second, it needs a special case for `lambda < 0` (which seems like it could benefit from a special case anyway?).

Which of these three does nvFuser execute the fastest? Do all of these propagate NaNs correctly?

thunder/tests/opinfos.py
beverlylytle: Clearer?

thunder/torch/__init__.py
beverlylytle: I've been playing with timeit to measure the various implementations proposed here, though I found no clear winner.  I'd be happy to learn about other tools/processes that would help in this.

Since PyTorch doesn't support cfloat for hardshrink, I suppose that Thunder shouldn't either.

Any inequality comparing Nan to a constant is False, therefore any of the `where` statements should have `a` as the "else" value.  I guess this is why the reference PyTorch inequality is directed the way it is.

torch.hardshrink is defined for lambd <= 0 and is equivalent to the identity function.

thunder/tests/opinfos.py
```diff
@@ -1730,6 +1730,35 @@ def gen(op, device, dtype, requires_grad):
...
+            sample_input_generator=get_elementwise_unary_with_kwargs_generator([kwargs]),
+            torch_reference=_elementwise_unary_torch(torch.nn.functional.hardshrink),
+            # fdm.jvp, which is used in test_vjp_correctness, behaves badly at jump discontinuties of the partial derviatives
+            singularity_fn=singularity_fn,
+            test_directives=(),
+        )
+        elementwise_unary_ops.append(hardshrink_opinfo)
+
+
+hardshrink_opinfo_factory([None, 0.25, -0.1])
```
mruberry, line 1759: Nice refactor and comment. If we see other functions like hardshrink where the singularity depends on more of the input then we could generalize how singularity functions work (by giving them all the input)

thunder/torch/__init__.py
mruberry: Sounds great. Then I suggest being consistent with the PyTorch reference implementation since it has nice theoretical speed improvements.

thunder/tests/opinfos.py
mruberry: Yep!

mruberry (approved):

PR #1518, opened 2024-12-05
construct complex return values in epilogue

t-vi:
This is needed to support more complex output types (eg for HF transformers).


lantiga (approved):
Looks great, that's how it was intended : )

t-vi:
still some corner cases to fix.

PR #1520, opened 2024-12-05
add logsigmoid op

beverlylytle:
<details>

</details>

## What does this PR do?

Add support for the `logsigmoid` activation function.

## PR review

## Did you have fun?


thunder/torch/__init__.py
```diff
@@ -1812,6 +1812,14 @@ def leaky_relu(a: TensorProxy, /, negative_slope: float = 0.01, inplace: bool =
 _inplace_to_out_of_place[leaky_relu] = leaky_relu, 2
 
 
+@torchsymbol(torch.nn.functional.logsigmoid, is_method=False)
+def logsigmoid(a: TensorProxy, /):
+    return log(sigmoid(a))
```
mruberry, line 1817: In SciPy, the function that Pytorch calls `logsigmoid` is called `log_expit`:

https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.log_expit.html

and the implementation is:

https://github.com/scipy/scipy/blob/b589120c24e45baf03aabeb7fcb5e68d5f2c4de3/scipy/special/xsf/log_exp.h#L60

In PyTorch the CUDA implementation is:

https://github.com/pytorch/pytorch/blob/12b8c2fd8b91fb54abb50b25f3618bd5150125aa/aten/src/ATen/native/cuda/ActivationLogSigmoidKernel.cu#L25

What do you think of those formulations, and what about PyTorch's custom logsigmoid backward?

https://github.com/pytorch/pytorch/blob/12b8c2fd8b91fb54abb50b25f3618bd5150125aa/aten/src/ATen/native/cuda/ActivationLogSigmoidKernel.cu#L43

If we implement logsigmoid like SciPy and PyTorch, would a custom logsigmoid backward reduce memory use or make forward->backward faster than the autogenerated backward?

thunder/tests/opinfos.py
```diff
@@ -1678,6 +1678,23 @@ def gen(op, device, dtype, requires_grad):
 elementwise_unary_ops.append(leaky_relu_opinfo)
 
 
+logsigmoid_opinfo = OpInfo(
+    ltorch.logsigmoid,
+    dtypes=(datatypes.floating,),
+    sample_input_generator=elementwise_unary_generator,
+    torch_reference=torch.nn.functional.logsigmoid,
+    test_directives=(
+        # test tols are too tight for these half precision tests
```
mruberry, line 1687: The test tolerances can be relaxed by using a custom comparator, like here:

https://github.com/Lightning-AI/lightning-thunder/blob/9c84e235d6b6deabb62a9bfd21b9ecc21f6d18f2/thunder/tests/opinfos.py#L1815

mruberry (commented):
Exciting!

I made some suggestions for your review, @beverlylytle. I hope you find looking into implementation speed and memory use as interesting as I do!

thunder/tests/opinfos.py
beverlylytle: Yes, I can do that, but it seems liked the tolerances must be relatively high.  I copied this directive from the sigmoid opinfo figuring that it was decided that it was preferable to skip the test rather than use such large tolerances.

Interestingly, the naive implementation of `log(sigmoid(a)` passes the tests with tolerances of (1e-2, 1e-2) while the SciPy-like implementation of `where(a < 0, -log1p(exp(-a)), a - log1p(exp(a)))` requires (1e-1, 1e-1).  This is unexpected as the justification for the SciPy implementation is greater accuracy, but maybe that doesn't necessarily apply uniformly across datatypes.

mruberry: That makes sense! I believe SciPy's implementation (and PyTorch's implementation) are designed to improve numerical accuracy, and we are testing for consistency with PyTorch's implementation

thunder/torch/__init__.py
beverlylytle: While it doesn't look like much, there was an awful lot of trial and error that went into this revision.  

I think I understand how the scipy version of the forward impl improves the efficiency and accuracy.  I'm not sure I understand how a custom impl for the backward version is a major improvement on what would be automatically generated from this the scipy-like forward impl, though, the trace does look "tighter".

thunder/torch/__init__.py
```diff
@@ -1812,6 +1812,25 @@ def leaky_relu(a: TensorProxy, /, negative_slope: float = 0.01, inplace: bool =
...
 
+@torchsymbol(torch.nn.functional.logsigmoid, is_method=False)
+def logsigmoid(a: TensorProxy, /) -> TensorLike:
+    return where(a > 0, -log1p(exp(-a)), a - log1p(exp(a)))
+
+
+_inplace_to_out_of_place[logsigmoid] = logsigmoid, -1
+
+
+# @torchsymbol("log_sigmoid_backward", id="log_sigmoid_backward")
```
beverlylytle, line 1823: Why, oh why, oh why does test_grad fail on the cpu and only the cpu when this line is uncommented?

mruberry: I'm not sure what the torchsymbol addition actually does in this case, but as an fyi here are the logsigmoid backward implementations in PyTorch:

https://github.com/pytorch/pytorch/blob/0ddb33ba2299542f9558d949aa482a9ca30ceb30/aten/src/ATen/native/cuda/ActivationLogSigmoidKernel.cu#L43

https://github.com/pytorch/pytorch/blob/0ddb33ba2299542f9558d949aa482a9ca30ceb30/aten/src/ATen/native/cpu/Activation.cpp#L94

beverlylytle: Yes, I was looking at those, and also:

https://github.com/pytorch/pytorch/blob/fd65bd755d30f69783f82c75f9f341c1bbd4614a/torch/_decomp/decompositions.py#L332

That's why there's this extra unused parameter in the signature (for `buffer` which is also unused in PyTorch), and why I register the backward op with 
```
log_sigmoid_backward = _register_torch_operation(
    "torch.ops.aten.log_sigmoid_backward", like=ltorch.log_sigmoid_backward
)
```

I also don't understand what the annotation does here, but it looks like what was done for `log_softmax_backward` [here](https://github.com/Lightning-AI/lightning-thunder/blob/f89cecac3b8a7e1e67913b1f0b53fa4c3b8322e8/thunder/torch/__init__.py#L4963), which has backward implementations in PyTorch too:

https://github.com/pytorch/pytorch/blob/fd65bd755d30f69783f82c75f9f341c1bbd4614a/torch/_decomp/decompositions.py#L854

I was also looking at `_nll_loss_backward_impl` and wondering about [this comment](https://github.com/Lightning-AI/lightning-thunder/blob/f89cecac3b8a7e1e67913b1f0b53fa4c3b8322e8/thunder/executors/torchex.py#L1612).

mruberry: Aha. I'm guessing the `torchsymbol` decorator registers the operator as a symbol so it can be consumed directly by an executor later. That comment probably says that the `impl` would ideally not be defined in the torch executor but available in the torch language definition, like where the log sigmoid is in this PR. 

Is `torch.ops.aten.log_sigmoid_backward` actually getting called as expected if you look at the traces? Do you think the forward->backward computations are actually different (when trying them) or do you think the test has a problem?

beverlylytle: Running the following program:
```
import thunder
import torch

def f(x):
    return torch.nn.functional.logsigmoid(x)

x = torch.rand((1024,1024,4), device='cpu', requires_grad=True)
jf = thunder.jit(f)
jf(x)
print(thunder.last_backward_traces(jf)[-1])
```
 with the annotation produces:
```
import torch

@torch.no_grad()
@no_autocast
  # cotangents: "Collection"
  C0, _, = saved_for_backward
  del saved_for_backward
  t18, = cotangents
  del cotangents
  x, = C0
  clear_mutable_collection(C0)
  del C0
  t27 = torch.torch.ops.aten.log_sigmoid_backward(t18, x, x)  # t27: "cpu f32[1024, 1024, 4]"
    # t27 = ltorch.log_sigmoid_backward(t18, x, x)  # t27: "cpu f32[1024, 1024, 4]"
      # t37 = ltorch.abs(x)  # t37: "cpu f32[1024, 1024, 4]"
        # t37 = prims.abs(x)  # t37: "cpu f32[1024, 1024, 4]"
      # t38 = ltorch.neg(t37)  # t38: "cpu f32[1024, 1024, 4]"
        # t38 = prims.neg(t37)  # t38: "cpu f32[1024, 1024, 4]"
      # t39 = ltorch.exp(t38)  # t39: "cpu f32[1024, 1024, 4]"
        # t39 = prims.exp(t38)  # t39: "cpu f32[1024, 1024, 4]"
      # t40 = ltorch.add(1, t39, alpha=1)  # t40: "cpu f32[1024, 1024, 4]"
        # t40 = prims.add(1.0, t39)  # t40: "cpu f32[1024, 1024, 4]"
      # t41 = ltorch.true_divide(t39, t40)  # t41: "cpu f32[1024, 1024, 4]"
        # t41 = prims.div(t39, t40)  # t41: "cpu f32[1024, 1024, 4]"
      # t42 = ltorch.gt(x, 0)  # t42: "cpu b8[1024, 1024, 4]"
        # t42 = prims.gt(x, 0.0)  # t42: "cpu b8[1024, 1024, 4]"
      # t43 = ltorch.sub(1, t41, alpha=1)  # t43: "cpu f32[1024, 1024, 4]"
        # t43 = prims.sub(1.0, t41)  # t43: "cpu f32[1024, 1024, 4]"
      # t44 = ltorch.where(t42, t41, t43)  # t44: "cpu f32[1024, 1024, 4]"
        # t44 = prims.where(t42, t41, t43)  # t44: "cpu f32[1024, 1024, 4]"
      # t27 = ltorch.mul(t18, t44)  # t27: "cpu f32[1024, 1024, 4]"
        # t27 = prims.mul(t18, t44)  # t27: "cpu f32[1024, 1024, 4]"
  del t18, x
  return (t27,)]
```

While the results of `test_grad` are
```
>       comp(J_u_v, u_J_star_v)
E       AssertionError: Scalars are not close!
E
E       Expected 0.0716214267429005 but got 0.22672571966112479.
E       Absolute difference: 0.15510429291822428 (up to 1e-07 allowed)
E       Relative difference: 2.1656130011903043 (up to 1e-07 allowed)

...

FAILED thunder/tests/test_grad.py::test_vjp_correctness_logsigmoid_torch_cpu_thunder.dtypes.float64 - AssertionError: Scalars are not close!
```

Without the annotation:
```
import torch

@torch.no_grad()
@no_autocast
  # cotangents: "Collection"
  C0, _, = saved_for_backward
  del saved_for_backward
  t18, = cotangents
  del cotangents
  x, = C0
  clear_mutable_collection(C0)
  del C0
  t19 = torch.abs(x)  # t19: "cpu f32[1024, 1024, 4]"
    # t19 = ltorch.abs(x)  # t19: "cpu f32[1024, 1024, 4]"
      # t19 = prims.abs(x)  # t19: "cpu f32[1024, 1024, 4]"
  t24 = torch.gt(x, 0)  # t24: "cpu b8[1024, 1024, 4]"
    # t24 = ltorch.gt(x, 0)  # t24: "cpu b8[1024, 1024, 4]"
      # t24 = prims.gt(x, 0.0)  # t24: "cpu b8[1024, 1024, 4]"
  del x
  t20 = torch.neg(t19)  # t20: "cpu f32[1024, 1024, 4]"
    # t20 = ltorch.neg(t19)  # t20: "cpu f32[1024, 1024, 4]"
      # t20 = prims.neg(t19)  # t20: "cpu f32[1024, 1024, 4]"
  del t19
  t21 = torch.exp(t20)  # t21: "cpu f32[1024, 1024, 4]"
    # t21 = ltorch.exp(t20)  # t21: "cpu f32[1024, 1024, 4]"
      # t21 = prims.exp(t20)  # t21: "cpu f32[1024, 1024, 4]"
  del t20
  t22 = torch.add(1, t21, alpha=1)  # t22: "cpu f32[1024, 1024, 4]"
    # t22 = ltorch.add(1, t21, alpha=1)  # t22: "cpu f32[1024, 1024, 4]"
      # t22 = prims.add(1.0, t21)  # t22: "cpu f32[1024, 1024, 4]"
  t23 = torch.true_divide(t21, t22)  # t23: "cpu f32[1024, 1024, 4]"
    # t23 = ltorch.true_divide(t21, t22)  # t23: "cpu f32[1024, 1024, 4]"
      # t23 = prims.div(t21, t22)  # t23: "cpu f32[1024, 1024, 4]"
  del t21, t22
  t25 = torch.sub(1, t23, alpha=1)  # t25: "cpu f32[1024, 1024, 4]"
    # t25 = ltorch.sub(1, t23, alpha=1)  # t25: "cpu f32[1024, 1024, 4]"
      # t25 = prims.sub(1.0, t23)  # t25: "cpu f32[1024, 1024, 4]"
  t26 = torch.where(t24, t23, t25)  # t26: "cpu f32[1024, 1024, 4]"
    # t26 = ltorch.where(t24, t23, t25)  # t26: "cpu f32[1024, 1024, 4]"
      # t26 = prims.where(t24, t23, t25)  # t26: "cpu f32[1024, 1024, 4]"
  del t24, t23, t25
  t27 = torch.mul(t18, t26)  # t27: "cpu f32[1024, 1024, 4]"
    # t27 = ltorch.mul(t18, t26)  # t27: "cpu f32[1024, 1024, 4]"
      # t27 = prims.mul(t18, t26)  # t27: "cpu f32[1024, 1024, 4]"
  del t18, t26
  return (t27,)]
```

The issue is not a misapplication of `fdm`.   It produces exactly the results I expect.  Testing with a random 1 element tensor shows that the vjp of logsigmoid on the cpu is off.  Meaning it's not some cumulative discrepancy for large tensors, nor an issue with a singularity.  

If I use the one line implementation (without the `abs`) of `log_sigmoid_backward`, traces of the annotated and unannotated version change as one would expect, with `test_grad` still failing for the decorated version.

mruberry: That's so interesting! So it looks like the logsigmoid foward and backward are being called as expected, and we 

So if you just forward+backward without going through test vjp on some examples, how do the grads compare between the two? Are there any obvious visual differences for small values? (Also, are the forward traces what you'd expect?) What if you explicitly test calling logsigmoid + logsigmoid_backward in eager PyTorch? How does that compare to what eager is actually doing?

mruberry: One idea that shouldn't make a difference, by the way, would be to write the fwd+bwd as a single function, like this one: https://github.com/Lightning-AI/lightning-thunder/blob/96486e81cf8b5c5a90faa3b2a529669d74d59658/thunder/core/transforms.py#L682

```
from thunder.torch import logsigmoid, log_sigmoid_backward
  fwd = logsigmoid(a)
  g = get_grad(fwd)
  ga = log_sigmoid_backward(g, a, _)

    g = get_grad(fwd)
    g_converted = prims.convert_element_type(g, dtypes.to_dtype(a))
    put_grad(a, g_converted)

    return fwd
```

thunder/core/transforms.py
```diff
@@ -1435,6 +1435,25 @@ def _copy_with_setitem_grad(a: TensorProxy, index, value: Number | TensorProxy):
...
+
+def _log_sigmoid_grad(
+    a: TensorProxy,
+) -> TensorProxy:
+    from thunder.torch import abs, exp, log_sigmoid_backward, logsigmoid
+
+    fwd = logsigmoid(a)
+
+    g = get_grad(fwd)
+    z = exp(-abs(a)) if a.device.type == "cpu" else full_like(a, fill_value=0)
```
mruberry, line 1447: Would you add a note explaining this divergence?

Follow-up question: when executing on CUDA (as we so often do), do you actually have to pass a tensor to `log_sigmoid_backward`? And if so, does the tensor actually have to have the same shape as `a`?

That is, could we write this like:

```
if a.device.type == 'cpu':
  # NOTE PyTorch's CPU computation for logsigmoid's grad uses an additional "buffer" tensor, see [link]
  z = exp(-abs(a))
  a_grad = log_sigmoid_backward(g, a, z)
else:
  # NOTE a is on a CUDA device (or some other yet unsupported device)
  a_grad = log_sigmoid_backward(g, a, None)
  OR
  a_grad = log_sigmoid_backward(g, a, zeros_like((0,))
```

Because then we wouldn't have to pay for the memory of the tensor full of zeros that will be ignored. 

From a DevX perspective, how did you find writing the grad as a single function instead of two functions?

thunder/torch/__init__.py
```diff
@@ -1812,6 +1812,25 @@ def leaky_relu(a: TensorProxy, /, negative_slope: float = 0.01, inplace: bool =
...
+def logsigmoid(a: TensorProxy, /) -> TensorLike:
+    return where(a > 0, -log1p(exp(-a)), a - log1p(exp(a)))
+
+
+_inplace_to_out_of_place[logsigmoid] = logsigmoid, -1
+
+
+@torchsymbol("log_sigmoid_backward", id="log_sigmoid_backward")
+def log_sigmoid_backward(g: TensorProxy, a: TensorProxy, buffer: TensorProxy) -> TensorLike:
+    # buffer is used by PyTorch in cpu-based calculations.  See
```
mruberry, line 1825: Awesome comment; I would just repeat it (or something like it) in the grad function

thunder/core/transforms.py
beverlylytle: Yes, a tensor is required, but not of a particular shape.  

I found the single function a little uncomfortable because there is magic happening here, and where there is magic, I fear dragons, peasant that I am.  I understand it improves efficiency in cases more involved than this one.  It also looks like there are three ways of accomplishing more or less the same thing and a not-immediately-obvious hierarchy between them: `grad_transform`, `register_grad` then `register_augmented_forward/register_backward`.  The "more or less" aspect of this is not good for DevX, as the subtleties between them take some thinking to figure out.

thunder/torch/__init__.py
```diff
@@ -1812,6 +1812,25 @@ def leaky_relu(a: TensorProxy, /, negative_slope: float = 0.01, inplace: bool =
 _inplace_to_out_of_place[leaky_relu] = leaky_relu, 2
 
 
+@torchsymbol(torch.nn.functional.logsigmoid, is_method=False)
+def logsigmoid(a: TensorProxy, /) -> TensorLike:
+    return where(a > 0, -log1p(exp(-a)), a - log1p(exp(a)))
+
+
+_inplace_to_out_of_place[logsigmoid] = logsigmoid, -1
```
crcrpar, line 1820: At glance logsigmoid doesn't seem to store the results into the inout tensor. Is where an in-place op?

crcrpar: Then I don't think this registration is necessary.

The second value is the index of a boolean arg indicating the op is inplace or not, if exists. So the comment here is misleading, looking at how it's used https://github.com/Lightning-AI/lightning-thunder/blob/d59f9ab16b8e7dd797e3c9849a72653634dd7142/thunder/core/functionalization.py#L320-L326
https://github.com/Lightning-AI/lightning-thunder/blob/d59f9ab16b8e7dd797e3c9849a72653634dd7142/thunder/torch/__init__.py#L94-L95

beverlylytle: ok, updated

thunder/torch/__init__.py
```diff
@@ -1812,6 +1812,19 @@ def leaky_relu(a: TensorProxy, /, negative_slope: float = 0.01, inplace: bool =
 _inplace_to_out_of_place[leaky_relu] = leaky_relu, 2
 
 
+@torchsymbol(torch.nn.functional.logsigmoid, is_method=False)
+def logsigmoid(a: TensorProxy, /) -> TensorLike:
+    return where(a > 0, -log1p(exp(-a)), a - log1p(exp(a)))
+
+
+@torchsymbol("log_sigmoid_backward", id="log_sigmoid_backward")
+def log_sigmoid_backward(g: TensorProxy, a: TensorProxy, buffer: TensorProxy) -> TensorLike:
```
mruberry, line 1821: A nitty follow-up to this PR would probably to be to have the PyTorch executor implement a custom grad transform for logsigmoid that added the buffer, and to have our default definition of the buffer fwd+bwd not use the buffer at all, based on the computation here that ignores the buffer argument

thunder/core/transforms.py
```diff
@@ -1435,6 +1435,31 @@ def _copy_with_setitem_grad(a: TensorProxy, index, value: Number | TensorProxy):
...
+
+    g = get_grad(fwd)
+    if a.device.type == "cpu":
+        # NOTE PyTorch's CPU computation for logsigmoid's grad uses an additional "buffer" tensor, see
+        # https://github.com/pytorch/pytorch/blob/7667235a23e2ffca4d32e6e16aa60a683418e159/torch/_decomp/decompositions.py#L332
+        z = exp(-abs(a))
+        a_grad = log_sigmoid_backward(g, a, z)
+    else:
+        # Here a placeholder tensor is provided.
+        a_grad = log_sigmoid_backward(g, a, full((1,), fill_value=0, device=a.device))
```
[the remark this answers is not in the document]
beverlylytle: Yes, `empty` works and yes, the tensors must be on the same device.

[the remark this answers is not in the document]
mruberry: Great; let's go with an empty tensor with no elements on the same device and get this landed.

In the future, we should do a follow-up PR that removes the buffer computation (at least for CUDA) from Thunder's general autograd formula and have the PyTorch executor use a custom autograd formula for logsigmoid.

Happy to talk more about this over VC.

mruberry (approved):
Rad!

I filed https://github.com/Lightning-AI/lightning-thunder/issues/1555 for avoiding the creation of the extra CUDA tensor by adding a custom autograd formula for logsigmoid that would only apply when using the PyTorch executor

thunder/executors/torchex.py
```diff
@@ -851,11 +855,33 @@ def _elementwise_unary_with_inplace_checker(a: TensorProxy, /, inplace: bool = F
...
+_register_elementwise_unary_implementation(ltorch.hardshrink, hardshrink, checker=_always_executable)
+_register_elementwise_unary_implementation(ltorch.hardswish, hardswish, checker=_elementwise_unary_with_inplace_checker)
 _register_elementwise_unary_implementation(ltorch.leaky_relu, leaky_relu, checker=_always_executable)
+_register_elementwise_unary_implementation(
+    ltorch.log_sigmoid_backward, log_sigmoid_backward, checker=_always_executable
+)
+# _register_elementwise_unary_implementation(ltorch.logsigmoid, logsigmoid)
+
+
+def log_sigmoid_grad_transform(a):
```
beverlylytle, line 867: @mruberry  do you mean something like this?

mruberry: Yes - but this would need to include both the CPU and CUDA paths because the PyTorch executor can run both CPU and CUDA operations (if it's set as the only executor, for example, it will run CUDA operations)

thunder/core/transforms.py
```diff
@@ -1435,6 +1436,32 @@ def _copy_with_setitem_grad(a: TensorProxy, index, value: Number | TensorProxy):
 
 register_grad(pids.COPY_WITH_SETITEM, _copy_with_setitem_grad)
 
+
+# def _log_sigmoid_grad(
```
mruberry, line 1440: For the follow-up, there's also the question of if we should have a custom grad for log_sigmoid generally, separate from the PyTorch executor's special grad function

thunder/executors/torchex.py
mruberry: (I would leave the follow-up for a separate PR)

PR #1256, opened 2024-10-03
Lookaside for `torch.ops.higher_order.autograd_function_apply`

crcrpar:
## What does this PR do?

As per #1248, the support of `torch.ops.higher_order.autograd_function_apply` would be a bit more flexible by tracing into both `fwd` and `bwd`.

cc @apaz-cli

PR #1526, opened 2024-12-07
spelling out `higher_order.autograd_function_apply`'s fwd and bwd in trace.

crcrpar:
## What does this PR do?

Alternative of #1256.
Instead of keeping the signature of `torch.ops.higher_order.autograd_function_apply(fwd, bwd, *fwd_args, **fwd_kwargs)` in a trace by introducing function objects, this PR traces the both fwd and bwd and writes the body of fwd into a trace like the lookaside for `torch.autograd.Function`.


t-vi (commented):
Thank you for making the non-hof-version work @crcrpar , I very much appreciate it.
Let's merge this now to unblock use and then we can see what to do with alternatives later if I can be convinced...

PR #1535, opened 2024-12-10
Add thunderfx API 

### LINKED ISSUE #1529: ThunderFX: Interface UX
It would be nice if instead of having to write

<pre class="ql-syntax" spellcheck="false">backend = ThunderCompiler()
opt_foo1 = torch.compile(foo, backend=backend)
</pre><p> </p><p>We instead allowed practitioners to write</p><p> </p>

<pre class="ql-syntax" spellcheck="false">from thunder import thunderfx

tfoo = thunderfx(foo)
</pre><p> </p><p>Notebooks like this are a good opportunity to identify areas for improvement in our UX like this. We can create the thunderfx function so it accepts <code>**kwargs</code> to pass to the torch.compile call, too. </p><p> </p><p>It'd be great to see a PR changing this so we can update this doc. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Lightning-AI/lightning-thunder/pull/1524/discussion/#comment-1873705406'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

_Originally posted by @mruberry in https://github.com/Lightning-AI/lightning-thunder/pull/1524#discussion_r1873705406_
            

cc @borda

kiya00:
<details>

</details>

## What does this PR do?

Fixes #1529.

### DIFF
--- a/thunder/dynamo/compiler.py
+++ b/thunder/dynamo/compiler.py
@@ -27,9 +27,7 @@ class ThunderCompiler:
         function.
 
         Keyword arguments:
-            thunder_options: a dictionary of options to pass to :func:`thunder.jit`. Besides all the arguments to :func:`thunder.jit`,
-                             it accepts ``torch_inductor_options`` which are passed to :func:`torch.compile` if part of the graph
-                             is not supported by thunder.
+            thunder_options: a dictionary of options to pass to :func:`thunder.jit`.
 
         Example:
             >>> import torch
@@ -60,11 +58,9 @@ class ThunderCompiler:
         # Ref to the documentation of `SubgraphInfo` to know more about the information it contains.
         self.subgraph_infos: list[SubgraphInfo] = []
 
-        torch_inductor_options = thunder_options.pop("torch_inductor_options", {})
-
         self.thunder_options = thunder_options
         self._thunder_jit = partial(jit, **thunder_options)
-        self._torch_compile = partial(torch.compile, **torch_inductor_options)
+        self._torch_compile = torch.compile
 
     def __call__(self, gm: torch.fx.GraphModule, sample_args: list[torch.SymInt, torch.Tensor]):
         gm = remove_empty_autocast(gm)
--- a/thunder/dynamo/compiler.py
+++ b/thunder/dynamo/compiler.py
@@ -3,6 +3,7 @@ from functools import partial
 from looseversion import LooseVersion
 from typing import TYPE_CHECKING
 import warnings
+import inspect
 
 import torch
 
@@ -10,10 +11,12 @@ from thunder.core.baseutils import run_once
 from thunder.core.utils import safe_zip
 from thunder.dynamo.utils import recompile_graph, remove_empty_autocast, reproducer, CompilerType
 from thunder.dynamo.splitter import _splitter
+from thunder.core.utils import check
 
 if TYPE_CHECKING:
     from thunder.dynamo.utils import SubgraphInfo
     from os import PathLike
+    from collections.abc import Callable
 
 
 @run_once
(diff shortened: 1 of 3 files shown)


thunder/dynamo/compiler.py
```diff
@@ -127,3 +130,32 @@ def save_reproducer_to_folder(self, reproducer_folder: str | PathLike, use_pytes
...
+    Keyword Args:
+        **kwargs: a dictionary of options to pass to :func:`torch.compile` and :func:`thunder.jit`.
+    Returns:
+        The compiled callable
+    """
+    import thunder
+
+    torch_compile_kwarg_names = inspect.getfullargspec(torch.compile).kwonlyargs
+    thunder_jit_kwarg_names = inspect.getfullargspec(thunder.jit).kwonlyargs
+    overlap = [kwarg_name for kwarg_name in thunder_jit_kwarg_names if kwarg_name in torch_compile_kwarg_names]
```
mruberry, line 148: This is a great check!

thunder/dynamo/compiler.py
```diff
@@ -127,3 +130,32 @@ def save_reproducer_to_folder(self, reproducer_folder: str | PathLike, use_pytes
...
+    check(
+        not overlap,
+        lambda: f"There are overlapping kwargs between thunder.jit and torch.compile: {overlap}",
+        ValueError,
+    )
+
+    torch_compile_options = {k: v for k, v in kwargs.items() if k in torch_compile_kwarg_names}
+    thunder_options = {k: v for k, v in kwargs.items() if k not in torch_compile_kwarg_names}
+
+    backend = ThunderCompiler(**thunder_options)
```
mruberry, line 158: When the `ThunderCompiler` class is created this way, it sets `self._torch_compile` here:

https://github.com/Lightning-AI/lightning-thunder/blob/6307021dbc7fa73f589ac72e2f3deea4c5cabb25/thunder/dynamo/compiler.py#L77

The `self._torch_compile` object is used here:

https://github.com/Lightning-AI/lightning-thunder/blob/6307021dbc7fa73f589ac72e2f3deea4c5cabb25/thunder/dynamo/compiler.py#L88

But with this approach, won't the torch.compile options not be passed to ThunderCompiler, so its internal `self._torch_compile` object won't be set properly?

Does the call to create the `ThunderCompiler` object need to look like `ThunderCompiler(thunder_options, torch_compile_options)`, and then internally it needs to call

```
self._torch_compile = partial(torch.compile, **torch_compile_options)
```

?

mruberry: (Maybe in the future we don't want the "outer" torch.compile kwargs to be the same as the "inner" torch.compile kwargs, so we might want to leave a note about that)

thunder/tests/test_dynamo.py
```diff
@@ -930,3 +930,25 @@ def check(file_name, cmd):
...
+    x = torch.randn(4, 4, device="cuda", requires_grad=True)
+    cfoo = thunderfx(foo)
+    cfoo(x)
+    thunder_compiled_fns = cfoo._backend.subgraph_infos[0].thunder_compiled_fns
+    assert len(thunder_compiled_fns) == 1
+    assert last_traces(thunder_compiled_fns[0])
+
+    from thunder.dev_utils.nvtx_profile_transform import NvtxProfileTransform
+
+    cfoo = thunderfx(foo, dynamic=True, transforms=[NvtxProfileTransform()])
```
mruberry, line 949: Very cool!

mruberry (commented):
Cool stuff, @kiya00!

I wrote a question about how kwargs propagate to the two torch.compile instances

thunder/dynamo/compiler.py
kiya00: > But with this approach, won't the torch.compile options not be passed to ThunderCompiler, so its internal self._torch_compile object won't be set properly?

I think currently the 2 groups of torch.compile kwargs can be different, one for the outside `torch.compile`, it's mainly used for the dynamo part I think, the other group is used in the inner `torch.compile` to handle the subgraph fall back to inductor

```
backend = ThunderCompiler(torch_inductor_options={dynamic=True})
cfunc = torch.compile(func, backend=backend, dynamic=False)
```
cc: @kshitij12345 do I understand it right?

mruberry: I understand, and I think you're right. But what does that look like with the new `thunderfx` function?

```
thunderfx(foo, full_graph=True, torch_inductor_options={'full_graph'=True})
```

?

That seems a little odd and hidden.

I would think we should do one of two things for now:

1. Set the torch.compile kwargs consistently (both "outer" and "inner" are the same)
2. Set the "inner" torch.compile kwargs however we like, without providing a mechanism for practitioners to set them (at this time)

What do you think, @kshitij12345?

kshitij12345: Right, now we do support `torch.compile` inside the `ThunderCompiler` have different arguments than the user one (outer). However, I don't think in practice, we had to set them differently yet.

So, I like the option 2 which seems less confusing.

NOTE - Even though we call `torch.compile` inside in `ThunderCompiler`, technically we just want something like `torch._inductor.compile_fx` (as the dynamo stage is already done).

kiya00: Then in that case the current code will work, the `torch.compile` kwargs passed to `thunder_fx` will only set the outer `torch.compile` and not affect the inner ones.

kiya00: @kshitij12345 Do we want to remove this `torch_inductor_options` option completely in the constructor of `ThunderCompiler`?

kshitij12345: Sure, we can remove it.

mruberry (approved):
Awesome! As a follow-up we should try to switch existing test cases that use the previous method of invoking thunderfx to this new one, too

PR #1538, opened 2024-12-11
[thunderfx] Add a test of custom `torch.autograd.Function`

crcrpar:
## What does this PR do?

Add a test case to track the progress of thunderfx.
Currently it seems that a custom `torch.autograd.Function` goes to the fallback path
even though it could be handled by `thunder.jit`.

### DIFF
--- a/thunder/tests/test_dynamo.py
+++ b/thunder/tests/test_dynamo.py
@@ -1,5 +1,4 @@
 import pytest
-import warnings
 import itertools
 import os
 from subprocess import run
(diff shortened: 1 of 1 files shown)


kshitij12345 (commented):
We already have a test for thunderFX and torch.autograd.Function interaction, verifying that it goes to fallback.

https://github.com/Lightning-AI/lightning-thunder/blob/673bdb90802a8e796810751ab30510952a9c8577/thunder/tests/test_dynamo.py#L268-L303

If anything is missing, we can probably update this test. Wdyt?

crcrpar:
The update is perfect. I overlooked it

Issue #1251, opened 2024-10-02
Don't use nvFuser for 1-op fusions of metadata manipulation ops


tfogal:
## 🚀 Feature

Sometimes we can end up with a fragment where a single lonesome op ends up in an nvFusion region. For example this program:

```
  class DynamoModule(torch.nn.Module):
    def forward(self, L_tensor_ : torch.Tensor):
        l_tensor_ = L_tensor_
        reduce = einops.einops.reduce(l_tensor_, 'b np sq sk -> (b np) sq sk', reduction = 'rearrange');  l_tensor_ = None
        return (reduce,)
```

produces this trace:

```
def computation(L_tensor_):
  # L_tensor_: "cuda:0 bf16[2, 40, 384, 384]"
  [tensor] = nvFusion0(L_tensor_)
    # tensor = prims.reshape(L_tensor_, (80, 384, 384))  # tensor: "cuda:0 bf16[80, 384, 384]"
  return {'output': (tensor,), 'flat_args': [L_tensor_], 'flat_output': (tensor,)}, ((), ())
```

Unfortunately it takes nvFuser 28 microseconds to do nothing:
![image](https://github.com/user-attachments/assets/54cc6ea0-fb28-455a-9239-13196c5f654e)

and we probably would have been better off just running things through torch.

[g51.py.txt](https://github.com/user-attachments/files/17235815/g51.py.txt) - full sample program/benchmark.

We can (and will) do better in nvFuser on this, but there's not much of a reason to call a kernel generator to do work that it cannot possibly do better than the status quo.

### Motivation

This is coming up through the ThunderFX path applied to the NeVA path.

### Pitch

Ideally we'd check if a single op 'view' or 'reshape' ends up being the entirety of a fusion group and reject it if so--causing the op to fallback to the torch executor.

### Alternatives

Of course, it might be better if we found a home for that single op in another fusion group, rather than just kicking it out of nvFuser entirely. It's not clear whether this case comes up in practice.

With the larger change of memory planning, we probably wouldn't need a solution as advocated here, as these single ops would either disappear (if they are aliases) or need a real copy anyway (in which case calling a kernel generator is probably about the same as not).

tfogal:
cc @kshitij12345 since i mentioned it.

don't go out of your way to check, but if you find this is relevant for Q4 models then please send me a ping.

tfogal:
@kshitij12345 I think you fixed this a while back under another bug.
[...]

PR #1559, opened 2024-12-16
[nvfuser] don't allow shape only nvfuser region

kshitij12345:
Related: https://github.com/Lightning-AI/lightning-thunder/issues/1251

Microbenchmark
```python
import torch
import torch.utils.benchmark
import thunder

# Seen in HF-Qwen2 Model
# [t6866] = nvFusion4(t6857)
#     # lora_res_12 = prims.reshape(t6857, (1, 4096, 16))  # lora_res_12: "cuda:0 bf16[1, 4096, 16]"
#     # t6866 = prims.reshape(lora_res_12, (4096, 16))  # t6866: "cuda:0 bf16[4096, 16]"

def fn(x):
    x = x.reshape(-1, 4096, 16)
    x = x.reshape(4096, 16)
    return x


x = torch.randn(4096, 16, device="cuda")
jfn = thunder.jit(fn, nv_enable_shape_only_fusion=False)

jfn(x)

timer = torch.utils.benchmark.Timer("jfn(x)", globals={"jfn": jfn, "x": x})

print(timer.blocked_autorange(min_run_time=1))
# print(thunder.last_traces(jfn)[-1])
```

Before PR
```python
<torch.utils.benchmark.utils.common.Measurement object at 0x7305591c54e0>
jfn(x)
  Median: 36.18 us
  3 measurements, 10000 runs per measurement, 1 thread
```

After PR
```python
<torch.utils.benchmark.utils.common.Measurement object at 0x7e533962cc10>
jfn(x)
  Median: 18.95 us
  IQR:    0.02 us (18.93 to 18.96)
  6 measurements, 10000 runs per measurement, 1 thread
```

TODO 
* [x] Add tests!
* [x] Check if CI works fine

### DIFF
--- a/thunder/core/prims.py
+++ b/thunder/core/prims.py
@@ -3585,7 +3585,7 @@ def stride_order_meta(a: TensorProxy, /, order: Sequence[int]) -> TensorProxy:
 # TODO Consider a more general stride manipulation primitive, like PyTorch's
 #   as_strided or set_strided operations
 # See clang.stride_order for this prim's documentation
-stride_order = make_prim(PrimIDs.STRIDE_ORDER, "stride_order", meta=stride_order_meta)
+stride_order = make_prim(PrimIDs.STRIDE_ORDER, "stride_order", meta=stride_order_meta, tags=(OpTags.SHAPE_OP,))
 
 #
 # Reduction prims
--- a/thunder/executors/nvfuserex_impl.py
+++ b/thunder/executors/nvfuserex_impl.py
@@ -470,6 +470,18 @@ class FusionDefinitionWrapper:
         return f"FusionDefinitionWrapper({self.name})"
 
 
+def all_tagged(bsym: BoundSymbol, tag: prims.OpTags) -> bool:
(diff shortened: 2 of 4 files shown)


t-vi (approved):
Seems good, thank you @kshitij12345 .

IvanYashchuk:
The pull request description mentions the `nv_enable_shape_only_fusion` option, but it's not added in the merged commit.

kshitij12345:
> The pull request description mentions the nv_enable_shape_only_fusion option, but it's not added in the merged commit.

Oops, I think I had added for benchmarking but removed it. I think in most scenarios, we won't want a Shape Op only nvFuser region. Do you think adding the option will be useful? Will send a patch accordingly. Thanks for catching this.

IvanYashchuk:
Yes, an option will be useful for ease of experimentation. There could be certain connectivity patterns of fusions allowing shape operations to be copied into other fusions and save on global memory tensors with fusions rematerialization pass as https://github.com/Lightning-AI/lightning-thunder/pull/1371 shows (if only nvfuser linear is enabled (no matmul) memory savings are achieved only if in addition there are shape only fusions in backward).

Issue #1379, opened 2024-10-31
Loss function backward output kept in memory longer than expected


riccardofelluga:
## 🐛 Bug

The outputs from the log_softmax_backward is kept in memory till the start of the next forward whereas it should be released after use in the backward trace. 

### To Reproduce

```python
import torch
import thunder

def foo(a):
    t0 = torch.nn.functional.relu(a)
    t1 = torch.nn.functional.relu(t0)
    t2 = torch.nn.functional.relu(t1)
    t3 = torch.nn.functional.relu(t2)
    return t3

x = torch.randn((256, 256), requires_grad=True, device="cuda")
y = torch.randn_like(x)

jmlp = thunder.jit(foo, executors=[thunder.pytorch_executor], enable_saved_for_backward_recomputation=True)

torch.cuda.memory._record_memory_history()

for i in range(3):
    ins = x.detach().requires_grad_(True)
    out = jmlp(ins)
    out = torch.nn.functional.cross_entropy(out, y, ignore_index=-1)
    out.backward()
    out.item()
    del out

torch.cuda.memory._dump_snapshot("cross_entropy_test.pickle")
```

<img width="559" alt="repro_snap" src="https://github.com/user-attachments/assets/aae23422-1931-49f2-bcf0-9d8bfa585ee2">

tfogal:
would this happen through thunderfx as well or is dynamo dealing with most of the saving tensors there?

riccardofelluga:
I've seen the same behavior with ThunderFX by modifying the previous snippet:

```python
backend = ThunderCompiler(executors=[thunder.pytorch_executor], enable_saved_for_backward_recomputation=True)
jmlp = torch._dynamo.optimize(backend=backend)(foo)
```

I think this is expected as it happens between fwd and bwd thunder traces so after the ThunderFX pass

kshitij12345:
I had a quick look at the backward traces with `enable_saved_for_backward_recomputation` being True and False

**Recomputation=True**: When recomputation is enabled, the earlier part of the backward trace is just recomputing and then post that we use the gradient from loss function (so it gets used later and deleted later, see `t65 = torch.where(t19, t10, 0)` in trace). Also,(from a quick glance) it seems that the recomputation performed at the earlier part is actually required in this order and can't be interleaved.

**Recomputation=False**: When recomputation is disabled, the first computation operation uses the gradient from loss and clears it. (See line `t65 = torch.where(t19, t10, 0)` in trace)

[...]

PR #1560, opened 2024-12-17
update checkpointing support for jit

t-vi:
I'll add switches and testing of checkpointing, but here is the material code changes.

### DIFF
--- a/thunder/benchmarks/benchmark_litgpt.py
+++ b/thunder/benchmarks/benchmark_litgpt.py
@@ -541,19 +541,15 @@ class Benchmark_litGPT:
         return model
 
     def setup_activation_checkpointing(self):
-        if "thunder" in self.compile and "dynamo" not in self.compile:
-            # checkpointing is an option to thunder.jit
-            return
-
         if any(isinstance(mod, CheckpointWrapper) for mod in self.model.modules()):
             warnings.warn(
                 "FSDP checkpointing is configured, but the model already contains checkpointed layers."
                 " Checkpointing will be ignored."
             )
             return
-
         check_fn = lambda submodule: isinstance(submodule, Block)
         apply_activation_checkpointing(self.model, checkpoint_wrapper_fn=checkpoint_wrapper, check_fn=check_fn)
+        print(self.model)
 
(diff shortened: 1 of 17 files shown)


t-vi:
There are a number of things still to be fixed.
There is a failure with the memory calculation because we don't update the initial collection proxy.
I'll look into these.

t-vi:
To my mind, the remaining missing bit is the handling of `uniform` -> `uniform_philox` conversion (currently handled by rematerialization) and the interaction with recomputation. I'll look into it. In the meantime, I'd be keen to hear complaints and/or success stories about memory / performance impact.

t-vi:
So one needs to enable checkpointing layers with compiler==thunder for this. Then the memory profile of the backward is still terrible:
Running @riccardofelluga 's benchmark:
Eager with checkpointing and 8 layers needs 12.59GB, thunder with checkpointing and 8 layers needs 38GB.
This is not the saved for backwards, which is (for thunder):  Saved for backward size: 1761.89 MiB Saved for backward number of tensors: 103

t-vi:
Doing the following:
- apply del last used to any trace of the last_backward_traces
- run the memory examine tool (so awesome @kiya00 !)

We see that we are doing much better at first than after transform for execution (this is for 4 layers), so we still have reordering that hurts us.
```
#the following trace uses ~9.84GB memory
# Constructed by Saved for backward remat trace (took 20.46 milliseconds)
#the following trace uses ~12.39GB memory
# Constructed by Transform for execution (took 886 milliseconds)
```
Note that the difference of 2.55GB is smaller than the difference between thunder and eager (5.82GB).

t-vi:
With the two latest bits

- dumb down fusion logic to not change ordering,
- splitting ThunderFunction to avoid the effect in #1379,

I have that

```
python thunder/benchmarks/benchmark_litgpt.py --model_name stablecode-completion-alpha-3b --compile thunder --checkpoint_activations True --low_precision_mode none --micro_batch_size 1 --n_layer 4
```

is on par (even a little below) the same with `--compile eager`.
I will be looking at the failing tests and splitting out some bits that can be handled independently.

t-vi:
This has been merged bit by bit, so it's obsolete.

PR #1573, opened 2024-12-19
Rename AdHocExecutor -> TemporaryExecutor and add documentation

IvanYashchuk:
We need a more precise name for an unnamed operator executor than "AdHoc," which can be ambiguous and difficult for new developers to understand. This class is used for dynamic/runtime registrations and manages one-off operations, and this should be reflected in the name to be self-documenting. I chose "Transient" because it communicates the executor's temporary/ephemeral nature.

Some other potential names based on functionality:
* DynamicOperatorExecutor - highlights dynamic nature
* TemporaryOperatorExecutor - emphasizes temporary aspect
* RuntimeOperatorExecutor - focuses on runtime registration

cc @borda @apaz-cli


t-vi:
I don't think "Operator" adds anything for clarity here.
Temporary sounds better than Transient to me.

IvanYashchuk:
Sure, let's rename it to Temporary.

The base class is Executor and its child OperatorExecutor keeps the "Executor" suffix. By this logic, any child of OperatorExecutor can maintain OperatorExecutor for consistency. TemporaryOperatorExecutor is not a FusionExecutor, but a TemporaryExecutor could also be a FusionExecutor.

mruberry:
This may be worth a brief online chat, too, because I'm curious about the details of this executor. It sounds interesting!

t-vi (approved):
Let's have it. Thank you @IvanYashchuk

PR #1576, opened 2024-12-20
[nvfuser executor] Allow sm89 for fp8 types

crcrpar:
## What does this PR do?

With NVIDIA/Fuser#3624, devices >= sm89 get allowed to use nvfuser executor for fp8.

### DIFF
--- a/thunder/executors/nvfuserex_impl.py
+++ b/thunder/executors/nvfuserex_impl.py
@@ -171,8 +171,13 @@ _low_precision_floats = (dtypes.float16, dtypes.float16_, dtypes.bfloat16, dtype
 
 
 def device_supports_fp8() -> bool:
+    cuda_version: LooseVersion | None = None
+    if torch.version.cuda is None:
+        return False
+    else:
+        cuda_version = LooseVersion(cuda_version)
     cuda_major, cuda_minor = torch.cuda.get_device_capability()
-    return (cuda_major, cuda_minor) >= (8, 9)
+    return (cuda_version >= LooseVersion("12.01") and (cuda_major, cuda_minor) >= (8, 9)) or (cuda_version >= LooseVersion("11.8") and cuda_major >= 9)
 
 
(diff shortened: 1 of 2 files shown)


crcrpar:
for me to be verbose, this waits on the mentioned nvfuser pr

crcrpar:
@jjsjann123 what would be the nvfuser version that ships the SM89 support?

jjsjann123:
> @jjsjann123 what would be the nvfuser version that ships the SM89 support?

You caught me! I forgot to bump nvfuser version in the PR. I'll go back and do that. So if you are making a version guard, make it 0.2.24 (nvfuser is currently at 0.2.23, and the PR is already in).

thunder/executors/nvfuserex_impl.py
```diff
@@ -161,8 +161,8 @@ def is_supported_devicetype(devicetype: DeviceType) -> bool:
 
 
 def device_supports_fp8() -> bool:
-    cuda_major, _ = torch.cuda.get_device_capability()
-    return cuda_major > 8
+    cuda_major, cuda_minor = torch.cuda.get_device_capability()
+    return (cuda_major, cuda_minor) >= (8, 9)
```
jjsjann123, line 165: I realize that this is an ugly bit...

I think the full logic here should copy this: https://github.com/NVIDIA/Fuser/blob/6fa084312d7eec5c69d59f3eb3cbdd9fa72a1600/csrc/device_lower/analysis/device_version.cpp#L24-L39

But that's a lot... We should have a generic API on nvfuser side that does `is_dtype_support_on_device(dtype, device_index)`

mruberry: A Python function exposed by nvFuser for this logic would be great!

mruberry: (Even if we don't do that in this PR, an issue for it would be great)

PR #1581, opened 2024-12-22
split ThunderFunction to deallocate grad_outs while computing backward

t-vi:
Fixes: #1379 

Unfortunately, this does not work when using Thunder inside a cuda graphs capture, but maybe this is not the ideal way of running Thunder anyways.


t-vi:
This seems to collide with some dynamo trickery, so we should only use this for the direct jit invocation for now.

lantiga (approved):
Looks great! It would be good to add a before/after memory usage on the PR for posterity.

PR #1583, opened 2024-12-23
[tensor wrapper subclass] Add Proxy, `prim`, and lookaside

crcrpar:
## What does this PR do?

Implement
- a new Proxy to express tensor wrapper subclasses
- a new prim and a new lookaside to support `MySubclass(...)` inside programs to `thunder.jit`

### DIFF
--- a/thunder/core/jit_ext.py
+++ b/thunder/core/jit_ext.py
@@ -63,6 +63,7 @@ from thunder.core.proxies import (
     StringProxy,
     TensorProxy,
     FutureTensorProxy,
+    SubclassTensorProxy,
     make_proxy_name,
     Variable,
     variableify,
@@ -928,6 +929,42 @@ def _general_jit_torch_checkpoint_lookaside(
 
     wrapped_thunder_function = wrap_const(thunder_function)
     return interpreter_needs_wrap(checkpoint)(wrapped_thunder_function, *args, **kwargs)
+
+
+@register_general_jit_lookaside(torch.Tensor._make_wrapper_subclass)
+def _make_wrapper_subclass(
+    cls: torch._C._TensorMeta,
+    size: Sequence[int],
+    strides: Sequence[int] | None = None,
+    storage_offset: int | None = None,
+    memory_format: torch.memory_format | None = None,
+    dtype: torch.dtype | None = None,
+    layout: torch.layout | None = torch.strided,
+    device: torch.device | None = None,
+    pin_memory: bool = False,
+    requires_grad: bool = False,
+    dispatch_sizes_strides_policy: str | None = None,
+    dispatch_device: bool = False,
+    dispatch_layout: bool = False,
+    _extra_dispatch_keys: torch.DispatchKeySet | None = None,
+    storage_size: int | None = None,
+):
+    ucls = unwrap(cls)
+    usize = unwrap(size)
+    udtype = unwrap(dtype)
+    udevice = unwrap(device)
+    urequires_grad = unwrap(requires_grad)
+
+    subclass = SubclassTensorProxy(
+        None,
+        shape=usize,
+        device=udevice,
+        dtype=udtype,
+        requires_grad=urequires_grad,
+        history=ProvenanceRecord(PseudoInst.LOOKASIDE, [cls.provenance]),
+        subclass_type=ucls,
+    )
+    return wrap(subclass, provenance=ProvenanceRecord(PseudoInst.LOOKASIDE, [cls.provenance]))
 
 
 # Adds proxy methods
--- a/thunder/core/prims.py
+++ b/thunder/core/prims.py
@@ -4073,7 +4073,6 @@ def tensor_subclass_ctor_meta(
         requires_grad=requires_grad,
         tensors=tensors,
         non_tensors=non_tensors,
-        history=[t.history for t in tensors],
     )
     return s
 
--- a/thunder/core/prims.py
+++ b/thunder/core/prims.py
@@ -1,3 +1,4 @@
+from __future__ import annotations
 from enum import auto, Enum
 from numbers import Number
 from functools import reduce, wraps
@@ -5,13 +6,17 @@ import operator
 import builtins
 import math
 from types import NoneType
-from typing import Union, Type, Any, List, Dict, Tuple, Optional
+from typing import Union, Type, Any, List, Dict, Tuple, Optional, TYPE_CHECKING
 from collections.abc import Callable
 from collections.abc import Callable, Hashable, Sequence
 
 import torch
 
 from thunder.core.langctxs import LanguageContext, register_langctx, Languages, langctx
+
+if TYPE_CHECKING:
+    from collections.abc import Iterable
+    from thunder.core.codeutils import ContextObject
 
 #
 # Creates and registers the torch language context
@@ -77,6 +82,7 @@ from thunder.core.proxies import (
     TupleProxy,
     AnyProxy,
     IntegerProxy,
+    SubclassTensorProxy,
 )
 import thunder.core.codeutils as codeutils
 from thunder.core.codeutils import Printable
@@ -272,6 +278,8 @@ class PrimIDs(Enum):
     COPY_ = auto()
     #
     SINK = auto()
+    # Tensor Subclasses methods
+    TENSOR_SUBCLASS_CTOR = auto()
 
 
 class OpTags(Enum):
@@ -3543,7 +3551,10 @@ transpose = make_prim(PrimIDs.TRANSPOSE, "transpose", meta=transpose_meta, tags=
 view = make_prim(PrimIDs.VIEW, "view", meta=reshape_meta, tags=(OpTags.SHAPE_OP,))
 
 
-def shallow_copy_meta(a: TensorProxy, /) -> TensorProxy:
+def shallow_copy_meta(a: TensorProxy | SubclassTensorProxy, /) -> TensorProxy:
+    if isinstance(a, SubclassTensorProxy):
+        # SubclassTensorProxy(like=...) would not copy some attrs such as `_tensors` while replace does.
+        return a.replace()
     return TensorProxy(like=a)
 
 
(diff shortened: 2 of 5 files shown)


thunder/executors/torchex.py
```diff
@@ -2220,3 +2220,37 @@ def _shape_impl(t):
...
+    new_imports = {t.__name__: t for t in filtered_types}
+    bsym._import_ctx.update(new_imports)
+
+
+tensor_subclass_ctor = ex.register_operator(
+    "tensor_subclass_ctor",
+    meta=prims.tensor_subclass_ctor,
+    fn=_tensor_subclass_ctor,
+    bind_postprocess=_bind_postprocess_of_tensor_subclass_ctor,
+    python_printer=prims.printer_of_tensor_subclass_ctor,
```
IvanYashchuk, line 2254: This is very clever. The `bind_postprocess` together with `python_printer` achieve the effect that we generate an execution Python program that looks like people would write it in a script
```py
t0 = ScaleTensorSubclass(x, scale)  # t0: "cpu f32[2, 2]"
```
even though `bsym.args` and `bsym.kwargs` are not just `x, scale`.

thunder/core/prims.py
```diff
@@ -4048,3 +4059,139 @@ def sink_meta(*args, **kwargs):
...
+                # note(crcrpar): maybe `thunder.core`?
+                and not t.__module__.startswith("thunder.")
+                and not t.__module__.startswith("torch.")
+            ),
+            types,
+        )
+    )
+
+
+def printer_of_tensor_subclass_ctor(
```
IvanYashchuk, line 4116: Currently, the execution and initial traces look the same even though they contain `prims.tensor_subclass_ctor` and `torchex.tensor_subclass_ctor` as `bsym.sym`. Could you modify the printer to distinguish between the two?

thunder/core/prims.py
```diff
@@ -4048,3 +4059,139 @@ def sink_meta(*args, **kwargs):
...
+        filtered_types += filter_types(types)
+    new_imports = {t.__name__: t for t in filtered_types}
+    bsym._import_ctx.update(new_imports)
+
+
+tensor_subclass_ctor = make_prim(
+    PrimIDs.TENSOR_SUBCLASS_CTOR,
+    "tensor_subclass_ctor",
+    meta=tensor_subclass_ctor_meta,
+    python_printer=printer_of_tensor_subclass_ctor,
```
IvanYashchuk, line 4195: Is the initial trace with this printer able to work with TensorProxy inputs? Here's an example for what I mean:
```py
In [1]: import torch

In [2]: import thunder

In [3]: a = torch.randn(3, 3)

In [4]: def f(x): return x.sin()

In [5]: jf = thunder.jit(f)

In [6]: jf(a);

In [7]: initial_trace = jf._lc_cs.last_traces[0]

In [8]: with thunder.core.trace.detached_trace():
   ...:     res = initial_trace.python_callable()(*initial_trace.args)
   ...: 

In [9]: res
Out[9]: (<TensorProxy(name="t0", dtype=thunder.dtypes.float32, shape=(3, 3))>,)
```
Here the initial trace is a valid Thunder program able to accept and return TensorProxies if there's an active trace (`detached_trace` in this case).

thunder/core/jit_ext.py
```diff
@@ -930,6 +931,42 @@ def thunder_function(*args, **kwargs):
     return interpreter_needs_wrap(checkpoint)(wrapped_thunder_function, *args, **kwargs)
 
 
+@register_general_jit_lookaside(torch.Tensor._make_wrapper_subclass)
```
IvanYashchuk, line 934: Is there a reason for this to be defined as jit_lookaside? It looks simple enough to be just in `thunder/torch/__init__.py` with the `@torchsymbol` decorator.

thunder/core/proxies.py
```diff
@@ -1880,6 +1880,166 @@ def real(self):
...
+                self.shape,
+                self.device,
+                self.dtype,
+                self.requires_grad,
+                self._tensors,
+                self._non_tensors,
+                output=self,
+            )
+            cls_module = inspect.getmodule(self._subclass_type)
+            bsym.sym = replace(bsym.sym, _module=cls_module)
```
IvanYashchuk, line 1952: Can you please remind me how `Symbol._module` is used? For hashing, printing, and what else?

Can this be part of `bind_postprocess`? `_subclass_type` is part of the arguments so the module is available after binding.

thunder/core/prims.py
```diff
@@ -4048,3 +4059,139 @@ def sink_meta(*args, **kwargs):
...
+    s = SubclassTensorProxy(
+        name,
+        subclass_type=cls,
+        shape=shape,
+        device=device,
+        dtype=dtype,
+        requires_grad=requires_grad,
+        tensors=tensors,
+        non_tensors=non_tensors,
+        history=[t.history for t in tensors],
```
IvanYashchuk, line 4076: What breaks without this?

thunder/core/jit_ext.py
```diff
@@ -930,6 +931,42 @@ def thunder_function(*args, **kwargs):
...
+
+    subclass = SubclassTensorProxy(
+        None,
+        shape=usize,
+        device=udevice,
+        dtype=udtype,
+        requires_grad=urequires_grad,
+        history=ProvenanceRecord(PseudoInst.LOOKASIDE, [cls.provenance]),
+        subclass_type=ucls,
+    )
```
IvanYashchuk, line 966: Why does this lookaside construct a proxy which has a bsym construction inside its init method instead of calling `prims.tensor_subclass_ctor`?
```suggestion
    subclass = prims.tensor_subclass_ctor(ucls, None, usize, udevice, udtype, urequires_grad, [], [])
```

Tests from this PR pass when I replaced `SubclassTensorProxy->prims.tensor_subclass_ctor` and removed bsym construction in the init method:
```diff
diff --git a/thunder/core/jit_ext.py b/thunder/core/jit_ext.py
index e8841954..e06b0ebd 100644
--- a/thunder/core/jit_ext.py
+++ b/thunder/core/jit_ext.py
@@ -955,15 +955,7 @@ def _make_wrapper_subclass(
     udevice = unwrap(device)
     urequires_grad = unwrap(requires_grad)
 
-    subclass = SubclassTensorProxy(
-        None,
-        shape=usize,
-        device=udevice,
-        dtype=udtype,
-        requires_grad=urequires_grad,
-        history=ProvenanceRecord(PseudoInst.LOOKASIDE, [cls.provenance]),
-        subclass_type=ucls,
-    )
+    subclass = prims.tensor_subclass_ctor(ucls, None, usize, udevice, udtype, urequires_grad, [], [])
     return wrap(subclass, provenance=ProvenanceRecord(PseudoInst.LOOKASIDE, [cls.provenance]))
 
 
diff --git a/thunder/core/proxies.py b/thunder/core/proxies.py
index c6c88656..424e60d6 100644
--- a/thunder/core/proxies.py
+++ b/thunder/core/proxies.py
@@ -1931,25 +1931,26 @@ class SubclassTensorProxy(TensorProxy):
             self._non_tensors = kwarg_non_tensors
             self._subclass_type = subclass_type
         else:
+            pass
             # TODO(crcrpar): Think about materializing `self` so that we can
             # call `__tensor_init__` to know each attribute names.
             from dataclasses import replace
             import inspect
             from thunder.core import prims
 
-            bsym = prims.tensor_subclass_ctor.bind(
-                self._subclass_type,
-                self.name,
-                self.shape,
-                self.device,
-                self.dtype,
-                self.requires_grad,
-                self._tensors,
-                self._non_tensors,
-                output=self,
-            )
-            cls_module = inspect.getmodule(self._subclass_type)
-            bsym.sym = replace(bsym.sym, _module=cls_module)
+            # bsym = prims.tensor_subclass_ctor.bind(
+            #     self._subclass_type,
+            #     self.name,
+            #     self.shape,
+            #     self.device,
+            #     self.dtype,
+            #     self.requires_grad,
+            #     self._tensors,
+            #     self._non_tensors,
+            #     output=self,
+            # )
+            # cls_module = inspect.getmodule(self._subclass_type)
+            # bsym.sym = replace(bsym.sym, _module=cls_module)
 
             # NOTE(crcrpar): A callable being `thunder.jit`ed can call `MySubclassTensor(...)`
             # inside of it either directly or indirectly: indirect way is to call it through
@@ -1959,11 +1960,11 @@ class SubclassTensorProxy(TensorProxy):
             # but not, otherwise. As [the lookasdie of `torch.autograd.Function`](
             # https://github.com/Lightning-AI/lightning-thunder/blob/3d42c10/thunder/core/jit_ext.py#L655)
             # puts the temporary scope to the current trace.
-            current_trace = get_tracectx()
-            if id(current_trace.bound_symbols) == id(cur_tail_scope := current_trace.scopes[-1]):
-                current_trace.add_bound_symbol(bsym)
-            else:
-                cur_tail_scope.append(bsym)
+            # current_trace = get_tracectx()
+            # if id(current_trace.bound_symbols) == id(cur_tail_scope := current_trace.scopes[-1]):
+            #     current_trace.add_bound_symbol(bsym)
+            # else:
+            #     cur_tail_scope.append(bsym)
 
             if not self._tensors and not self._non_tensors:
                 for a in tensors:
```

thunder/core/proxies.py
```diff
@@ -1880,6 +1880,166 @@ def real(self):
...
+            self._non_tensors = kwarg_non_tensors
+            self._subclass_type = subclass_type
+        else:
+            # TODO(crcrpar): Think about materializing `self` so that we can
+            # call `__tensor_init__` to know each attribute names.
+            from dataclasses import replace
+            import inspect
+            from thunder.core import prims
+
+            bsym = prims.tensor_subclass_ctor.bind(
```
IvanYashchuk, line 1940: Tests pass when this piece is removed and `prims.tensor_subclass_ctor` is used jit_ext.py. Is there a test that would require doing bsym construction here?

thunder/core/proxies.py
crcrpar: I'm embarrassingly losing my memory. One thing which might be related is 
https://github.com/Lightning-AI/lightning-thunder/blob/52ee541073ddf6dce78b77ddb272621f712e3c6a/thunder/core/symbol.py#L606-L641, more specifically, this else path, https://github.com/Lightning-AI/lightning-thunder/blob/52ee541073ddf6dce78b77ddb272621f712e3c6a/thunder/core/symbol.py#L629-L638 for which `Symbol._module` cannot be None unless we change the condition

thunder/core/prims.py
crcrpar: I've never thought of it. Will try it out

thunder/core/jit_ext.py
crcrpar: hmm, maybe no.
I originally tried to make it more graceful in a way but currently it's quite simple as you see.

thunder/core/prims.py
crcrpar: now it looks like a fallout of things I tried while working on understanding and implementing things for tensor wrapper subclasses. let me see...

thunder/core/prims.py
crcrpar: how would we want to distinguish them? With comments (a.k.a bsym.header)? or signature?

t-vi: Could the TorchEx version be using the subclass constructor directly or somesuch?
Or you do it with prims.tensor_subclass_ctor similar to how we have ltorch.foo und torch.foo for torch symbols.

thunder/core/jit_ext.py
crcrpar: If it works, then it'd simplify the codebase, which is nice.

Even though meticulous mode of mine want to say something.
After trying to remember things and give another consideration, I think it could be considered as an undesirable use of that prim in that we call it without any actual tensors and non-tensor values, unlike when we instantiate tensor wrapper subclasses. To some extent, I even think it'd be reasonable to check `tensors: list[Tensor]` is not empty in the meta function

PR #1588, opened 2024-12-27
clean up imports of `jit_ext.py`

crcrpar:
## What does this PR do?

some cleaning of jit_ext's imports

### DIFF
--- a/thunder/core/jit_ext.py
+++ b/thunder/core/jit_ext.py
@@ -1,120 +1,80 @@
-import thunder
+from __future__ import annotations
 import math
-from typing import Any, Optional, Dict, Tuple, Literal
-import builtins
+from typing import Any
 import collections
-from collections.abc import ValuesView, Iterable, Iterator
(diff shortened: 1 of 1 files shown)


thunder/core/jit_ext.py
```diff
@@ -1671,7 +1634,7 @@ def is_variableified_tensorproxy(v: Variable | Proxy) -> Proxy:
 
             prim(*args)
 
-        cache_info = thunder._get_cache_info()
+        cache_info = _get_cache_info()
```
t-vi, line 1637: What's the advantage here?
To my mind, the original version is clearer because you immediately see that it is a function from a different module (and which). I can see how the shorter way is important for things that get used a lot, but this seems distributing the information that this is a utility function from the `thunder/__init__.py` int two lines that are 300 lines apart.

t-vi (approved):
Thank you @crcrpar . I took the liberty of making it more about the top level import while rebasing.

Issue #1595, opened 2024-12-30
Name argument for Proxies is silently ignored for names that were used before


IvanYashchuk:
## 🐛 Bug

In the snippet below we create three TensorProxies with the same name and only the first object is created with the specified name:
```py
In [1]: from thunder.core.proxies import TensorProxy

In [2]: from thunder.core.devices import cpu

In [3]: from thunder.core.dtypes import float32

In [4]: from thunder.core.trace import set_tracectx, TraceCtx

In [5]: set_tracectx(TraceCtx(None))
Out[5]: <Token var=<ContextVar name='tracectx' at 0x7fbd9c7e73d0> at 0x7fbd8f69de00>

In [6]: TensorProxy(shape=(1,), device=cpu, dtype=float32, name="test", requires_grad=True)
Out[6]: <TensorProxy(name="test", dtype=thunder.dtypes.float32, shape=(1,))>

In [7]: TensorProxy(shape=(1,), device=cpu, dtype=float32, name="test", requires_grad=True) # 'test' is unused without a warning or error
Out[7]: <TensorProxy(name="t0", dtype=thunder.dtypes.float32, shape=(1,))>

In [8]: TensorProxy(shape=(1,), device=cpu, dtype=float32, name="test", requires_grad=True) # 'test' is unused without a warning or error
Out[8]: <TensorProxy(name="t1", dtype=thunder.dtypes.float32, shape=(1,))>
```

cc @apaz-cli

mruberry:
What would you like the behavior to be?

IvanYashchuk:
I prefer the specified name to be always used regardless of the ambient trace state. I prefer modular components with weak or no interactions and specialized checks on whether component ties are harmonious with each other.

Another possibility is to raise an error if the ambient trace has already generated a proxy with the requested name, according to the discussion in the linked PR this option is implemented in https://github.com/Lightning-AI/lightning-thunder/pull/1596.

PR #1596, opened 2024-12-30
Error out when specified name for Proxy creation is already used

IvanYashchuk:
Previously the provided name could be ignored if it was included in the `get_tracectx().names` set. The behavior is undocumented and may be confusing. With the change in this pull request, an error is thrown if the specified name is already used.

Fixes https://github.com/Lightning-AI/lightning-thunder/issues/1595.


IvanYashchuk:
There are lots of things to fix from the CI.

t-vi (changes requested):
As far as I know, this has been the behaviour forever (at least for all of 2024), and our code uses the current behaviour throughout.
If we are going to change things around naming and proxies, we should have traces own proxies.

I'm quite tired of bandaid fixes, having met numerous of them during my exploration of memory use.

lantiga:
I second that, I think it’s a good time to finally do this and making the overall structure easier to reason about.

@IvanYashchuk what’s your angle on moving to traces owning proxies?

thunder/core/proxies.py
```diff
@@ -71,17 +71,6 @@ def unvariableify(x: Any) -> Any:
     return x
 
 
-# Tries to register a proxy name and returns a boolean indicating success
-def register_proxy_name(name: None | str = None):
```
mruberry, line 75: What happens if two proxies share a trace and a name? Wouldn't the Python program use the same name to refer to both of them?

IvanYashchuk:
> what’s your angle on moving to traces owning proxies?

I don't know what it would mean. Do you know if there is a summary somewhere?

IvanYashchuk:
> As far as I know, this has been the behaviour forever (at least for all of 2024), and our code uses the current behaviour throughout.

Our code won't use the current behavior when this PR is merged and all CI checks pass.

> If we are going to change things around naming and proxies, we should have traces own proxies.

I don't know what "traces own proxies" means or what problems it can fix.

> I'm quite tired of bandaid fixes, having met numerous of them during my exploration of memory use.

It's a fix for a bug that lived in the codebase for a bit too long than it should have. I want to decrease the number of assumptions Thunder developers need to be aware of. This particular PR would enable removing the following workarounds:
1. https://github.com/Lightning-AI/lightning-thunder/blob/ad790ae0d5eefce9d1d8a719d7aba48dd5226eb9/thunder/core/symbol.py#L333-L335
2. https://github.com/Lightning-AI/lightning-thunder/blob/ad790ae0d5eefce9d1d8a719d7aba48dd5226eb9/thunder/core/trace_interpreter.py#L123-L128

thunder/core/proxies.py
IvanYashchuk: Yes it would. When two proxies with the same name are used in a trace the Python program is generated as prescribed by the order of bound symbols:
```py
In [1]: from thunder.core.devices import cpu

In [2]: from thunder.core.dtypes import float32

In [3]: from thunder.core.proxies import TensorProxy

In [4]: p_test0 = TensorProxy(name="test", shape=(1,), device=cpu, dtype=float32)

In [5]: p_test1 = TensorProxy(name="test", shape=(1,), device=cpu, dtype=float32)

In [6]: a = TensorProxy(name="name", shape=(1,), device=cpu, dtype=float32)

In [7]: from thunder.core.prims import exp

In [8]: b1 = exp.bind(a, output=p_test1)

In [9]: b0 = exp.bind(a, output=p_test0)

In [10]: bsyms = [b0, b1]

In [11]: from thunder.core.trace import TraceCtx

In [12]: t = TraceCtx(None)

In [13]: t.bound_symbols = bsyms

In [14]: t
Out[14]: 
import thunder
import torch

@torch.no_grad()
@no_autocast
# No signature available
  test = prims.exp(name)  # test: "cpu f32[1]"
  test = prims.exp(name)  # test: "cpu f32[1]"
```

mruberry: Isn't it a problem if names can collide like that?

IvanYashchuk: Well, of course, it can be problematic, but it's the same case when a single proxy is used for both bound symbols.

t-vi:
I don't know what I can say here:
- Currently, proxies get renamed when the name is already used in the proxy,
- after not renaming, it clashes.

t-vi:
So I can agree that silently renaming is not good. If avoiding that is the main issue, we could raise an error?

IvanYashchuk:
> So I can agree that silently renaming is not good. If avoiding that is the main issue, we could raise an error?

Let's raise an error (done in https://github.com/Lightning-AI/lightning-thunder/pull/1596/commits/63505b0a482b2780d2f3c74fc3aaf9cbbb40d425).

IvanYashchuk:
> So I can agree that silently renaming is not good. If avoiding that is the main issue, we could raise an error?

Right, we can't allow this behavior if we agree it's not intentional. Let's raise an error (done in https://github.com/Lightning-AI/lightning-thunder/pull/1596/commits/63505b0a482b2780d2f3c74fc3aaf9cbbb40d425).

t-vi (approved):
Thank you @IvanYashchuk @mruberry

t-vi:
I think there still is a case to fix of using a fixed name, but other than that looks great.

Issue #1577, opened 2024-12-20
Consider adding is_leaf attribute to TensorProxies


beverlylytle:
In the context of [raising errors on inplace operations on leaf tensors requiring grad](https://github.com/Lightning-AI/lightning-thunder/pull/1458#discussion_r1889899879), it may be beneficial to add an `is_leaf` attribute to `TensorProxies` in a manner similar to the addition of the attribute `requires_grad`.

Issue #1570, opened 2024-12-18
"requires_grad" attribute on intermediate TensorProxies is unused and misleading


IvanYashchuk:
_Originally posted by @IvanYashchuk in https://github.com/Lightning-AI/lightning-thunder/pull/1563#discussion_r1890146557_

`requires_grad` of intermediate TensorProxies is ignored in our automatic differentiation code because we haven't done the work of properly threading this property through all computations.
We should remove the ability to query `.requires_grad` from intermediate TensorProxies completely to avoid similar bugs in the future. This can be achieved by introducing a separate "InputTensorProxy" which has this attribute and removing it from the regular TensorProxy.

IvanYashchuk:
Remove it and restore it when it's correct. Today requires_grad is checked and used only for inputs of computation traces and ignored otherwise. Currently, the implemented logic is that every intermediate requires a gradient. The result of intermediate gradient calculations may be DCE'd if it remains unused and is not an output of the backward trace. The output of the backward trace is set to use `None` for every input tensor which doesn't require a gradient.

PyTorch has at least two ways of stopping gradient propagation:
1. setting `requires_grad=False` on an intermediate tensor
2. using `.detach()` on an intermediate tensor

We should support tracing both and a special operator to act as a gradient sink could be "detach". It should be possible to coerce the `tensor.requires_grad` assignment to False to `detach`. I don't think setting `tensor.requires_grad=True` does anything useful except for undoing the False assignment.

t-vi:
We're not there yet, but there may be cases where we want to create new autograd leaves when we look beyond tracing models.

PR #1599, opened 2024-12-30
Remove requires_grad from regular TensorProxy and add specialized TensorProxy with requires_grad and is_leaf attributes

IvanYashchuk:
PyTorch tensors have attributes that affect the computation and Thunder should mirror those in TensorProxies. However, some of the attributes (`requires_grad`, `is_leaf`, `stride`) are tricky to model precisely and propagate the values in Thunder's meta functions. This PR removes the possibility of querying and specifying `requires_grad` on regular TensorProxies which are used for intermediate values and reintroduces the attribute with a special class `RuntimeTensorProxy` that should be constructed through a corresponding real PyTorch tensor.
A new attribute is added, `is_leaf`, which is useful for raising errors when in-place copies should be disallowed (https://github.com/Lightning-AI/lightning-thunder/issues/1577, https://github.com/Lightning-AI/lightning-thunder/pull/1458).
It's not part of this PR but an extension to query static strides information will be easy to add in the future.

Ref https://github.com/Lightning-AI/lightning-thunder/issues/1577, https://github.com/Lightning-AI/lightning-thunder/issues/1570.

### DIFF
--- a/thunder/core/prims.py
+++ b/thunder/core/prims.py
@@ -2846,7 +2846,7 @@ def _randn_meta(
     utils.check_type(dtype, dtypes.dtype)
     utils.check_type(shape, tuple)
     utils.check_valid_shape(shape)
-    return TensorProxy(shape=shape, device=device, dtype=dtype, requires_grad=False)
+    return TensorProxy(shape=shape, device=device, dtype=dtype)
 
 
 randn = make_prim(PrimIDs.RANDN, "randn", meta=_randn_meta)
--- a/thunder/core/prims.py
+++ b/thunder/core/prims.py
@@ -2642,8 +2642,7 @@ def _where_meta(pred: Number | TensorProxy, a: Number | TensorProxy, b: Number |
     shapes = tuple(x.shape for x in (pred, a, b) if isinstance(x, TensorProxy))
     resultshape = shapes[0]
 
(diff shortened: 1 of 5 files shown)


thunder/core/proxies.py
```diff
@@ -1868,6 +1844,52 @@ def real(self):
         return method(self)
 
 
+class RuntimeTensorProxy(TensorProxy):
+    def __init__(
```
mruberry, line 1848: Would you add a docstring explaining this class's use vs. TensorProxy?

thunder/tests/test_networks.py
```diff
@@ -419,10 +419,6 @@ def test_thunderfx_mistral_nemo_small():
     assert mdl._backend.subgraph_infos, "Should have at least 1 subgraph"
 
 
-@pytest.mark.skipif(
```
mruberry, line 422: This change seems unrelated to the rest of the PR - maybe it should be in a separate PR?

IvanYashchuk: Right, it should be merged separately. It slipped through as I was testing whether HF models work.

thunder/core/prims.py
IvanYashchuk, line 1: Changes in this file should be submitted in a separate PR.

t-vi:
In particular, I feel that this is kind of change is extremely tricky w.r.t. the Thunder Python interpreter, which uses TensorProxies as, well, a proxy for Tensors.

PR #1600, opened 2024-12-30
TE: fix failing intermediate sharding test

kshitij12345:
Fixes the failing test - `test_fsdp_transformer_engine`  with intermediate sharding.

https://github.com/Lightning-AI/lightning-thunder/blob/ad790ae0d5eefce9d1d8a719d7aba48dd5226eb9/thunder/tests/distributed/test_fsdp.py#L1078

Context - 

For intermediate sharding, TransformerEngine shards the tensor and updates an internal attribute `_data` (which holds the data) on `Float8Tensor`, so this doesn't actually change the shape of `Float8Tensor` object but only updates the `_data` with the sharded buffer. 

See the links below -

https://github.com/NVIDIA/TransformerEngine/blob/838345eba4fdd2a169dd9e087d39c30a360e684a/transformer_engine/pytorch/module/linear.py#L318-L323
https://github.com/NVIDIA/TransformerEngine/blob/838345eba4fdd2a169dd9e087d39c30a360e684a/transformer_engine/pytorch/distributed.py#L910-L914

However, thunder calls `detach` in `ThunderFunction` on saved tensors (to avoid leak, see https://github.com/Lightning-AI/lightning-thunder/pull/1352)

https://github.com/Lightning-AI/lightning-thunder/blob/ad790ae0d5eefce9d1d8a719d7aba48dd5226eb9/thunder/executors/torch_autograd.py#L80-L85

Calling this detach returns a new `Float8Tensor` object which has the shape of the sharded tensor (problem is `TransformerEngine` doesn't expect this to happen between saving the tensors and gathering them in backward, as during the gather it only updates the `_data` on the `Float8Tensor` object, assuming it wasn't changed previously). 

Solution - We only call `detach` if `tensor` is a view determined by `tensor._base is not None`. 

NOTE - Above solution works as TransformerEngine makes sure that it doesn't pass a view as saved tensor (when using intermediate sharding) .
https://github.com/NVIDIA/TransformerEngine/blob/838345eba4fdd2a169dd9e087d39c30a360e684a/transformer_engine/pytorch/utils.py#L189-L216


thunder/executors/torch_autograd.py
```diff
@@ -80,7 +80,8 @@ def forward(
         def detach_if_tensor(t):
             # Some operations may claim to return Tensor (as per their meta function)
             # but may return None at Runtime (eg. noticed this for sdpa)
-            if isinstance(t, torch.Tensor):
+            if isinstance(t, torch.Tensor) and t._base is not None:
+                # Only detach if the Tensor is a view.
```
t-vi, line 84:
```suggestion
                # Only detach if the Tensor is a view.
                # This is needed because TransformerEngine can create (non-view) tensors that have different
                # metadata on the `t.detach()` output than on `t`. (They should not, but what can we do.)
                # See https://github.com/Lightning-AI/lightning-thunder/pull/1600 for details.
```

t-vi (approved):
Thank you @kshitij12345 . The reason to have it is decidedly terrible, but hey.

kshitij12345:
> Is this solution meant to be general even if TE is not used?

Yes, this should work with or without TE. The reason we call `detach` was to avoid saving view of the output, with this PR we call detach only on tensors which are view, so I presume this should work. Do you have a minor repro for the same? I am curious to know why this PR leads to a leak. Thanks!

Ref to PR that introduced calling `detach` - https://github.com/Lightning-AI/lightning-thunder/pull/1352

riccardofelluga:
Here is a full repro:

```python
import thunder
import torch
import torch.nn as nn

from transformers import AutoConfig
from transformers.models.qwen2.modeling_qwen2 import Qwen2Attention

torch.manual_seed(0)

configuration = AutoConfig.from_pretrained(
    "Qwen/Qwen2.5-7B-Instruct",
    num_hidden_layers=1,
)
configuration.hidden_size = configuration.num_attention_heads

with torch.device("cuda"):
    model = Qwen2Attention(configuration, layer_idx=0).to(torch.bfloat16)

backend = ThunderCompiler(executors=("torchcompile",))
compiled_model = torch.compile(model, backend=backend, fullgraph=True)

hidden_states = torch.randn(1, 4096, 28, dtype=torch.bfloat16, device="cuda")
attention_mask = None
position_ids = torch.randint(0, configuration.vocab_size, (1, configuration.max_position_embeddings), device="cuda")
past_key_value = None
cache_position = torch.randint(0, configuration.vocab_size, (1, configuration.max_position_embeddings), device="cuda")
position_embeddings = (torch.randn(1, 4096, 2, dtype=torch.bfloat16, device="cuda"), torch.randn(1, 4096, 2, dtype=torch.bfloat16, device="cuda"))
output_attentions = False,
use_cache = False

def run_forward_backward():
    compiled_output = compiled_model(
        hidden_states,
        position_embeddings,
        attention_mask,
        past_key_value,
        cache_position,
        position_ids = position_ids,
        output_attentions = output_attentions,
        use_cache = use_cache
        )
    print("Peak memory", torch.cuda.max_memory_allocated() / 1e6)

for i in range(5):
    run_forward_backward()

```

Peak memory will continue to increase as the iterations go.

kshitij12345:
Running the above repro with latest internal image, I am seeing -
```python
TypeError: Qwen2Attention.forward() got multiple values for argument 'position_ids'
```

Also, we have a test for previously fixed leak which seems to be fine, so maybe this leak could be due to a different interaction - https://github.com/Lightning-AI/lightning-thunder/blob/d2b5dfd59986316fb7bd6a8bbfd797464333cf70/thunder/tests/test_core.py#L3056

> Could the t._base check be done closer to TE logic instead of in the autograd tho?

Maybe, we will tag (mark) the tensors saved from TE at runtime such that `detach` is not called on them. But I would like understand first what is happening here.

riccardofelluga:
My guess is that it's about the version of HF transformers. Try to update it to >`4.48.0`

kshitij12345:
Updating to `4.48`, I can repro the leak, thanks!

PR #1608, opened 2025-01-06
fuse only consecutive symbols

t-vi:
```
python thunder/benchmarks/benchmark_litgpt.py --model_name stablecode-completion-alpha-3b --compile thunder --checkpoint_activations True --low_precision_mode none --micro_batch_size 1 --n_layer 4 --max_iters 3 --warmup_iters 2
```

On my GPU (3080):
Before this PR: Average iter time: 1271.63 ms Memory used: 13.44 GB
After this PR: Average iter time: 1267.20 ms Memory used: 13.27 GB

While this is not huge, this becomes a lot more when checkpointing is not "recompute everything in backward" (16GB to 11.14GB at some stage in optimizing the above benchmark). I would book the timing as "stable".

The core problem here is the uncontrolled reordering without regard to memory allocation. We could reintroduce the more elaborate algorithm when we figure out controlling the memory impact.

On slack, @tfogal reported:

> With microsoft/Phi-3-mini-128k-instruct and mistralai/Mistral-Nemo-Base-2407 there was no memory use difference.
> With Qwen/Qwen2.5-7B-Instruct this saved 5.1% memory (~1.8Gb).


t-vi:
With just this patch without the checkpointing changes, the vanilla JIT does not see much change in qwen memory use (probably expected), and a slight (0.5%) regression in perf, which I would chalk up as noise.

```
python thunder/benchmarks/benchmark_litgpt.py --model_name Qwen2.5-7B --compile thunder --checkpoint_activations True --low_precision_mode none --micro_batch_size 1 --n_layer 4 --max_iters 3 --warmup_iters 2 --block_size 4096
```
(I added block size.)
Average iter time: 785.11 ms, Memory used: 18.78 GB
Average iter time: 788.74 ms, Memory used: 18.76 GB

lantiga (approved):
Stamped! Unblocking the next steps.

PR #1609, opened 2025-01-06
harden scope stack

t-vi:
let's be a bit more consistent in how we use the scope stack.

- use push_scope / pop_scope and not assign the stack
- enforce that the bottom-most scope are the trace's bound symbols


t-vi:
Oops. I wanted to auto-merge. sorry about that.

Issue #1575, opened 2024-12-19
thunderfx : detecting parameters and buffers on thunderfx path


kshitij12345:
The FXGraph provided by Dynamo takes in Parameters and Buffers as arguments, however `thunder.jit` currently only determines a TensorProxy to be a parameter if it is unpacked from a Module. So, on thunderfx path, we don't tag these parameters with STATIC_MEMORY_LOCATION, leading to problem with CUDAGraphTransform and ExtraionOnlyPrologueTransform which depend on these tags.

https://github.com/Lightning-AI/lightning-thunder/blob/35ca2e9c20fdc2a2a58b7c3106ac8c617710b5a5/thunder/core/jit_ext.py#L1493-L1500

**Potential Solution for Parameters** : For parameters, maybe `thunder.jit` tag Proxies based on `isinstance(obj, nn.Parameter)`.

Sample:
```python
import torch

def backend(gm, sample_arg):
    gm.print_readable()
    print(sample_arg)
    return gm

model = torch.nn.Linear(2, 2)

cmodel = torch.compile(model, backend=backend)
cmodel(torch.randn(1, 2))

```

Output
```python
class GraphModule(torch.nn.Module):
    def forward(self, L_fn_parameters_weight_: "f32[2, 2]", L_fn_parameters_bias_: "f32[2]", L_args_0_: "f32[1, 2]"):
        l_fn_parameters_weight_ = L_fn_parameters_weight_
        l_fn_parameters_bias_ = L_fn_parameters_bias_
        l_args_0_ = L_args_0_
        
         # File: /home/kkalambarkar/git/pytorch/torch/_dynamo/external_utils.py:31 in inner, code: return fn(*args, **kwargs)
        linear: "f32[1, 2]" = torch._C._nn.linear(l_args_0_, l_fn_parameters_weight_, l_fn_parameters_bias_);  l_args_0_ = l_fn_parameters_weight_ = l_fn_parameters_bias_ = None
        return (linear,)
        
[Parameter containing:
tensor([[ 0.6478,  0.6590],
        [-0.5319, -0.3303]], requires_grad=True), Parameter containing:
tensor([0.3209, 0.6565], requires_grad=True), tensor([[ 0.7282, -0.3549]])]
{}
```

kshitij12345:
Example of ExtrationOnlyPrologueTransform not working

```python

import torch
import thunder
import thunder.transforms
[...]

PR #1611, opened 2025-01-07
[WIP] jit : update how nn.Parameters are detected and tagged with STATIC_MEMORY_LOCATION tag

kshitij12345:
Related #1575

Checking to see if this breaks CI

### DIFF
--- a/thunder/core/jit_ext.py
+++ b/thunder/core/jit_ext.py
@@ -276,6 +276,13 @@ class JitCtx:
                 p = p_new
             else:
                 p_orig = p
+
+            # We assume that a Parameter's underlying storage won't be changed.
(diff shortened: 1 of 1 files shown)


t-vi (approved):
Seems terrible, but if it unblocks you.

kshitij12345:
> Seems terrible, but if it unblocks you.

Do you have another fix in mind? Or is it the fact that we can't see nn.Module via thunderfx path that is bad?

mruberry:
Tagging parameters makes a lot of sense to me. They're parameters for a reason.

In the future we may also consider what happens if someone calls thunderfx on a module. We could detect that a module was passed to the call before passing it to torch.compile. However, there's no requirement that someone pass the module itself for optimization.

PR #1613, opened 2025-01-07
get_and_update_rng_state is a random op

t-vi:
...in the sense that it needs the RNG state and thus cannot be rematerialized.

### DIFF
--- a/thunder/core/prims.py
+++ b/thunder/core/prims.py
@@ -2795,6 +2795,7 @@ get_and_update_rng_state = make_prim(
     PrimIDs.GET_AND_UPDATE_RNG_STATE,
     "get_and_update_rng_state",
     meta=_get_and_update_rng_state_meta,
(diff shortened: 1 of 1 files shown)


mruberry (approved):
A comment for this one would be nice since it's not the most straightforward deduction

lantiga (approved):

thunder/core/prims.py
```diff
@@ -2795,6 +2795,7 @@ def _get_and_update_rng_state_meta(
     PrimIDs.GET_AND_UPDATE_RNG_STATE,
     "get_and_update_rng_state",
     meta=_get_and_update_rng_state_meta,
+    tags=(OpTags.RANDOM_OP,),
```
t-vi, line 2798:
```suggestion
    tags=(OpTags.RANDOM_OP,),  # RANDOM_OP here is a short hand for "uses / modifies the random state"
```

PR #1614, opened 2025-01-07
Add a new dynamo.last_traces entry point.

tfogal:
I am constantly rewriting this so I thought upstreaming something made sense.

Not married to the implementation, but @t-vi @IvanYashchuk curious for your thoughts on having an explicitly entry point for this?

## What does this PR do?

It adds a new entry point which iterates over all the ThunderFX subgraphs and grabs the trace information.

## PR review

### DIFF
--- a/thunder/dynamo/compiler.py
+++ b/thunder/dynamo/compiler.py
@@ -145,7 +145,7 @@ def thunderfx(fn: Callable, /, **kwargs) -> Callable:
     compiled = torch.compile(fn, backend=backend, **torch_compile_options)
 
     # We return this object instead of just the raw `compiled` Callable so that
-    # we have a place to hang the `last_traces` property.
+    # we have a place to hang the `last_*traces` properties.
     class CompiledObject:
         def __init__(self, be, func: Callable):
             self._backend = backend
--- a/thunder/dynamo/compiler.py
+++ b/thunder/dynamo/compiler.py
@@ -143,5 +143,36 @@ def thunderfx(fn: Callable, /, **kwargs) -> Callable:
 
     backend = ThunderCompiler(**thunder_options)
     compiled = torch.compile(fn, backend=backend, **torch_compile_options)
-    compiled._backend = backend
-    return compiled
+
+    # We return this object instead of just the raw `compiled` Callable so that
+    # we have a place to hang the `last_traces` property.
+    class CompiledObject:
+        def __init__(self, be, func: Callable):
+            self._backend = backend
+            self._func = func
+
+        def __call__(self, *args, **kwargs):
+            return self._func(*args, **kwargs)
+
+        @property
+        def last_traces(self) -> [Trace]:
+            """
+            Get the Thunder traces for all subgraphs of a ThunderFX callable.
+
+            .. note:: The object must have been invoked before calling this
+                      function.
+            """
+            rv: [Trace] = []
+            for sinfo in self._backend.subgraph_infos:
+                for th_fqn in sinfo.thunder_compiled_fns:
+                    trcs = thunder.last_traces(th_fqn)
+                    if trcs is not None and trcs != []:
+                        rv.append(trcs[-1])
+                    del trcs
+
+                    trcs_bw = thunder.last_backward_traces(th_fqn)
+                    if trcs_bw is not None and trcs_bw != []:
+                        rv.append(trcs_bw[-1])
+
+    c = CompiledObject(backend, compiled)
+    return c
--- a/thunder/dynamo/compiler.py
+++ b/thunder/dynamo/compiler.py
@@ -157,7 +157,8 @@ def thunderfx(fn: Callable, /, **kwargs) -> Callable:
         @property
         def last_traces(self) -> [Trace]:
             """
-            Get the Thunder traces for all subgraphs of a ThunderFX callable.
+            Get the Thunder traces for all the forward subgraphs of a ThunderFX
+            callable.
 
             .. note:: The object must have been invoked before calling this
                       function.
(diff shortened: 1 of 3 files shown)


thunder/dynamo/utils.py
```diff
@@ -842,3 +843,36 @@ def torch_inductor(fn, inputs):
 
         if not use_pytest_benchmark:
             print(f"\ntest_{graph_name}()", file=f)
+
+
+def last_traces(c: Callable) -> str:
```
t-vi, line 848: The main question I would have is whether it'd be better to return a list of lists of traces or something instead of a string.
We seem to use last traces a lot for inspecting the trace objects.
That said, given that it is entirly inside thunder.dynamo, it is more a curiosity than a preference.

tfogal: Yeah, that's a good question!
I'll let Ivan respond too before I update it.

kiya00: All traces are combined into a single string when dealing with multiple graphs, is this acceptable?

tfogal: Ivan and I chatted offline and he convinced me that a property would be better here :-).
Also, based on your feedback @t-vi I updated it to return a list of traces.

> All traces are combined ... is this acceptable?

Yeah I think this is the heart of TomV's question as well. So now we have a list such that this information isn't necessarily lost.
Some use cases will still want to use the `subgraph_infos` path to get e.g. the split reasons, or if you only care about the 3rd graph, etc., but I think that's fine; these aren't mutually exclusive.

thunder/dynamo/compiler.py
```diff
@@ -143,5 +143,37 @@ def thunderfx(fn: Callable, /, **kwargs) -> Callable:
...
+    class CompiledObject:
+        def __init__(self, be, func: Callable):
+            self._backend = backend
+            self._func = func
+
+        def __call__(self, *args, **kwargs):
+            return self._func(*args, **kwargs)
+
+        @property
+        def last_traces(self) -> [Trace]:
```
kshitij12345, line 158: I think it'd be nice to get the `backward` traces in a separate list. This would make it easier to inspect just the backward traces. Wdyt?

tfogal: 🤷 don't have a strong opinion.

Can I get a +1 from someone else? Happy to implement it if so.

riccardofelluga: +1! so it's similar UX to the way we get the traces out of the jit

tfogal: ack, will do!

thunder/dynamo/__init__.py
```diff
@@ -1,4 +1,3 @@
 from thunder.dynamo.compiler import ThunderCompiler, thunderfx
 
-
```
IvanYashchuk, line 3: This change is unnecessary.

thunder/dynamo/utils.py
```diff
@@ -16,6 +16,7 @@
 from thunder.torch import _torch_to_thunder_function_map
 from thunder.torch.langctx import torchctx
 from thunder.core.utils import check
+import thunder
```
IvanYashchuk, line 19: Is it a leftover from some experimentation? This import is unused in the file.

thunder/tests/test_dynamo.py
```diff
@@ -10,6 +10,7 @@
 import torch.nn.functional as F
 from looseversion import LooseVersion
 
+import thunder
```
IvanYashchuk, line 13: This import seems unused.

IvanYashchuk (commented):
Let's merge after @kshitij12345 approves it.

thunder/dynamo/compiler.py
```diff
@@ -143,5 +143,51 @@ def thunderfx(fn: Callable, /, **kwargs) -> Callable:
...
+            """
+            Get the Thunder traces for all the forward subgraphs of a ThunderFX
+            callable.
+
+            .. note:: The object must have been invoked before calling this
```
kshitij12345, line 163: Can we issue a warning as well if this is not the case?

thunder/dynamo/compiler.py
```diff
@@ -143,5 +143,51 @@ def thunderfx(fn: Callable, /, **kwargs) -> Callable:
...
+                    del trcs
+            return rv
+
+        @property
+        def last_backward_traces(self) -> [Trace]:
+            """
+            Get the Thunder traces for all the backward subgraphs of a
+            ThunderFX callable.
+
+            .. note:: The object must have been invoked before calling this
```
kshitij12345, line 181: Can we issue a warning as well if this is not the case?

thunder/dynamo/compiler.py
```diff
@@ -143,5 +143,51 @@ def thunderfx(fn: Callable, /, **kwargs) -> Callable:
...
+            callable.
+
+            .. note:: The object must have been invoked before calling this
+                      function.
+            """
+            rv: [Trace] = []
+            for sinfo in self._backend.subgraph_infos:
+                for th_fqn in sinfo.thunder_compiled_fns:
+                    trcs = thunder.last_traces(th_fqn)
+                    if trcs is not None and trcs != []:
```
kshitij12345, line 170: Curious to know if there are cases where `trcs` could be None or []. I think `trcs` should always be populated.

https://github.com/Lightning-AI/lightning-thunder/blob/21d93191e6629696b9cb875fedd9dc218c58e761/thunder/__init__.py#L827-L839

kshitij12345 (approved):
Overall LGTM, just a couple of comments and should clean-up the changes pointed by @IvanYashchuk 

Thank you @tfogal

thunder/dynamo/compiler.py
tfogal: `None` is clearly handled (i'll remove that), but an empty list happens in Alexandros' benchmark scripts for one or more of the 3 models; i frequently modify that code in _nemo.py to account for this.

thunder/dynamo/utils.py
tfogal: yes, will fix

thunder/dynamo/__init__.py
tfogal: ack, thanks, fixed.

rewrote history to do so, though; note the diff is respun. sorry about that.

thunder/dynamo/compiler.py
tfogal: yes, added

t-vi (approved):
Stamping based on @kshitij12345 's review. Thank you @tfogal @kshitij12345

Issue #1627, opened 2025-01-09
Triu operator which can be executed via nvFuser


protonu:
## 🚀 Feature

We want to implement a Triu operator which can be executed by nvFuser
### Motivation

We want to improve performance via nvFuser execution

### Additional context

We can implement Triu like the current implementation of Tril
https://github.com/Lightning-AI/lightning-thunder/blob/c5c0b77dd9c385a2fd0dacb4abdf85dbfb33d540/thunder/torch/__init__.py#L2392

Or we can execute if calling the Triu exposed in nvFuser 
(https://github.com/NVIDIA/Fuser/pull/3637)

cc @t-vi

mruberry:
@beverlylytle — would you be interested in adding `triu`? It and `tril` are interesting operations

protonu:
I added to PR (https://github.com/Lightning-AI/lightning-thunder/pull/1631) to address this issue.

protonu:
Closed by PR https://github.com/Lightning-AI/lightning-thunder/pull/1631.
Thanks @mruberry @tfogal @beverlylytle @nikitaved !

PR #1631, opened 2025-01-10
Adding a decomposition and a test for triu

protonu:
This PR decomposes `triu` in the same manner as `tril`. This also extends `test_ops.py` to cover `triu`

</details>

## What does this PR do?

Fixes # (issue).
https://github.com/Lightning-AI/lightning-thunder/issues/1627

## PR review


thunder/tests/opinfos.py
```diff
@@ -2913,6 +2913,30 @@ def tril_sample_generator(op, device, dtype, requires_grad, **kwargs):
 )
 conditional_and_mask_ops.append(tril_opinfo)
 
+triu_opinfo = OpInfo(
+    ltorch.triu,
+    sample_input_generator=tril_sample_generator,
+    torch_reference=torch.triu,
+    test_directives=(
+        # Not all PyTorch versions support complex32 tril
```
beverlylytle, line 2921: Copypasta: tril -> triu (here and below).

thunder/torch/__init__.py
```diff
@@ -2409,6 +2409,29 @@ def tril_(a: TensorLike, /, diagonal: int = 0, *, fill_value: None | Number = No
     return prims.copy_(tril(a, diagonal, fill_value=fill_value), a)
 
 
+# NOTE triu is the same as tril except that we modify the inequality to return the upper triangluar
```
beverlylytle, line 2412: typo: triangular

beverlylytle (commented):
Thanks!

This looks good so far, but the operator also needs to be registered in the torch executor.  See [here](https://github.com/Lightning-AI/lightning-thunder/blob/6708b160f8557156167043fb2b9232029513c09a/thunder/executors/torchex.py#L1101) and [here](https://github.com/Lightning-AI/lightning-thunder/blob/6708b160f8557156167043fb2b9232029513c09a/thunder/executors/torchex.py#L1127) for how it is done for `tril`.

protonu:
> Thanks!
> 
> This looks good so far, but the operator also needs to be registered in the torch executor. See [here](https://github.com/Lightning-AI/lightning-thunder/blob/6708b160f8557156167043fb2b9232029513c09a/thunder/executors/torchex.py#L1101) and [here](https://github.com/Lightning-AI/lightning-thunder/blob/6708b160f8557156167043fb2b9232029513c09a/thunder/executors/torchex.py#L1127) for how it is done for `tril`.

Thanks! Please let me know if I need to register this anywhere else.

thunder/executors/torchex.py
```diff
@@ -1133,6 +1134,15 @@ def _tril_transform(a: TensorLike, /, diagonal: int = 0, *, fill_value: None | N
     return tril(a, diagonal)
 
 
+# NOTE PyTorch's triu like tril does not have a fill_value parameter
+def _triu_checker(a: TensorLike, /, diagonal: int = 0, *, fill_value: None | Number = None) -> bool:
+    return fill_value is None
+
+
+def _triu_transform(a: TensorLike, /, diagonal: int = 0, *, fill_value: None | Number = None) -> TensorLike:
```
beverlylytle, line 1142: Great!  I think the last thing is that the implementation needs to be registered with this checker and transform like line 1150.

protonu: Thanks, I should have caught that

thunder/torch/__init__.py
```diff
@@ -2409,6 +2409,29 @@ def tril_(a: TensorLike, /, diagonal: int = 0, *, fill_value: None | Number = No
     return prims.copy_(tril(a, diagonal, fill_value=fill_value), a)
 
 
+# NOTE triu is the same as tril except that we modify the inequality to return the upper triangular
+# NOTE matrix instead of the lower triangular matrix.
+@torchsymbol(torch.triu, is_method=True)
+def triu(a: TensorLike, /, diagonal: int = 0, *, fill_value: None | Number = None) -> TensorLike:
+    utils.check(a.ndim >= 2, lambda: f"triu: a ({a.ndim=}) must have at least two dimensions")
```
nikitaved, line 2416: Alternatively, `triu` is `tril` applied to a transposed input.

protonu: That sounds good - I'll add that to the comment.

nikitaved: But that means that that you can implement `triu` as `tril`. One only has to be careful, because then, if decomposed like that, the result is going to be column-major, and not row-major. But, given that there is a transpose prim, flips from contiguous to column-major should not be a problem? CC @mruberry

thunder/torch/__init__.py
```diff
@@ -2409,6 +2409,29 @@ def tril_(a: TensorLike, /, diagonal: int = 0, *, fill_value: None | Number = No
     return prims.copy_(tril(a, diagonal, fill_value=fill_value), a)
 
 
+# NOTE triu is the same as tril except that we modify the inequality to return the upper triangular
+# NOTE matrix instead of the lower triangular matrix.Alternatively, triu is tril applied to a transposed input.
```
nikitaved, line 2413: Sorry for the confusion, `triu(x) = tril(x.mT).mT`.

thunder/torch/__init__.py
protonu: I'd prefer implement it the way it's done in PyTorch:
https://github.com/pytorch/pytorch/blob/b8f383107eebd9495a0f132d58a970e178e15930/torch/_refs/__init__.py#L5819

thunder/torch/__init__.py
protonu: I'll remove this from the comment to prevent confusion.

thunder/torch/__init__.py
mruberry: While it's technically neat to reuse `tril` to implement `triu`, I think the straightforward implementation here that is consistent with PyTorch's decomposition should be OK for now

mruberry (commented):
Let's give the tests a chance to run

mruberry (approved):
Tests look good; thanks for adding `triu`, @protonu!

mruberry (approved):
Tests look good; thanks for adding `triu`, @protonu!

Issue #1646, opened 2025-01-15
Perf Regression : SDPA is recomputed in backward


kshitij12345:
Since https://github.com/Lightning-AI/lightning-thunder/pull/1615, recomputing in backward is on by default.

https://github.com/Lightning-AI/lightning-thunder/blob/21d93191e6629696b9cb875fedd9dc218c58e761/thunder/core/transforms.py#L3160-L3164

This means that we may end up recomputing compute heavy ops like SDPA leading to perf regression.

Running `python thunder/benchmarks/benchmark_litgpt.py --model_name Llama-2-7b-hf --compile dynamo_thunder --micro_batch_size 1 --dump_thunder_traces true`

Current Main
```python
Model name: Llama-2-7b-hf
Seq Length: 4096
Micro BS: 1
Global BS: 1
Number of Layers: 32
Number of parameters: 6.74B
Distributed Mode: none
Compiler: dynamo_thunder
Low Precision Mode: none
Average iter time: 350.15 ms
Memory used: 64.17 GB
Tokens/s: 11697.73
Tokens/s/GPU: 11697.73
TFLOP/s: 539.09
```

With patch to disable recomputing SDPA
```python
Model name: Llama-2-7b-hf
Seq Length: 4096
Micro BS: 1
Global BS: 1
Number of Layers: 32
Number of parameters: 6.74B
Distributed Mode: none
Compiler: dynamo_thunder
Low Precision Mode: none
Average iter time: 341.78 ms
Memory used: 64.18 GB
Tokens/s: 11988.92
Tokens/s/GPU: 11988.92
TFLOP/s: 552.51
```

I think we should test recomputing for performance and memory usage on multiple models and then turn it on by default based on the numbers.

Also, we should form a list of compute heavy operations for which recomputing might not be best choice and tag them accordingly.

<details>

<summary> Patch to disable recomputing for SDPA </summary>

```patch
diff --git a/thunder/executors/cudnnex.py b/thunder/executors/cudnnex.py
index 409c1672..53ce186a 100644
--- a/thunder/executors/cudnnex.py
+++ b/thunder/executors/cudnnex.py
@@ -79,6 +79,7 @@ import thunder.core.dtypes as dtypes
 from thunder.torch import TensorLike
 from thunder.core.compile_data import get_compile_option
 from thunder.core.proxies import Proxy, TensorProxy
+import thunder
 
 
 from thunder.core.transforms import (
[...]

PR #1648, opened 2025-01-16
don't auto-recompute attention or linear

t-vi:
Fixes: #1646 

Thank you @kshitij12345 for the detailed issue.

### DIFF
--- a/thunder/core/prims.py
+++ b/thunder/core/prims.py
@@ -293,6 +293,8 @@ class OpTags(Enum):
     CTX_MANAGER_ENTER_EXIT_OP = auto()
     # Label to explicitly disable an operation from recomputing in backward - see function `recompute_saved_for_backward`.
     DONT_RECOMPUTE_IN_BACKWARD = auto()
+    # Don't automatically tag operation to be recomputed in backward
+    DONT_AUTO_RECOMPUTE_IN_BACKWARD = auto()
 
 
 # TODO RC1 Document this function and describe the parts of a primitive
@@ -3780,7 +3782,9 @@ def linear_meta(a: TensorProxy, w: TensorProxy, bias: None | TensorProxy) -> Ten
     return TensorProxy(shape=out_shape, device=a.device, dtype=dtype, requires_grad=requires_grad)
 
 
-linear = make_prim(PrimIDs.LINEAR, "linear", meta=linear_meta, tags=(OpTags.MATMUL_OP,))
+linear = make_prim(
+    PrimIDs.LINEAR, "linear", meta=linear_meta, tags=(OpTags.MATMUL_OP, OpTags.DONT_AUTO_RECOMPUTE_IN_BACKWARD)
+)
 
 
 def matmul_meta(a: TensorProxy, b: TensorProxy, /) -> TensorProxy:
@@ -3839,7 +3843,9 @@ def matmul_meta(a: TensorProxy, b: TensorProxy, /) -> TensorProxy:
     return TensorProxy(like=a, shape=shape)
 
 
-matmul = make_prim(PrimIDs.MATMUL, "matmul", meta=matmul_meta, tags=(OpTags.MATMUL_OP,))
+matmul = make_prim(
+    PrimIDs.MATMUL, "matmul", meta=matmul_meta, tags=(OpTags.MATMUL_OP, OpTags.DONT_AUTO_RECOMPUTE_IN_BACKWARD)
+)
 
 #
 # NN prims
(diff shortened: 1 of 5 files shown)


thunder/torch/__init__.py
```diff
@@ -5206,7 +5206,7 @@ def mse_loss(
 
 # TODO Add annotations
 # NOTE The scale parameter is kwarg-only in PyTorch
-@torchsymbol(torch.nn.functional.scaled_dot_product_attention)
```
kshitij12345, line 5209: I was wondering if just tagging `torch.nn.functional.sdpa` should be sufficient or is it necessary to tag the cudnn and sdpaex implementation as well.

I am not sure `recompute_saved_for_backward` is called before or after the transform_for_execution pass

kshitij12345 (commented):
Overall looks good, just have one question. Also, we should  add a test with a simple 2 layer model to verify sdpa or linear is not recomputed in backward.

Thank you @t-vi

thunder/torch/__init__.py
t-vi: Maybe it would work to check the original bsym instead (but so to my mind, we might also have a transformation replacing the SDPA and then we need the tag on the thing that is written by the transform).

t-vi: Changed this to use the original bsym but left the tags on the implementation bsym, too.

thunder/tests/test_networks.py
```diff
@@ -581,4 +581,12 @@ def forward_backward_peak(m, inp):
     mem_thunder = forward_backward_peak(jm, inp)
     mem_eager = forward_backward_peak(m, inp)
 
+    # assert that attention is not automatically recomputed, see
+    # https://github.com/Lightning-AI/lightning-thunder/issues/1646
+    assert not {
+        bsym.sym.name
+        for bsym in thunder.last_backward_traces(jm)[-1].bound_symbols
+        if "attention" in bsym.sym.name and "forward" in bsym.sym.name
+    }
```
kshitij12345, line 590: I think this should also consider `if "sdpa" in bsym.sym.name and "fwd" in bsym.sym.name` as `cudnn_sdpa_fwd` could show up as well because it is one of the default executors if available (or test can just remove this executor while passing to `thunder.jit`).


```python
    # assert that attention is not automatically recomputed, see
    # https://github.com/Lightning-AI/lightning-thunder/issues/1646
    assert not {
        bsym.sym.name
        for bsym in thunder.last_backward_traces(jm)[-1].bound_symbols
        if ("attention" in bsym.sym.name and "forward" in bsym.sym.name) or \
            ("sdpa" in bsym.sym.name and "fwd" in bsym.sym.name)
    }
```

kshitij12345 (approved):
LGTM, thank you @t-vi

thunder/tests/test_networks.py
t-vi: Great catch, I switched to `or`ing the attention/sdpa and forward/fwd separately before `and`ing.

lantiga (approved):

IvanYashchuk:
That's a good quick fix, but can we revert to the previous default behavior: don't do any recomputation except for fused operations? Current logic doesn't use the information on whether the operation will be fused. To overwrite the default rule we can propagate `torch.utils.checkpoint` tags or have opt-in automatic passes.

thunder/core/prims.py
```diff
@@ -293,6 +293,8 @@ class OpTags(Enum):
     CTX_MANAGER_ENTER_EXIT_OP = auto()
     # Label to explicitly disable an operation from recomputing in backward - see function `recompute_saved_for_backward`.
     DONT_RECOMPUTE_IN_BACKWARD = auto()
+    # Don't automatically tag operation to be recomputed in backward
+    DONT_AUTO_RECOMPUTE_IN_BACKWARD = auto()
```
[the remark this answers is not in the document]
t-vi: Now that you ask, I think the MATMUL and SDPA and then having enumerations in rematerialize is misguided.
To my mind, the property of the operators is "they're too expensive to compute twice" without the user explicitly asking more than "they're a matmul" or "they're a SDPA".

t-vi:
> That's a good quick fix, but can we revert to the previous default behavior: don't do any recomputation except for fused operations? Current logic doesn't use the information on whether the operation will be fused. To overwrite the default rule we can propagate torch.utils.checkpoint tags or have opt-in automatic passes.

Maybe, I had that as one of the options in the issue, we went for this for now.

To my mind there are multiple parts:
- we want to be sure that we don't cause memory regressions,
- I think the rematerialization for forward and backward eventually needs to work without creating the joint trace.

PR #1652, opened 2025-01-16
Check inplace leafs redo

beverlylytle:
Reverts #1651 which is the reversion of #1458, with a fix for `triu`.


t-vi (approved):
Let's relanding. Thank you for updating @beverlylytle

Issue #1658, opened 2025-01-17
Cannot turn of rematerialization for an individual lang op with hand specified grad transform


kevinstephano:
*Note*: If you have a model or program that is not supported yet but should be, please use the program coverage template.

### To Reproduce
Patch to apply:
```python
diff --git a/thunder/core/rematerialization.py b/thunder/core/rematerialization.py
index 37ebd5a1..0a11772d 100644
--- a/thunder/core/rematerialization.py
+++ b/thunder/core/rematerialization.py
@@ -755,6 +755,7 @@ def rematerialize_forward_and_backward(fw_trace: TraceCtx, bw_trace: TraceCtx) -


 def replace_uniform(trace: TraceCtx) -> TraceCtx:
+    return trace
     """For better rematerialization, replace the uniform operator with the stateless uniform_philox operator and manually update the RNG state."""
     start_time_ns = time.perf_counter_ns()
     from thunder.core.trace import VariableInterface
```
Patch to apply:
```python
diff --git a/thunder/core/transforms.py b/thunder/core/transforms.py
index 9deed5fe..5676eead 100644
--- a/thunder/core/transforms.py
+++ b/thunder/core/transforms.py
@@ -1460,6 +1460,49 @@ def _log_sigmoid_grad(

 register_grad("torch.nn.functional.logsigmoid", _log_sigmoid_grad)

+def _dropout_grad(a: TensorProxy, /, p: NumberLike = 0.5, training: bool = True, inplace: bool = False) -> TensorProxy:
+    #assert False, "I am here!"
+
+    if inplace:
+        raise NotImplementedError("Only inplace=False is currently supported in dropout")
+
+    if not training:
+        fwd = a
+        g = get_grad(fwd)
+        put_grad(a, g)
+        return fwd
+
+    utils.check(
+        p <= 1 and p >= 0,
+        lambda: f"Dropout probability has to be between 0 and 1, but got, {p}",
+    )
+
+    fwd: TensorProxy
+
+    if p == 1:
+        fwd = zeros_like(a)
+        put_grad(a, zeros_like(a))
+        return fwd
+    if p == 0:
+        fwd = a
+        g = get_grad(fwd)
+        put_grad(a, g)
+        return fwd
+
+    scale = 1 / (1 - p)
+    r = clang.uniform_like(a, 0.0, 1.0)
+    dropout_mask = r < (1 - p)
+
+    scaled_dropout_mask = scale * dropout_mask
[...]

PR #1659, opened 2025-01-18
Disable enable_saved_for_backward_recomputation path by default

IvanYashchuk:
Let's wait a bit before using the changes introduced in #1615 by default. The feature is available by passing `enable_saved_for_backward_recomputation=True` to thunder.jit.

Disabling auto-recomputation as suggested in https://github.com/Lightning-AI/lightning-thunder/issues/1658#issuecomment-2599598645.

### DIFF
--- a/thunder/core/transforms.py
+++ b/thunder/core/transforms.py
@@ -3160,8 +3160,6 @@ def forward_and_backward_from_trace(trace: Trace, torch_autograd=False) -> Forwa
     enable_saved_for_backward_recomputation: None | bool = get_compile_option(
         "enable_saved_for_backward_recomputation", "Enable save for backward tensors recomputation."
     )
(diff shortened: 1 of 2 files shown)


t-vi (changes requested):
No. We disable the auto-recomputation, which is in the setting of RECOMPUTE_IN_BACKWARD in trace_interpreter.
The enable_saved_for_backward_recomputation flag needs to go as soon as @riccardofelluga does not need it anymore.
Note that enable_saved_for_backward_recomputation is a noop when the option is not given and no tensor has that tag.

t-vi:
I would prefer to have #1661 instead. Does that work for you?

IvanYashchuk:
Yes, it works. I didn't know where the code for the pass to be used only for checkpointed operators lives. Thank you.

PR #1660, opened 2025-01-18
Add an option to be able to disable replace_uniform

IvanYashchuk:
Currently, Thunder decomposes Dropout into recomputation operations and rematerializes it in the nvFuser fusions in backward computations. 

For easier experimentation what would the performance look like without dropout recomputation an option to disable the pass is added in this PR.

It's one of the requests from https://github.com/Lightning-AI/lightning-thunder/issues/1658.

cc @t-vi


t-vi:
What could be an UX alternative here? I think we should find better ways than setting options for everything.

t-vi (approved):
I would really, really, really not want the options to be part of our user experience. Let us merge this now to unblock #1658 quickly, but if you actually need this, we need a better solution to activate.

PR #1667, opened 2025-01-21
increase tolerance on bf16 layernorm

t-vi:
the results seems to have changed in PyTorch nightly, so we get CI flakes, hopefully this avoids them.


t-vi:
Merging to fix CI.

PR #1669, opened 2025-01-21
remove experimental dev option for recomputation

t-vi:
> We don't want options, we just want sane defaults.

This served well for starting checkpointing but now tag-based checkpointing / recomputation seems to be the way.


thunder/core/transforms.py
```diff
@@ -3191,22 +3171,6 @@ def recompute_saved_for_backward(fwd_trace: Trace, bwd_trace: Trace) -> tuple[Tr
...
     old_saved_for_bwd = OrderedSet(variableify(j) for j in saved_for_bw)
 
-    remat_policy: None | Callable[[set[Variable]], set[Variable]] = get_compile_option(
-        "recomputation_policy",
-        "A callable that accepts a set of variables and returns a set of the variables that are allowed to be recomputed from the forward in the backward trace. The compile option `enable_saved_for_backward_recomputation` needs to be true for this policy to take effect.",
-    )
-
-    enable_saved_for_backward_recomputation: None | bool = get_compile_option(
-        "enable_saved_for_backward_recomputation", "Enable save for backward tensors recomputation."
-    )
```
riccardofelluga, line 3201: I thought that it would have been helpful to keep the switch to enable or disable the recomputation of the saved for backward tensors. What is the rationale behind this choice?

t-vi: If you don't have tennsors tagged for recomputation, you don't get recomputation.
If you have tensors tagged for recomputation, you get that.

The thought here is that you there should ideally be a single source of truth.
If you want to disable recomputation that you otherwise capture, add a transform that removes all recompute tags from the tensors.

Issue #1531, opened 2024-12-09
ExtractionOnlyPrologueTransform for dropping checks form the prologue


t-vi:
The prologue can take a considerable amount of time.

Currently, the prologue does two things:
1) extract values for the computation trace,
2) check shapes and other metadata.

In situations where the user controls the environment, it makes sense to let them take control and offer to skip 2.

It would be nice to have a ExtractionOnlyPrologueTransform . For bonus points, use the [`TraceSubstitutionProcessor`](https://github.com/Lightning-AI/lightning-thunder/blob/087637f0e443ae08824f23940f3e02dfcf8e02fc/thunder/core/trace_interpreter.py#L209).

PR #1564, opened 2024-12-17
Add a transform to remove param and buffer shape check from prologue

kshitij12345:
Related #1542 #1531 

Adds a Transform which removes `check_shape_and_metadata` from prologue for any proxy tagged with STATIC_MEMORY_LOCATION (which mainly includes module parameters and buffers currently).

Example (have attached the sample trace)
```python

import torch
from thunder.tests.litgpt_model import Config
from litgpt.model import GPT
import thunder
import thunder.transforms
import thunder.transforms.extraction_only_prologue_transform

model = GPT.from_name("llama1-like", n_layer=1)
x = torch.randint(model.max_seq_length, (2, 5))
cmodel = thunder.jit(model, transforms=[thunder.transforms.extraction_only_prologue_transform.ExtractionOnlyPrologueTransform()])
_ = cmodel(x)
prl_trc = thunder.last_prologue_traces(cmodel)[-1]

with open("prologue_trc.py", "w") as f:
    f.write(str(prl_trc))


```

[prologue_with_checks_trc.py.txt](https://github.com/user-attachments/files/18169022/prologue_with_checks_trc.py.txt)
[prologue_without_checks_trc.py.txt](https://github.com/user-attachments/files/18169023/prologue_without_checks_trc.py.txt)

Benchmark
```python
import torch
import torch.utils.benchmark
from thunder.tests.litgpt_model import Config
from litgpt.model import GPT
import thunder
import thunder.transforms
import thunder.transforms.extraction_only_prologue_transform

cfg = Config("Llama3-8b")
with torch.device("cuda"):
    model = GPT(cfg)
    x = torch.randint(cfg.vocab_size, (2, 5))

cmodel = thunder.jit(model, transforms=[thunder.transforms.extraction_only_prologue_transform.ExtractionOnlyPrologueTransform()])
# cmodel = thunder.jit(model)
_ = cmodel(x)

comp_and_input_fn = cmodel._lc_cd._get_computation_and_inputs
cd = cmodel._lc_cd
cs = cmodel._lc_cs
cache_entry, inps, pro_to_epi = comp_and_input_fn(x)
prologue_fn = cache_entry.prologue_fn

# Mock the cache_info (required by prologue)
cache_info = {}
cache_info["default_dtype"] = torch.float32
cache_info["default_device"] = torch.device("cpu")
cache_info["is_autocast_enabled"] = False
cache_info["no_grad_sync"] = False
cache_info["alias_tensor_indices"] = ''
[...]

PR #1671, opened 2025-01-21
add PrunePrologueChecks transform

t-vi:
Fixes: #1531

With apologies to @kshitij12345 , I just saw the issue in the open issues list and totally forgot about his prior work.

@kshitij12345 made a similar transform in #1564 , but this one has some advantages:
- also removes non-tensor checks (for attributes like "in_dimension" etc.),
- works for buffers, too,
- has an option to remove all checks.

### DIFF
--- a/thunder/transforms/prune_prologue_checks.py
+++ b/thunder/transforms/prune_prologue_checks.py
@@ -1,0 +1,56 @@
+import thunder
+from thunder.core import prims as prims
+
+
+class PrunePrologueChecks(thunder.core.transform_common.Transform):
+    """A Transform to prune Prologue checks
+
+    This transform removes prologue checks and can be applied when the user
+    controls the environment enough to ensure that these checks would always
+    succeed. By default, we remove all checks that use the module.
+    """
+
+    def __init__(self, prune_all_checks=False):
+        self.prune_all_checks = prune_all_checks
+
+    def transform_traces_pre_prologue(self, prologue_trace, compute_trace, epilogue_trace, **kwargs):
+        def is_check(bsym):
+            return bsym.sym in {
+                prims.check_tensor_shape_and_metadata,
+                prims.check_string_value,
+                prims.check_number_type_and_value,
+                prims.check_len,
+                prims.check_literal_like,
+            }
+
+        if not self.prune_all_checks:
+            bsyms_to_skip = set()
+            module_member_names = set()
+            seen_names = set()
+
+            for bsym in prologue_trace.bound_symbols:
+                if bsym.sym in {prims.unpack_trivial, prims.unpack_cache_info}:
+                    continue
+                elif all(i.name in module_member_names for i in bsym.flat_proxy_args):
+                    for o in bsym.flat_proxy_outs:
+                        module_member_names.add(o.name)
+                    if is_check(bsym):
+                        bsyms_to_skip.add(bsym)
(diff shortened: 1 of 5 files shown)


thunder/transforms/prune_prologue_checks.py
```diff
@@ -0,0 +1,56 @@
...
+                prims.check_string_value,
+                prims.check_number_type_and_value,
+                prims.check_len,
+                prims.check_literal_like,
+            }
+
+        if not self.prune_all_checks:
+            bsyms_to_skip = set()
+            module_member_names = set()
+            seen_names = set()
```
kshitij12345, line 29: unused variable

thunder/transforms/prune_prologue_checks.py
```diff
@@ -0,0 +1,56 @@
...
+
+            for bsym in prologue_trace.bound_symbols:
+                if bsym.sym in {prims.unpack_trivial, prims.unpack_cache_info}:
+                    continue
+                elif all(i.name in module_member_names for i in bsym.flat_proxy_args):
```
kshitij12345, line 34: I am a bit confused how this will work. `module_member_names` is empty at the beginning. So, this branch could be true if `bsym.flat_proxy_args = []` as `all([]) -> True`. I am not sure how else would it be True otherwise (most likely I am missing something).

thunder/transforms/prune_prologue_checks.py
```diff
@@ -0,0 +1,56 @@
...
+                    for o in bsym.flat_proxy_outs:
+                        module_member_names.add(o.name)
+                    if is_check(bsym):
+                        bsyms_to_skip.add(bsym)
+                else:
+                    print(bsym)
```
kshitij12345, line 40: Stray print

thunder/transforms/prune_prologue_checks.py
t-vi: Yes, and this is precisely the case for unpack_function_obj, but I should add a comment.

thunder/transforms/prune_prologue_checks.py
t-vi: Thank you

thunder/transforms/prune_prologue_checks.py
```diff
@@ -0,0 +1,56 @@
...
+            bsyms_to_skip = set()
+            module_member_names = set()
+
+            for bsym in prologue_trace.bound_symbols:
+                if bsym.sym in {prims.unpack_trivial, prims.unpack_cache_info}:
+                    # These don't have inputs but need to be skipped to not trigger false positives
+                    continue
+                elif all(i.name in module_member_names for i in bsym.flat_proxy_args):
+                    # This has the special case of no proxy inputs, which is the case for unpack_function_obj,
+                    # the root of module_member_names
```
kshitij12345, line 36: Do you think we should add an `assert bsym.sym == unpack_function_obj`?

kshitij12345 (approved):
LGTM, thanks @t-vi

thunder/transforms/prune_prologue_checks.py
t-vi: all not empty, but yes, fair.

lantiga (approved):

lantiga (approved):

PR #1674, opened 2025-01-22
Embedding support for nvfuser in inference

Priya2698:
This PR enables nvfuser to accept the embedding operator for the forward pass.
Since Thunder uses `prims.TAKE` instead of `prims.EMBEDDING` in certain cases, the nvfuser operator is also registered against `ltorch.embedding`. For backward pass, a separate operator would need to be registered against `ltorch.embedding_backward`.

### DIFF
--- a/thunder/executors/nvfuserex_impl.py
+++ b/thunder/executors/nvfuserex_impl.py
@@ -2595,15 +2595,16 @@ ex.register_supported(
     grad_transform=scaled_dot_product_flash_attention_grad,
 )
 
+
 def _embedding_check(
-  input: TensorProxy, 
-  weight: TensorProxy,
-  padding_idx: None | int,
-  max_norm: None | float,
-  norm_type: None | float,
-  scale_grad_by_freq: None | bool,
-  sparse: None | bool,
-  ) -> bool:
+    input: TensorProxy,
+    weight: TensorProxy,
+    padding_idx: None | int,
+    max_norm: None | float,
+    norm_type: None | float,
+    scale_grad_by_freq: None | bool,
+    sparse: None | bool,
+) -> bool:
     if nvfuser_version() < LooseVersion("0.2.25"):
         return False
     enable_embedding: None | bool = get_compile_option("nv_enable_embedding", "Enable nvFuser embedding.")
@@ -2616,7 +2617,7 @@ def _embedding_check(
 
 
 def embedding(
-    input: TensorProxy, 
+    input: TensorProxy,
     weight: TensorProxy,
     padding_idx: None | int = None,
     max_norm: None | float = None,
@@ -2633,6 +2634,7 @@ def embedding(
         nv_inp = getnv(inp, fd, lc_to_nv_map) if inp is not None else None
         nv_inputs.append(nv_inp)
     return fd.ops.embedding_fwd(*nv_inputs)
+
 
 register_supported(PrimIDs.EMBEDDING, embedding, _embedding_check)
 register_supported(ltorch.embedding, embedding, _embedding_check)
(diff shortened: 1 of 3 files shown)


thunder/tests/test_nvfuser.py
```diff
@@ -1157,3 +1158,34 @@ def fn(x):
 
     # Make sure there is a fusion symbol.
     assert any(bsym.sym.is_fusion for bsym in fwd_trace.bound_symbols)
+
+
+@instantiate(
+    dtypes=(thunder.float16,),
```
riccardofelluga, line 1164: What dtypes are supported by the embedding op?

Priya2698: Updated it to include both float16 and bfloat. There is no restriction on the dtype of weights. Inputs are indices and should be integers.

Priya2698: The CI is not using nvfuser version 0.2.25, so the embedding test is skipped. I am not sure how to update that.

t-vi: The CI uses the nvfuser_cu121_torch25 packages on pip for the stable PT and builds NVFuser from source for the unstable one. There have been slight hickups building PT from source in the last two days, but it should pick it up.

thunder/tests/test_nvfuser.py
```diff
@@ -1157,3 +1158,34 @@ def fn(x):
...
+
+
+@instantiate(
+    dtypes=(thunder.float16, thunder.bfloat16),
+    devicetypes=(devices.DeviceType.CUDA,),
+    executors=(nvFuserExecutor,),
+    decorators=(
+        pytest.mark.skipif(
+            nvfuser_version() is None or nvfuser_version() < LooseVersion("0.2.25"),
+            reason="Requires nvFuser version 0.2.25 or later",
```
mruberry, line 1170: Nice comment

thunder/tests/test_nvfuser.py
```diff
@@ -1157,3 +1158,34 @@ def fn(x):
...
+    executor,
+    device: str,
+    dtype: dtypes.dtype,
+):
+
+    def embedding_fn(inputs):
+        return torch.nn.functional.embedding(*inputs)
+
+    for sample in embedding_opinfo.sample_inputs(device, dtype):
+        compiled_func = thunder.jit(embedding_fn, executors_list=executor.executors_list(), nv_enable_embedding=True)
```
mruberry, line 1184: Nit: `compiled_func` could be created outside the loop since it isn't sample-dependent

thunder/tests/test_nvfuser.py
```diff
@@ -1157,3 +1158,34 @@ def fn(x):
...
+        out = compiled_func(sample.args)
+        expected_out = torch.nn.functional.embedding(*sample.args)
+        fwd_trace = thunder.last_traces(compiled_func)[-1]
```
mruberry, line 1187: `fwd_trace` and `fwd_fusion` aren't actually used by the test -- probably just some leftover debugging code?

thunder/executors/nvfuserex_impl.py
```diff
@@ -2593,3 +2594,47 @@ def scaled_dot_product_flash_attention_grad(
...
+    if not enable_embedding:
+        return False
+    # Verify input and weight are supported tensors.
+    if not are_supported_tensors(input, weight) or (weight.ndim != 2):
+        return False
+    return True
+
+
+def embedding(
+    input: TensorProxy,
```
mruberry, line 2620: Nice annotations

mruberry (approved):

PR #1677, opened 2025-01-22
don't create dataclass instances in dataclass_new meta

t-vi:
Dataclasses can do wild things in their `__post_init__`, so it is better to be careful there.

### DIFF
--- a/thunder/core/prims.py
+++ b/thunder/core/prims.py
@@ -1295,7 +1295,8 @@ pack_setitem = make_prim(
 
 
 def python_dataclass_new_meta(typ, **kwargs):
-    return AnyProxy(typ(**kwargs))
+    # we are cheating here, but instantiating a dataclass can be wild. typ(**kwargs)
+    return AnyProxy(None)
 
 
 def python_dataclass_new_printer(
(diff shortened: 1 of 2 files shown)


thunder/core/prims.py
```diff
@@ -1295,7 +1295,8 @@ def pack_setitem_impl(o: Any, key: Any, v: Any) -> None:
 
 
 def python_dataclass_new_meta(typ, **kwargs):
-    return AnyProxy(typ(**kwargs))
+    # we are cheating here, but instantiating a dataclass can be wild. typ(**kwargs)
+    return AnyProxy(None)
```
riccardofelluga, line 1299: What were we passing as `kwargs`? are they not needed anymore?

t-vi: The inputs to the dataclass. We want them (and record them in the symbol), but we don't use them right now.
The trouble is that HF has dataclasses that also want to be dicts and if they only get one argument.
An alternative could be not allow calling the symbol and requiring to .bind it instead, but I don't see the advantage right now.

lantiga (approved):

PR #1681, opened 2025-01-22
make no_sync behave like in pytorch

t-vi:
FSDP seems to not work, even though I think it should result in exactly the same calls as before...

(I also need to take out the debug prints after I figure out what's going on.)


t-vi:
The problem here is that the sync needs to happen after (or during?) the corresponding backward. I'll fix that.

Issue #1687, opened 2025-01-24
Memory leak from `connect_to_autograd`


riccardofelluga:
## 🐛 Bug

Initially found as leak when running HF Qwen2Attention [here](https://github.com/Lightning-AI/lightning-thunder/pull/1600#issuecomment-2609307220) but narrowed down by @kshitij12345 to a problem with the interaction between #1600 and #1581 in comment [here](https://github.com/Lightning-AI/lightning-thunder/pull/1600#issuecomment-2609848480)

```python

import torch
from thunder.executors.torch_autograd import connect_to_autograd

def foo():
    def backward(*args):
        return None

    x = torch.randn(1024, 1024, requires_grad=True, device="cuda")
    o = x.sin()

    connect_to_autograd(backward_fn=backward, flat_args=(x,), flat_output=(o,), saved_tensors=(o,), saved_other=(), return_none_instead_of_grads=True)

for i in range(5):
    foo()
    print(torch.cuda.max_memory_allocated())

```

which outputs:

```shell
8388608
16777216
25165824
33554432
41943040
```

### Environment

- transformers>=4.48.0

cc @lantiga

t-vi:
Thank you @kshitij12345 @riccardofelluga

PR #1688, opened 2025-01-24
Fix memory leak in saved tensors

t-vi:
Fixes: #1687

Basically we need to move the saved tensors to the function that outputs tensors that might also be saved.

All debugging credit belongs to @kshitij12345  @riccardofelluga .


lantiga (approved):
Great find, thank you

PR #1689, opened 2025-01-24
Remove prims.embedding and prims.embedding_backward

IvanYashchuk:
We don't need `prims.embedding` and `prims.embedding_backward` because we can simply use `thunder.torch.embedding` and `thunder.torch.embedding_backward` as our transformation targets. When we added these primitives we didn't have a good idea how multi-level symbols can be targeted.

With this PR `thunder.torch.embedding` is transformed directly to `torch.nn.functional.embedding` for execution. While before this PR `thunder.torch.embedding` is decomposed into `prims.embedding` which is then transformed into `torch.nn.functional.embedding` (and similarly for `embedding_backward`).

Testing:
```
pytest thunder/tests/test_grad.py -k "embedding" -vvv
pytest thunder/tests/test_ops.py -k "embedding" -vvv
```

### DIFF
--- a/thunder/core/prims.py
+++ b/thunder/core/prims.py
@@ -265,8 +265,6 @@ class PrimIDs(Enum):
     MATMUL = auto()
     # NN prims (Experimental!)
     CONVOLUTION = auto()
-    EMBEDDING = auto()
-    EMBEDDING_BACKWARD = auto()
     LINEAR = auto()
     PAD = auto()
     # Memory access methods
@@ -4077,39 +4075,6 @@ convolution = make_prim(
     "convolution",
     meta=convolution_meta,
 )
-
-
-def embedding_meta(
-    a: TensorProxy, /, weight, *, padding_idx=-1, max_norm=None, norm_type=2.0, scale_grad_by_freq=False, sparse=False
-) -> TensorProxy:
-    # TODO: canonicalize and validating padding idx with weight.shape[0]
-
-    if max_norm is not None:
-        raise NotImplementedError
-
-    utils.check(a.dtype == dtypes.int64, lambda: f"Expected a.dtype={a.dtype} to be int64")
-    utils.check(weight.ndim == 2, lambda: f"Expected weight (weight.shape={weight.shape} to be a matrix)")
-
-    shape = list(a.shape)
-    shape.append(weight.shape[1])
-
-    return TensorProxy(like=weight, shape=shape)
-
-
-embedding = make_prim(PrimIDs.EMBEDDING, "embedding", meta=embedding_meta)
-
-
-# TODO Update this so it's not a prim
-# TODO Add annotations
-# TODO Review requires_grad=False -- what about double backward?
-# TODO Once we have fusible index_put we can implement it using primitives
-# For now we just use the PyTorch implementation
-def embedding_backward_meta(grad, indices, num_weights, padding_idx, scale_grad_by_freq, sparse):
-    shape = (num_weights, grad.shape[-1])
-    return TensorProxy(shape=shape, device=grad.device, dtype=grad.dtype, requires_grad=False)
-
-
-embedding_backward = make_prim(PrimIDs.EMBEDDING_BACKWARD, "embedding_backward", meta=embedding_backward_meta)
 
 
 def copy__meta(
--- a/thunder/executors/nvfuserex_impl.py
+++ b/thunder/executors/nvfuserex_impl.py
@@ -2633,5 +2633,4 @@ def embedding(
     return fd.ops.embedding_fwd(*nv_inputs)
 
 
-register_supported(PrimIDs.EMBEDDING, embedding, _embedding_check)
 register_supported(ltorch.embedding, embedding, _embedding_check)
--- a/thunder/torch/__init__.py
+++ b/thunder/torch/__init__.py
@@ -4710,17 +4710,17 @@ def embedding(
 ) -> TensorLike:
     # TODO: add embedding_renorm_ so we can remove embedding prim
     # NOTE: padding_idx has impact on backward and is not supported by take
-    if max_norm is not None or padding_idx is not None:
+    if max_norm is not None:
+        raise NotImplementedError("max_norm argument is currently not support. Please create an issue detailing your use case.")
+
+    if padding_idx is not None:
         padding_idx = padding_idx if padding_idx is not None else -1
-        return prims.embedding(
-            a,
-            weight,
-            padding_idx=padding_idx,
-            max_norm=max_norm,
-            norm_type=norm_type,
-            scale_grad_by_freq=scale_grad_by_freq,
-            sparse=sparse,
-        )
+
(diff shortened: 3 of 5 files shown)


mruberry:
This is interesting.

Isn't this regressing support for calling `torch.embedding` with the `max_norm` or `padding_idx` arguments, though?

thunder/core/prims.py
```diff
@@ -4079,39 +4077,6 @@ def maybe_expand_seq(seq, ndim):
 )
 
 
-def embedding_meta(
-    a: TensorProxy, /, weight, *, padding_idx=-1, max_norm=None, norm_type=2.0, scale_grad_by_freq=False, sparse=False
-) -> TensorProxy:
-    # TODO: canonicalize and validating padding idx with weight.shape[0]
-
-    if max_norm is not None:
-        raise NotImplementedError
```
IvanYashchuk, line 4088: The `max_norm` argument was not supported. I don't know the reason for this. I think it should work and I'll update OpInfo samples to test it.

riccardofelluga:
A wait then we might want to change this too 

https://github.com/Lightning-AI/lightning-thunder/blob/52ee541073ddf6dce78b77ddb272621f712e3c6a/thunder/executors/nvfuserex_impl.py#L2636-L2637

thunder/executors/nvfuserex_impl.py
```diff
@@ -2633,5 +2633,4 @@ def embedding(
     return fd.ops.embedding_fwd(*nv_inputs)
 
 
-register_supported(PrimIDs.EMBEDDING, embedding, _embedding_check)
```
IvanYashchuk, line 2636: Re https://github.com/Lightning-AI/lightning-thunder/pull/1689#issuecomment-2613090315
Thank you, Riccardo, the requested change is part of the PR.

thunder/core/prims.py
IvanYashchuk: `max_norm` argument when active modifies the `weight` argument inplace. That's probably the reason this argument raised NotImplementedError. I'll keep the behavior as is in this PR.
Created an issue for tracking completion https://github.com/Lightning-AI/lightning-thunder/issues/1699.

thunder/torch/__init__.py
```diff
@@ -4710,18 +4710,20 @@ def embedding(
...
-        padding_idx = padding_idx if padding_idx is not None else -1
-        return prims.embedding(
-            a,
-            weight,
-            padding_idx=padding_idx,
-            max_norm=max_norm,
-            norm_type=norm_type,
-            scale_grad_by_freq=scale_grad_by_freq,
-            sparse=sparse,
+    if max_norm is not None:
```
mruberry, line 4713: Please also add a comment linking to the issue you filed

IvanYashchuk:
@mruberry, could you please take another look?

mruberry:
RMSNorm failure looks flaky:

```
2025-02-05T15:48:31.4628903Z [31mFAILED[0m thunder/tests/test_ops.py::[1mtest_core_vs_torch_consistency_rms_norm_nvfuser_cuda_thunder.dtypes.float16[0m - AssertionError: Tensor-likes are not close!
2025-02-05T15:48:31.4629274Z 
2025-02-05T15:48:31.4629647Z Mismatched elements: 1 / 1 (100.0%)
2025-02-05T15:48:31.4630080Z Greatest absolute difference: 0.203125 at index (0,) (up to 0.01 allowed)
2025-02-05T15:48:31.4630577Z Greatest relative difference: 0.0252685546875 at index (0,) (up to 0.01 allowed)
```

The absolute difference is quite large. fyi @kevinstephano -- I think you're looking at RMSNorm now?

thunder/torch/__init__.py
```diff
@@ -4710,18 +4710,21 @@ def embedding(
...
         )
 
+    if padding_idx is not None:
+        padding_idx = padding_idx if padding_idx is not None else -1
+
+        utils.check(a.dtype == dtypes.int64, lambda: f"Expected a.dtype={a.dtype} to be int64")
+        utils.check(weight.ndim == 2, lambda: f"Expected weight (weight.shape={weight.shape} to be a matrix)")
+        shape = list(a.shape)
+        shape.append(weight.shape[1])
+        return TensorProxy(like=weight, shape=shape)
```
mruberry, line 4726: Maybe want to add a comment for the direct construction of a tensorproxy here instead of calling an operator that would create one -- I'm curious, too

thunder/torch/__init__.py
```diff
@@ -4733,10 +4736,12 @@ def embedding(
...
 
 
+# TODO Once we have fusible index_put we can implement it using primitives
+# For now we just use the PyTorch implementation
 @torchsymbol(torch.ops.aten.embedding_backward)
 def embedding_backward(grad, indices, num_weights, padding_idx, scale_grad_by_freq, sparse):
-    result = prims.embedding_backward(grad, indices, num_weights, padding_idx, scale_grad_by_freq, sparse)
-    return result
+    shape = (num_weights, grad.shape[-1])
+    return TensorProxy(shape=shape, device=grad.device, dtype=grad.dtype)
```
mruberry, line 4744: Same question here about directly constructing and return a TensorProxy over calling some op that would create one

I guess I'm a little confused about operations in the torch namespace directly constructing TensorProxies, as that seems more like a primitives thing? Curious to hear your thoughts

thunder/torch/__init__.py
IvanYashchuk: `thunder.torch. embedding` is the operator we're calling to create the proxy.

thunder/torch/__init__.py
IvanYashchuk: We started Thunder thinking that the terminal level should be the `thunder.prims` namespace. It was before moving to the multi-level representation of PyTorch operators, there was just a jump from PyTorch straight to prims, I think. We also didn't have a direct lowering from `thunder.torch` to `torch` at the time. In similar cases, like this indirection is unnecessary, and instead of forming proxies in the meta function of `prims.embedding_backward` we can do that directly in `thunder.torch.embedding_backward`. The subsymbols of this operation would be empty and this is the indicator to treat it as a primitive both for execution and for transformations.

mruberry: My guess is that this approach will work today but will require revisiting in the future. If we start to use the primitive language context as a signal that we should do some special account for shape constraints, for example (fyi @jjsjann123)

jjsjann123: Effectively this is expanding the set of `terminal level symbols` to be `prims` + `symbols without prim decomposition`?

I see the benefit of eliminating one level of indirection.
i.e. effectively this PR just lifts (or lowers) these torch.nn.functional.symbols to be our `expanded prims`. I don't have a strong opinion against it.

> My guess is that this approach will work today but will require revisiting in the future.

I'm curious what's the concern you have @mruberry ? Are you just worried about this making it too easy to smuggle `effective prims` into the code base and it's hard to police that?

PR #1693, opened 2025-01-24
Backward transform dependency fix

jjsjann123:
This PR is to address grad transformation where we handle duplicate shape queries.

e.g. in a trace like
```
  (i0, i1, i3) = prims.shape(t0)  # query 1
  # ... some other operation use i0, i1, i3
  # ... unary op producing t1 from t0
  (i0, i1, i3) = prims.shape(t0)  # query 2
```

Two things patched:
1. `find_producer_symbols` needs to mark `prims.shape(t0)  # query 1` as producer of `i0, i1, i3`. i.e. we should mark the first encounter as the original order in the trace, otherwise, we'll break the data dependency. The existing implementation would mark the producer as the last one in the trace.

2. Grad transform function: `make_aug_forward_and_backward(bsym: BoundSymbol)` needs to run dce on the joint trace. Otherwise we'll have duplicate symbols like above, which triggers redefinition check.


mruberry:
@IvanYashchuk Would you like to review this?

thunder/core/prims.py
```diff
@@ -1818,6 +1818,7 @@ def _put_grad_meta(grad_for: Number | NumberProxy | TensorProxy, grad: Number |
     PrimIDs.PUT_GRAD,
     "put_grad",
     meta=_put_grad_meta,
+    tags=(OpTags.DONT_DCE,),
```
mruberry, line 1821: This might need a comment

thunder/core/utils.py
```diff
@@ -1128,11 +1128,21 @@ def find_producer_symbols(trace: TraceCtx, proxies: Sequence[Proxy], stop_proxie
...
             for arg in p.flat_args:
-                arg_name = arg.name if isinstance(arg, Proxy) else None
+                if not isinstance(arg, Proxy):
+                    continue
+                arg_name = arg.name
                 if arg_name not in map(lambda x: x.name, stop_proxies) and arg_name not in seen:
                     queue.append(arg)
                     seen.add(arg_name)
-    original_order = {bsym: i for i, bsym in enumerate(trace.bound_symbols)}
+    original_order = dict()
```
mruberry, line 1137: fyi: `= {}` is equivalent to `= dict()`

thunder/core/utils.py
```diff
@@ -1128,11 +1128,21 @@ def find_producer_symbols(trace: TraceCtx, proxies: Sequence[Proxy], stop_proxie
...
+    for i, bsym in enumerate(trace.bound_symbols):
+        # Don't overwrite the order if it's already encountered. This is necessary for
```
mruberry, line 1139: This seems like a good fix and comment, but the comment assumes the reader knows the previous code. I think we could add a little more context here, like "Creates a map from ..."

Like, what's the point of ordering, or why might there be duplicate bsyms? 

Also, would the description of what this function returns benefit from being updated to explain that the bsyms in the tuple are presented in order, and are the "first" bsyms that produce a proxy if there are multiple such symbols?

thunder/core/utils.py
jjsjann123: Out of curiosity, is `= {}` preferred over `dict()`?

mruberry: Either is fine

thunder/tests/test_transforms.py
```diff
@@ -807,3 +807,66 @@ def _count_shape_query(trace):
...
+    expected.sum().backward()
+
+    assert_close(actual, expected)
+    assert_close(a.grad, a_ref.grad)
+    assert_close(b.grad, b_ref.grad)
+    assert thunder.cache_misses(jfoo) == 1
+    assert thunder.cache_hits(jfoo) == 1
+
+
+def test_cache_symbolic_values_grad_unsqueeze():
```
jjsjann123, line 854: this test is coming from #1696

thunder/core/vjp_utils.py
```diff
@@ -60,6 +61,8 @@ def make_aug_forward_and_backward(bsym: BoundSymbol) -> tuple[Callable, Callable
         return cached_result
 
     joint_trace = thunder.trace(inline_trace=False, use_dce=False)(joint_forward_backward, *bsym.args, **bsym.kwargs)
+    # dce is necessary to remove duplicated shape queries, otherwise the trace might overwritten NumberProxy variables
+    joint_trace = dce(joint_trace)
```
IvanYashchuk, line 65: We have a `use_dce` flag that could be used here for some reason. Conveniently, it defaults to `True`, so there's no need even to spell it out here.
```suggestion
    # dce is necessary to remove duplicated shape queries, otherwise the trace might overwritten NumberProxy variables
    joint_trace = thunder.trace(inline_trace=False, use_dce=True)(joint_forward_backward, *bsym.args, **bsym.kwargs)
```

thunder/core/prims.py
```diff
@@ -1814,10 +1814,13 @@ def _put_grad_meta(grad_for: Number | NumberProxy | TensorProxy, grad: Number |
     return None
 
 
+# PUT_GRAD is a sink node with side effects that updates Tensor.grad. It needs
+# DONT_DCE tag to avoid removal in DCE pass.
```
IvanYashchuk, line 1818: Thanks for the explanation! This could be the reason `use_dce=False` was set in vjp_utils.py file

IvanYashchuk (approved):
Looks good to me!

thunder/core/vjp_utils.py
jjsjann123: Thanks a lot, somehow I totally missed that arg in trace. :bow:

t-vi:
We can have this now, but how do we expect this to work with transforms?

t-vi (approved):
Stamping to unblock @jjsjann123

jjsjann123:
> We can have this now, but how do we expect this to work with transforms?

I think dce is the proper solution.

My expectation is that, when a new trace is constructed (or if a mutation like subsymbols being flattened), running a dce to remove the duplicates seems like a reasonable~ish solution to clean up the graph.
I feel for other transforms, where SSA is not a requirement, we don't really have a problem with it.

In case I'm not reading your questions properly, feel free to elaborate on your concerns. :bow:

PR #1673, opened 2025-01-21
fixing dce pass to remove duplicate shape queries.

jjsjann123:
## What does this PR do?

dce pass doesn't remove duplicate `prims.shape` node.

```
  # /volume/ttt.py:9:       shape_0 = x.shape
  (i0, i1, i2, i3) = prims.shape(x)
  # /volume/ttt.py:10:      shape_1 = x.shape
  (i0, i1, i2, i3) = prims.shape(x)
```
becomes
```
  # /volume/ttt.py:9:       shape_0 = x.shape
  (i0, i1, i2, i3) = prims.shape(x)
  # /volume/ttt.py:10:      shape_1 = x.shape
  (_, _, _, _) = prims.shape(x)
```
This PR patches it to instead emit
```
  # /volume/ttt.py:9:       shape_0 = x.shape
  (i0, i1, i2, i3) = prims.shape(x)
```

Helps #1579 .

Note that I don't think this fixes the whole issue, but I think it's a more consistent behavior, since dce is expected to remove un-used operations.

PR #1696, opened 2025-01-25
patching shape query in TensorProxy to avoid prims

jjsjann123:
We should avoid calling `TensorProxy.shape` directly during trace transformations, but use `TensorProxy._shape` instead.

The difference is that `shape` would leave a `prims.shape` symbol in the trace, if under a tracing context. Which is the case during grad transform. 

Without this change, the trace will be corrupted with `prims.shape` calling on non-existing TensorProxy.

### DIFF
--- a/thunder/common.py
+++ b/thunder/common.py
@@ -392,8 +392,12 @@ def _unpack_inputs(fn, tracectx: TraceCtx, args, kwargs, *, rename_proxies: bool
 
 # TODO Update cacheable types
 def _make_subkey_for(x: Any) -> tuple[bool, None | tuple]:
-    if isinstance(x, (torch.Tensor, TensorProxy)):
+    if isinstance(x, torch.Tensor):
         return True, (type(x), x.shape, x.device, x.dtype, x.requires_grad)
+
+    if isinstance(x, TensorProxy):
+        # calling _shape instead shape to avoid leaving a prims.shape in the trace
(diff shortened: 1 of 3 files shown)


mruberry:
This is really interesting. fyi @t-vi and @IvanYashchuk 

As a note for the future, when we're transforming traces, maybe we shouldn't generally be operating within the trace context, to avoid accidentally putting operations into the trace. I guess if we did that then calling `.shape` on these tensors would throw an error that we're calling a tracing operation outside a trace context? If it doesn't, then maybe it should? Otherwise it would be easy for a developer to accidentally write `.shape` when they mean `._shape`. 

I also wonder if, in the future, we should make these queries more explicit. For example, instead of telling people to query `._shape` to avoid appearing in a trace (which seems totally fine) we could have properties like `.shape_notrace` or a utility library like with functions like `notrace_shape(x)` to acquire metadata without appearing in a trace.

I don't think we have to pursue either of these ideas in this PR, of course.

mruberry (approved):

jjsjann123:
err.. I should have made it clear that these three PRs are stacked. so they really should be merged in the order of #1673 #1693 #1696

IvanYashchuk:
> err.. I should have made it clear that these three PRs are stacked. so they really should be merged in the order of #1673 #1693 #1696

This happens 😅 I usually keep the stacked PRs in the draft mode to prevent merges

PR #1697, opened 2025-01-27
update _div_prim_impl

### LINKED ISSUE #1695: prims.div - Fails with torch_compile executor and incorrect torchex implementation
```python
import torch
import thunder

def fn(x, y):
    return thunder.prims.div(x, y) + x

jfn = thunder.jit(fn, executors=[torch_compile_ex,])
# jfn = thunder.jit(fn)  # Works
jfn(torch.ones(3, 3), torch.ones(3, 3))
print(thunder.last_traces(jfn))
```

<details>

<summary> Error </summary>

```python
torch._dynamo.exc.Unsupported: call_method UserDefinedObjectVariable(set) __contains__ [UserDefinedObjectVariable(floating)] {}

from user code:
   File "thunder.to_be_compiled_3", line 9, in to_be_compiled
    t2 = torch_div_prim_impl(x, y)  # t2: "cpu f32[3, 3]"
  File "/home/kkalambarkar/lightning-thunder/thunder/executors/torchex.py", line 931, in _div_prim_impl
    if dtypes.is_exact_dtype(to_dtype(a.dtype)) and dtypes.is_exact_dtype(to_dtype(a.dtype)):
  File "/home/kkalambarkar/lightning-thunder/thunder/core/dtypes.py", line 468, in is_integer_dtype
    return dtype in integer_dtypes

```

</details>

Found while running 

```python
import torch
from thunder.dev_utils.utils import _benchmark_fusion_region_with_nvfuser_and_torch_compile

model_id = "microsoft/Phi-3-mini-128k-instruct"
configuration = AutoConfig.from_pretrained(
    model_id,
    # Scaled down for testing
    vocab_size=16,
    pad_token_id=15,
    max_position_embeddings=32,
    num_hidden_layers=1,
)
configuration.hidden_size = configuration.num_attention_heads
with torch.device("cuda"):
    model = AutoModelForCausalLM.from_config(configuration).to(torch.bfloat16)

model = thunderfx(model, nv_store_fusion_inputs=True)

input_ids = torch.randint(0, configuration.vocab_size, (1, configuration.max_position_embeddings), device="cuda")
model(input_ids=input_ids, labels=input_ids)

trc = model.last_traces[0]

for bsym in trc.bound_symbols:
    if bsym.sym.is_fusion and "nvFusion" in bsym.sym.name:
        benchmark_comparison_data = _benchmark_fusion_region_with_nvfuser_and_torch_compile(bsym)

```

kshitij12345:
Fixes #1695

### DIFF
--- a/thunder/executors/torchex.py
+++ b/thunder/executors/torchex.py
@@ -928,7 +928,15 @@ def _add_transform(
 
 # Maps exact inputs to truncation division
 def _div_prim_impl(a: Number | torch.Tensor, b: Number | torch.Tensor) -> torch.Tensor:
-    if dtypes.is_exact_dtype(to_dtype(a.dtype)) and dtypes.is_exact_dtype(to_dtype(a.dtype)):
+    def is_exact_number_or_exact_tensor(x):
+        if isinstance(x, (int, bool)):
+            return True
+        # We use PyTorch's dtype attribute (instead of `dtypes.is_exact_dtype`) so that torch.compile on prims.div works.
+        elif isinstance(x, torch.Tensor) and (not x.dtype.is_complex and not x.dtype.is_floating_point):
+            return True
+        return False
+
+    if is_exact_number_or_exact_tensor(a) and is_exact_number_or_exact_tensor(b):
         return torch.div(a, b, rounding_mode="trunc")
 
     return torch.true_divide(a, b)
--- a/thunder/tests/opinfos.py
+++ b/thunder/tests/opinfos.py
@@ -2527,6 +2527,7 @@ def div_sample_generator(op, device, dtype, requires_grad, **kwargs):
             # numerator, denominator, rounding_mode
             yield SampleInput(make(shape_a), make(shape_b), rounding_mode=rounding_mode)
             yield SampleInput(make(shape_a), number(), rounding_mode=rounding_mode)
+            yield SampleInput(number(), make(shape_a), rounding_mode=rounding_mode)
 
 
 div_opinfo = OpInfo(
@@ -2547,6 +2548,7 @@ div_opinfo = OpInfo(
             pytest.mark.xfail,
             "test_core_vs_torch_consistency",
             executors=("nvfuser",),
+            dtypes=(datatypes.bool8, datatypes.bfloat16, datatypes.float16, datatypes.float32),
         ),
         DecorateInfo(pytest.mark.xfail, "test_vjp_correctness"),
     ),


thunder/executors/torchex.py
```diff
@@ -928,7 +928,15 @@ def _add_transform(
 
 # Maps exact inputs to truncation division
 def _div_prim_impl(a: Number | torch.Tensor, b: Number | torch.Tensor) -> torch.Tensor:
-    if dtypes.is_exact_dtype(to_dtype(a.dtype)) and dtypes.is_exact_dtype(to_dtype(a.dtype)):
+    def is_exact_number_or_exact_tensor(x):
+        if isinstance(x, (int, bool)):
+            return True
+        # We use PyTorch's dtype attribute (instead of `dtypes.is_exact_dtype`) so that torch.compile on prims.div works.
```
mruberry, line 934: Cool; in the future we might want to refactor this into a helper for other functions, too.

fyi @jjsjann123 because with this check may need to account for number proxies in the future, too

thunder/tests/opinfos.py
```diff
@@ -2547,6 +2548,7 @@ def div_sample_generator(op, device, dtype, requires_grad, **kwargs):
             pytest.mark.xfail,
             "test_core_vs_torch_consistency",
             executors=("nvfuser",),
+            dtypes=(datatypes.bool8, datatypes.bfloat16, datatypes.float16, datatypes.float32),
```
mruberry, line 2551: Thanks for adding this line. We should probably follow-up on enabling these tests! It's worrying that nvfuser and torch div appear to be inconsistent (as best our tests can tell). Maybe we have a singularity issue?

fyi @beverlylytle -- this might be interesting for you to take a look at

mruberry:
I don't think we test running div with the torch.compile executor in our CI, but as long as it works for you locally this change looks correct to me!

mruberry (approved):

thunder/tests/opinfos.py
beverlylytle: I took a look at these tests, and I don't see a singularity issue.  Instead I see two unrelated issues.  

First, Thunder produces a float64 result from float32 input whereas torch maintains the type.  I don't understand where the conversion is happening.  

Second, one of the sample inputs has a float (ie, not a tensor) as the denominator.  The float16 version of this fails with "Tensor-likes are not close".  If the denominator is replaced by a tensor with empty shape and dtype float16, it passes.

mruberry: Interesting! Would you please file issues for both of those?

The producing a float64 result from a float32 input is concerning, and could cause considerable addition memory usage and slowdown. It'd be great to figure out what's going on there! Maybe we're mapping to the wrong nvfuser operation?

PR #1694, opened 2025-01-24
[WIP] Test out thunder.jit w/ NeMo models.

tfogal:
## What does this PR do?

Demonstrates how to test thunder.jit with NeMo models.

## PR review

This isn't meant to be merged or reviewed. If we actually want to do this, someone would want to *extend* the testing to test *both* `thunder.jit` *and* `thunder.dynamo.thunderfx`, not remove the thunderfx testing (as done here).

## Did you have fun?

With this group, always :-)

PR #1698, opened 2025-01-27
bump transformers version

riccardofelluga:
try to bump transformers version to lates as the lib changed a lot from 4.46.x

I considered using the >= symbol but maybe better to stay safe and keep 48.1 for now, let me know in case of other opinions

fyi cc. @kshitij12345 @IvanYashchuk 

cc @borda

### DIFF
--- a/requirements/test.txt
+++ b/requirements/test.txt
@@ -18,7 +18,7 @@ pandas # thunder/benchmarks/test_benchmark_litgpt.py
 xlsxwriter # thunder/benchmarks/test_benchmark_litgpt.py
 jsonargparse # thunder/benchmarks/benchmark_litgpt.py
 bitsandbytes==0.42.0  # fixed version!
-transformers==4.46.3 # for test_networks.py
+transformers==4.48.1 # for test_networks.py
 
 # Installs JAX on Linux and MacOS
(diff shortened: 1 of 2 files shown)


riccardofelluga:
Looks like to be the same issue as in #1694

t-vi:
Seems that this is funny to us (because it creates a new Sequential)
```
File [/home/zeus/miniconda3/envs/cloudspace/lib/python3.10/site-packages/transformers/models/llama/modeling_llama.py:572](https://jupyterlab-01jfey9fhyngvg2ds2ck9gc3jd.studio.lightning.ai/home/zeus/miniconda3/envs/cloudspace/lib/python3.10/site-packages/transformers/models/llama/modeling_llama.py#line=571), in LlamaModel.forward()
    569 all_hidden_states = () if output_hidden_states else None
    570 all_self_attns = () if output_attentions else None
--> 572 for decoder_layer in self.layers[: self.config.num_hidden_layers]:
    573     if output_hidden_states:
    574         all_hidden_states += (hidden_states,)
```

t-vi:
This fails on the HF with dynamic cache, which is advertised as not compile compatible (and suffers from the max position being an int, so without symbolic caching, it would not be useable).
We can mark the `test_hf_llama` as skip and bump the version.

t-vi (approved):

PR #1701, opened 2025-01-27
Reduce `test_thunderfx_mistral_nemo_small` model size

riccardofelluga:
Reducing the models size to speedup CI and make it work in environments with constrained memory size.

cc. @tfogal since you added this, I think this change should be fine since with this test we are verifying functionality more than memory footprint. Let me know what you think. 

With this change the peak memory of the test is ~2.6GB instead of ~14GB

### DIFF
--- a/thunder/tests/test_networks.py
+++ b/thunder/tests/test_networks.py
@@ -392,14 +392,14 @@ def test_thunderfx_mistral_nemo_small():
     config = transformers.models.mistral.configuration_mistral.MistralConfig(
         num_hidden_layers=1,
         torch_dtype=torch.bfloat16,
-        max_position_embeddings=1024,
+        max_position_embeddings=64,
         architectures=["MistralForCausalLM"],
-        hidden_size=5120,
+        hidden_size=1024,
         rms_norm_eps=1e-05,
         rope_theta=1000000.0,
         sliding_window=None,
         vocab_size=131072,
-        head_dim=128,
+        head_dim=32,
         _name_or_path=model_id,
     )
     model = transformers.AutoModelForCausalLM.from_config(config, trust_remote_code=False)


thunder/tests/test_networks.py
```diff
@@ -392,14 +392,14 @@ def test_thunderfx_mistral_nemo_small():
...
+        max_position_embeddings=64,
         architectures=["MistralForCausalLM"],
-        hidden_size=5120,
+        hidden_size=1024,
         rms_norm_eps=1e-05,
         rope_theta=1000000.0,
         sliding_window=None,
         vocab_size=131072,
-        head_dim=128,
+        head_dim=32,
```
IvanYashchuk, line 402: Is the cuDNN SDPA kernel still used with this change?

thunder/tests/test_networks.py
```diff
@@ -392,14 +392,14 @@ def test_thunderfx_mistral_nemo_small():
...
         torch_dtype=torch.bfloat16,
-        max_position_embeddings=1024,
+        max_position_embeddings=64,
         architectures=["MistralForCausalLM"],
-        hidden_size=5120,
+        hidden_size=1024,
         rms_norm_eps=1e-05,
         rope_theta=1000000.0,
         sliding_window=None,
         vocab_size=131072,
```
IvanYashchuk, line 401: If you'd like to reduce the memory usage even further you may want to scale down vocab_size.

tfogal (approved):
Ivan's got a good point that we should double check that a fast kernel SDPA is still in use, but otherwise LGTM. Indeed this is just to test functionality, it's not meant as a benchmark.

thunder/tests/test_networks.py
riccardofelluga: It does not show up in the trace after the change, but also it is not showing up in the trace even without this change. Looking at the checker with the original model sizes, I get `cudnn._compiled_module.cudnnGraphNotSupportedError: No valid engine configs for Matmul_MUL_ADD_Reduction_SUB_EXP_Reduction_LOG_ADD_DIV_Matmul_` so it seems that even tho the input shapes pass all the tests, when it comes to create the cudnn graph, the backend refuses.

tfogal: In that case I think we should merge as-is. It is working in the actual nemo bench repository so it might just be that we had already reduced the sizes here such that it no longer applies.

but ping @vedaanta not high priority but you might consider adjusting the checker at some point

IvanYashchuk:
Certainly!

PR #1704, opened 2025-01-27
[DO NOT REVIEW][SMOKE TEST] Skip prims shape in grad transform

jjsjann123:


thunder/core/transform_common.py
```diff
@@ -123,7 +123,7 @@ def keep_or_swap(p):
         if not isinstance(p, NumberProxyInterface):
             return p
         if p.name in seen:
-            return p.value  # don't make it a duplicate
+            return None  # don't make it a duplicate
```
jjsjann123, line 126: note for myself, I needed this to avoid an error in rematerialization pass. I'll work on a separate PR with a repro when I get that.

jjsjann123: errrrr... this one doesn't seem to be working any more... I'm seeing the assert in rematerialization.py again.

jjsjann123:
```
import thunder

def foo(a, b):
    return a + b

import torch

a = torch.randn(1, 32, 232, 232)
a.requires_grad_()
b = torch.randn(1, 1, 232, 232)
#b.requires_grad_()

out = jfoo(a, b)
```

took me a while to get a repro here.
I think the issue is coming from saving for backward not properly identifying which gradient path isn't required. So after dce kicks in, the saved_for_backward is inconsistent... now I feel that I just missed a dce somewhere.

jjsjann123: trying my luck with #1725

jjsjann123:
closing inactive PRs.

PR #1707, opened 2025-01-28
Adds FXGraphReport, FXReport

### LINKED ISSUE #1700: [Reporting Tool] Modular report classes for saving non-Thunder-specific repro scripts
For the next PR, I think it'd be useful to create an FXReport object. What I'm thinking is:

- Create a new function called `fx_report` (not Thunder-specific)
- Return a FXReport object from `fx_report`
- The FXReport object has a sequence of FXGraphReport objects, one per fx graph in the original program
- Each FXGraphReport has a method to write a reproducer script
  - The method should be able to easily select whether the reproducer uses PyTorch eager, torch.compile, or thunderfx (with some supported options)
  - For total control, the method should allow writing an arbitrary executor string and allow for extending what's imported with arbitrary import strings, in case someone wants to use a very special version of thunderfx or torch.compile or PyTorch eager
- Each FXGraphReport should allow its FXGraph's inputs to be queried (they can just be FakeTensors for now)

So someone could do something like:

```
report: FXReport = fx_report(foo)

fxgr: FXGraphReport
for fxgr in report.graph_reports:
  print(fxgr.inputs)
  fxgr.write_eager_reproducer(filename0)
  fxgr.write_torchcompile_reproducer(filename1)
  fxgr.write_thunder_reproducer(filename2)
  fxgr.write_reproducer(filename3, executor_string="my_crazy_executor", import_strings=("import my_crazy_module", "from my_crazy_module import my_crazy_executor"))
```

And that would generate four reproducers for each of the FX graphs in the program, each one using the different requested executor.

Writing the reproducers directly like this may not be super interesting in isolation, but once that's working we can add more of the data that `thunderfx_report` makes available, and that will become really interesting really fast. It would also be great if the existing reporting functionality, like `thunderfx_report` could be updated to use `FXGraphReport` and `FXReport` objects internally for consistency. Using modular Python objects for the report data and analysis should make it easier for us to develop and for others to extend.

What are your thoughts, @kiya00?

_Originally posted by @mruberry in https://github.com/Lightning-AI/lightning-thunder/issues/1636#issuecomment-2610898585_

kiya00:
<details>

</details>

## What does this PR do?

Fixes #1700.

This PR introduces `FXGraphReport`, `FXReport`, `fx_report`, example uses:
```py
import tempfile
import torch
from thunder.dynamo.report import fx_report
def model(x):
    return x * 2
report = fx_report(model, torch.randn((2, 2), requires_grad=True, device="cuda"), compile_options={"dynamic": False})
print(len(report.fx_graph_reports))
with tempfile.TemporaryDirectory() as tmpdir:
    for graph_report in report.fx_graph_reports:
        graph_report.write_eager_repro(tmpdir)
        graph_report.write_thunder_repro(tmpdir, use_benchmark=True)
        graph_report.write_inductor_repro(tmpdir)
        # Executes using the user-defined executor.
        my_executor = "partial(thunder.jit, transforms=[NvtxProfileTransform()], executors=[nvfuser_executor])"
        my_imports = [
            "import thunder",
            "from thunder.dev_utils.nvtx_profile_transform import NvtxProfileTransform",
            "from thunder import nvfuser_executor",
            "from functools import partial",
        ]
        graph_report.write_repro(
            tmpdir, f"{graph_report.graph_name}_mythunder_repro.py", executor_str=my_executor, import_str=my_imports
        )
        graph_report.write_benchmark_repro(
            tmpdir,
            f"{graph_report.graph_name}_mythunder_benchmark.py",
            executor_name_str=["mythunder"],
            executor_str=[my_executor],
            import_str=my_imports,
        )
```
The repro script can be called with `python graph0_repro_thunder.py --compute_type forward+backward --check_consistency True`
The benchmark script can be called with `pytest graph0_benchmark_thunder.py --benchmark-timer=torch.utils.benchmark.utils.timer.timer --benchmark-warmup=on --benchmark-group-by=param:compute_type -k backward`
A rough outline of the script is shown in `lightning-thunder/thunder/dynamo/repro_script_template.py`

`ThunderCompiler.save_reproducer_to_folder` and `thunderfx_report` is updated to use the above APIs. Example uses:
```py
from thunder.dynamo.report import thunderfx_report
x = torch.randn(4, 4, device="cuda", requires_grad=True)
thunderfx_report(model, x, folder_path=tmp_path)
```
The report will be printed to the screen, the generated scripts will be saved to the input `folder_path`
(Note that the scripts used in `thunderfx_report` are saved by `compiled_fn._backend.save_reproducer_to_folder`)
```
The input callable can be successfully executed by ThunderFX.

Verifying consistency between Thunder and Torch eager ...
[graph0_thunder_0_repro] Consistency check succeeded

Analyzing performance through benchmarking, this might take a moment...
graph0_thunder_0_benchmark:
Running 8 items in this shard
........                                                                                                                [100%]

------------------------------------------------------------------------------ benchmark 'compute_type=ComputeType.TRAINING_BACKWARD': 4 tests -------------------------------------------------------------------------------
Name (time in us)                                          Min                 Max                Mean            StdDev              Median               IQR            Outliers  OPS (Kops/s)            Rounds  Iterations
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_graph0_thunder_0[backward-eager]                  31.4999 (1.0)       62.4359 (1.0)       32.8305 (1.0)      3.0221 (1.0)       32.0636 (1.0)      0.4669 (1.0)       130;177       30.4595 (1.0)        2485          13
test_graph0_thunder_0[backward-torch_inductor]         63.3261 (2.01)      94.4822 (1.51)      66.0851 (2.01)     3.9774 (1.32)      64.9091 (2.02)     1.1842 (2.54)       85;173       15.1320 (0.50)       1591          10
test_graph0_thunder_0[backward-thunder]               127.2262 (4.04)     173.3258 (2.78)     130.7810 (3.98)     4.5011 (1.49)     129.4591 (4.04)     1.7901 (3.83)        61;99        7.6464 (0.25)        783          10
test_graph0_thunder_0[backward-thunder_cudagraph]     131.2922 (4.17)     187.8136 (3.01)     135.0093 (4.11)     4.9375 (1.63)     133.6817 (4.17)     1.8639 (3.99)        46;60        7.4069 (0.24)        762          10
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

------------------------------------------------------------------------------- benchmark 'compute_type=ComputeType.TRAINING_FORWARD': 4 tests ------------------------------------------------------------------------------
Name (time in us)                                         Min                 Max                Mean            StdDev              Median               IQR            Outliers  OPS (Kops/s)            Rounds  Iterations
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_graph0_thunder_0[forward-eager]                   4.6208 (1.0)       10.1274 (1.0)        4.9058 (1.0)      0.5871 (1.0)        4.7395 (1.0)      0.1296 (1.0)       145;152      203.8405 (1.0)        2092         100
test_graph0_thunder_0[forward-torch_inductor]         16.5303 (3.58)      26.3296 (2.60)      17.3497 (3.54)     1.5808 (2.69)      16.9633 (3.58)     0.2652 (2.05)      126;217       57.6379 (0.28)       2540          24
test_graph0_thunder_0[forward-thunder]               120.1964 (26.01)    145.9336 (14.41)    124.4323 (25.36)    4.2887 (7.30)     122.9047 (25.93)    3.4189 (26.39)       65;51        8.0365 (0.04)        834          10
test_graph0_thunder_0[forward-thunder_cudagraph]     120.2525 (26.02)    154.6226 (15.27)    124.9182 (25.46)    4.5630 (7.77)     123.7054 (26.10)    1.8510 (14.29)       53;71        8.0052 (0.04)        829          10
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

Legend:
  Outliers: 1 Standard Deviation from Mean; 1.5 IQR (InterQuartile Range) from 1st Quartile and 3rd Quartile.
  OPS: Operations Per Second, computed as 1 / Mean
8 passed, 56928 warnings in 52.48s


Max allocated CUDA memory usage:
graph0_thunder_0[forward-thunder]: 9.765625e-07 GB
graph0_thunder_0[forward-torch_inductor]: 9.765625e-07 GB
graph0_thunder_0[forward-eager]: 9.765625e-07 GB
graph0_thunder_0[forward-thunder_cudagraph]: 9.765625e-07 GB


graph0_thunder_0[backward-thunder]: 2.9296875e-06 GB
graph0_thunder_0[backward-torch_inductor]: 2.9296875e-06 GB
graph0_thunder_0[backward-eager]: 2.9296875e-06 GB
graph0_thunder_0[backward-thunder_cudagraph]: 2.9296875e-06 GB
```

## PR review

## Did you have fun?


kiya00:
Hi @mruberry , I added some interfaces, mainly the base class `FXGraphReportBase` and the subclasses `FXGraphBenchmarkReport` and `FXGraphReport` that override the method `write_repro(folder: str | PathLike, custom_executor_name: str, custom_import_str: str,  custom_executor_str: str, save_input_tensor: bool, inputs: Sequence[torch.Tensor | ExampleInputMetaData] = None,)` to generate repro/benchmark script. An example of its usage can be found in the test case, what do you think about the interfaces? would you mind review this interface part first and then I'll update the current `ThunderCompiler.save_reproducer_to_folder` to use these new interfaces.

<details>
<summary>the generated repro script `graph0_mythunder.py`:</summary>

```
from math import inf
from math import nan
NoneType = type(None)
import torch
from torch import device
import torch.fx._pytree as fx_pytree
import torch.utils._pytree as pytree
import thunder
from thunder.dev_utils.nvtx_profile_transform import NvtxProfileTransform
from thunder import nvfuser_executor
from thunder.transforms.cudagraph import CUDAGraphTransform


import argparse

parser = argparse.ArgumentParser(description="Script for executing an FX graph with specified configurations.")

parser.add_argument(
    "--check_consistency",
    type=bool,
    default=False,
    help="Whether to check consistency (default: False)"
)
parser.add_argument(
    "--compute_type",
    type=str,
    choices=["forward", "forward+backward"],
    default="forward",
    help="Type of computation to perform (forward, forward+backward)"
)
args = parser.parse_args()
compute_type = args.compute_type
check_acc = args.check_consistency

def test_graph0():
    class DynamoModule(torch.nn.Module):
      def forward(self, L_x_ : torch.Tensor):
          l_x_ = L_x_
          y = l_x_.sin();  l_x_ = None
          return (y,)

    inputs = [
        torch.testing.make_tensor((4, 4), dtype=torch.float32,  device='cuda:0', requires_grad=True, low=None, high=None,).as_strided((4, 4), (4, 1)),

    ]

    model = DynamoModule()
    from thunder.dynamo.report import run_repro1

    compiled_model = thunder.jit(model, transforms=[NvtxProfileTransform(),CUDAGraphTransform()], executors=[nvfuser_executor])
    result = run_repro1(compiled_model, compute_type, *inputs)
    if check_acc:
        eager_result = run_repro1(model, compute_type, *inputs)
        for (compute_t, eager_v), (_, cur_v) in zip(eager_result.items(), result.items()):
            torch.testing.assert_close(eager_v, cur_v, msg=lambda e : f'{compute_t}: {e}')


test_graph0()

```

</details>

<details>
<summary>the generated benchmark script `graph0_mythunder.py`</summary>

```
# NOTE: This script requires `pytest-benchmark==4.0.0` to be installed.
# To execute the script, run `pytest graph0.py --benchmark-timer=torch.utils.benchmark.utils.timer.timer --benchmark-warmup=on --benchmark-group-by=param:compute_type`
# To check the peak allocated CUDA memory, use --benchmark-json=json_file_name and look at the "max_allocated_memory_MB" field in the json file
# To run tests for a specific compute_type, use the pytest `-k` option.
# For example:
#   - `-k "forward"` will run only the forward pass.
#
# Available options:
#   - compute_type: "forward", "backward"

import pytest
from thunder.benchmarks.targets import parametrize_compute_type_only_training, benchmark_for_compute_type
from math import inf
from math import nan
NoneType = type(None)
import torch
from torch import device
import torch.fx._pytree as fx_pytree
import torch.utils._pytree as pytree
import thunder
from thunder.dev_utils.nvtx_profile_transform import NvtxProfileTransform
from thunder import nvfuser_executor
from thunder.transforms.cudagraph import CUDAGraphTransform


@parametrize_compute_type_only_training
def test_graph0(benchmark, compute_type):
    class DynamoModule(torch.nn.Module):
      def forward(self, L_x_ : torch.Tensor):
          l_x_ = L_x_
          y = l_x_.sin();  l_x_ = None
          return (y,)

    inputs = [
        torch.testing.make_tensor((4, 4), dtype=torch.float32,  device='cuda:0', requires_grad=True, low=None, high=None,).as_strided((4, 4), (4, 1)),

    ]

    model = DynamoModule()
    compiled_model = thunder.jit(model, transforms=[NvtxProfileTransform(),CUDAGraphTransform()], executors=[nvfuser_executor])
    benchmark_for_compute_type(compute_type, benchmark, compiled_model, inputs, {})
```

</details>

And one thing we like about using pytest benchmark is its ability to generate a performance comparison table across multiple executors. So I'm also thinking about extending `FXGraphBenchmarkReport.write_repro(...)` to accept a list of executors so that all the executors can be run in a single benchmark script to give the tabular like the previous PR showed

thunder/dynamo/report.py
```diff
@@ -166,3 +198,235 @@ def print_sorted_memory_info(compute_t: str):
 
             print_sorted_memory_info("forward")
             print_sorted_memory_info("backward")
+
+
+class FXGraphReportBase:
+    def __init__(self, graph: torch.fx.GraphModule, graph_name: str):
```
mruberry, line 204: Cool!

Please also add a comment describing the class like

```
class FXGraphReportBase:
  """
  Wraps an FX Graph module with additional metadata to facilitate reproductions and benchmarks.
  """
```

thunder/dynamo/report.py
```diff
@@ -166,3 +198,235 @@ def print_sorted_memory_info(compute_t: str):
...
+        thunder_import_str = ["import thunder"]
+        self.write_repro(folder, "thunder", thunder_import_str, thunder_compile_str)
+
+    def write_inductor_repro(self, folder):
+        inductor_compile_str = "compiled_model=torch.compile(model)"
+        inductor_import_str = ["import torch"]
+        self.write_repro(folder, "torchcompile", inductor_import_str, inductor_compile_str)
+
+    @abstractmethod
+    def write_repro(
```
mruberry, line 227: Cool -- can the more specific "write...repro" functions call this more general one?

thunder/dynamo/report.py
```diff
@@ -166,3 +198,235 @@ def print_sorted_memory_info(compute_t: str):
...
+
+    def write_eager_repro(self, folder):
+        self.write_repro(folder, "eager", custom_executor_str="compiled_model=model")
+
+    def write_thunder_repro(self, folder):
+        thunder_compile_str = "compiled_model=thunder.jit(model)"
+        thunder_import_str = ["import thunder"]
+        self.write_repro(folder, "thunder", thunder_import_str, thunder_compile_str)
+
+    def write_inductor_repro(self, folder):
```
mruberry, line 221: Does the written reproducer run fwd + bwd or only fwd?

In general it seems tricky to know what kind of bwd to run if an fxgraph produces multiple tensors. What does the current script do in this case? Maybe we could create an extensibility point to specify the bwd behavior to test?

thunder/dynamo/report.py
```diff
@@ -166,3 +198,235 @@ def print_sorted_memory_info(compute_t: str):
...
+    @abstractmethod
+    def write_repro(
+        self,
+        folder: str | PathLike,
+        custom_executor_name: str,
+        custom_import_str: str,
+        custom_executor_str: str,
+        save_input_tensor: bool,
+        inputs: Sequence[torch.Tensor | ExampleInputMetaData],
+    ): ...
```
mruberry, line 235: This will definitely need some documentation with an example or two ;)

thunder/dynamo/report.py
```diff
@@ -166,3 +198,235 @@ def print_sorted_memory_info(compute_t: str):
...
+        self.write_repro(folder, "torchcompile", inductor_import_str, inductor_compile_str)
+
+    @abstractmethod
+    def write_repro(
+        self,
+        folder: str | PathLike,
+        custom_executor_name: str,
+        custom_import_str: str,
+        custom_executor_str: str,
+        save_input_tensor: bool,
```
mruberry, line 233: Maybe `serialize_tensor_inputs`?

thunder/dynamo/report.py
```diff
@@ -166,3 +198,235 @@ def print_sorted_memory_info(compute_t: str):
...
+        inputs: Sequence[torch.Tensor | ExampleInputMetaData],
+    ): ...
+
+
+class FXGraphBenchmarkReport(FXGraphReportBase):
+    def write_repro(
```
mruberry, line 239: Comment string explaining this class, too

thunder/dynamo/report.py
```diff
@@ -166,3 +198,235 @@ def print_sorted_memory_info(compute_t: str):
...
+Environment information get from `torch.utils.collect_env.get_pretty_env_info()`:
+{torch_env}
+
+Versions of Thunder related libraries:
+{thunder_pkgs}"""
+'''
+        del torch_env, thunder_pkgs
+
+        code_str = ""
+        code_str += f"""# NOTE: This script requires `pytest-benchmark==4.0.0` to be installed.
```
mruberry, line 268: This is totally fine, but this might be more readable as a string template. See https://docs.python.org/3/library/string.html#string-formatting. Then you could write:

```
code_template = (
"""
This is my template string. Replace like {this}.
"""
)
code_template.format(this=4)
# This is my template string. Replace like 4. 
```

The template could also be defined outside the class so it doesn't interrupt the function's logic

thunder/dynamo/report.py
```diff
@@ -166,3 +198,235 @@ def print_sorted_memory_info(compute_t: str):
...
+        else:
+            func_str += f"""benchmark_for_compute_type(compute_type, benchmark, compiled_model, inputs, {{}})"""
+
+        with open(folder / f"{self.graph_name}_{custom_executor_name}.py", "w") as f:
+            print(comment_str, file=f)
+            print(code_str, file=f)
+            print(_addindent(func_str, 4), file=f)
+
+
+class FXGraphReport(FXGraphReportBase):
```
mruberry, line 317: I wouldn't make these subclasses of `FXGraphReportBase`. Because they don't have additional data, I think all three classes can just be `FXGraphReport` and then `write_repro` can become `write_benchmark` or `write_repro` and its arguments can change

thunder/dynamo/report.py
```diff
@@ -166,3 +198,235 @@ def print_sorted_memory_info(compute_t: str):
...
+
+        with open(folder / f"{self.graph_name}_{custom_executor_name}.py", "w") as f:
+            print(comment_str, file=f)
+            print(code_str, file=f)
+            print(_addindent(func_str, 4), file=f)
+            print(f"\ntest_{self.graph_name}()", file=f)
+
+
+class FXReport:
+    def __init__(self, graph_report_cls: type[FXGraphReportBase], graphs: list[torch.fx.GraphModule]):
```
mruberry, line 413: Comment string for this class

mruberry: Following the above thinking about merging the `FXGraphReport` classes, I think this can just accept a list of fx graphs, and not have to take a class argument

thunder/dynamo/report.py
```diff
@@ -166,3 +198,235 @@ def print_sorted_memory_info(compute_t: str):
...
+        self.fx_graph_reports: list[FXGraphReportBase] = [
+            graph_report_cls(g, f"graph{idx}") for idx, g in enumerate(graphs)
+        ]
```
mruberry, line 416: In the future, `FXReport` could have methods to facilitate the creation of reports over all its graphs

mruberry:
Great start! I made some comments inline for you review. Once those have been resolved let's get this merged, and then we can focus on the next extension.

In the next PR, I think what we'll want to do next is make the timing data programmatically available, so that the timing data can be queried directly to produce interesting reports from Python. That could be a good opportunity to review how we're timing, too.

thunder/dynamo/report.py
kiya00: Currently, in the repro I use https://github.com/Lightning-AI/lightning-thunder/blob/4f4d63249569b82e087689a38bc1eec0743eb1f8/thunder/dynamo/report.py#L19-L35 for the backward calculation and checks consistency;
In the benchmarking I use the https://github.com/Lightning-AI/lightning-thunder/blob/4f4d63249569b82e087689a38bc1eec0743eb1f8/thunder/benchmarks/targets.py#L84-L92 to benchmark the time used for fwd/bwd calculation
>In the next PR, I think what we'll want to do next is make the timing data programmatically available, so that the timing data can be queried directly to produce interesting reports from Python. 

For now the timer used is `--benchmark-timer=torch.utils.benchmark.utils.timer.timer` and can be replace by other timer as we want. The timing data can be accessed by using the benchmark fixture directly or save it in a json file and process the json file (thunderfx report accesses the timing data this way)

thunder/dynamo/report.py
kiya00: yes, they call the more general one, and when people want to pass in their own compiled function and import strs, it's also the general one is used

mruberry:
Updates sound great; let me know when this is ready to be reviewed!

kiya00:
Hi @mruberry @IvanYashchuk it's ready to review, thank you in advance!

thunder/dynamo/report.py
```diff
@@ -145,24 +140,324 @@ def thunderfx_report(
...
+        executor_str: list[str],
+        import_str: list[str] = None,
+        serialize_inputs: bool = False,
+        inputs: Sequence[torch.Tensor | ExampleInputMetaData] = None,
+        **kwargs,
+    ) -> None:
+        """
+        Generates a benchmark reproduction script for a given FX graph module using various executors and writes it to the specified file.
+
+        Args:
```
mruberry, line 268: Nice comment

mruberry (approved):
Cool!

fyi @t-vi it looks like there are some CI issues that will prevent this from being auto-merged

@kiya00 This is great! Now we need to target timing data.

There are two immediate next options I can think of:

One option is to make thunder segmentation and fusion information programmatically available. This could be done by having a function like `analyze_with_thunder` that takes an FXGraphReport and creates a ThunderFXGraphReport, which would be a subclass of FXGraphReport. The ThunderFXGraphReport would have additional ThunderSegmentReport classes, and maybe ThunderFusionReport classes, so that the segments and fusions could be iterated over. Then the ability to write reproducers from the segments alone and from the fusions alone could be added to those classes (reusing the existing functionality as much as possible). That would allow for the creation of files that reproduce or benchmark segments and fusions. Each nvFusion could then be compared to executing with torch.compile instead, for example.

I think the second option is to focus on making the timing data programmatically available at the FX level. Basically running the benchmark and returning its data as a Python object. In the future we should let callers specify their own timers, but a first PR can just pick a timer to use. 

Whichever is done first, the second one can probably be done next. The timings don't need to be perfect at first, but it will be exciting to work on the proper timing of fx graphs, segments, and fusions in the near future. @kshitij12345 did some work attempting to compare nvfuser and torch.compile fusions, but the nvFuser team and mixology teams may have some other ideas for timing, too.

Once we have programmatically available timings for fx graphs, segments, and fusions, then I think it's time to revisit running these reports as part of the reports that Mixology produces.

How does that sound, @kiya00?

mruberry (approved):
Cool!

fyi @t-vi it looks like there are some CI issues that will prevent this from being auto-merged

@kiya00 This is great! Now we need to target timing data.

There are two immediate next options I can think of:

One option is to make thunder segmentation and fusion information programmatically available. This could be done by having a function like `analyze_with_thunder` that takes an FXGraphReport and creates a ThunderFXGraphReport, which would be a subclass of FXGraphReport. The ThunderFXGraphReport would have additional ThunderSegmentReport classes, and maybe ThunderFusionReport classes, so that the segments and fusions could be iterated over. Then the ability to write reproducers from the segments alone and from the fusions alone could be added to those classes (reusing the existing functionality as much as possible). That would allow for the creation of files that reproduce or benchmark segments and fusions. Each nvFusion could then be compared to executing with torch.compile instead, for example.

I think the second option is to focus on making the timing data programmatically available at the FX level. Basically running the benchmark and returning its data as a Python object. In the future we should let callers specify their own timers, but a first PR can just pick a timer to use. 

Whichever is done first, the second one can probably be done next. The timings don't need to be perfect at first, but it will be exciting to work on the proper timing of fx graphs, segments, and fusions in the near future. @kshitij12345 did some work attempting to compare nvfuser and torch.compile fusions, but the nvFuser team and mixology teams may have some other ideas for timing, too.

Once we have programmatically available timings for fx graphs, segments, and fusions, then I think it's time to revisit running these reports as part of the reports that Mixology produces.

How does that sound, @kiya00?

thunder/dynamo/report.py
```diff
@@ -145,24 +140,324 @@ def thunderfx_report(
...
+            executor_str=executor_str,
+            graph_name=self.graph_name,
+            extra_comment_str=extra_comment_str,
+        )
+
+        with open(folder / file_name, "w") as f:
+            print(code_str, file=f)
+
+
+class FXReport:
```
IvanYashchuk, line 388: I suggest adding a useful method for interactive exploration, the `__repr__` method, with the number of FXGraphReports this class stores. Something like "<FXGraphReport with NUMBER FXGraphReports accessible via .fx_graph_reports>"

thunder/dynamo/report.py
```diff
@@ -145,24 +140,324 @@ def thunderfx_report(
...
+            executor_names=executor_names_str,
+            compute_type_decorator=compute_type_decorator,
+            extra_comment_str=extra_comment_str,
+        )
+        with open(folder / file_name, "w") as f:
+            print(code_str, file=f)
+
+    def write_repro(
+        self,
+        folder: str | PathLike,
```
IvanYashchuk, line 331: The example from the PR description suggests to use `TemporaryDirectory` but I get an error when I tried the following:
```py
tmpdir = tempfile.TemporaryDirectory()
graph_report.write_eager_repro(tmpdir)
```
```py
TypeError: expected str, bytes or os.PathLike object, not TemporaryDirectory
```
Using `tmpdir.name` instead works. Why is there no error when TemporaryDirectory is used as a context manager?

thunder/dynamo/report.py
```diff
@@ -145,24 +140,324 @@ def thunderfx_report(
...
+                    tmpdir,
+                    f"{graph_report.graph_name}_mythunder_benchmark.py",
+                    executor_name_str=["mythunder"],
+                    executor_str=[my_executor],
+                    import_str=my_imports,
+                )
+    """
+    graphs = []
+
+    def helper_backend(gm, example_inputs):
```
IvanYashchuk, line 453: Do `example_inputs` remain unused because all needed metadata is available on the GraphModule object itself?

thunder/dynamo/report.py
kiya00: I didn't noticed this, I try it and find the context manager convert it to str
```
>>> with tempfile.TemporaryDirectory() as tmpdir:
...     print(type(tmpdir))
...
<class 'str'>
```
The example from the PR is taken from the docstr, TemporaryDirectory is only used to delete the created folder. and we actually want the input folder to be str, bytes or os.PathLike object (that pathlib.Path can handle)

thunder/dynamo/report.py
kiya00: yes, the input metadata is inferred from the placeholder node in the FX Graph so the example inputs is not necessary

thunder/dynamo/report.py
IvanYashchuk: Thank you for checking! It's interesting that when used as a context manager the object is converted to a string. No need to support the `TemporaryDirectory` directly.

PR #1712, opened 2025-01-29
Grad transform patch

jjsjann123:
This PR refactors grad transform to add a `_grad_skip_ids`, which allows us to skip symbols that does not participate in grad transformation.

The main motivation is to ensure that `prims.shape` does not break the saved for backward in the transform.

### DIFF
--- a/thunder/core/transforms.py
+++ b/thunder/core/transforms.py
@@ -619,13 +619,22 @@ def clear_grads(module: torch.nn.Module) -> None:
         b.grad = None
 
 
+# Grad functions for prims
+from thunder.core.prims import PrimIDs as pids, get_grad, put_grad
+
+
 _grad_fn_map: dict[Any, Callable] = {}
 
+_grad_skip_ids: set[PrimIDs] = set()
 
-def register_grad(sym_or_id: Symbol | Any, gradfn: Callable) -> None:
+
+def register_grad(sym_or_id: Symbol | Any, gradfn: Callable, skip: bool = False) -> None:
     id: Any = sym_or_id
     if isinstance(sym_or_id, Symbol):
         id = sym_or_id.id
+
+    if skip:
+        _grad_skip_ids.insert(id)
 
     # The gradfn are expected to be written in terms of torch functions by
     # default even if the original forward function could be written in terms of
@@ -635,10 +644,6 @@ def register_grad(sym_or_id: Symbol | Any, gradfn: Callable) -> None:
     # appropriate language context that will take precedence over the default
     # torch language context.
     _grad_fn_map[id] = langctx(Languages.TORCH)(gradfn)
-
-
-# Grad functions for prims
-from thunder.core.prims import PrimIDs as pids, get_grad, put_grad
 
 
 # A generalization of prims.put_grad to pytrees
@@ -658,13 +663,13 @@ def put_grads(a, g):
 #
 
 # NOTE prims.unpack_empty_dict creates no grad associations
-register_grad(pids.UNPACK_EMPTY_DICT, prims.unpack_empty_dict)
+register_grad(pids.UNPACK_EMPTY_DICT, prims.unpack_empty_dict, True)
 
 # NOTE prims.unpack_key creates no grad associations
-register_grad(pids.UNPACK_KEY, prims.unpack_key)
+register_grad(pids.UNPACK_KEY, prims.unpack_key, True)
 
 # NOTE prims.unpack_sequence creates no grad associations
-register_grad(pids.UNPACK_SEQUENCE, prims.unpack_sequence)
+register_grad(pids.UNPACK_SEQUENCE, prims.unpack_sequence, True)
 
 
 #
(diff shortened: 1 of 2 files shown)


thunder/core/transforms.py
```diff
@@ -1401,10 +1406,12 @@ def _maximum_grad(a: TensorProxy, b: TensorProxy, /):
...
 
 # This operation creates no grad associations
-register_grad(pids.ARGMAX, prims.argmax)
+register_grad(pids.ARGMAX, prims.argmax, True)
 
 # This operation creates no grad associations
-register_grad(pids.SHAPE, prims.shape)
+# NOTE we need to skip pids.SHAPE in grad transform, because `prims.shape` creates multiple outputs, which
+# breaks the assumption the in the pass that each symbol has a single output
+register_grad(pids.SHAPE, prims.shape, True)
```
jjsjann123, line 1414: I only needed this change for `prims.shape`, but I think it could/should apply to any operation that doesn't require gradient.

thunder/tests/test_transforms.py
```diff
@@ -855,3 +853,24 @@ def foo(x):
...
+
+
+def test_cache_symbolic_values_grad_shape():
+    def foo(a):
+        b = torch.arange(0, 12, 1).reshape(1, 1, 3, 4)
+        a = a.unsqueeze(0)
+        a = a.unsqueeze(1)
+        return a + b
+
+    jfoo = thunder.jit(foo, cache="symbolic values")
```
jjsjann123, line 865: for `a = a.unsqueeze(0)`. The grad transforms function `deconstruct_forward_env_for_backward` runs on the trace

```
@torch.no_grad()
@no_autocast
def unsqueeze(a, dim):
  # a: "cpu f32[1, [IntegerProxy name=i0, value=3, static=CONSTRAINT.CONSTRAINABLE], [IntegerProxy name=i1, value=4, static=CONSTRAINT.CONSTRAINABLE]]"
  (_, i0, i1) = prims.shape(a)
  (_, _, _) = prims.shape(a)
  (_, _, _) = prims.shape(a)
  t8 = prims.broadcast_in_dim(a, [1, 1, i0, i1], [0, 2, 3])  # t8: "cpu f32[1, 1, [IntegerProxy name=i0, value=3, static=CONSTRAINT.CONSTRAINABLE], [IntegerProxy name=i1, value=4, static=CONSTRAINT.CONSTRAINABLE]]"
  return t8
```

The transform handles `saved for backwards` with the assumption that each symbol produces only a single output. https://github.com/Lightning-AI/lightning-thunder/blob/4f4d63249569b82e087689a38bc1eec0743eb1f8/thunder/core/transforms.py#L2313

which is not true for `prims.shape`.

Given that `prims.shape` doesn't participate in grad at all, I decided to skip it, instead of refactoring the transform to handle multiple outputs.

jjsjann123:
hmm. seems like we are hitting a real issue.

```
FAILED thunder/tests/test_networks.py::test_hf_kvcache - thunder.core.interpreter.InterpreterError: Encountered exception RuntimeError: dictionary changed size during iteration while tracing LlamaForCausalLM(
```

jjsjann123:
not sure if it's caused by this PR, since we did clear CI before the merge. I'll double check that.

thunder/core/transforms.py
```diff
@@ -2293,6 +2300,8 @@ def iter_bound_symbols(bound_symbols):
     for symbol in bound_symbols:
         if symbol.sym.id in trace_interpreter_skip_list:
```
IvanYashchuk, line 2301: We have one skip list here, `trace_interpreter_skip_list`, instead of adding a new one the existing list should be reused.

IvanYashchuk (commented):
The current pull request proposes to skip certain operations that do nothing in backward completely. However, we don't want to skip the forward computation because the result may have proxies that could be consumed later. An alternative would be to extend the existing code without adding more complexity to `register_grad`. There's `is_constant_for_vjp(bsym: BoundSymbol) -> bool` function
https://github.com/Lightning-AI/lightning-thunder/blob/40f79727c419d3acf83cd75b51fa867810e23a46/thunder/core/transforms.py#L2505
that could be modified instead (if output contains no differentiable proxies then it's considered as constant computation with no backward computation). I think adding an additional condition to that function, `not any(isinstance(arg, (FloatProxy, TensorProxy)) for arg in symbol.flat_outs)`, should be enough to enable the tests from this pull request.

thunder/core/transforms.py
jjsjann123: Thanks. That's really helpful.

sounds like `is_constant_for_vjp` is actually what I wanted.  i.e. it skips the backward but keeps forward as is?

I don't think `trace_interpreter_skip_list` is the right list though. It's used in trace_interpreter.py as well, which is outside the context of grad. But since you pointed out `is_constant_for_vjp`, I think `trace_interpreter_skip_list` is the right list to add them. I'll refactor this.

jjsjann123:
> The current pull request proposes to skip certain operations that do nothing in backward completely. However, we don't want to skip the forward computation because the result may have proxies that could be consumed later. 

Yeah that sounds about right.
I'm surprised that CI didn't caught any issue with that. :cry:

jjsjann123:
Should close this in favor of #1834

PR #1719, opened 2025-01-30
Add benchmark isolation option

riccardofelluga:
To reduce noise when running multiple benchmarks one after the other, I added the option to run the benchmarks in sub-processes by using the flag `--isolate-benchmarks` when running benchmarks from the `targets.py` suite. 

More information on the usage can be found in the documentation.

cc @borda @lantiga @crcrpar


t-vi (approved):
Thank you @riccardofelluga

PR #1721, opened 2025-01-30
TE: update for TE release_v2.0 support

kshitij12345:
Fixes https://github.com/Lightning-AI/lightning-thunder/pull/1737

Update to support https://github.com/NVIDIA/TransformerEngine/tree/release_v2.0

Tested with existing tests on RTX6000, H100 and B200.

NOTE - This update drops support for TE 1.x releases.


thunder/tests/distributed/test_fsdp.py
```diff
@@ -841,19 +851,20 @@ def _test_fsdp_transformer_engine_bucketing(input_data):
         torch.cuda.set_device(rank)
 
         # data
-        batch_size = 2
-        max_seq_len = 32
-        vocab_size = 32
```
kshitij12345, line 846: Before this TE [PR](https://github.com/NVIDIA/TransformerEngine/pull/1437), this was the minimum requirement for shape.

thunder/tests/distributed/test_ddp.py
```diff
@@ -448,42 +456,41 @@ def is_same_across_ranks(t):
...
-                is_same_across_ranks(thunder_fp8_meta["scaling_bwd"].scale)
-                is_same_across_ranks(thunder_fp8_meta["scaling_bwd"].scale_inv)
-                is_same_across_ranks(thunder_fp8_meta["scaling_bwd"].amax_history)
-            except Exception as e:
-                # Return exceptions only for rank==0
-                if rank == 0:
-                    comparison_exceptions.append(e)
+    if not isinstance(
+        get_default_fp8_recipe(), MXFP8BlockScaling
+    ):  # MXFP8BlockScaling recipe doesn't have state like scale, amax_history.
```
IvanYashchuk, line 461: Where is the block scaling state stored? On tensor objects?

thunder/tests/distributed/test_ddp.py
```diff
@@ -366,7 +371,10 @@ def _test_ddp_transformer_engine(input_data):
     torch.cuda.set_device(rank)
 
     dim = 256
-    n_iter = 10
+    # Running more iterations leads to `nan` for both eager and thunder
+    # with BlockScaling.
+    # Potentially because we are training on dummy data and task
+    n_iter = 5
```
IvanYashchuk, line 377: The main goal of scaling is to avoid overflows. Is the behavior expected? Could there be a bug in the TransformerEngine implementation?

IvanYashchuk (approved):
Great! I only have a few clarifying questions.

Note to other reviewers: this PR is not tested in CI (https://github.com/Lightning-AI/lightning-thunder/issues/196).

thunder/executors/transformer_engineex.py
```diff
@@ -197,14 +194,18 @@ def forward(self, inp, weight, bias, is_first_microbatch: bool | None = None, is
         # so under grad mode we enable grad for input tensors
         # Ref: https://github.com/NVIDIA/TransformerEngine/blob/b957aa475bcbcf22405381d18bd7fefe4fb6b171/transformer_engine/pytorch/module/linear.py#L264
         grad_ctx = enable_grad(*tensor_inputs) if is_grad_enabled else nullcontext()
-        with grad_ctx, self.prepare_forward(inp, is_first_microbatch) as inp:
+        with grad_ctx, self.prepare_forward(inp) as inp:
```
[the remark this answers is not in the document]
kshitij12345: > Is is_first_microbatch removed from prepare_forward?
Yes it has been removed from `prepare_forward`

https://github.com/NVIDIA/TransformerEngine/blob/884248f5788ad1d5561a55878cd772866efc46a8/transformer_engine/pytorch/module/base.py#L793-L799

> Do you know if there is any special handling of the first microbatch somewhere else in the code?

It is still present in the codebase, but we never set it in trace during execution (from the time transformer_engine_executor was added). Will remove it to avoid confusion.

thunder/executors/transformer_engineex.py
```diff
@@ -411,7 +375,10 @@ def _te_functional_linear_backward_meta(
 LINEAR_CALLS_COUNTER = 0
 
 if TE_AVAILABLE:
-    _DEFAULT_RECIPE = recipe.DelayedScaling()
+    # Recipe is chosen based on hardware platform
+    # For H100 or lower, it returns DelayedScaling recipe.
+    # For B200, it returns MXFP8BlockScaling recipe.
```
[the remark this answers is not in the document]
kshitij12345: Yes, essentially 
```python
def get_device_compute_capability() -> Tuple[int, int]:
    """CUDA compute capability of current GPU"""
    props = torch.cuda.get_device_properties(torch.cuda.current_device())
    return (props.major, props.minor)
```

and 
```python
def get_default_fp8_recipe() -> Recipe:
    """FP8 recipe with default args."""
    if get_device_compute_capability() >= (10, 0):  # blackwell and above
        return MXFP8BlockScaling()
    return DelayedScaling()
```


https://github.com/NVIDIA/TransformerEngine/blob/884248f5788ad1d5561a55878cd772866efc46a8/transformer_engine/pytorch/utils.py#L41-L44

https://github.com/NVIDIA/TransformerEngine/blob/884248f5788ad1d5561a55878cd772866efc46a8/transformer_engine/pytorch/fp8.py#L46-L50

thunder/tests/distributed/test_ddp.py
kshitij12345: I will need to check this separately. Ideally, BlockScaling should be more stable, but since we train a dummy task with dummy data, I haven't investigated where and how the weights blow up.

thunder/tests/distributed/test_ddp.py
kshitij12345: For BlockScaling recipe, the scales are stored on the quantized tensor objects. Also, worth noting that unlike DelayedScaling, there is no need to preserve state across the iteration as scaling factors are computed dynamically.

https://github.com/NVIDIA/TransformerEngine/blob/884248f5788ad1d5561a55878cd772866efc46a8/transformer_engine/pytorch/tensor/_internal/mxfp8_tensor_base.py#L46-L64

https://github.com/NVIDIA/TransformerEngine/blob/884248f5788ad1d5561a55878cd772866efc46a8/transformer_engine/pytorch/tensor/mxfp8_tensor.py#L129-L153

thunder/tests/distributed/test_fsdp.py
kshitij12345: Right, we should update this based on the recipe. Thanks!

kshitij12345: Reference for MXFP8 checks - https://github.com/NVIDIA/TransformerEngine/blob/884248f5788ad1d5561a55878cd772866efc46a8/transformer_engine/pytorch/csrc/extensions/quantizer.cpp#L177-L180

t-vi (approved):
Awesome, thank you @kshitij12345 @IvanYashchuk

PR #1725, opened 2025-01-31
fixing `TensorProxy.__init__(like=a)` to accept `a.requires_grad`

jjsjann123:
Previously we were discarding `requires_grad` from `like`. Since the default for `requires_grad=False` precede the properties from `like`.

This breaks `requires_grad` propagation in prims. Since many prim.meta functions use `TensorProxy(like=a)` to produce output(s) without explicitly passing `requires_grad` arg. e.g. https://github.com/Lightning-AI/lightning-thunder/blob/183a08811afc0fa0984085aea0dc5db925374514/thunder/core/prims.py#L3922


IvanYashchuk:
I plan to propose removing all propagation of requires_grad because it's not tested and not used anywhere (https://github.com/Lightning-AI/lightning-thunder/pull/1599).

jjsjann123:
sounds like we have a different plan for how grad is handled. closing this from [the comment](https://github.com/Lightning-AI/lightning-thunder/issues/1733#issuecomment-2639996480) and #1599.

Issue #1724, opened 2025-01-30
dividing a float16 tensor by a python float is inaccurate with nvfuser


beverlylytle:
The sample case of `test_core_vs_torch_consistency_div_nvfuser_cuda_thunder.dtypes.float16` which uses a python float as the denominator fails with numerical inaccuracy.  If the same sample is run with a tensor with empty shape, the test passes. Specifically, if `.item()` is deleted from [this line](https://github.com/Lightning-AI/lightning-thunder/blob/880419df25024b52fe8674fd2ef305348a056a62/thunder/tests/opinfos.py#L74), the test passes.

mruberry:
@beverlylytle, thank you for filing this issue. I know you're not suggesting otherwise, but if anyone else is reading this, it's OK to have a sampleinput that tests fp16 tensor / pyfloat, because that's a real supported input to div. We should figure out what's going on and see if we can fix it (or maybe just widen the test's tolerances)

PR #1734, opened 2025-02-01
fixing dtype promotion in where

### LINKED ISSUE #1723: div with nvfuser returns incorrect dtype
`test_core_vs_torch_consistency_div_nvfuser_cuda_thunder.dtypes.float32`, which tests the elementwise division of tensors of dtype float32 with nvfuser, fails with the torch result of dtype float32 while the Thunder result is float64.  

Thunder should also return float32.

cc @tfogal

jjsjann123:
Fixes #1723 

nvfuser type promotion rule for `where` is different from thunder. We could accidentally create output tensor in double as noted in https://github.com/Lightning-AI/lightning-thunder/issues/1723#issuecomment-2628768859.


thunder/executors/nvfuserex_impl.py
```diff
@@ -1953,7 +1953,11 @@ def where(
     nva = getnv(a, fd, lc_to_nv_map)
     nvb = getnv(b, fd, lc_to_nv_map)
 
-    return fd.ops.where(nvpred, nva, nvb)
+    # Determines result dtype
+    numbertype, tensordtype = utils.check_same_dtype(a, b)
+    dtype = tensordtype if tensordtype is not None else numbertype
```
jjsjann123, line 1958: technically I think we only needed the cast when both inputs are scalars.

thunder/tests/opinfos.py
```diff
@@ -2548,7 +2548,7 @@ def div_sample_generator(op, device, dtype, requires_grad, **kwargs):
             pytest.mark.xfail,
             "test_core_vs_torch_consistency",
             executors=("nvfuser",),
-            dtypes=(datatypes.bool8, datatypes.bfloat16, datatypes.float16, datatypes.float32),
+            dtypes=(datatypes.bool8,),
```
beverlylytle, line 2551: I'm not sure that we can delete the float16 case from this directive.  It's a bit flaky, failing 3/50 times with this change on my machine.   There's already [#1724](https://github.com/Lightning-AI/lightning-thunder/issues/1724) to track this. 

Would you mind also removing the TODO comment above?

beverlylytle (approved):
Nice! thanks!

kshitij12345 (changes requested):
IIUC, Current fix works because there is another `where` later with `tensor` inputs which correctly casts the output later.

I think we should also handle the following case (which still fails).

- Fusion with only `prims.where(cond, 1., 0.)` symbol
```python
import thunder
import thunder.examine
import torch

def fn(c, x, y):
    return torch.where(c, x, y)

x = 1.
y = 0.

print(fn(torch.rand(3, 3, device='cuda') > 0.5, x, y).dtype)  # torch.float

jfn = thunder.jit(fn, nv_store_fusion_inputs=True)
print(jfn(torch.rand(3, 3, device='cuda') > 0.5, x, y).dtype)  # torch.double

# trc = thunder.last_traces(jfn)[-1]

# print(trc)
# fusion_syms = thunder.examine.get_fusion_symbols(trc)

# bsym = fusion_syms[0]

# repro = thunder.examine.get_nvfuser_repro(trc, bsym.sym.name)
# print(repro)

```

thunder/executors/nvfuserex_impl.py
mruberry: This seems like a fine workaround, but should this eventually be fixed in nvfuser itself? Maybe we should get an issue to track that

mruberry:
> IIUC, Current fix works because there is another `where` later with `tensor` inputs which correctly casts the output later.
> 
> I think we should also handle the following case (which still fails).
> 
> * Fusion with only `prims.where(cond, 1., 0.)` symbol
> 
> ```python
> import thunder
> import thunder.examine
> import torch
> 
> def fn(c, x, y):
>     return torch.where(c, x, y)
> 
> x = 1.
> y = 0.
> 
> print(fn(torch.rand(3, 3, device='cuda') > 0.5, x, y).dtype)  # torch.float
> 
> jfn = thunder.jit(fn, nv_store_fusion_inputs=True)
> print(jfn(torch.rand(3, 3, device='cuda') > 0.5, x, y).dtype)  # torch.double
> 
> # trc = thunder.last_traces(jfn)[-1]
> 
> # print(trc)
> # fusion_syms = thunder.examine.get_fusion_symbols(trc)
> 
> # bsym = fusion_syms[0]
> 
> # repro = thunder.examine.get_nvfuser_repro(trc, bsym.sym.name)
> # print(repro)
> ```

Interesting! I'm surprised this case still fails with this fix. Maybe we need more tests cases for nvfuser and where, too

thunder/executors/nvfuserex_impl.py
jjsjann123: Just to clarify, the root cause is that nvfuser doesn't have `float` scalar input, but rather only allows `double`.

So not really a bug but rather a feature request. I'll open an issue to track that.

mruberry: haha -- however nvfuser wants to track the issue is fine

jjsjann123:
Thx for the repro.

I see what the issue is now. nvfuser executor maps dtypes differently for number types. so I just need another condition to convert weak to strong type when output is a tensor proxy.

jjsjann123:
@kshitij12345 verified the fix on your repro. Added tests in opinfo as well.

Thanks a lot for pointing this out. :bow:

mruberry:
@jjsjann123 Looks like the CI issues are related

thunder/executors/nvfuserex_impl.py
```diff
@@ -1953,7 +1953,16 @@ def where(
...
+
+    # NOTE: for scalar inputs, dtype mapping is different. e.g. float -> double. We convert dtypes to strong
+    # type if the output is supposed to be a tensor proxy
```
mruberry, line 1961: A link the nvfuser issue would be nice

thunder/tests/opinfos.py
```diff
@@ -2548,7 +2547,7 @@ def div_sample_generator(op, device, dtype, requires_grad, **kwargs):
             pytest.mark.xfail,
             "test_core_vs_torch_consistency",
             executors=("nvfuser",),
-            dtypes=(datatypes.bool8, datatypes.bfloat16, datatypes.float16, datatypes.float32),
+            dtypes=(datatypes.bool8, datatypes.float16),
```
mruberry, line 2550: Why does bfloat16 and bool still fail?

thunder/tests/opinfos.py
```diff
@@ -2733,6 +2732,15 @@ def where_error_generator(op, device, dtype=torch.float32, **kwargs):
     )
     yield (SampleInput(make(2, 1, 2)), NotImplementedError, err_msg)
 
+    # generate scalar inputs
+    dtypes = [float, int, bool, complex]
+
+    for dtype in dtypes:
```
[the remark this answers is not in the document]
jjsjann123: oops. good catch...

thunder/executors/nvfuserex_impl.py
jjsjann123: https://github.com/NVIDIA/Fuser/issues/3816

thunder/executors/nvfuserex_impl.py
jjsjann123: linked nvfuser issue already.

kshitij12345 (approved):
LGTM, thank you @jjsjann123!

jjsjann123:
code is good for review again. cc'ing @mruberry

jjsjann123:
> @jjsjann123 Looks like the CI issues are related

CI issue was because we try to run grad on the newly added scalar inputs, which doesn't have grad function.

mruberry (approved):
Cool! Nice comments

thunder/executors/nvfuserex_impl.py
```diff
@@ -1965,7 +1965,18 @@ def where(
...
+    if any(map(lambda x: isinstance(x, TensorProxy), (pred, a, b))):
+        dtype = dtypes.to_strong_dtype(dtype)
+
+    return fd.ops.cast(fd.ops.where(nvpred, nva, nvb), lcdtype_to_nvdtype(dtype))
```
IvanYashchuk, line 1979: This cast is inserted unconditionally. Is nvFuser or CUDA compiler capable of removing the cast when it's redundant?

jjsjann123: nvfuser is statically typed, so we are not inserting an explicit `cast` here. https://github.com/NVIDIA/Fuser/blob/ff91f2534f4efba6b21957e96e7e0d9c1737bfb5/csrc/ops/arith.cpp#L32-L35

PR #1735, opened 2025-02-03
warn if triton @jit or @autotune throws errors during import but continue

t-vi:
needed for the docs build in the CI


t-vi:
Merging to fix CI

PR #1742, opened 2025-02-04
type annotation minor change: `[Trace]` -> `list[Trace]`

crcrpar:
## What does this PR do?

As per title, as I've not often seen `[Trace]`


t-vi (approved):
Thank you @crcrpar

PR #1745, opened 2025-02-05
use `var is None` as `None` check

crcrpar:
## What does this PR do?

As per title.

cc @kiya00, as you seem to be the main contributor of this file


kiya00:
Thank you @crcrpar for the fix

mruberry (approved):

PR #1747, opened 2025-02-05
[Thunderfx report] Adds Thunder segmentation and fusion information

### LINKED ISSUE #1741: [Reporting tool] Adds the Thunder segmentation and fusion information to the FXReport
There are two immediate next options I can think of:

One option is to make thunder segmentation and fusion information programmatically available. This could be done by having a function like `analyze_with_thunder` that takes an FXGraphReport and creates a ThunderFXGraphReport, which would be a subclass of FXGraphReport. The ThunderFXGraphReport would have additional ThunderSegmentReport classes, and maybe ThunderFusionReport classes, so that the segments and fusions could be iterated over. Then the ability to write reproducers from the segments alone and from the fusions alone could be added to those classes (reusing the existing functionality as much as possible). That would allow for the creation of files that reproduce or benchmark segments and fusions. Each nvFusion could then be compared to executing with torch.compile instead, for example.

I think the second option is to focus on making the timing data programmatically available at the FX level. Basically running the benchmark and returning its data as a Python object. In the future we should let callers specify their own timers, but a first PR can just pick a timer to use. 

Whichever is done first, the second one can probably be done next. The timings don't need to be perfect at first, but it will be exciting to work on the proper timing of fx graphs, segments, and fusions in the near future. @kshitij12345 did some work attempting to compare nvfuser and torch.compile fusions, but the nvFuser team and mixology teams may have some other ideas for timing, too.

Once we have programmatically available timings for fx graphs, segments, and fusions, then I think it's time to revisit running these reports as part of the reports that Mixology produces.

How does that sound, @kiya00?

_Originally posted by @mruberry in https://github.com/Lightning-AI/lightning-thunder/pull/1707#pullrequestreview-2587842277_
            

cc @crcrpar

kiya00:
<details>

</details>

## What does this PR do?

Fixes #1741 .

adds `ThunderFXGraphReport, ThunderSegmentGraphReport, ThunderFusionReport, analyze_with_thunder` to support query thunder split graph and nvfusion information
An example:
```
import tempfile
from pathlib import Path
import torch
from thunder.dynamo.report import (
    fx_report, FXReport, ThunderFXGraphReport, FXGraphReport, 
    ThunderSplitGraphReport, ThunderFusionReport, analyze_thunder_splits
)
from pathlib import Path

x = torch.ones(2, 2, device="cuda", requires_grad=True)

# Dynamo segments `foo` into two graphs. Each graph contains one Thunder-split graph, 
# and each Thunder-split graph has one nvFusion region.
def foo(x):
    x = x.exp()
    torch._dynamo.graph_break()
    y = torch.sinc(x) + torch.cos(x)
    return y + 1

# If using `torch.compile` alone, you can stop here and query the reports in `FXReport`. 
# For more details, see the example in :func:`fx_report`.
results: FXReport = fx_report(foo, x)

tmp_path = Path("tmp_repro")
fx_graph_report: FXGraphReport
for idx, fx_graph_report in enumerate(results.fx_graph_reports):
    # `ThunderFXGraphReport` extends `FXGraphReport`, providing the ability to save 
    # reproduction/benchmark scripts for the original FX graph. Additionally, it 
    # includes information about Thunder-split subgraphs in `subgraph_reports`.
    thunder_fx_graph_report: ThunderFXGraphReport = analyze_thunder_splits(fx_graph_report)
    # Saves a reproduction script for the original FX graph
    thunder_fx_graph_report.write_thunder_repro(tmp_path)

    thunder_split_report: ThunderSplitGraphReport
    for thunder_split_report in thunder_fx_graph_report.subgraph_reports:
        split_folder = tmp_path / str(idx)
        thunder_split_report.write_eager_repro(split_folder)
        thunder_split_report.write_thunder_repro(split_folder)
        thunder_split_report.write_inductor_repro(split_folder)

        # If you are only interested in the Thunder-split FX graph, you can stop here. 
        # If you want to inspect Thunder traces and nvFusion regions further, explicitly call 
        # `ThunderSplitGraphReport.create_fusion_reports()` and analyze as shown below.
        thunder_split_report.create_fusion_reports()
        print(f"fwd_trace:\n{thunder_split_report.fwd_trc}\n")
        print(f"bwd_trace:\n{thunder_split_report.bwd_trc}\n")
        nvfusion_report: ThunderFusionReport
        for nvfusion_report in thunder_split_report.fusion_reports:
            nvfusion_report.write_nvfuser_repro(split_folder / "nvfusion")
            nvfusion_report.write_inductor_repro(split_folder / "nvfusion")
            bench_data = nvfusion_report.run_benchmark()
            print(bench_data)
            print("---"*10)
```
generates files:
```
.
├── 0
│   ├── graph0_thunder_0_repro_eager.py
│   ├── graph0_thunder_0_repro_thunder.py
│   ├── graph0_thunder_0_repro_torchcompile.py
│   └── nvfusion
│       ├── graph0_thunder_0_nvFusion0_backward_repro_inductor.py
│       ├── graph0_thunder_0_nvFusion0_backward_repro_nvfuser.py
│       ├── graph0_thunder_0_nvFusion0_forward_repro_inductor.py
│       └── graph0_thunder_0_nvFusion0_forward_repro_nvfuser.py
├── 1
│   ├── graph1_thunder_1_repro_eager.py
│   ├── graph1_thunder_1_repro_thunder.py
│   ├── graph1_thunder_1_repro_torchcompile.py
│   └── nvfusion
│       ├── graph1_thunder_1_nvFusion0_backward_repro_inductor.py
│       ├── graph1_thunder_1_nvFusion0_backward_repro_nvfuser.py
│       ├── graph1_thunder_1_nvFusion0_forward_repro_inductor.py
│       └── graph1_thunder_1_nvFusion0_forward_repro_nvfuser.py
├── graph0_repro_thunder.py
└── graph1_repro_thunder.py

```
The nvfusion repro script is taken from nvfuser's `repro_script_for`

## PR review

## Did you have fun?

### DIFF
--- a/thunder/dynamo/report.py
+++ b/thunder/dynamo/report.py
@@ -29,6 +29,7 @@ if TYPE_CHECKING:
     from collections.abc import Sequence
 
     from thunder.dynamo.utils import ExampleInputMetaData
+    from thunder.dev_utils.utils import BenchmarkComparisonData
 
 
 def run_backward(fn, *args, **kwargs):
@@ -473,7 +474,7 @@ def fx_report(fn: Callable, *args, compile_options: dict = None, **kwargs) -> FX
     return FXReport(graphs)
 
 
-class ThunderSegmentGraphReport(FXGraphReport):
+class ThunderSplitGraphReport(FXGraphReport):
     def __init__(self, graph, graph_name, compiled_fn, example_input, thunder_options: dict, split_reason: str):
         super().__init__(graph, graph_name)
         self.compiled_fn = compiled_fn
@@ -567,7 +568,7 @@ class ThunderFusionReport:
         comment_str = f'"""\n{self.nvfusion_bsym}\n"""'
 
         if file_name == None:
-            file_name = f"{self.name}_repro.py"
+            file_name = f"{self.name}_repro_nvfuser.py"
         with open(folder / file_name, "w") as f:
             print(comment_str, file=f)
             print(repro_code, file=f)
@@ -575,21 +576,28 @@ class ThunderFusionReport:
     def get_inputs(self):
         return self.nvfusion_bsym._call_ctx[self.nvfusion_bsym.sym.name].last_inputs
 
-    def run_inductor_repro(self):
+    def write_inductor_repro(self, folder, file_name=None):
         from thunder.executors.torch_compile import make_compiled
+        from thunder.dynamo.repro_script_template import template_bsym_torch_compile
+        from thunder.core.utils import create_python_callable_from_bsym
 
-        callable_for_torchcompile = make_compiled(
-            self.nvfusion_bsym.subsymbols, self.nvfusion_bsym.flat_args, self.nvfusion_bsym.flat_outs
-        )
+        folder = Path(folder)
+        folder.mkdir(exist_ok=True, parents=True)
+        python_func = create_python_callable_from_bsym(self.nvfusion_bsym)
+        nvfusion_name = self.nvfusion_bsym.sym.name
+
         inputs = self.get_inputs()
-        callable_for_torchcompile(*inputs)
+        inputs = "[" + "".join(arg_like(inp) for inp in inputs) + "]"
+        program = template_bsym_torch_compile.format(python_func=python_func, func_name=nvfusion_name, inputs=inputs)
+        if file_name == None:
+            file_name = f"{self.name}_repro_inductor.py"
+        with open(folder / file_name, "w") as f:
+            f.write(program)
 
-    def run_benchmark(self):
+    def run_benchmark(self) -> BenchmarkComparisonData:
         from thunder.dev_utils.utils import _benchmark_fusion_region_with_nvfuser_and_torch_compile
 
-        benchmark_comparison_data = _benchmark_fusion_region_with_nvfuser_and_torch_compile(self.nvfusion_bsym)
-        print(self.nvfusion_bsym)
-        print(benchmark_comparison_data)
+        return _benchmark_fusion_region_with_nvfuser_and_torch_compile(self.nvfusion_bsym)
 
 
 class ThunderFXGraphReport(FXGraphReport):
@@ -598,10 +606,10 @@ class ThunderFXGraphReport(FXGraphReport):
         self.subgraph_info = subgraph_info
         self.thunder_options = thunder_options
 
-        self.subgraph_reports: list[ThunderSegmentGraphReport] = []
+        self.subgraph_reports: list[ThunderSplitGraphReport] = []
         self._create_subgraph_reports()
 
-    def get_thunder_module_name(self):
+    def _get_thunder_module_name(self):
         thunder_module_names = []
         for node in self.subgraph_info.split_graph_module.graph.nodes:
             target = node.target
@@ -610,7 +618,7 @@ class ThunderFXGraphReport(FXGraphReport):
         return thunder_module_names
 
     def _create_subgraph_reports(self):
-        thunder_module_names = self.get_thunder_module_name()
+        thunder_module_names = self._get_thunder_module_name()
         original_modules_to_thunder_modules = (
             [m, compiled_m]
             for m, compiled_m in self.subgraph_info.submodule_to_compiled_functions.items()
@@ -623,7 +631,7 @@ class ThunderFXGraphReport(FXGraphReport):
             thunder_module_names, original_modules_to_thunder_modules, example_inputs
         ):
             self.subgraph_reports.append(
-                ThunderSegmentGraphReport(
+                ThunderSplitGraphReport(
                     sub_gm_pair[0],
                     name,
                     sub_gm_pair[1].compiled_fn,
--- a/thunder/dynamo/report.py
+++ b/thunder/dynamo/report.py
@@ -501,7 +501,7 @@ class ThunderSplitGraphReport(FXGraphReport):
         fwd_trc: The forward trace, available only after calling :meth:`_create_thunder_traces`.
         bwd_trc: The backward trace, available only after calling :meth:`_create_thunder_traces`.
 
-    For an example, see the documentation of :func:`analyze_with_thunder`.
+    For an example, see the documentation of :func:`analyze_thunder_splits`.
     """
 
     def __init__(
@@ -534,6 +534,10 @@ class ThunderSplitGraphReport(FXGraphReport):
         self.bwd_trc = last_backward_traces(self.compiled_fn)[-1]
 
     def create_fusion_reports(self):
+        """
+        Runs the Thunder-compiled function to obtain the nvFusion definition
+        and generate the :class:`ThunderFusionReport` instance based on it.
+        """
         self._create_thunder_traces()
         for trace, prefix in [(self.fwd_trc, "forward"), (self.bwd_trc, "backward")]:
             for bsym in trace.bound_symbols:
@@ -593,7 +597,7 @@ class ThunderFusionReport:
         nvfusion_bsym (BoundSymbol): The symbolic representation of the nvFusion region.
         name (str): The name of the fusion region.
 
-    For an example, see the documentation of :func:`analyze_with_thunder`.
+    For an example, see the documentation of :func:`analyze_thunder_splits`.
     """
 
     def __init__(self, bsym: BoundSymbol, name: str):
@@ -665,7 +669,7 @@ class ThunderFXGraphReport(FXGraphReport):
         subgraph_reports (list[ThunderSplitGraphReport]): A list of reports for each
             Thunder-split FX graph. For more details, see :class:`ThunderSplitGraphReport`.
 
-    For an example, see the documentation of :func:`analyze_with_thunder`.
+    For an example, see the documentation of :func:`analyze_thunder_splits`.
     """
 
     def __init__(self, gm: torch.fx.GraphModule, gm_name: str, thunder_options: dict):
@@ -715,7 +719,7 @@ class ThunderFXGraphReport(FXGraphReport):
             )
 
 
-def analyze_with_thunder(
+def analyze_thunder_splits(
     report: FXGraphReport,
     **thunder_options,
 ) -> ThunderFXGraphReport:
@@ -733,7 +737,7 @@ def analyze_with_thunder(
         import torch
         from thunder.dynamo.report import (
             fx_report, FXReport, ThunderFXGraphReport, FXGraphReport,
-            ThunderSplitGraphReport, ThunderFusionReport, analyze_with_thunder
+            ThunderSplitGraphReport, ThunderFusionReport, analyze_thunder_splits
         )
         from pathlib import Path
 
@@ -758,7 +762,7 @@ def analyze_with_thunder(
                 # `ThunderFXGraphReport` extends `FXGraphReport`, providing the ability to save
                 # reproduction/benchmark scripts for the original FX graph. Additionally, it
                 # includes information about Thunder-split subgraphs in `subgraph_reports`.
-                thunder_fx_graph_report: ThunderFXGraphReport = analyze_with_thunder(fx_graph_report)
+                thunder_fx_graph_report: ThunderFXGraphReport = analyze_thunder_splits(fx_graph_report)
                 # Saves a reproduction script for the original FX graph
                 thunder_fx_graph_report.write_thunder_repro(tmp_path)
 
--- a/thunder/dynamo/report.py
+++ b/thunder/dynamo/report.py
@@ -474,11 +474,32 @@ def fx_report(fn: Callable, *args, compile_options: dict = None, **kwargs) -> FX
 
 
 class ThunderSegmentGraphReport(FXGraphReport):
-    def __init__(self, graph, graph_name, example_input, thunder_options: dict, split_reason: str):
+    def __init__(self, graph, graph_name, compiled_fn, example_input, thunder_options: dict, split_reason: str):
         super().__init__(graph, graph_name)
+        self.compiled_fn = compiled_fn
         self.example_input = example_input
         self.thunder_options = thunder_options
         self.split_reason = split_reason
+
+        self.fusion_reports: list[ThunderFusionReport] = []
+        self.fwd_trc = None
+        self.bwd_trc = None
+
+    def create_thunder_traces(self):
+        from thunder import last_traces, last_backward_traces
+
+        example_inputs = eval(f"[\n{chr(10).join(arg_like(a) for a in self.example_input)}]")
+        # executes to get the trace
+        run_backward(self.compiled_fn, *example_inputs)
+        self.fwd_trc = last_traces(self.compiled_fn)[-1]
+        self.bwd_trc = last_backward_traces(self.compiled_fn)[-1]
+
+    def create_fusion_reports(self):
+        self.create_thunder_traces()
+        for trace, prefix in [(self.fwd_trc, "forward"), (self.bwd_trc, "backward")]:
+            for bsym in trace.bound_symbols:
+                if bsym.sym.is_fusion and "nvFusion" in bsym.sym.name:
+                    self.fusion_reports.append(ThunderFusionReport(bsym, f"{prefix}_{bsym.sym.name}"))
 
     def write_thunder_repro(self, folder, use_benchmark=False, serialize_inputs=False):
         thunder_ex_str = (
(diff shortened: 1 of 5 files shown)


mruberry (commented):
Exciting! Looking forward to more comments on the classes and functions. Your PR descriptions always do an excellent job of showcasing the work.

Overall this looks really great. Let's keep the focus on the new classes and methods in this PR, and then let's follow-up in another PR on the details of the timing information. @kshitij12345 has been reviewing various timers carefully and can probably help, too

kshitij12345:
> but getting the nvfusion subsymbols to run with torch.compile in a standalone script is a bit tricky

Regarding this, recently I was looking for something similar and this is roughly what seemed to work.

```python
import torch
import thunder
from thunder.dev_utils.utils import _benchmark_fusion_region_with_nvfuser_and_torch_compile
from thunder.core.trace import TraceCtx, tracectx
from thunder.core.codeutils import SigInfo
from thunder.dynamo.utils import arg_like

def fn(x):
    y = (x * x).sum()
    z = x @ x
    return z.sin() + y.cos()

jfn = thunder.jit(fn, nv_store_fusion_inputs=True)

jfn(torch.randn(16, 16, device="cuda"))

trc = thunder.last_traces(jfn)[-1]

def create_python_callable_from_bsym(bsym):
    trace = TraceCtx()
    si = SigInfo(bsym.sym.name)
    si.args = [(v.name, None) for v in bsym.flat_args]
    trace._siginfo = si
    trace.siginfo()
    # trace.args = bsym.flat_args
    trace.bound_symbols = list(bsym.subsymbols)

    with tracectx(trace):
        thunder.prims.python_return(bsym.output)

    
    return trace.python(include_decorators=False)

template_torch_compile = '''
{python_func}

from thunder.executors.torch_compile import make_compiled as make_torch_compile_callable
import thunder.examine

# inputs
x = torch.randn(16, 16, device="cuda")
inputs = {inputs}


jfn = thunder.jit({func_name})
o = jfn(*inputs)

trc = thunder.last_traces(jfn)[-1]
fusion_symbols = thunder.examine.get_fusion_symbols(trc)
assert len(fusion_symbols) == 1
bsym = fusion_symbols[0]

# print(bsym)  # Verified visually it looks the same
torch_compiled_callable = make_torch_compile_callable(bsym.subsymbols, bsym.flat_args, bsym.flat_outs)
o = torch_compiled_callable(*inputs)
'''

for bsym in trc.bound_symbols:
    if bsym.sym.is_fusion and "nvFusion" in bsym.sym.name:
        python_func = create_python_callable_from_bsym(bsym)
        with open("torch_compile_repro.py", "w") as f:
            nvfuser_callable = bsym._call_ctx[bsym.sym.name]
            inputs = nvfuser_callable.last_inputs
            inputs = "[" + "".join(arg_like(inp) for inp in inputs) + "]"
            program = template_torch_compile.format(python_func=python_func, func_name=bsym.sym.name, inputs=inputs)
            f.write(program)
        break

```

Generated Repro

<details>


```python

import thunder
import torch

def nvFusion0(x):
  # /opt/pytorch/lightning-thunder/test.py:9: 	    y = (x * x).sum()
  t0 = prims.mul(x, x)  # t0: "cuda:0 f32[16, 16]"
  y = prims.sum(t0, (0, 1))  # y: "cuda:0 f32[]"
  return [y]

from thunder.executors.torch_compile import make_compiled as make_torch_compile_callable
import thunder.examine

# inputs
x = torch.randn(16, 16, device="cuda")
inputs = [torch.testing.make_tensor((16, 16), dtype=torch.float32,  device='cuda:0', requires_grad=False, low=-2.9306769371032715, high=2.9876084327697754,).as_strided(torch.Size([16, 16]), (16, 1)),]


jfn = thunder.jit(nvFusion0)
o = jfn(*inputs)

trc = thunder.last_traces(jfn)[-1]
fusion_symbols = thunder.examine.get_fusion_symbols(trc)
assert len(fusion_symbols) == 1
bsym = fusion_symbols[0]

# print(bsym)  # Verified visually it looks the same
torch_compiled_callable = make_torch_compile_callable(bsym.subsymbols, bsym.flat_args, bsym.flat_outs)
o = torch_compiled_callable(*inputs)

```

</details>

The repro takes a roundabout way to create a torch.compile callable for a nvFusion but it seems to work. Maybe this could help come up with a cleaner approach.

mruberry:
Starting with @kshitij12345's technique and throwing an exception if a problem is encountered sounds like a great start. fyi @IvanYashchuk. I'm sure over time we can make the approach @kshitij12345 recommends more modular. It seems like it has the right concepts

thunder/dynamo/report.py
```diff
@@ -464,3 +471,194 @@ def helper_backend(gm, example_inputs):
...
+            self.nvfusion_bsym.subsymbols, self.nvfusion_bsym.flat_args, self.nvfusion_bsym.flat_outs
+        )
+        inputs = self.get_inputs()
+        callable_for_torchcompile(*inputs)
+
+    def run_benchmark(self):
+        from thunder.dev_utils.utils import _benchmark_fusion_region_with_nvfuser_and_torch_compile
+
+        benchmark_comparison_data = _benchmark_fusion_region_with_nvfuser_and_torch_compile(self.nvfusion_bsym)
+        print(self.nvfusion_bsym)
```
mruberry, line 591: Overall the classes and methods look pretty reasonable.

The big change I'd like to see, if not in this PR then in the next one, is that conversion of these prints to returning Python objects which can be programmatically queried for the desired data. Of course, those objects should also be printable, if someone wants to print them

thunder/dynamo/report.py
```diff
@@ -464,3 +471,194 @@ def helper_backend(gm, example_inputs):
...
+        self.compiled_fn = compiled_fn
+        self.example_input = example_input
+        self.thunder_options = thunder_options
+        self.split_reason = split_reason
+
+        self.fusion_reports: list[ThunderFusionReport] = []
+        self.fwd_trc = None
+        self.bwd_trc = None
+
+    def create_thunder_traces(self):
```
kshitij12345, line 488: Looks like this is supposed to be a private method. In that case, it would be nice to prefix with `_`.

thunder/dynamo/report.py
```diff
@@ -464,3 +471,194 @@ def helper_backend(gm, example_inputs):
...
+        from thunder import last_traces, last_backward_traces
+
+        example_inputs = eval(f"[\n{chr(10).join(arg_like(a) for a in self.example_input)}]")
+        # executes to get the trace
+        run_backward(self.compiled_fn, *example_inputs)
+        self.fwd_trc = last_traces(self.compiled_fn)[-1]
+        self.bwd_trc = last_backward_traces(self.compiled_fn)[-1]
+
+    def create_fusion_reports(self):
```
kshitij12345, line 497: It would nice to have a docstring for this. I couldn't infer from the name that this will populate `fusion_reports` attribute.

thunder/dynamo/report.py
```diff
@@ -464,3 +471,194 @@ def helper_backend(gm, example_inputs):
...
+
+    def _create_subgraph_reports(self):
+        thunder_module_names = self.get_thunder_module_name()
+        original_modules_to_thunder_modules = (
+            [m, compiled_m]
+            for m, compiled_m in self.subgraph_info.submodule_to_compiled_functions.items()
+            if compiled_m.compiler == CompilerType.THUNDER
+        )
+        example_inputs = self.subgraph_info.thunder_compiled_fns_example_inputs
+        self.split_reason = get_split_reasons_string(self.subgraph_info)
```
kshitij12345, line 620: Does it have to be an attribute on `self`? I think a normal variable should be sufficient here.

thunder/dynamo/report.py
```diff
@@ -464,3 +471,194 @@ def helper_backend(gm, example_inputs):
...
+
+    def run_benchmark(self):
+        from thunder.dev_utils.utils import _benchmark_fusion_region_with_nvfuser_and_torch_compile
+
+        benchmark_comparison_data = _benchmark_fusion_region_with_nvfuser_and_torch_compile(self.nvfusion_bsym)
+        print(self.nvfusion_bsym)
+        print(benchmark_comparison_data)
+
+
+class ThunderFXGraphReport(FXGraphReport):
```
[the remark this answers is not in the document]
kshitij12345: We should add a docstring here as mentioned in the `analyze_with_thunder` docstring -> For more details, see :class:`ThunderFXGraphReport`

thunder/dynamo/report.py
```diff
@@ -464,3 +471,194 @@ def helper_backend(gm, example_inputs):
...
+        self.bwd_trc = last_backward_traces(self.compiled_fn)[-1]
+
+    def create_fusion_reports(self):
+        self.create_thunder_traces()
+        for trace, prefix in [(self.fwd_trc, "forward"), (self.bwd_trc, "backward")]:
+            for bsym in trace.bound_symbols:
+                if bsym.sym.is_fusion and "nvFusion" in bsym.sym.name:
+                    self.fusion_reports.append(ThunderFusionReport(bsym, f"{prefix}_{bsym.sym.name}"))
+
+    def write_thunder_repro(self, folder, use_benchmark=False, serialize_inputs=False):
```
kshitij12345, line 504: Just confirming if we also need to override `write_inductor_repro` and `write_eager_repro`.

thunder/dynamo/report.py
```diff
@@ -464,3 +471,194 @@ def helper_backend(gm, example_inputs):
     compiled(*args, **kwargs)
 
     return FXReport(graphs)
+
+
+class ThunderSegmentGraphReport(FXGraphReport):
```
kshitij12345, line 476: It would be nice to have a docstring since term `ThunderSegment` is probably being used first time here.

thunder/dynamo/report.py
```diff
@@ -464,3 +471,194 @@ def helper_backend(gm, example_inputs):
...
+class ThunderFXGraphReport(FXGraphReport):
+    def __init__(self, gm, gm_name, subgraph_info, thunder_options):
+        super().__init__(gm, gm_name)
+        self.subgraph_info = subgraph_info
+        self.thunder_options = thunder_options
+
+        self.subgraph_reports: list[ThunderSegmentGraphReport] = []
+        self._create_subgraph_reports()
+
+    def get_thunder_module_name(self):
```
kshitij12345, line 604: Name should probably be prefixed with `_`.

thunder/dynamo/report.py
```diff
@@ -464,3 +471,194 @@ def helper_backend(gm, example_inputs):
...
+                    name,
+                    sub_gm_pair[1].compiled_fn,
+                    example_input,
+                    self.thunder_options,
+                    self.split_reason,
+                )
+            )
+
+
+def analyze_with_thunder(
```
kshitij12345, line 637: Should it be named as `analyze_thunder_segments/regions` or something along that lines.

Because `analyze_with_thunder` seemed to imply we will be using `thunder` for analysis but IIUC this will return `FXGraphReport` object for thunder regions of the FXGraph from `report` (and probably not use thunder directly for anything).

kshitij12345 (commented):
Overall this looks good to me. I just have few suggestions and questions regarding naming. Thanks @kiya00

thunder/dynamo/report.py
kiya00: we don't need to override inductor/eager repro, the `write_thunder_repro` is overridden because we need to add the split reason and the specific thunder.jit command which doesn't exist in `FXGraphReport`

thunder/dynamo/report.py
kiya00: I rename it to `ThunderSplitGraphReport` because it seems we use segment for Dynamo and split for Thunder in the code base

thunder/dynamo/report.py
kshitij12345: Got it, thanks for clarification.

thunder/dynamo/report.py
kiya00: It's a bit confusing here, `analyze_with_thunder` doesn't return `FXGraphReport`, it returns `ThunderFXGraphReport` which extends `FXGraphReport`, providing the ability to save reproduction/benchmark scripts for the original FX graph(I add this description in docstr). Additionally, it includes information about Thunder-split subgraphs `ThunderSplitGraphReport`.
we can alternatively make `analyze_with_thunder` return a list of `ThunderSplitGraphReport` but I guess it's no harm to add this container class `ThunderFXGraphReport` to hold the list of `ThunderSplitGraphReport`

kshitij12345: Sorry I think I wasn't very clear. The signature of the function looks good to me in terms of what it returns.

I was just wondering about the name of the function `analyze_with_thunder`. Because, based on the name it seems that we will be using thunder to analyse the `FXGraph`. But the function is analysing ThunderSplit's from the FXGraphReport it takes as input, so does the name like `analyze_thunder_splits` or something along that line seem more descriptive of what this function does. Wdyt?

kiya00: Oh yes, I get it now, `analyze_thunder_splits` is a better name

thunder/dynamo/report.py
kiya00: For now I just modified it to return the benchmark data that @kshitij12345 added

thunder/dynamo/repro_script_template.py
```diff
@@ -113,3 +113,31 @@ def test_{graph_name}(benchmark, executor, compute_type):
...
+
+
+bsym_torch_compile_repro_template = """
+{python_func}
+
+from thunder.executors.torch_compile import make_compiled as make_torch_compile_callable
+import thunder.examine
+
+# inputs
+x = torch.randn(16, 16, device="cuda")
```
kshitij12345, line 125: This looks to be an unused variable.

kiya00: Thank you! I missed that. I'll remove it.

kiya00:
Hi @mruberry @IvanYashchuk @kshitij12345 it's ready to review
In the updates I use the snippet from @kshitij12345 to save the inductor repro script of nvFusion bsyms, and integrate the `_benchmark_fusion_region_with_nvfuser_and_torch_compile` for benchmarking nvFusion

thunder/dynamo/report.py
```diff
@@ -464,3 +479,314 @@ def helper_backend(gm, example_inputs):
...
+        )
+        if file_name == None:
+            file_name = f"{self.name}_repro_inductor.py"
+        with open(folder / file_name, "w") as f:
+            f.write(program)
+
+    def run_benchmark(self) -> BenchmarkComparisonData:
+        from thunder.dev_utils.utils import _benchmark_fusion_region_with_nvfuser_and_torch_compile
+
+        return _benchmark_fusion_region_with_nvfuser_and_torch_compile(self.nvfusion_bsym)
```
mruberry, line 656: We'll want to focus here in the next PR. Fyi @kshitij12345. A couple notes:

- We should probably improve our modularity so that we can just pass a callable to a timer class to generate timings, instead of timing both torch.compile and nvfuser simultaneously and explicitly
- We should use @kshitij12345's learnings from his analysis of kernel times to implement a kernel timer class or function that's a subclass of either our own timer class or some generic timer class so it can accept arguments like warmups and # iterations
- We should also have a wall timer class or function
- We should probably report statistics of the non-warmup runs performed, like mean, median, and stdev of the distribution (which I suppose requires assuming the distribution is a normal distribution)

One coding-style note is that in general we may want to separate the generation of dataclasses and the dataclasses themselves. For example, creating a `ThunderFXGraphReport` object currently triggers computation by calling `_create_subgraph_reports`. This is OK, but a more common pattern is to have the factory function, like `analyze_thunder_splits`, do the analysis and then stuff the data into the `ThunderGraphReport` object.

mruberry (approved):
Cool! This is a great step forward in that it represents the remaining objects we're currently interested in: segments and fusions. Fantastic.

I wrote a comment about what I think the next step is — improving the modularity and quality of our benchmark numbers. @kshitij12345 should be able to help there.

Once we have the timings working for fx graphs, segments, and fusions as expected, then I think we'll just need a few refinements to begin using this tool in our benchmark runs. In particular, we should review how correctness issues are reported, how we can identify significant performance deviations and only generate reports when they're found, and how we can summarize split reasons. 

How does that sound, @kiya00?

mruberry (approved):
Cool! This is a great step forward in that it represents the remaining objects we're currently interested in: segments and fusions. Fantastic.

I wrote a comment about what I think the next step is — improving the modularity and quality of our benchmark numbers. @kshitij12345 should be able to help there.

Once we have the timings working for fx graphs, segments, and fusions as expected, then I think we'll just need a few refinements to begin using this tool in our benchmark runs. In particular, we should review how correctness issues are reported, how we can identify significant performance deviations and only generate reports when they're found, and how we can summarize split reasons. 

How does that sound, @kiya00?

mruberry:
Conflicts will need to be resolved before this can be merged

kiya00:
Hi @mruberry , thanks for the suggestions. The code in the constructor of `ThunderFXGraphReport` is moved into the factory function `analyze_thunder_splits`.
And for the timer thing I'll think about it and ask @kshitij12345 in Monday and get back to you.

PR #1753, opened 2025-02-07
thunderFX - fix leak caused due to holding exception object

### LINKED ISSUE #1750: Memory leak when using ThunderFX in a part of the MoE layer code.

### To Reproduce

```
import torch

# Adapted from
# https://github.com/sgl-project/sglang/blob/04d8cd20883937a2b9cb1e23b192b534a64170cc/python/sglang/srt/layers/moe/fused_moe_native.py#L15
def fused_moe_def(x, w1, w2,) -> torch.Tensor:
    topk_weights = torch.tensor([[0.1736, 0.1356, 0.1349, 0.1239, 0.1196, 0.1119, 0.1095, 0.0910]], device='cuda:0')
    topk_ids = torch.tensor([[ 82,  97, 152,  26, 145, 211, 255, 242]], device='cuda:0')
    w13_weights = w1[topk_ids]
    w1_weights, w3_weights = torch.chunk(w13_weights, 2, dim=2)
    w2_weights = w2[topk_ids]
    x1 = torch.einsum("ti,taoi -> tao", x, w1_weights)
    x1 = torch.nn.functional.silu(x1)
    x3 = torch.einsum("ti, taoi -> tao", x, w3_weights)
    expert_outs = torch.einsum("tao, taio -> tai", (x1 * x3), w2_weights)
    return torch.einsum("tai,ta -> ti", expert_outs, topk_weights.to(expert_outs.dtype))

def init(
    batch_size: int,
) -> None:
    num_tokens = batch_size
    num_experts=256
    shard_intermediate_size=2304
    hidden_size=7168
    dtype = torch.bfloat16

    x = torch.randn(num_tokens, hidden_size, dtype=dtype, device="cuda")
    w1 = torch.randn(num_experts, shard_intermediate_size, hidden_size, dtype=dtype, device="cuda")
    w2 = torch.randn(
        num_experts, hidden_size, shard_intermediate_size // 2, dtype=dtype, device="cuda"
    )
    return x, w1, w2


def test_deepseek_sglang_moe():
    args = init(batch_size=1)
    fn = thunderfx(fused_moe_def)
    # fn = torch.compile(fused_moe_def)
    fn(*args)

for i in range(4):
    # torch.cuda.memory._record_memory_history()
    test_deepseek_sglang_moe()
    torch.cuda.synchronize()
    print(torch.cuda.max_memory_allocated())
    print(torch.cuda.max_memory_reserved())
    # torch.cuda.memory._dump_snapshot("moe_snapshot.pickle")
```

ThunderFX outputs:
```
13080177152
13082034176
25763766784
25765609472
38447356416
38449184768
torch.OutOfMemoryError: CUDA out of memory. 
```
torch.compile outputs:
```
12956370944
12970885120
12956370944
12970885120
12956370944
12970885120
12956370944
12970885120
```

Found when adding the microbenchmark of MoE layer with deepseek-ai/DeepSeek-R1 config (https://github.com/Lightning-AI/lightning-thunder/pull/1746)

kshitij12345:
Fixes #1750 

The leak occurs because when we see an unsupported thunder operation, we save the exception object in SplitReason 

https://github.com/Lightning-AI/lightning-thunder/blob/f15512a9a4d96dea600866e0e92a30c9a3919700/thunder/dynamo/utils.py#L259-L263

This is a problem because Exception will hold onto frame objects which hold on to locals and also by some interaction leading to a ref cycle (see below).

Fix - We break the cycle by saving the str representation of the exception instead of the exception object. This should be enough to tell what caused the exception.

Ref Cycle Image

<details>

cycle

</details>

### DIFF
--- a/thunder/tests/test_dynamo.py
+++ b/thunder/tests/test_dynamo.py
@@ -10,6 +10,7 @@ import torch.nn as nn
 import torch.nn.functional as F
 from looseversion import LooseVersion
 from unittest.mock import patch
+import weakref
 
 from thunder import dtypes
 from thunder.dynamo import thunderfx
@@ -1123,3 +1124,30 @@ def test_fxreport(executor, device: str, dtype: dtypes.dtype, use_benchmark, tmp
 
     for file in py_files:
         check(tmp_path / file, cmd)
+
+
+def test_leak_on_unsupported_thunder_operator():
+    # This test is to check the fix for a previous leak
+    # which was caused by holding onto the
+    # exception object in split_reason.
+
+    def unsupported_op_fn(w1) -> torch.Tensor:
+        topk_ids = torch.tensor([[0, 1]])
+        # This indexing is not supported by thunder and get's passed to inductor.
+        w13_weights = w1[topk_ids]
+        return w13_weights + 1
+
+    def call_thunderfx_on_leaking_fn():
+        w1 = torch.randn(16, 16, 32, dtype=torch.bfloat16)
+        fn = thunderfx(unsupported_op_fn)
+        fn(w1)
+
+        # There should be two thunder traces because of the split caused by indexing.
+        # In future when thunder supports this indexing, the test will fail here
+        # as we won't be checking for the leak in case of unsupported operation.
+        assert len(fn.last_traces) == 2
+
+        return weakref.ref(w1)
+
+    t_weak_ref = call_thunderfx_on_leaking_fn()
+    assert t_weak_ref() is None
--- a/thunder/dynamo/utils.py
+++ b/thunder/dynamo/utils.py
@@ -73,12 +73,12 @@ class SplitReason:
     Attributes:
         reason_type: Reason for the split.
         info: String with details of what caused the split.
-        exception: Exception if there was any.
+        exception: String with details of exception if there was any.
     """
 
     reason_type: SplitReasonType
     info: str | None
-    exception: Exception | None = None
+    exception: str | None = None
 
 
 @dataclasses.dataclass(frozen=True)
@@ -248,7 +248,7 @@ def try_execute_thunder_symbol(thunder_symbol: Symbol, node: torch.fx.Node) -> t
             return False, SplitReason(
                 SplitReasonType.EXCEPTION_PROXY_THUNDER_OP,
                 f"Failed while creating proxy for node with name: {node.name} and target: {node.target}, see exception field",
-                exception=e,
+                exception=str(e),
             )
 
         # We need to be under trace context to generate proxies.
@@ -259,7 +259,7 @@ def try_execute_thunder_symbol(thunder_symbol: Symbol, node: torch.fx.Node) -> t
                 return False, SplitReason(
                     SplitReasonType.EXCEPTION_META_THUNDER_OP,
                     f"Failed while running meta for node with name: {node.name} and target: {node.target}, see exception field",
-                    exception=e,
+                    exception=str(e),
                 )
 
         # Execution with proxies was successful.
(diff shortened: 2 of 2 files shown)


kiya00 (approved):
Thank you @kshitij12345 for the quick fix!

thunder/tests/test_dynamo.py
```diff
@@ -1123,3 +1124,30 @@ def check(file_name, cmd):
...
+        fn(w1)
+
+        # There should be two thunder traces because of the split caused by indexing.
+        # In future when thunder supports this indexing, the test will fail here
+        # as we won't be checking for the leak in case of unsupported operation.
+        assert len(fn.last_traces) == 2
+
+        return weakref.ref(w1)
+
+    t_weak_ref = call_thunderfx_on_leaking_fn()
```
mruberry, line 1152: This seems OK since it's passing CI, although in the future I'm not sure if we can rely on CPython to delete `w1` without calling https://docs.python.org/3/library/gc.html#gc.collect explicitly

mruberry: We may want to do something like call https://docs.python.org/3/library/gc.html#gc.get_referents on the reports

mruberry (approved):
I mention this in my comments, but in the future we may want to more explicitly validate whether the object in question refers to something it shouldn't instead of relying on CPython's garbage collection behavior (or we may want to more explicitly trigger such behavior)

kshitij12345:
For the test, we are relying on the object to be freed by reference count going to zero and not by the `gc` being called. So, I don't think we need the `gc.collect` call or worry about whether it is implicitly triggered or not.

Eg.
```python
import gc
import weakref
import torch
gc.disable()

# Relying on reference counting.
def fn():
    a = torch.randn(3, 3)
    return weakref.ref(a)

weakref_a = fn()

# This is ok even though we manually disabled gc.
assert weakref_a() is None
```

Relying on gc to clear the cycle
```python
import gc
import weakref
import torch
gc.disable()

# Relying on GC
def fn():
    a = torch.randn(3, 3)
    l1 = [a]
    l2 = []

    # Create a cycle.
    l1.append(l2)
    l2.append(l1)
    return weakref.ref(a)

weakref_a = fn()

# gc.collect()  # Calling this will free `a`

# a will still be alive as in this case,
# gc needs to break the cycle before
# `a` can be freed.
assert weakref_a() is None  # This will fail as gc was disabled.

```

mruberry:
> For the test, we are relying on the object to be freed by reference count going to zero and not by the `gc` being called. So, I don't think we need the `gc.collect` call or worry about whether it is implicitly triggered or not.
> 
> Eg.
> 
> ```python
> import gc
> import weakref
> import torch
> gc.disable()
> 
> # Relying on reference counting.
> def fn():
>     a = torch.randn(3, 3)
>     return weakref.ref(a)
> 
> weakref_a = fn()
> 
> # This is ok even though we manually disabled gc.
> assert weakref_a() is None
> ```
> 
> Relying on gc to clear the cycle
> 
> ```python
> import gc
> import weakref
> import torch
> gc.disable()
> 
> # Relying on GC
> def fn():
>     a = torch.randn(3, 3)
>     l1 = [a]
>     l2 = []
> 
>     # Create a cycle.
>     l1.append(l2)
>     l2.append(l1)
>     return weakref.ref(a)
> 
> weakref_a = fn()
> 
> # gc.collect()  # Calling this will free `a`
> 
> # a will still be alive as in this case,
> # gc needs to break the cycle before
> # `a` can be freed.
> assert weakref_a() is None  # This will fail as gc was disabled.
> ```

Good point! Sounds good. In the future we may still want to develop a helper to test if an object x is holding a reference to another object y, but definitely not needed for now.

Issue #1732, opened 2025-02-01
Grad Transform generates inconsistent saved_for_backward between forward and backward trace.


jjsjann123:
## bug

In branch #1712 

```
import thunder

def foo(a, b):
    return a * b

import torch

a = torch.randn(1, 32, 232, 232)
a.requires_grad_()
b = torch.randn(1, 1, 232, 232)
#b.requires_grad_()  # adding this fixes the issue. So does this small code change [here](https://github.com/Lightning-AI/lightning-thunder/pull/1725/files#diff-dffa7275f3b675c42b33a9f91f1259dd4167631b3b6318ea03981cfbb1f87073R78-R79) 

jfoo = thunder.jit(foo, cache="symbolic values")
out = jfoo(a, b)
out.sum().backward()
```

Hitting the issue
```
Traceback (most recent call last):
  File "/volume/thunder_bwd6.py", line 21, in <module>
    out = jfoo(a, b)
          ^^^^^^^^^^
  File "/opt/pytorch/lightning-thunder/thunder/__init__.py", line 741, in wrapped
    return fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^
  File "/opt/pytorch/lightning-thunder/thunder/__init__.py", line 777, in fn_
    cache_entry, inps, pro_to_epi = get_computation_and_inputs(*args, **kwargs)
                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/pytorch/lightning-thunder/thunder/__init__.py", line 723, in wrapped
    cache_entry, inps, pro_to_epi = get_computation_and_inputs_fn(*args, **kwargs)
                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/pytorch/lightning-thunder/thunder/core/langctxs.py", line 136, in _fn
    result = fn(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^
  File "/opt/pytorch/lightning-thunder/thunder/__init__.py", line 235, in cache_info_wrapper
    res = fn(*args, **kwargs)
          ^^^^^^^^^^^^^^^^^^^
  File "/opt/pytorch/lightning-thunder/thunder/__init__.py", line 629, in get_computation_and_inputs
    computation_trc, backward_trc = split_forward_backward(computation_trc, cd, cs, *inps)
[...]

PR #1754, opened 2025-02-07
Simplify and fix match_fw_and_bw_saved_for_bw_proxies implementation

IvanYashchuk:
This change is needed to unblock @jjsjann123 for https://github.com/Lightning-AI/lightning-thunder/issues/1732.

The previous implementation uses both tensors and non-tensors saved for backward in constructing `old_saved_for_backward_fw` but in construction of the mirrored object from the backward trace non-tensors are ignored leading to an error that @jjsjann123 observed while working on https://github.com/Lightning-AI/lightning-thunder/issues/1732.

cc @ali-alshaar7

PR #1755, opened 2025-02-07
Update vjp implementation to attach residuals to all outputs

IvanYashchuk:
**Base PR: https://github.com/Lightning-AI/lightning-thunder/pull/1754, I'm keeping this PR in Draft for stacking PRs**

The current code assumes that the first output of all BoundSymbols will be a Proxy with a `.name` attribute. It's not always the case when DCE might replace unused outputs with `None` or some other objects are returned from Symbols.

The important changes are:

* skipping symbols with no proxies in the output to attach additional metadata for propagation (using is_literal, the approach is similar to the attempt in https://github.com/Lightning-AI/lightning-thunder/pull/1712)
* instead of specialing on the assumption that the first output of bound symbols is always a proxy object we actually find the first output, if all outputs are non-proxies they should have been skipped due to changes from the previous bullet point

I added the test from the issue description of https://github.com/Lightning-AI/lightning-thunder/issues/1732.

Fixes https://github.com/Lightning-AI/lightning-thunder/issues/1732.

### DIFF
--- a/thunder/tests/test_grad.py
+++ b/thunder/tests/test_grad.py
@@ -1934,6 +1934,18 @@ def test_adhoc_executor_grad(executor, device, _):
     torch.testing.assert_close(actual_gr, expected_gr)
 
 
+# See https://github.com/Lightning-AI/lightning-thunder/issues/1732
+def test_symbolic_shape_for_backward_issue_1732():
+    @partial(thunder.jit, cache="symbolic values")
+    def f(a, b):
+        return a * b
+
+    a = make_tensor((1, 32, 232, 232), device="cpu", dtype=torch.float32, requires_grad=True)
+    b = make_tensor((1, 1, 232, 232), device="cpu", dtype=torch.float32, requires_grad=False)
+    out = f(a, b)
+    out.backward(torch.ones_like(out))
+
+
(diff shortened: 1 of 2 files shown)


thunder/tests/test_grad.py
```diff
@@ -1934,6 +1934,18 @@ def func(x):
     torch.testing.assert_close(actual_gr, expected_gr)
 
 
+# See https://github.com/Lightning-AI/lightning-thunder/issues/1732
+def test_symbolic_shape_for_backward_issue_1732():
```
IvanYashchuk, line 1938: @jjsjann123, is there any other test you tried that didn't work?

IvanYashchuk:
This is ready for review, the draft status is only to prevent merges into the previous PR.

IvanYashchuk:
Moved https://github.com/Lightning-AI/lightning-thunder/pull/1755/commits/7a2f7f7d192d428219d858241ba68124b720eaba and https://github.com/Lightning-AI/lightning-thunder/pull/1755/commits/b50f9466f98b8871873bd4dbca952d647a53b171 to a separate PR https://github.com/Lightning-AI/lightning-thunder/pull/1834.

PR #1759, opened 2025-02-10
Adds the timer based on torch.profiler.profile

### LINKED ISSUE #1757: [ThunderFX report] Provides modular Timer interface
We'll want to focus here in the next PR. Fyi @kshitij12345. A couple notes:

- We should probably improve our modularity so that we can just pass a callable to a timer class to generate timings, instead of timing both torch.compile and nvfuser simultaneously and explicitly
- We should use @kshitij12345's learnings from his analysis of kernel times to implement a kernel timer class or function that's a subclass of either our own timer class or some generic timer class so it can accept arguments like warmups and # iterations
- We should also have a wall timer class or function
- We should probably report statistics of the non-warmup runs performed, like mean, median, and stdev of the distribution (which I suppose requires assuming the distribution is a normal distribution)

_Originally posted by @mruberry in https://github.com/Lightning-AI/lightning-thunder/pull/1747#discussion_r1946843339_

kiya00:
<details>

</details>

## What does this PR do?

Fixes #1757.

Adds the `CompileSpecificationInterface`, `TimerInterface` as discussed in https://github.com/Lightning-AI/lightning-thunder/issues/1757#issuecomment-2648620364
Adds the APIs that take the `CompileSpecificationInterface`, `TimerInterface`  instance as inputs: 
The programmatical interfaces in FXGraphReport: `run_benchmark`, `run_repro`, and the APIs to save the corresponding scripts: `write_benchmark`, `write_repro_v2`

NOTE the previously implemented function for writing pytest-benchmark scripts has been renamed to `write_pytest_benchmark`. The old method `write_repro(..., executor_str: str, import_str: list[str])` , which differs from the new one `write_repro_v2(...,compile_fn: CompileSpecificationInterface)`) is still there and are used by the series of `write_xxx_repro` functions.
 

TODO:
- Currently we use the `TorchProfileTimer` modified from https://github.com/NVIDIA/Fuser/blob/212ac38e08c47251356e0f0ee8f48e21a12b2293/benchmarks/python/core.py#L141 as the inner timing function in `torch.utils.benchmark.Timer` to measure the CUDA kernel time.
Will collaborate with the nvFuser team to unify this implementation, and also check if the perf number is right (possible issue
mentioned in https://github.com/Lightning-AI/lightning-thunder/issues/1757#issuecomment-2648792618)

Example usage:
```
thunderjit = ThunderCompileSpecification(executors=[torch_compile_cat_ex, nvfuser_executor])
fxgraph_report.run_repro(thunderjit, check_consistency=True)
fxgraph_report.write_repro_v2(folder_path, thunderjit, check_consistency=True)
fxgraph_report.run_benchmark(thunderjit, WallTime)
fxgraph_report.write_benchmark(folder_path, thunderjit, KernelTime)
```

## PR review

## Did you have fun?


kiya00:
Hi @mruberry , I had some draft code here, and explained in the PR description, it's not exactly the way we planned [here](https://github.com/Lightning-AI/lightning-thunder/issues/1757#issuecomment-2648620364), but it's kind of easy to implement, WDYT?

thunder/dynamo/report.py
```diff
@@ -607,6 +615,11 @@ def __init__(self, bsym: BoundSymbol, name: str):
     def __repr__(self):
         return f"<ThunderFusionReport of bound symbol\n{self.nvfusion_bsym}>"
 
+    def benchmark(self, fn: Callable, timer_fn: Callable):
+        compiled_fn = fn(self.nvfusion_bsym)
+        inputs = self.get_inputs()
+        return timer_fn("compiled_fn(*inputs)", globals={"compiled_fn": compiled_fn, "inputs": inputs})
```
mruberry, line 621: Interesting that the timer takes a string to execute and not a Python callable -- I'm guessing that's just the typical interface for timers?

thunder/dynamo/report.py
```diff
@@ -607,6 +615,11 @@ def __init__(self, bsym: BoundSymbol, name: str):
     def __repr__(self):
         return f"<ThunderFusionReport of bound symbol\n{self.nvfusion_bsym}>"
 
+    def benchmark(self, fn: Callable, timer_fn: Callable):
```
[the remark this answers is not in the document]
mruberry: My thinking here is that it'd be nice to standardize the representation of these segments as FX graphs, and then standardize how each is benchmarked

thunder/dynamo/timer.py
```diff
@@ -0,0 +1,74 @@
+import torch
+from torch.profiler import profile, ProfilerActivity
+from torch.utils.benchmark import Timer, timer
+from torch.utils.benchmark.utils.common import Measurement
+
+
+class TorchProfileTimer:
```
mruberry, line 7: Comment string for this class

thunder/dynamo/timer.py
```diff
@@ -0,0 +1,74 @@
...
+        """
+        from torch.autograd import DeviceType
+
+        elapsed_cuda_time = 0
+        has_cuda_event = False
+        for event in prof_averages:
+            if event.device_type != DeviceType.CUDA:
+                continue
+            has_cuda_event = True
+            # Re: torch profiler API changes in https://github.com/pytorch/pytorch/pull/123247
```
mruberry, line 27: Nice comment

thunder/dynamo/timer.py
```diff
@@ -0,0 +1,74 @@
...
+            elapsed_cuda_time = (
+                elapsed_cuda_time + event.self_device_time_total
+                if hasattr(event, "self_device_time_total")
+                else event.self_cuda_time_total
+            )
+        # assert has_cuda_event, "No CUDA events found"
+        # print(has_cuda_event)
+        # print(elapsed_cuda_time)
+        return elapsed_cuda_time / 1e6
```
mruberry, line 36: Add a comment for this conversion with the from and to units

thunder/dynamo/timer.py
```diff
@@ -0,0 +1,74 @@
...
+    def __init__(self):
+        self.prof = profile(activities=[ProfilerActivity.CUDA])
+        self.current_time = 0.0
+
+    def _get_kernel_time(self, prof_averages: torch.autograd.profiler_util.EventList) -> float:
+        """
+        Arguments:
+            prof_averages: Output of self.prof.key_averages()
+        Returns:
+            time_value: Elapsed CUDA time in seconds.
```
mruberry, line 17: Seconds seems like a large unit to use; what about microseconds?

thunder/dynamo/timer.py
```diff
@@ -0,0 +1,74 @@
...
+        # assert has_cuda_event, "No CUDA events found"
+        # print(has_cuda_event)
+        # print(elapsed_cuda_time)
+        return elapsed_cuda_time / 1e6
+
+    def _increment_global_time(self, elapsed_time: float) -> None:
+        self.current_time += elapsed_time
+
+    def __call__(self):
+        try:
```
mruberry, line 42: Interesting! What could cause an `AssertionError` here?

thunder/dynamo/timer.py
```diff
@@ -0,0 +1,74 @@
...
+
+        prof_averages = self.prof.key_averages()
+        elapsed_cuda_time = self._get_kernel_time(prof_averages)
+        self._increment_global_time(elapsed_cuda_time)
+        # Clear the internal profiler object to avoid accumulating function events and then restart the profiler
+        # See PR: https://github.com/pytorch/pytorch/pull/125510
+        # print(prof.key_averages().table(sort_by="self_cuda_time_total", row_limit=5))
+        self.prof.profiler = None
+
+        return self.current_time
```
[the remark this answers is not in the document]
kiya00: this is because `torch.utils.benchmark.Timer` requires the custom timer to be a callable and returns the current time. And with it the `torch.utils.benchmark.Timer` will return a `Measurement` that contains the statistics, like mean, median, iqr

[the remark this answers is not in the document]
kiya00: ah yes, I just saw it

thunder/dynamo/timer.py
kiya00: This is actually taken from nvfuser benchmark, I think instead of the triton timer it should be no problem if we use the same way as nvfuser did
https://github.com/NVIDIA/Fuser/blob/212ac38e08c47251356e0f0ee8f48e21a12b2293/benchmarks/python/core.py#L141-L163

thunder/dynamo/timer.py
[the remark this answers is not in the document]
mruberry: I think this concept of a timer might initially confuse readers (like myself) who aren't very familiar with extending Python's timing infrastructure, so it might help to point to https://github.com/python/cpython/blob/247b50dec8af47ed8a80069117e07b7139f9d54f/Lib/timeit.py#L67 to explain how these timers work and their interface

thunder/dynamo/timer.py
kiya00: This is for the first time the `stop` will raise error, so that the profiler start, and if it's already start, the `stop` will be called.
the timer is used in the torch benchmark like:
```
timer()
for loop:
     fn()
timer()
```

kiya00: I think in nvfuser python benchmark it uses pytest-benchmark with a custom benchmark fixture, and inside the fixture there's the custom timer which also needs to be a callable returns current time. What we used here is to take the same timer in the nvfuser but use it in the pytorch benchmark

thunder/dynamo/report.py
```diff
@@ -607,6 +615,11 @@ def __init__(self, bsym: BoundSymbol, name: str):
...
+        compiled_fn = fn(self.nvfusion_bsym)
+        inputs = self.get_inputs()
+        return timer_fn("compiled_fn(*inputs)", globals={"compiled_fn": compiled_fn, "inputs": inputs})
+
     def write_nvfuser_repro(self, folder, file_name=None):
```
mruberry, line 623: It would be really nice if these methods could be inherited from a common base, as the comment above mentions. Maybe `FXGraphReport` is that common base, or maybe they need a new base class, like `FXGraphSegment`, and then the `FXGraphReport` and `ThunderFusionReport` classes can define methods that customize how their reproductions are created?

mruberry:
Hey @kiya00!



I think it's OK for this PR to focus on programmatically benchmarking the various FX graph segments we're interested in, while improving the modularity of how we represent and manipulate those segments. I made a few comments about how we might reuse more code and create more consistency between these classes. I look forward to hearing your thoughts on it!

thunder/tests/test_dynamo.py
```diff
@@ -1192,3 +1192,77 @@ def check(file_name, cmd):
...
+
+    with FusionDefinition() as fd:
+        nvfuser_fusion_id2(fd)
+
+    inputs = [
+        torch.testing.make_tensor((2, 2), dtype=torch.float32, device="cuda:0"),
+        torch.testing.make_tensor((2, 2), dtype=torch.float32, device="cuda:0"),
+    ]
+
+    kernel_time(stmt="fd.execute(inputs)", globals={"fd": fd, "inputs": inputs})
```
mruberry, line 1223: Is there some validation we can do on the returned objects?

thunder/dynamo/timer.py
mruberry: Sounds like a great opportunity for a comment explaining what's happening ;)

thunder/dynamo/report.py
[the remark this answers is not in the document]
kiya00: it's like the `ThunderFXGraphReport` `ThunderSplitGraphReport` both inherit `FXGraphReport` because they both have an fx graph and the methods are around handling fx graph, but for `ThunderFusionReport` there's only thunder bound symbol, currently it's hard to convert the bound symbol to an fx graph. The methods are handling the bsym, so the signature of their methods are quite different, only some of the function names are the same. So in this situation, do we still create a common interface class?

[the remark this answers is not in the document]
mruberry: This is such an interesting problem!

Developing a technology that could convert a bound symbol to an FX graph would be great. I agree that we don't have to do that in this PR. If we don't pursue that in this PR, then I'd just add a TODO to the `ThunderFusionReport` class explaining that it's different, and in the future we might convert bound symbols to fx graphs and make it consistent with the other classes.

If we do want to pursue converting a bound symbol to an FX graph, I wonder if we could try something like this:

1. Take the trace constructed in `create_python_callable_from_bsym`
2. Transform it for execution by the PyTorch and Python executors
3. Use fx to create an fx graph by tracing through that callable

One issue with this approach is that the bound symbol may contain an operation with no PyTorch analogue, but that seems unlikely when dealing with nvFuser fusions. 

What do you think, @kiya00, @kshitij12345 ?

thunder/dynamo/report.py
```diff
@@ -399,6 +399,14 @@ def write_repro(
         with open(folder / file_name, "w") as f:
             print(code_str, file=f)
 
+    def benchmark(self, fn: Callable, time_fn: Callable):
```
kshitij12345, line 402: I think it'd be better to call `fn` -> `compile_fn` so that it is a bit clear.

thunder/dynamo/report.py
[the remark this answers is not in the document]
kiya00: I added a TODO comment about converting bsym to FX graph for now

thunder/dynamo/timer.py
kiya00: For now this is copied from the nvfuser member functions in the benchmark fixture class and used to form a timer class here, as @kshitij12345 suggested, will coordinate with nvFuser team to unify this implementation, and add comments explaining it.

kiya00:
Hi @mruberry , I’ve added the CompileSpecification and TimerSpecification parts (as detailed in the updated PR description). With these APIs in place, it should now be straightforward to implement the automatic testing report you mentioned.

thunder/dynamo/benchmark_utils.py
```diff
@@ -0,0 +1,275 @@
...
+    def import_str(self):
+        return ["import thunder"]
+
+
+class TorchCompileSpecification(CompileSpecificationInterface):
+    """
+    A compile specification for :func:`torch.compile`.
+    """
+
+    def __init__(self, **kwargs):
```
mruberry, line 72: Excellent work accepting the kwargs here

thunder/dynamo/benchmark_utils.py
```diff
@@ -0,0 +1,275 @@
...
+    """
+
+    def compile(self, fn):
+        return fn
+
+    def to_source(self, fn_name):
+        return fn_name
+
+
+class BoundSymbolNvfuserSpecification(CompileSpecificationInterface):
```
mruberry, line 98: Good special casing this for now and good comment with a TODO

thunder/dynamo/benchmark_utils.py
```diff
@@ -0,0 +1,275 @@
...
+    # Returns the torch compile callable from the nvFuser bound symbol.
+    # See the TODO in :class:`thunder.dynamo.report.ThunderFusionReport` for more details.
+    def compile(self, bsym):
+        from thunder.executors.torch_compile import make_compiled
+
+        return make_compiled(bsym.subsymbols, bsym.flat_args, bsym.flat_outs)
+
+
+# NOTE: This class is modified from https://github.com/NVIDIA/Fuser/blob/212ac38e08c47251356e0f0ee8f48e21a12b2293/benchmarks/python/core.py#L141
+# TODO: Collaborate with the nvFuser team to unify this implementation,
```
mruberry, line 115: Nice TODO

thunder/dynamo/benchmark_utils.py
```diff
@@ -0,0 +1,275 @@
...
+            )
+
+        return elapsed_cuda_time / 1e6
+
+    def _increment_global_time(self, elapsed_time: float) -> None:
+        self.current_time += elapsed_time
+
+    def __call__(self):
+        """
+        Custom torchprofiler-based timer used by pytest-benchmark.
```
mruberry, line 151: Nice comment

thunder/dynamo/benchmark_utils.py
```diff
@@ -0,0 +1,275 @@
...
+        t = TorchBenchmarkTimer(stmt=stmt, setup=setup, globals=globals)
+        return t.blocked_autorange(min_run_time=min_run_time)
+
+    @staticmethod
+    def import_str():
+        return ["from thunder.dynamo.benchmark_utils import WallTime"]
+
+    @staticmethod
+    def to_source(fn_name="compiled_model", inputs_name="inputs"):
+        return f'WallTime.time("{fn_name}(*{inputs_name})", globals={{"{fn_name}":{fn_name}, "{inputs_name}": {inputs_name}}})'
```
mruberry, line 251: Very cool ability to use the wall timer in a script

thunder/dynamo/report.py
```diff
@@ -399,6 +401,200 @@ def write_repro(
...
+
+    def run_repro(
+        self,
+        compile_fn: CompileSpecificationInterface,
+        check_consistency=False,
+        inputs: Sequence[torch.Tensor | ExampleInputMetaData] = None,
+    ):
+        compiled_model = compile_fn.compile(self.graph)
+        inputs = self.get_input_metadata()
+        example_inputs = eval(f"[\n{chr(10).join(arg_like(a) for a in inputs)}]")
```
mruberry, line 451: TODO -- there's probably a nicer way to generate the example inputs that doesn't involve evaluating a string. If string evaluation is ideal, however, a comment explaining why would be nice

thunder/dynamo/report.py
```diff
@@ -399,6 +401,200 @@ def write_repro(
...
+        compile_str = compile_fn.to_source("model")
+        code_str += textwrap.indent(f"compiled_model = {compile_str}\n", "    ")
+
+        if forward_only:
+            run_str = "result = compiled_model(*inputs)\n"
+        else:
+            run_str = "from thunder.dynamo.report import run_forward_backward\nfwd_result, grads = run_forward_backward(compiled_model, *inputs)\n"
+        code_str += textwrap.indent(run_str, "    ")
+
+        if check_consistency:
```
mruberry, line 510: Nice gradual extension of the script

thunder/dynamo/report.py
```diff
@@ -399,6 +401,200 @@ def write_repro(
...
+        if file_name is None:
+            file_name = f"{self.graph_name}.py"
+        with open(folder / file_name, "w") as f:
+            print(code_str, file=f)
+
+    def run_benchmark(self, compile_fn: CompileSpecificationInterface, time_fn: TimerInterface):
+        compiled_fn = compile_fn.compile(self.graph)
+        inputs = self.get_input_metadata()
+        forward_only = not any(hasattr(arg, "requires_grad") and arg.requires_grad for arg in inputs)
+        example_inputs = eval(f"[\n{chr(10).join(arg_like(a) for a in inputs)}]")
```
mruberry, line 529: Maybe this should be a helper since it appears in a couple places -- maybe same with the `forward_only` check

thunder/dynamo/report.py
```diff
@@ -399,6 +401,200 @@ def write_repro(
...
+        fwd_measurement = time_fn.time(
+            "compiled_fn(*example_inputs)", globals={"compiled_fn": compiled_fn, "example_inputs": example_inputs}
+        )
+        if forward_only:
+            return fwd_measurement
```
mruberry, line 534: Suggestion: `fwd_measurement, None` so that the return type is always a tuple with two values (in general it's nice when functions have consistent return types)

thunder/dynamo/report.py
```diff
@@ -399,6 +401,200 @@ def write_repro(
...
+            inputs (Sequence[torch.Tensor | ExampleInputMetaData], optional):
+                The input tensors or metadata for the FX graph module. Defaults to None.
+                If not provided, inputs will be inferred from the placeholders in the FX graph.
+            **kwargs: Additional arguments for customization.
+        """
+        folder = Path(folder)
+        folder.mkdir(exist_ok=True, parents=True)
+        if inputs == None:
+            inputs = self.get_input_metadata()
+        forward_only = not any(hasattr(arg, "requires_grad") and arg.requires_grad for arg in inputs)
```
mruberry, line 573: Seems like there should be an opportunity to reuse more code with `write_repro_v2`

thunder/dynamo/report.py
```diff
@@ -399,6 +401,200 @@ def write_repro(
...
+    # forward
+    fwd_measurement = {fwd_timing_str}
+"""
+        if not forward_only:
+            code_str = f"""{code_str}
+    # backward
+    from thunder.benchmarks.targets import backward_only
+    backward_fn, backward_setup = backward_only(compiled_model, *inputs)
+    backward_args = backward_setup()
+    bwd_measurement = {bwd_timing_str}
```
mruberry, line 589: If a practitioner runs this script, will they see anything happen? Maybe it should print some information about the `fwd_measurement` and the `bwd_measurement` so that by running the script someone can see the summary immediately? This can all be done in the future. Also in the future if the script can contain a message then maybe the message should be printed, too, so just by running the script an engineer would see a description of why the script was created and what it should be showing, then the benchmarks

thunder/dynamo/report.py
```diff
@@ -586,6 +782,10 @@ def write_thunder_repro(self, folder, use_benchmark=False, serialize_inputs=Fals
         )
 
 
+# TODO: `ThunderFusionReport` is expected to inherit from `FXGraphReport` for consistency.
```
mruberry, line 785: Nice comment

mruberry (approved):
Great!

Let's get this merged. I made some comments for possible follow-up tweaks and revisions.

In addition to those, I think it's time we start trying to deploy this script. What do you think about creating a function that will do the following, @kiya00:

- Given a callable and inputs to it...
- Try to run that callable eagerly and record success/failure
- Try to run that callable with torch.compile and record the generated fx graphs, record whether torch.compile succeeded or failed
- For each generated fx graph, see where thunder can run it, and record success/failure
- For each generated fx graph, compare the speed at which torch.compile and thunder runs it
  - I think we can use a comparison based on an absolute and relative tolerance, like for testing whether tensors are "close" (see the formula for https://pytorch.org/docs/stable/testing.html)
- If the comparison exceeds the requested threshold, create a benchmark script that reproduces the benchmarks
  - Ideally the benchmarking script would include the time comparison using the specified atol and rtol
  - Ideally the benchmarking script would include a programmatically generated message that explains what was originally found (e.g. "observed nvFuser time was X seconds, torch.compile time was Y seconds" 

This proposal only suggests comparisons for each FX graph. If you'd like to also compare segments and/or fusions that'd be great. We should probably include an option to configure which comparisons are made (e.g. `compare_fusions: bool = True`)

Maybe we can show how the script works on an example callable and then work with our partners on the Mixology team to run it as part of their automated analysis.

What do you think, @kiya00?

PR #1763, opened 2025-02-11
Revert "fuse only consecutive symbols (#1608)"

IvanYashchuk:
This reverts commit f2d715240555ca787fbdbba6f42c6f9d422ae0c3.

Testing CI. I will add the description later.

### DIFF
--- a/thunder/executors/data_dependent_partition.py
+++ b/thunder/executors/data_dependent_partition.py
@@ -1,0 +1,303 @@
+from __future__ import annotations
+
+from typing import List, Set
+from collections.abc import Callable
+from copy import copy
+from itertools import chain
+
+import thunder.core.utils as utils
+from thunder.core.trace import TraceCtx
+from thunder.core.symbol import BoundSymbol
+from thunder.core.proxies import variableify, Proxy
+import thunder.core.prims as prims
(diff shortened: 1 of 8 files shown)


t-vi:
I'm all for increasing the amount of fusion we do, bringing back the stuff that reodered things without regard to memory consumption, is not the solution, though.

IvanYashchuk:
What is the benchmark command for testing memory consumption differences?

We need to restore to the state when we had better performance on the benchmarks we're tracking and then improve memory consumption without regressing on performance for the rest of the benchmark configurations.

t-vi:
It was using the benchmark code with various parameters, in particular with and without checkpointing.

And in my experience, the code "randomly" (as in probably determined somehow, but not reliably controlled) reordered operations without regard to memory consumption.
We cannot reorder operations without regard to memory consumption. Sure, it might be better if you get lucky, but it will be arbitrarily bad for other use cases.
I'm not sure why we're trying to return to the brittleness that this thing was.

IvanYashchuk:
@t-vi, is there anything else you would like to see implemented before merging this PR?

t-vi (approved):
Thank you @IvanYashchuk  Let's merge keeping it off for the non-dynamo path now, too.

PR #1765, opened 2025-02-14
Add fusion_type compile option for controlling the fusion logic

IvanYashchuk:
This PR adds an option for choosing a fusion type (requested in https://github.com/Lightning-AI/lightning-thunder/pull/1763#issuecomment-2654504891). Currently, there are two fusion types:
1. `"dataflow"` - fusions are formed based on dataflow and merged horizontally and vertically,
2. `"consecutive"` - fusions are formed if two symbols are fusible preserving the input order (added in https://github.com/Lightning-AI/lightning-thunder/pull/1608).

`thunder.dynamo.thunderfx` and `thunder.jit` have different defaults, `thunderfx` uses the "dataflow" fusion type (as it was before https://github.com/Lightning-AI/lightning-thunder/commit/f2d715240555ca787fbdbba6f42c6f9d422ae0c3) and `jit` uses the "consecutive" fusion type introduced in https://github.com/Lightning-AI/lightning-thunder/commit/f2d715240555ca787fbdbba6f42c6f9d422ae0c3.

### DIFF
--- a/docs/source/intermediate/benchmarking.rst
+++ b/docs/source/intermediate/benchmarking.rst
@@ -272,3 +272,8 @@ As seen earlier, it's possible to write benchmarks for models and not just stand
       benchmark_for_compute_type(compute_type, benchmark, fn, *args, **kwargs)
 
 And that's as simple as that! Just add the decorator ``@parametrize_compute_type`` after your parametrization, add the ``compute_type`` argument, and use ``benchmark_for_compute_type`` to call the benchmark function.
+
+Isolate benchmarks to avoid OutOfMemory errors
+----------------------------------------------
+
+When running multiple benchmarks in sequence, ``pytest`` does not always do a good job cleaning up, and sometimes it happens that, while they work when called standalone, benchmarks fail anyway. The main problem we observed is that memory is not entirely freed before running the next benchmark, therefore the option ``--isolate-benchmarks`` comes in rescue. It will separate the benchmark runs, creating a sub-process for each benchmark configuration and run them one after the other. Logs of failures will be saved in the ``failed_benchmarks_logs`` folder and benchmark results will be saved in the form of json in the ``benchmarks_reports`` folder unless the ``THUNDER_BENCH_DIR`` environment variable is specified.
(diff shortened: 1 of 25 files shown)


IvanYashchuk:
This pull request is ready for review. I'm keeping the draft mode to prevent accidental merges into the wrong base branch.

t-vi:
Please make it a boolean option. Thank you.

t-vi:
I think we are having similar discussions for just about any option.
There are two things that make me prefer a boolean option:
- To my mind, having keyword options like this is a discoverability problem.
- I don't think it is good to have many codepaths through the compiler, so I would prefer to keep the option space small.

My concern about the algorithm you prefer to be the default with the dynamo frontend is that it appears to reoder operations without regard to memory requirements and just happens to produce a happier outcome on some models and seems to be producing very high memory requirements in some checkpointing cases for us. As such it would be preferable if the non-dynamo-frontend codepath to use the consecutive merging exclusively until it is amended by a memory-aware reordering of operations. Whether to capture this in a boolean option set by the dynamo frontend call to the jit or just hardcode based on the frontend option is OK.

I understand that your vision for the dynamo frontend is different from that, so if you deal with that however you want, it will be fine.

IvanYashchuk:
I understand your concerns about the default choice for ThunderFX being suboptimal for certain types of computations. We are aware of the limitations.

> As such it would be preferable if the non-dynamo-frontend codepath to use the consecutive merging exclusively until it is amended by a memory-aware reordering of operations.

It's done so in the current PR. `thunder.jit` by default uses the `fusion_type="consecutive"` option. 

I'm sorry I don't get your request. I thought you were asking instead of "fusion_type=string" something like "enable_dataflow_fusion=True" (which is still a keyword argument to the `thunder.jit` call). But then you state one reason to prefer a boolean option is that keyword arguments have a discoverability problem. I thought "boolean" refers to the type of the value passed as a keyword argument to `thunder.jit`. Could you please clarify what change you would like to see?

t-vi:
Let's leave it as is then.
Can you please resolve the conflicts so we can merge?

t-vi (approved):
Thank you @IvanYashchuk

Issue #703, opened 2024-07-03
vjp correctness fails for sdpa manual grad forward sdpa


t-vi:
From the CI run in an intermediate version of #691 :

FAILED thunder/tests/test_grad.py::test_vjp_correctness_sdpa_manual_grad_forward_scaled_dot_product_attention_nvfuser_cuda_thunder.dtypes.float16 - AssertionError: Tensor-likes are not close!

Mismatched elements: 19624 / 245760 (8.0%)
Greatest absolute difference: 0.125 at index (1, 1, 60, 64) (up to 1e-05 allowed)
Greatest relative difference: inf at index (0, 0, 54, 6) (up to 0.001 allowed)
FAILED thunder/tests/test_grad.py::test_vjp_correctness_sdpa_manual_grad_forward_scaled_dot_product_attention_nvfuser_cuda_thunder.dtypes.bfloat16 - AssertionError: Tensor-likes are not close!

Mismatched elements: 7563 / 180224 (4.2%)
Greatest absolute difference: 1.046875 at index (6, 0, 86, 73) (up to 1e-05 allowed)
Greatest relative difference: inf at index (0, 1, 22, 61) (up to 0.016 allowed)
= 2 failed, 4667 passed, 823 skipped, 108 xfailed, 96 xpassed, 119791 warnings in 594.94s (0:09:54) =

@vedaanta I tentatively assigned it to you because #691 is your PR...

tfogal:
triage review:
- could actually be a pytorch bug

IvanYashchuk:
@riccardofelluga, when you have time could you please check that the same backend is chosen in Thunder and PyTorch for the failing cases? Maybe the test inputs should be modified to have better-behaving values for numerical testing.

Recently the failure was observed in https://github.com/NVIDIA/Fuser/issues/3905.

riccardofelluga:
It looks like that there is no difference between the two tests. They both call the same sdpa function when given identical inputs. Moreover, removing the skip I cannot seem to be able to reproduce the flakiness

riccardofelluga:
Upon further inspection, out of 300 runs 1 of them resulted in a numerical mismatch. By capturing that input and feeding it back to the test however the mismatch does not appear either. It looks like that this is coming from some op being not deterministic. Since the only failed test was on the inputs with the `scale` kwarg, my recommendation here would be to change it to a number that can be casted to float16/bfloat16 instead of `0.123`.

[...]

PR #1767, opened 2025-02-16
Mark `test_vjp_correctness_sdpa_manual` with non-strict xfail instead of version check

IvanYashchuk:
Mark `test_vjp_correctness_sdpa_manual` with non-strict xfail instead of version check. The current nightly PyTorch version is 2.7 and the flakiness may re-emerge as seen in https://github.com/NVIDIA/Fuser/issues/3905.

The numerical flakiness of this test is tracked in https://github.com/Lightning-AI/lightning-thunder/issues/703.

cc @borda


t-vi (approved):
Thank you Ivan

Issue #1521, opened 2024-12-05
Repro function saved from FX graph is segmented again when passed back to `torch.compile`


kiya00:
*Note*: If you have a model or program that is not supported yet but should be, please use the program coverage template.

## 🐛 Bug
In the saved reproducer function, since it's already processed by dynamo, it's possible that there are some dynamo operators such as `torch.amp.autocast_mode._enter_autocast`. If we pass this repro function into torch.compile again, it creates graph-break on this kind of operator, which can result in slower perf for torch.compile case.

### To Reproduce

```
import torch
from thunder.dynamo import ThunderCompiler

def test_autocast():
    x = torch.rand(2, 2, device="cuda", requires_grad=True)

    backend = ThunderCompiler()

    def func(x):
        x = x + 2
        with torch.autocast("cuda"):
            y = torch.sin(x)
            return y

    cfunc = torch.compile(func, backend=backend)
    actual = cfunc(x)
    backend.save_reproducer_to_folder("repro")
```
Run the above script it produces repro function as follows, we modifies it to print out the graph break information:
```
import torch
def test_graph0_thunder_0():
    class DynamoModule(torch.nn.Module):
      def forward(self, l_x_ : torch.Tensor):
          x = l_x_ + 2;  l_x_ = None
          _enter_autocast = torch.amp.autocast_mode._enter_autocast('cuda', None, True, None)
          y = torch.sin(x);  x = None
          _exit_autocast = torch.amp.autocast_mode._exit_autocast(_enter_autocast);  _enter_autocast = _exit_autocast = None
          return y

    inputs = [
        torch.testing.make_tensor((2, 2), dtype=torch.float32,  device='cuda:0', requires_grad=True, low=0.32686540484428406, high=0.8776439428329468,).as_strided((2, 2), (2, 1)),

    ]

    mod = DynamoModule()
    exp = torch._dynamo.explain(mod)(*inputs)
    print(exp)
```

```
Graph Count: 2
Graph Break Count: 1
Op Count: 2
Break Reasons:
  Break Reason 1:
    Reason: torch.* op returned non-Tensor autocast call_function <function _enter_autocast at 0x76ad30d6d8a0>
    User Stack:
      <FrameSummary file /wayan/lightning-thunder/tmp.py, line 30 in forward>
  Break Reason 2:
    Reason: call_function args: AutocastModeVariable()
    User Stack:
[...]

PR #1773, opened 2025-02-17
[ThunderFX report] Adds `thunderfx_benchmark_report` to do some automatic performance analysis on the FX graph

kiya00:
<details>

</details>

## What does this PR do?

This PR:
1. Adds `thunderfx_benchmark_report` function that uses the previously implemented reporting APIs to analyze the the input callable, see the function comment for details
<details>
<summary>example outputs</summary>

```
The input callable can be successfully executed by torch.compile.

graph0: Split Information:
The original graph is not split, and is entirely run by Thunder.


graph0_thunder_0 can be successfully executed by Thunder

Benchmark walltime for **graph0_thunder_0 forward** requires investigation: thunder(132.469 us) and torchcompile(28.990 us) is not close (rtol=0.1, atol=0.0)
Benchmark walltime for **graph0_thunder_0 backward** requires investigation: thunder(98.921 us) and torchcompile(34.108 us) is not close (rtol=0.1, atol=0.0)
The scripts are saved: graph0_thunder_0_torchcompile_walltime.py, graph0_thunder_0_thunder_walltime.py


Benchmark kerneltime ran successfully on **graph0_thunder_0 forward**: thunder(955.523 ns) , torchcompile(1.002 us)
Benchmark kerneltime ran successfully on **graph0_thunder_0 backward**: thunder(965.957 ns) , torchcompile(1.012 us)


Benchmark walltime for **graph0_thunder_0_nvFusion0_forward** requires investigation: nvfuser(22.932 us) and torchcompile(20.361 us) is not close (rtol=0.1, atol=0.0)
The scripts are saved: graph0_thunder_0_nvFusion0_forward_torchcompile_walltime.py, graph0_thunder_0_nvFusion0_forward_nvfuser_walltime.py


Benchmark kerneltime ran successfully on **graph0_thunder_0_nvFusion0_forward**: nvfuser(952.301 ns) , torchcompile(998.566 ns)


Benchmark walltime for **graph0_thunder_0_nvFusion0_backward** requires investigation: nvfuser(26.004 us) and torchcompile(21.240 us) is not close (rtol=0.1, atol=0.0)
The scripts are saved: graph0_thunder_0_nvFusion0_backward_torchcompile_walltime.py, graph0_thunder_0_nvFusion0_backward_nvfuser_walltime.py


Benchmark kerneltime ran successfully on **graph0_thunder_0_nvFusion0_backward**: nvfuser(962.118 ns) , torchcompile(1.007 us)

```

</details>

2. Adds `TorchInductorSpecification` which uses only TorchInductor without TorchDynamo, the reason is mentioned here https://github.com/Lightning-AI/lightning-thunder/issues/1521
3. Removes the previous write_repro interface, keeping only the one that takes "CompileSpecification" and timer function as inputs
4. Makes some revisions according to the previous review comments
5. Adds the WAR for Triton error:
  <details>
  <summary>repro script</summary>
  
  ```
  import torch
  
  def test_graph0():
      class DynamoModule(torch.nn.Module):
          def forward(self, L_x_ : torch.Tensor):
              l_x_ = L_x_
              x = l_x_.exp();  l_x_ = None
              return (x,)
          
  
      inputs = [
          torch.testing.make_tensor((2, 2), dtype=torch.float32,  device='cuda:0', requires_grad=True),
      ]
  
      model = DynamoModule()
      compiled_model = torch.compile(model, )
      from thunder.dynamo.report import run_forward_backward
  
      result = compiled_model(*inputs)
  
      output_grads = [torch.ones_like(r) for r in result]
  
      torch.autograd.backward(result, output_grads)
      print(result)
  
  # the WAR: https://github.com/pytorch/pytorch/issues/124565
  # torch.empty(1, device='cuda', requires_grad=True).backward()
  test_graph0()
  ```
  
  <\details>
## PR review

## Did you have fun?

### DIFF
--- a/thunder/dynamo/benchmark_utils.py
+++ b/thunder/dynamo/benchmark_utils.py
@@ -105,7 +105,9 @@ class TorchEagerSpecification(CompileSpecificationInterface):
 
 class TorchInductorSpecification(CompileSpecificationInterface):
     """
-    A compile specification for :func:`torch.compile`.
+    A compilation specification for using TorchInductor without TorchDynamo.
+    For details on why compilation without TorchDynamo is needed, see:
+    https://github.com/Lightning-AI/lightning-thunder/issues/1521
     """
 
     def __init__(self, inputs, specification_name="inductor_backend"):
--- a/thunder/dynamo/benchmark_utils.py
+++ b/thunder/dynamo/benchmark_utils.py
@@ -131,7 +131,7 @@ class TorchInductorSpecification(CompileSpecificationInterface):
 
 
 class BoundSymbolNvfuserSpecification(CompileSpecificationInterface):
-    def __init__(self, specification_name="nvfusersymbol_nvfuser"):
+    def __init__(self, specification_name="nvfuser"):
         self.name: str = specification_name
 
     # Returns the nvFuser callable from the nvFuser bound symbol.
@@ -141,7 +141,7 @@ class BoundSymbolNvfuserSpecification(CompileSpecificationInterface):
 
 
 class BoundSymbolTorchCompileSpecification(CompileSpecificationInterface):
-    def __init__(self, specification_name="nvfusersymbol_torchcompile"):
+    def __init__(self, specification_name="torchcompile"):
         self.name: str = specification_name
 
     # Returns the torch compile callable from the nvFuser bound symbol.
--- a/thunder/dynamo/benchmark_utils.py
+++ b/thunder/dynamo/benchmark_utils.py
@@ -48,7 +48,8 @@ class ThunderCompileSpecification(CompileSpecificationInterface):
         compiled_fn = spec.compile(my_function)
     """
 
-    def __init__(self, **kwargs):
+    def __init__(self, specification_name="thunder", **kwargs):
+        self.name = specification_name
         self.thunder_options: dict = kwargs
 
     def compile(self, fn):
@@ -69,7 +70,8 @@ class TorchCompileSpecification(CompileSpecificationInterface):
     A compile specification for :func:`torch.compile`.
     """
 
-    def __init__(self, **kwargs):
+    def __init__(self, specification_name="torchcompile", **kwargs):
+        self.name = specification_name
         self.torch_compile_options: dict = kwargs
 
     def compile(self, fn):
@@ -91,6 +93,9 @@ class TorchEagerSpecification(CompileSpecificationInterface):
     A compile specification for Torch eager mode, which returns the callable unchanged.
     """
 
+    def __init__(self, specification_name="torcheager"):
+        self.name = specification_name
+
     def compile(self, fn):
         return fn
 
@@ -103,7 +108,8 @@ class TorchInductorSpecification(CompileSpecificationInterface):
     A compile specification for :func:`torch.compile`.
     """
 
-    def __init__(self, inputs):
+    def __init__(self, inputs, specification_name="inductor_backend"):
+        self.name: str = specification_name
         self.inputs: list = inputs
 
     @staticmethod
@@ -125,6 +131,9 @@ class TorchInductorSpecification(CompileSpecificationInterface):
 
 
 class BoundSymbolNvfuserSpecification(CompileSpecificationInterface):
+    def __init__(self, specification_name="nvfusersymbol_nvfuser"):
+        self.name: str = specification_name
+
     # Returns the nvFuser callable from the nvFuser bound symbol.
     # See the TODO in :class:`thunder.dynamo.report.ThunderFusionReport` for more details.
     def compile(self, nvfusion_bsym):
@@ -132,6 +141,9 @@ class BoundSymbolNvfuserSpecification(CompileSpecificationInterface):
 
 
 class BoundSymbolTorchCompileSpecification(CompileSpecificationInterface):
+    def __init__(self, specification_name="nvfusersymbol_torchcompile"):
+        self.name: str = specification_name
+
     # Returns the torch compile callable from the nvFuser bound symbol.
     # See the TODO in :class:`thunder.dynamo.report.ThunderFusionReport` for more details.
     def compile(self, bsym):
--- a/thunder/dynamo/benchmark_utils.py
+++ b/thunder/dynamo/benchmark_utils.py
@@ -247,7 +247,7 @@ class WallTime(TimerInterface):
         return ["from thunder.dynamo.benchmark_utils import WallTime"]
 
     @staticmethod
-    def to_source(fn_name="compiled_model", inputs_name="inputs"):
+    def to_source(fn_name, inputs_name):
         return f'WallTime.time("{fn_name}(*{inputs_name})", globals={{"{fn_name}":{fn_name}, "{inputs_name}": {inputs_name}}})'
 
 
@@ -271,5 +271,5 @@ class KernelTime(TimerInterface):
         return ["from thunder.dynamo.benchmark_utils import KernelTime"]
 
     @staticmethod
-    def to_source(fn_name="compiled_model", inputs_name="inputs"):
+    def to_source(fn_name, inputs_name):
         return f'KernelTime.time("{fn_name}(*{inputs_name})", globals={{"{fn_name}":{fn_name}, "{inputs_name}": {inputs_name}}})'
--- a/thunder/dynamo/benchmark_utils.py
+++ b/thunder/dynamo/benchmark_utils.py
@@ -80,7 +80,10 @@ class TorchCompileSpecification(CompileSpecificationInterface):
         return f"torch.compile({fn_name}, {kwargs_str})"
 
     def import_str(self):
-        return ["import torch"]
+        # WAR for triton error https://github.com/pytorch/pytorch/issues/124565
+        comment_str = '# Workaround for "RuntimeError: Triton Error [CUDA]: an illegal memory access was encountered"\n# https://github.com/pytorch/pytorch/issues/124565'
+        code = f'if torch.cuda.is_available:\n    torch.empty(1, device="cuda", requires_grad=True).backward()'
+        return ["import torch", comment_str, code]
 
 
 class TorchEagerSpecification(CompileSpecificationInterface):
@@ -93,6 +96,32 @@ class TorchEagerSpecification(CompileSpecificationInterface):
 
     def to_source(self, fn_name):
         return fn_name
+
+
+class TorchInductorSpecification(CompileSpecificationInterface):
+    """
+    A compile specification for :func:`torch.compile`.
+    """
+
+    def __init__(self, inputs):
+        self.inputs: list = inputs
+
+    @staticmethod
+    def torch_inductor(fn, inputs):
+        from torch._inductor import compile as inductor_compile
+        from torch.fx import symbolic_trace
+
+        fx_graph = symbolic_trace(fn)
+        return inductor_compile(fx_graph, inputs)
+
+    def compile(self, fn):
+        return self.torch_inductor(fn, self.inputs)
+
+    def to_source(self, fn_name):
+        return f"TorchInductorSpecification.torch_inductor({fn_name}, inputs)"
+
+    def import_str(self):
+        return ["import torch", "from thunder.dynamo.benchmark_utils import TorchInductorSpecification"]
 
 
 class BoundSymbolNvfuserSpecification(CompileSpecificationInterface):
--- a/thunder/dynamo/report.py
+++ b/thunder/dynamo/report.py
@@ -9,6 +9,7 @@ from typing import TYPE_CHECKING
 import textwrap
 
 import torch
+from torch.utils.benchmark.utils.common import select_unit as select_time_unit
 from thunder.core.pytree import tree_flatten
 from thunder.core.utils import sequencify, create_python_callable_from_bsym
 from thunder.dynamo.compiler import thunderfx
@@ -61,6 +62,8 @@ def run_forward_backward(fn, *args, **kwargs):
 
     forward_inputs = tree_flatten((args, kwargs))[0]
     forward_inputs = list(filter(lambda x: isinstance(x, torch.Tensor) and x.requires_grad, forward_inputs))
+    if not forward_inputs:
+        return result, None
     differentiable_tensor_result = list(filter(lambda x: isinstance(x, torch.Tensor) and x.requires_grad, result))
 
     output_grads = []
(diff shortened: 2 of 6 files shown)


thunder/dynamo/benchmark_utils.py
```diff
@@ -80,29 +82,70 @@ def to_source(self, fn_name):
         return f"torch.compile({fn_name}, {kwargs_str})"
 
     def import_str(self):
-        return ["import torch"]
+        # WAR for triton error https://github.com/pytorch/pytorch/issues/124565
+        comment_str = '# Workaround for "RuntimeError: Triton Error [CUDA]: an illegal memory access was encountered"\n# https://github.com/pytorch/pytorch/issues/124565'
```
mruberry, line 86: Couple nits about this:

- What's the expected return from `import_str` now? 
- Why not one string for the comment and the code in the returned list?

thunder/dynamo/benchmark_utils.py
```diff
@@ -80,29 +82,70 @@ def to_source(self, fn_name):
...
     def compile(self, fn):
         return fn
 
     def to_source(self, fn_name):
         return fn_name
 
 
+class TorchInductorSpecification(CompileSpecificationInterface):
+    """
+    A compilation specification for using TorchInductor without TorchDynamo.
```
mruberry, line 108: Cool!

thunder/dynamo/benchmark_utils.py
mruberry: Reading more -- I think this makes sense. This will put the comment and the tensor initialization among the imports, right? That seems fine.

thunder/dynamo/utils.py
```diff
@@ -100,6 +100,14 @@ class ExampleInputMetaData:
     def stride(self) -> list[int]:
         return self.strides
 
+    def is_contiguous(self) -> bool:
```
mruberry, line 103: Instead of computing this, why not initialize this class with the contiguity information obtained from `torch.Tensor.is_contiguous`?

thunder/dynamo/utils.py
```diff
@@ -478,9 +486,11 @@ def _get_example_input_tensor_metadata(t: torch.Tensor) -> ExampleInputMetaData:
...
     from thunder.tests.make_tensor import make_tensor
 
-    return make_tensor(t.storage_shape, dtype=t.dtype, device=t.device, requires_grad=t.requires_grad).as_strided(
-        t.shape, t.stride()
-    )
+    torch_tensor = make_tensor(t.storage_shape, dtype=t.dtype, device=t.device, requires_grad=t.requires_grad)
+    if t.is_contiguous():
+        return torch_tensor
+
+    return torch_tensor.as_strided(t.shape, t.stride())
```
mruberry, line 493: Nice detailed improvement!

Maybe we should have allowed specifying strides as part of the `make_tensor` UX. Oh well.

Two suggestions:

- Why not always call this instead of checking for contiguity?
- Instead of calling `as_strided`, which will create a view of the tensor, `set_` can modify the strides inplace: https://pytorch.org/docs/stable/generated/torch.Tensor.set_.html#torch-tensor-set

So maybe 

```
torch_tensor = make_tensor(...)
torch.tensor.set_(stride=t.stride())
```

Would you also add a TODO that we should consider memory alignment in the future. If the original tensor was a view, for example, its storage offset may not be zero, and it's possible that the first element of a tensor with small elements may not be aligned to a multiple of 32

thunder/dynamo/utils.py
```diff
@@ -613,24 +623,44 @@ def remove_empty_autocast(graph_module: torch.fx.GraphModule) -> torch.fx.GraphM
     return empty_autocast_removed_graph_module
 
 
-def arg_like_tensor(arg: torch.Tensor | ExampleInputMetaData):
-    """Creates a new argument like the given tensor or tensor metadata"""
+def _get_example_input_tensor_metadata(t: torch.Tensor) -> ExampleInputMetaData:
     min_val = None
     max_val = None
+    if not isinstance(t, FakeTensor) and t.device.type != "meta" and t.numel() != 0:
```
mruberry, line 629: Nice min-max extraction

thunder/tests/test_dynamo.py
```diff
@@ -1056,6 +1056,12 @@ def foo(x):
     assert out.device.type == "meta"
 
 
+def run_script(file_name, cmd):
```
mruberry, line 1059: Nit: comment for this function would be nice

thunder/dynamo/report.py
```diff
@@ -1042,3 +980,153 @@ def foo(x):
...
+    folder_path: str | PathLike,
+    thunder_compile_kwargs: dict = None,
+    compare_fusion: bool = False,
+    print_traces: bool = False,
+    rtol=0.5,
+    atol=0.0,
+    **kwargs,
+):
+    """
+    A utility function that analyzes the runnability and performance benchmarks of each FX graph.
```
mruberry, line 997: Wahoo!

thunder/dynamo/report.py
```diff
@@ -1042,3 +980,153 @@ def foo(x):
...
+
+
+def thunderfx_benchmark_report(
+    fn: Callable,
+    *args,
+    folder_path: str | PathLike,
+    thunder_compile_kwargs: dict = None,
+    compare_fusion: bool = False,
+    print_traces: bool = False,
+    rtol=0.5,
```
mruberry, line 992: This is great; we can work out the defaults later.

One thing we should be clear on, however, is the units for atol. Is `atol=1` an absolute tolerance of 1 second, or 1 microsecond...?

thunder/dynamo/report.py
```diff
@@ -1042,3 +980,153 @@ def foo(x):
...
+    - If it fails, attempts to run it eagerly.
+    - If eager execution also fails, an error is printed, and the function returns.
+    - If eager execution succeeds, the analysis continues.
+
+    2. Collects all ThunderFX subgraphs and verifies whether each subgraph can be successfully executed by Thunder.
+
+    3. For each subgraph:
+    - Compares wall time and kernel time between `torch.compile` and Thunder.
+    - Reports performance metrics and saves the benchmark script in `folder_path` if the difference exceeds the tolerance (`rtol`, `atol`).
+    - Uses `math.isclose` for tolerance checks.
```
mruberry, line 1009: Excellent choice and great documentation

thunder/dynamo/report.py
```diff
@@ -1042,3 +980,153 @@ def foo(x):
...
+    def check_timing_bsym(report, compile_fn1, compile_fn2, timer_fn, timer_name: str):
+        graph_name = report.name
+        measure1 = report.run_benchmark(compile_fn1, timer_fn)
+        measure2 = report.run_benchmark(compile_fn2, timer_fn)
+
+        ret = check_threshold_log(
+            measure1.median, measure2.median, compile_fn1.name, compile_fn2.name, graph_name, timer_name
+        )
+        if not ret[0]:
+            extra_comment = f"Benchmark results:\n{ret[1]}\n"
```
mruberry, line 1092: Extra comments are cool!

thunder/dynamo/report.py
```diff
@@ -1042,3 +980,153 @@ def foo(x):
...
+
+    def check_threshold_log(a: float, b: float, a_name: str, b_name: str, test_name: str, timer_name: str):
+        a_time_str = get_pretty_time_str(a)
+        b_time_str = get_pretty_time_str(b)
+        if not check_threshold(a, b, rtol=rtol, atol=atol):
+            log_str = f"Benchmark {timer_name} for **{test_name}** requires investigation: {b_name}({b_time_str}) and {a_name}({a_time_str}) is not close (rtol={rtol}, atol={atol})"
+            print(log_str)
+            return False, log_str
+        log_str = f"Benchmark {timer_name} ran successfully on **{test_name}**: {b_name}({b_time_str}) , {a_name}({a_time_str})"
+        print(log_str)
```
mruberry, line 1050: Looks like there's still some debug printing happening here, which is totally fine because this PR is in draft

thunder/dynamo/report.py
```diff
@@ -1042,3 +980,153 @@ def foo(x):
...
+
+    results = fx_report(fn, *args, compile_options=thunder_compile_kwargs, **kwargs)
+    if thunder_compile_kwargs is None:
+        thunder_compile_kwargs = {}
+    thunderjit_specification = ThunderCompileSpecification(**thunder_compile_kwargs)
+    torchcompile = TorchCompileSpecification()
+    for idx, fx_graph_report in enumerate(results.fx_graph_reports):
+        thunder_fx_graph_report = analyze_thunder_splits(fx_graph_report, **thunder_compile_kwargs)
+        print(f"\n{thunder_fx_graph_report.graph_name}: {thunder_fx_graph_report.split_reason}\n")
+        for split_report in thunder_fx_graph_report.subgraph_reports:
```
mruberry, line 1108: Very cool; definitely some exciting TODOs here for analyzing the segmentation information and summarizing why segmentation occurred.

A separate TODO is probably making information on the operators used programmatically available. Like being able to query if `torch.add` is in an fx graph, so we can filter for graphs that contain particular calls. And information about operations not executed by thunder, and operations executed by thunder but using "fallback" logic.

mruberry (commented):
Exciting stuff! I made a few small suggestions for your review, @kiya00

thunder/dynamo/report.py
kiya00: it's in second, follows the torch benchmark's Measurement (uses timeit.timeit)

thunder/dynamo/benchmark_utils.py
kiya00: >This will put the comment and the tensor initialization among the imports, right?

Ah, yes, here borrows the `import_str` to insert the  WAR part

thunder/dynamo/report.py
kiya00: This is printed when calling `thunderfx_benchmark_report`, like:
```
Benchmark walltime for **graph0_thunder_0 forward** requires investigation: thunder(132.469 us) and torchcompile(28.990 us) is not close (rtol=0.1, atol=0.0)
Benchmark walltime for **graph0_thunder_0 backward** requires investigation: thunder(98.921 us) and torchcompile(34.108 us) is not close (rtol=0.1, atol=0.0)
The scripts are saved: graph0_thunder_0_torchcompile_walltime.py, graph0_thunder_0_thunder_walltime.py


Benchmark kerneltime ran successfully on **graph0_thunder_0 forward**: thunder(955.523 ns) , torchcompile(1.002 us)
Benchmark kerneltime ran successfully on **graph0_thunder_0 backward**: thunder(965.957 ns) , torchcompile(1.012 us)

```
and the information like `Benchmark walltime for **graph0_thunder_0 backward** requires investigation: thunder(98.921 us) and torchcompile(34.108 us) is not close (rtol=0.1, atol=0.0)` is also saved in the benchmark script as extra comments

thunder/dynamo/utils.py
kiya00: Thanks for the suggestions, I changed as_strided to set_ and added the storage_offset

thunder/dynamo/report.py
mruberry: Sounds great!

kiya00:
Hi @mruberry , I met the problem that the original model runs successfully but the current report interface runs OOM on the model (that's a problem, we should be able to run the same data size with the report right?), because the report looks like:
```
def report(model, inputs...):
    try to run torch.compile
    # loop subgraphs
    reports = fx_report(model)
    for sub_report in reports.sub_reports:
        thunder_report = analyse_thunder_splits(sub_report)
        for split_report in thunder_report.sub_reports:
            ###this part creates new random input tensors, it causes OOM, needs to free the original model and inputs to get enough space
            split_report.run_benchmark(torchcompile)  
            split_report.run_benchmark(thunder)
            compare and save the benchmark script if necessary
```
but it's hard to free the model and inputs inside the function. I'm thinking of providing some utility function and let the user to free the model and inputs between the `get_subgraph_reports` and `performance_analyse`, like:
```
def get_subgraph_reports(model, inputs,...):
    report_list = []
    reports = fx_report(model)
    ...
    return report_list

# user code:
model = GPT(config)
inputs = make_tensor(...)
reports = get_subgraph_reports(model, inputs)
del model, input
performance_analyse(reports)
```

mruberry:
> ```
> def report(model, inputs...):
>     try to run torch.compile
>     # loop subgraphs
>     reports = fx_report(model)
>     for sub_report in reports.sub_reports:
>         thunder_report = analyse_thunder_splits(sub_report)
>         for split_report in thunder_report.sub_reports:
>             ###this part creates new random input tensors, it causes OOM, needs to free the original model and inputs to get enough space
>             split_report.run_benchmark(torchcompile)  
>             split_report.run_benchmark(thunder)
>             compare and save the benchmark script if necessary
> ```
> 
> but it's hard to free the model and inputs inside the function. I'm thinking of providing some utility function and let the user to free the model and inputs between the `get_subgraph_reports` and `performance_analyse`, like:
> 
> ```
> def get_subgraph_reports(model, inputs,...):
>     report_list = []
>     reports = fx_report(model)
>     ...
>     return report_list
> 
> # user code:
> model = GPT(config)
> inputs = make_tensor(...)
> reports = get_subgraph_reports(model, inputs)
> del model, input
> performance_analyse(reports)
> ```
> 
> WDYT?

It makes a lot of sense that we want to get rid of the model and the inputs before starting our analysis, since each report can synthetically create the appropriate inputs and be run in isolation. We can also implement this pattern ourselves in the UX we expose to generate the reports.

Why a helper function vs. just documenting that if OOM is encountered the practitioner should try freeing other cuda tensors?

kiya00:
Hi @mruberry , it's ready to review, I keep the `thunderfx_benchmark_report` for easy use and give comments about if it goes OOM user needs to delete the model and inputs manually
https://github.com/Lightning-AI/lightning-thunder/blob/ea56d4e87645f1c45c96de24e94e806efae0aea3/thunder/dynamo/report.py#L1055-L1076

thunder/dynamo/benchmark_utils.py
```diff
@@ -48,7 +48,8 @@ class ThunderCompileSpecification(CompileSpecificationInterface):
         compiled_fn = spec.compile(my_function)
     """
 
-    def __init__(self, **kwargs):
+    def __init__(self, specification_name="thunder", **kwargs):
```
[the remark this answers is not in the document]
kiya00: It might be possible that we check different executors like with sdpa, or with CUDAgraph transform, then here can have names like "thunder_sdpa", "thunder_cudagraph", and it can be used as a suffix of the saved script so they don't overwrite each other

mruberry (approved):
Improvements look great! While there will continue to be refinements and features, let's start using this and get customer feedback.

While getting customer feedback there are still a variety of extensions that could be implemented. What are your thoughts, @kiya00? Would it be interesting to look at measure memory usage, or maybe producing statistics about what operators are used and how they're executed, or would you like to start working on creating more isolated reproductions of correctness issues (and then performance issues), or summarize thunder split reasons?

I think the most interesting would be the creation of more isolated reproductions, but they're all good directions to go in.

PR #1780, opened 2025-02-18
Introduce update_aliases

beverlylytle:
In an abstract sense, Thunder views pytorch programs as a directed acyclic graph of nodes representing elemental functions with edges between those nodes representing composition of functions.  Execution of the program is then can be viewed as the waterfall execution of the nodes.  If it is assumed that the execution of each node has no side-effects, then executing the nodes in the order coming from any topological sorting will be valid.  However, in-place operations do have side-effects, and reordering the execution of such nodes may yield incorrect results.  These side-effects may be propagated further than expected by views of tensors.  This PR introduces a new elemental function `update_aliases` which introduces new edges into the DAG.   When inserted judiciously, these new edges will prevent the reordering of the execution of nodes involving tensors which have been affected by in-place operations.  The in-place operations are left... in place.  Because of this and nvfuser's current approach to in-place operations, new fusion breaks are introduced.  

Currently, this change is hidden behind the compile option flag `skip_inplace_alias_updates` which defaults to `True`.  It has only been tested with `skip_inplace_functionalization=True` as the resolution of the two approaches has not been fleshed out.  Only the forward execution is tested here, but I have hopes (and some experimental evidence) that the backward execution largely works as well.


thunder/core/transforms.py
```diff
@@ -2299,6 +2312,19 @@ def iter_bound_symbols(bound_symbols):
             yield symbol
 
 
+def get_first_proxy(proxies) -> Proxy | None:
```
IvanYashchuk, line 2315: So we need to merge https://github.com/Lightning-AI/lightning-thunder/pull/1755 first. I'll check if it's easy to pull it out from the current PR stack.

thunder/extend/__init__.py
```diff
@@ -102,6 +102,12 @@ def can_execute(self, bsym: BoundSymbol) -> bool:
     def can_fuse(self, bsym: BoundSymbol) -> bool:
         sym = bsym.sym
 
+        # !!! this is probably horribly wrong, please help
```
IvanYashchuk, line 105: One example where we need to fuse copies is efficient batch_norm https://github.com/Lightning-AI/lightning-thunder/blob/672886db0581f895b6c3791053aea4e92cc827ce/thunder/benchmarks/targets.py#L265

What problems did you see?

thunder/core/transforms.py
```diff
@@ -756,6 +756,17 @@ def _cat_prim_grad(tensors: list[TensorProxy], /, dim: int) -> TensorProxy:
 register_grad(pids.CAT, _cat_prim_grad)
 
 
+def _update_aliases_prim_grad(tensors: tuple[TensorProxy]) -> tuple[TensorProxy]:
```
crcrpar, line 759: nitpicking

```suggestion
def _update_aliases_prim_grad(tensors: tuple[TensorProxy, ...]) -> tuple[TensorProxy, ...]:
```

thunder/core/update_aliases.py
```diff
@@ -0,0 +1,107 @@
...
+
+def _is_inplace_op(bsym):
+    return bsym.sym.tags and prims.OpTags.IN_PLACE in bsym.sym.tags
+
+
+def _is_view_creation_op(bsym):
+    import thunder.torch as ltorch
+
+    # !!! check on _syms_that_may_return_views?
+    return bsym.sym in ltorch._syms_returning_views
```
crcrpar, line 40: I guess this might be orthogonal to this PR, but what's the approach we're to take for the ops listed in `_syms_that_may_return_views`?

thunder/core/update_aliases.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+def _involves_viewed_args(bsym, viewed):
+    if bsym.sym.id == prims.PrimIDs.RETURN:
+        return False
+    return any(isinstance(p, TensorProxy) and variableify(p) in viewed for p in bsym.flat_proxy_args)
+
+
+def insert_alias_updates(computation_trace: Trace) -> list[Trace]:
```
crcrpar, line 49: nitpicking

```suggestion
def insert_alias_updates(computation_trace: Trace) -> Trace:
```

thunder/core/update_aliases.py
```diff
@@ -0,0 +1,107 @@
...
+    swap_map = dict()
+    bsyms = []
+
+    # First pass: identify views, their originals, and operands involved in inplace ops
+    view_groups = []
+    inplace_inputs = set()
+    for bsym in computation_trace.bound_symbols:
+        if _is_inplace_op(bsym) or _is_view_creation_op(bsym):
+            # only interested in the input which is modified by the inplace op
+            in_tensor = list(map(variableify, filter(lambda p: isinstance(p, TensorProxy), bsym.flat_proxy_args)))[0]
```
crcrpar, line 59: In my gut feeling for most ops this would be equivalent to `variableify(bsym.flat_proxy_args[0]`. Correct me if I'm wrong.
I guess this could be simpler if we have a map from op to index of tensor arg to be modified.
What do you think?

thunder/core/update_aliases.py
```diff
@@ -0,0 +1,107 @@
...
+            if _is_inplace_op(bsym):
+                in_tensors = {list(in_tensors)[0]}
+            else:
+                in_tensors = set(in_tensors)
+            out_tensors = set(map(variableify, filter(lambda p: isinstance(p, TensorProxy), bsym.flat_proxy_outs)))
+            encountered.update(in_tensors)
+            group = next((g for g in view_groups if any(g.intersection(in_tensors))), None)
+            if group is None:
+                # this is a view creation with operands that are not involved in any inplace ops
+                bsyms.append(bsym.from_bsym_swap_proxies(swap_map))
```
crcrpar, line 87: out of curiosity, would bsym get its outputs updated here?

thunder/core/update_aliases.py
```diff
@@ -0,0 +1,107 @@
...
+            views_encountered = group.intersection(encountered)
+            new_aliases = _get_new_aliases(views_encountered, computation_trace)
+
+            update_bsym, swap_map = _get_update_bsym(views_encountered, swap_map, new_aliases)
+            bsyms.append(update_bsym)
+            encountered.update(out_tensors)
+            new_bsym = bsym.from_bsym_swap_proxies(swap_map)
+            if _is_inplace_op(bsym):
+                #  This relies on these being one element sets
+                swap_map[in_tensors.pop()] = unvariableify(out_tensors.pop())
```
crcrpar, line 98: I'm reading your changes from top, so far I'm not quote following the point of having one `in_tensor` as a set of one element. Would there be some (future) use cases or extensions of it?

crcrpar (commented):
Most comments are questions to help me understand the PR, itself. Generally looks nice :)

thunder/core/update_aliases.py
beverlylytle: Yes, you are right about the equivalence to `variableify(bsym.flat_proxy_args[0]`.

This does assume that the first argument to an in-place op or a view creation op is the thing needing to be aliased.  If this is the case, then I think what's done here is simplest.  If it is not the case, then having such a map would be the way to go.  I think having such a map pushes the complexity to another place, and makes adding operators more difficult.

thunder/core/update_aliases.py
beverlylytle: No, the output of this as-of-yet untouched bsym would not be present in the swap_map.  The swap_map only has old aliases and the input tensors of in-place operations as keys.

thunder/core/update_aliases.py
beverlylytle: In the case that the bsym has views as arguments, in_tensors may have more than one element.  For example:
```
def f(x, y):
    a = x.view(-1)
    b = a.mul(y)
    return b
```
Then the bsym for mul would have a and y as in_tensors, and because a is a view, it will pass through the update_alias machine.  But in the line you reference, the assumption is that the bsym is an in-place op and so its input and its output should each have only one element.

thunder/extend/__init__.py
beverlylytle: Consider the nearly trivial
```
def foo(x, y):
   z = x + y
   z.add_(1)
   return z
```

If I restore the fusibility of copy_,  then I encounter the following:
```
An error occurred while defining nvFuser FusionDefinition None.
If you believe this is a bug or need assistance, please file an issue at https://github.com/NVIDIA/Fuser/issues/new
Here's a script to reproduce the error:
```python
# CUDA devices:
#  0: NVIDIA RTX 6000 Ada Generation
# torch version: 2.5.1+cu124
# cuda version: 12.4
# nvfuser version: 0.2.23+gitd53be45
import torch

def nvfuser_incomplete_fusion(fd : FusionDefinition) -> None :
    T0 = fd.define_tensor(shape=[4, 10], contiguity=[True, True], dtype=DataType.Float, is_cpu=False, stride_order=[1, 0])
    T1 = fd.define_tensor(shape=[4, 10], contiguity=[True, True], dtype=DataType.Float, is_cpu=False, stride_order=[1, 0])
    T2 = fd.ops.add(T0, T1)
    T3 = fd.ops.set(T2)
    S4 = fd.define_scalar(1.00000, dtype=DataType.Double)
    T5 = fd.ops.add(T3, S4)
    T6 = fd.ops.set(T5)
    fd.add_output(T6, T3)
    fd.add_output(T3)

    nvfuser_fusion_idNone(fd)

  File "/home/blytle/miniforge3/envs/thdrs/lib/python3.10/site-packages/nvfuser/__init__.py", line 105, in __exit__
    self._finalize_definition()
RuntimeError:  INTERNAL ASSERT FAILED at "/workspace/Fuser/csrc/python_frontend/fusion_state.cpp":141, please report a bug with repro script to NVFuser at https://github.com/NVIDIA/Fuser/issues. 
Detected exception while building Fusion Ir. The failing RecordFunctor is: fd.add_output(T6, T3)
NvFuser error message is:  INTERNAL ASSERT FAILED at "/workspace/Fuser/csrc/fusion.cpp":784, please report a bug with repro script to NVFuser at https://github.com/NVIDIA/Fuser/issues. alias source can only be a fusion input
```

Outside of this function, it's invisible whether the `add` is in-place or not.  And that's how Thunder handles it currently, the in-place-ness disappears:
```
# CUDA devices:
#  0: NVIDIA RTX 6000 Ada Generation
# torch version: 2.5.1+cu124
# cuda version: 12.4
# nvfuser version: 0.2.23+gitd53be45
import torch

def nvfuser_fusion_id0(fd : FusionDefinition) -> None :
    T0 = fd.define_tensor(shape=[4, 10], contiguity=[True, True], dtype=DataType.Float, is_cpu=False, stride_order=[1, 0])
    T1 = fd.define_tensor(shape=[4, 10], contiguity=[True, True], dtype=DataType.Float, is_cpu=False, stride_order=[1, 0])
    T2 = fd.ops.add(T0, T1)
    S3 = fd.define_scalar(1.00000, dtype=DataType.Double)
    T4 = fd.ops.add(T2, S3)
    fd.add_output(T4)

    nvfuser_fusion_id0(fd)
```
So the approach of "preserving the in-place operations" doesn't work without some breaking up of the fusion regions. Breaking at every instance of copy_ is too severe though.   If there is a chain of in-place operations (eg `x.add(1).mul_(5)`), it would be terribly inefficient and this where the existing in-place functionalization excels.

beverlylytle: In the even simpler example
```
def foo(x):
   x.add_(1)
   return x
```
transforming this to
```
def foo(x):
    x_alias = update_aliases((x,))
    x_alias.add_(1)
    return x_alias
```
is already problematic with nvfuser without introducing more fusion breaks because x_alias is not input to the fusion.:
```
def nvfuser_incomplete_fusion(fd : FusionDefinition) -> None :
    T0 = fd.define_tensor(shape=[4, 10], contiguity=[True, True], dtype=DataType.Float, is_cpu=False, stride_order=[1, 0])
    T1 = fd.ops.set(T0)
    S2 = fd.define_scalar(1.00000, dtype=DataType.Double)
    T3 = fd.ops.add(T1, S2)
    T4 = fd.ops.set(T3)
    fd.add_output(T4, T1)
    fd.add_output(T1)
```

thunder/core/update_aliases.py
crcrpar: > This does assume that the first argument to an in-place op or a view creation op is the thing needing to be aliased. If this is the case, then I think what's done here is simplest.

Agreed. Okay let's keep it.

some clarification: The reason I asked is that we already have a map which is tad similar to what I had in my mind -- https://github.com/Lightning-AI/lightning-thunder/blob/92a6adb63ddcf36e3910a55fc5f7a02cec5bee8c/thunder/torch/__init__.py#L93-L95. So I thought this line could be written as e.g. `variableify(bsym.flat_proxy_args[some_map.get(bsym.sym, 0)])`

thunder/core/update_aliases.py
crcrpar: understood. How about calling `.from_swap_proxies` with `skip_output=True`?
https://github.com/Lightning-AI/lightning-thunder/blob/92a6adb63ddcf36e3910a55fc5f7a02cec5bee8c/thunder/core/symbol.py#L439-L445

thunder/executors/nvfuserex_impl.py
```diff
@@ -838,6 +838,11 @@ def map_redundant(x: Any) -> Any:
         )
         return cse_trace
 
+    def can_fuse(self, bsym: BoundSymbol) -> bool:
+        if bsym.sym.id == prims.PrimIDs.UPDATE_ALIASES:
+            return False
+        return super().can_fuse(bsym)
```
beverlylytle, line 844: This is a new attempt at a way to break up fusion regions.  This is more conservative than the previous approach in that it only applies to nvfuser's executor.  

This approach addresses two of the issues I was encountering.  The first issue is that `add_ouput` can only have input to the fusion definition as its second argument.  Breaking the `update_aliases` and the subsequent in-place op over two fds allows this to happen.  The second issue is that since `update_aliases` is translated into a collection of individual `fd.ops.set` calls, there is no artificial dependency introduce in the fusion definition.  Breaking the fusion regions at this point does introduce those dependencies.  

If there were some fd op that could introduce those dependencies, there could be fewer fusion regions if `can_fuse` could know whether the `update_aliases` bsym precedes an in-place op, and only then introduce a fusion break.

thunder/tests/test_update_aliases.py
```diff
@@ -0,0 +1,217 @@
...
+    sample = next(op.sample_inputs(device, dtype))
+
+    # polygamma expects an int as its first argument and a tensor as its second but
+    # torch.Tensor.polygamma_ wants opposite; tensor first, int second.
+    # ref:
+    # - https://pytorch.org/docs/stable/special.html#torch.special.polygamma
+    # - https://pytorch.org/docs/stable/generated/torch.Tensor.polygamma_.html
+    args = list(sample.args)
+    if op.name == "polygamma_":
+        args[0], args[1] = args[1], args[0]
```
IvanYashchuk, line 28: It's a bug in Thunder's `polygamma_`:
https://github.com/Lightning-AI/lightning-thunder/blob/aefafe93f1b859451e3d6120d2775bb699f28372/thunder/torch/__init__.py#L2212
The order of arguments should be the same as in PyTorch.

thunder/executors/nvfuserex_impl.py
IvanYashchuk: Commenting out the line with `register_supported(PrimIDs.UPDATE_ALIASES, update_aliases, _elementwise_nnary_check)` would have the same effect: update_aliases is not fused into nvFuser regions.

> The first issue is that `add_ouput` can only have input to the fusion definition as its second argument.

Could you please create a test case (that can be decorated with xfail) for this example? If I understand correctly the issue is that we can't tell to nvFuser "write result to this intermediate variable". This is probably the case when we need to use functionalization to fuse current breaks later.

thunder/tests/test_update_aliases.py
```diff
@@ -0,0 +1,217 @@
...
+        torch.testing.assert_close(actual, expected)
+
+
+@instantiate(
+    devicetypes=(devices.DeviceType.CPU,),
+    dtypes=NOTHING,
+)
+def test_aliased_input(executor, device, dtype):
+    def f(x, y, z):
+        return y.exp_().add(x) + z.exp()  # Fails on CUDA because operations have been reordered.
```
IvanYashchuk, line 201: Here's the execution trace:
```py
def computation(x, y, z):
  # x: "cuda:0 f32[2, 1, 2]"
  # y: "cuda:0 f32[2, 1, 2]"
  # z: "cuda:0 f32[1, 2, 2]"
  (t9,) = update_aliases((y,))
  del y
  [t1, t6] = nvFusion0(t9, x, z)
    # t0 = prims.exp(t9)  # t0: "cuda:0 f32[2, 1, 2]"
    # t1 = prims.copy_(t0, t9, grad_enabled=True)  # t1: "cuda:0 f32[2, 1, 2]"
    # t2 = prims.add(t1, x)  # t2: "cuda:0 f32[2, 1, 2]"
    # t3 = prims.exp(z)  # t3: "cuda:0 f32[1, 2, 2]"
    # t4 = prims.broadcast_in_dim(t2, (2, 2, 2), (0, 1, 2))  # t4: "cuda:0 f32[2, 2, 2]"
    # t5 = prims.broadcast_in_dim(t3, (2, 2, 2), (0, 1, 2))  # t5: "cuda:0 f32[2, 2, 2]"
    # t6 = prims.add(t4, t5)  # t6: "cuda:0 f32[2, 2, 2]"
  del t9
  return (t6,)
```
The program seems correct. Let's check out the nvFuser definition (`jfn._lc_cs.last_traces[-1].python_ctx()["nvFusion0"].last_used`):
```py
def nvfuser_fusion_id0(fd : FusionDefinition) -> None :
    T0 = fd.define_tensor(shape=[2, 1, 2], contiguity=[True, None, True], dtype=DataType.Float, is_cpu=False, stride_order=[2, 1, 0])
    T1 = fd.define_tensor(shape=[2, 1, 2], contiguity=[True, None, True], dtype=DataType.Float, is_cpu=False, stride_order=[2, 1, 0])
    T2 = fd.define_tensor(shape=[1, 2, 2], contiguity=[None, True, True], dtype=DataType.Float, is_cpu=False, stride_order=[2, 1, 0])
    T3 = fd.ops.exp(T0)
    T4 = fd.ops.set(T3)
    fd.add_output(T4, T0)
    T5 = fd.ops.add(T0, T1)
    T6 = fd.ops.exp(T2)
    T11 = fd.ops.broadcast_in_dim(T5, shape=[2, 2, 2], broadcast_dims=[0, 1, 2])
    T16 = fd.ops.broadcast_in_dim(T6, shape=[2, 2, 2], broadcast_dims=[0, 1, 2])
    T17 = fd.ops.add(T11, T16)
    fd.add_output(T0)
    fd.add_output(T17)
```
Could it be due to `fd.add_output(T0)`? Is it an nvFuser bug or a Thunder-translation-to-nvFuser bug?

Script to reproduce the error using the generated fusion:
```py
import torch

def nvfuser_fusion_id0(fd : FusionDefinition) -> None :
    T0 = fd.define_tensor(shape=[2, 1, 2], contiguity=[True, None, True], dtype=DataType.Float, is_cpu=False, stride_order=[2, 1, 0])
    T1 = fd.define_tensor(shape=[2, 1, 2], contiguity=[True, None, True], dtype=DataType.Float, is_cpu=False, stride_order=[2, 1, 0])
    T2 = fd.define_tensor(shape=[1, 2, 2], contiguity=[None, True, True], dtype=DataType.Float, is_cpu=False, stride_order=[2, 1, 0])
    T3 = fd.ops.exp(T0)
    T4 = fd.ops.set(T3)
    fd.add_output(T4, T0)
    T5 = fd.ops.add(T0, T1)
    T6 = fd.ops.exp(T2)
    T11 = fd.ops.broadcast_in_dim(T5, shape=[2, 2, 2], broadcast_dims=[0, 1, 2])
    T16 = fd.ops.broadcast_in_dim(T6, shape=[2, 2, 2], broadcast_dims=[0, 1, 2])
    T17 = fd.ops.add(T11, T16)
    fd.add_output(T0)
    fd.add_output(T17)

    nvfuser_fusion_id0(fd)

def f(x, y, z):
    return y, y.exp_().add(x) + z.exp()

a = make_tensor((2, 1, 2), dtype=torch.float32, device="cuda:0")
b = a.clone()
c = a.view(1, 2, 2)
a_ = a.clone().detach()
b_ = b.clone().detach()
c_ = c.clone().detach()
actual = fd.execute([a, b, c])
expected = f(a_, b_, c_)
torch.testing.assert_close(actual, expected)
torch.testing.assert_close(a, a_)
torch.testing.assert_close(b, b_)
torch.testing.assert_close(c, c_)
```

IvanYashchuk: There's a line in the nvFuser definition `T5 = fd.ops.add(T0, T1)` which should be `T5 = fd.ops.add(T4, T1)` or `T5 = fd.ops.add(T3, T1)`. It's happening because of this line:
https://github.com/Lightning-AI/lightning-thunder/blob/aefafe93f1b859451e3d6120d2775bb699f28372/thunder/executors/nvfuserex_impl.py#L2170
In order to generate nvFuser fusion definition with T3 or T4 in the add arguments we need to return either `nvcopy_from` or `alias_output`. With both options the test passes. I push the commit with `alias_output` return (https://github.com/Lightning-AI/lightning-thunder/pull/1780/commits/ba264c302651e9a00abcf04714336262ea3ce2ac).

thunder/tests/test_update_aliases.py
```diff
@@ -0,0 +1,216 @@
...
+        torch.testing.assert_close(actual, expected)
+
+
+@instantiate(
+    devicetypes=(devices.DeviceType.CPU,),
+    dtypes=NOTHING,
+)
+def test_inplace_on_chunk(executor, device, dtype):
+    def f(x):
+        y, z = x.chunk(2, dim=1)  # Fails on CUDA with stride issues
```
IvanYashchuk, line 97: Created https://github.com/NVIDIA/Fuser/issues/3957.

thunder/tests/test_update_aliases.py
beverlylytle: Fix in #1804

thunder/tests/test_update_aliases.py
```diff
@@ -0,0 +1,217 @@
...
+    expected = op.torch_reference(*args, **sample.kwargs)
+    torch.testing.assert_close(actual, expected, equal_nan=True)
+
+
+@instantiate(
+    devicetypes=(devices.DeviceType.CPU,),
+    dtypes=NOTHING,
+)
+def test_inplace_on_view(executor, device, dtype):
+    def f(x, _):
```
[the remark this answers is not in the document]
beverlylytle: Indeed.  I think what's happening here is that nvfuser's executor runs a dce pass over the bsyms.  The logic behind dce is focused on output of bsyms.  You'll notice that in the working case the last interesting line is `copy_(t31, t29, grad_enabled=True)` which is absent from the non-working case.  That line produces no output, and hence, is removed by nvfuser's fusion pass.  I believe this lack of output for copy_ is coming from the lack of output in setitem_

thunder/executors/nvfuserex_impl.py
beverlylytle: done and done.

thunder/tests/test_update_aliases.py
```diff
@@ -0,0 +1,217 @@
...
+        y = x.view((3, 2))
+        y[0][0] = 0  # Fails on CUDA
```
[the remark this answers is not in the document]
beverlylytle: done.

thunder/core/transforms.py
```diff
@@ -1636,7 +1647,7 @@ def zeros_like(x):
     prims.PrimIDs.LOG2: lambda x, g: g / (x * 0.6931471805599453),
     prims.PrimIDs.FMOD: lambda x, y, g: (g, -g * prims.trunc(x / y)),
     # The copy should not be differentiable. We return None to enable the generation of the backward graph through them.
```
crcrpar, line 1649: Given the change below, shouldn't this line get updated?

thunder/core/transforms.py
```diff
@@ -2290,6 +2301,19 @@ def iter_bound_symbols(bound_symbols):
             yield symbol
 
 
+def get_first_proxy(proxies) -> Proxy | None:
+    """Get the first proxy from a list of proxies.
+
+    Args:
+        proxies (List[Proxy]): List of proxies
```
crcrpar, line 2308: I'd prefer this style but it's up to you
```suggestion
def get_first_proxy(proxies: list[Proxy]) -> Proxy | None:
    """Get the first proxy from a list of proxies.

    Args:
        proxies: List of proxies
```

thunder/core/update_aliases.py
```diff
@@ -0,0 +1,106 @@
...
+                in_tensors = {in_tensor}
+            else:
+                in_tensors = set(in_tensors)
+            out_tensors = set(map(variableify, filter(lambda p: isinstance(p, TensorProxy), bsym.flat_proxy_outs)))
+            encountered.update(in_tensors)
+            group = set(reduce(set.union, filter(lambda g: any(g.intersection(in_tensors)), view_groups), set()))
+            if group is None:
+                # This is a view creation with operands that are not involved in any inplace ops.
+                bsyms.append(bsym.from_bsym_swap_proxies(swap_map, skip_output=True))
+                continue
```
crcrpar, line 87: I'm not quite sure about the appropriateness of the `None` check here as in my gut feeling `group` would be either an empty set or a non-empty set.

thunder/executors/nvfuserex_impl.py
```diff
@@ -1759,6 +1763,15 @@ def clone(a: TensorProxy, *, fd: FusionDefinition, lc_to_nv_map: dict) -> Any:
 
 register_supported(PrimIDs.CLONE, clone, _elementwise_unary_check)
 
+# update_aliases is disabled.  nvfuser does not support it.
+# def update_aliases(aliases: tuple[TensorProxy], *, fd: FusionDefinition, lc_to_nv_map: dict) -> Any:
+#     nvaliases = tuple(getnv(alias, fd, lc_to_nv_map) for alias in aliases)
+#     return tuple(fd.ops.set(nvalias) for nvalias in nvaliases)
+
+
+# register_supported(PrimIDs.UPDATE_ALIASES, update_aliases, _elementwise_nnary_check)
```
crcrpar, line 1772: It'd be great to put TODO comment and/or reference of issue (if exists) here

thunder/tests/test_update_aliases.py
crcrpar, line 1: While looking at this file, I got an evanescent question whether this file could be merged with https://github.com/Lightning-AI/lightning-thunder/blob/3e6ad5bc0f43c56f3c9ec34f6fd6d683b1a12693/thunder/tests/test_inplace_functionalization.py and it leads to a better structured tests.

What do you think? Are there tad many differences from the test to merge?

thunder/core/update_aliases.py
beverlylytle: Good catch.  I forgot to update this when I made an adjustment to how `group` is calculated.

thunder/executors/nvfuserex_impl.py
beverlylytle: There is no issue yet, in part because the requirements for nvfuser are not firm. But will add a TODO.

thunder/tests/test_update_aliases.py
beverlylytle: I find it easier to comprehend when the two different approaches addressed in two different files.  At least for the moment.  It's very easy to overlook a flag like `skip_inplace_alias_updates` to understand which of the two approaches is being applied in a given test.  At some point in the future (probably around the time that skip_inplace_alias_updates defaults to False, and both approaches are being used), I think it will makes sense to combine the two files into one test_inplace_support.py.

crcrpar: Understood. Let's have two files till things start settling down.

Thank you for perspicuous explanation. I also thought maybe the separate files would be easier to understand even though some parts look pretty similar.

thunder/core/rematerialization.py
```diff
@@ -308,8 +308,9 @@ def find_cut(
     required_consumer_symbols = tuple(
         utils.find_producer_symbols(consumer_trace, consumer.output, external_consumer_inputs)
     )
+    # !!! Filtering out the Nones coming from update_aliases + CDE
```
crcrpar, line 311: is this a typo?

```suggestion
    # !!! Filtering out the Nones coming from update_aliases + DCE
```

thunder/core/update_aliases.py
```diff
@@ -0,0 +1,106 @@
...
+def _is_inplace_op(bsym):
+    return (bsym.sym.tags and prims.OpTags.IN_PLACE in bsym.sym.tags) or (
+        bsym.subsymbols and bsym.subsymbols[-1].sym.id == prims.PrimIDs.COPY_
+    )
+
+
+def _is_view_creation_op(bsym):
+    import thunder.torch as ltorch
+
+    return bsym.sym in ltorch._syms_returning_views or bsym.sym in ltorch._syms_that_may_return_views
```
crcrpar, line 36: Would we happen to have corner cases to check behavior/outputs of `insert_alias_updates` when `_syms_that_may_return_views` ops return a new tensor not a view?

thunder/core/rematerialization.py
beverlylytle: Oh, yes, thanks.

thunder/core/update_aliases.py
beverlylytle: Perhaps eventually.  The worry for calling `update_aliases` too frequently would be that bsyms that could have otherwise been reordered no longer can, potentially resulting in slower performance.  The added complexity of dealing with these corner cases needs to be weighed against this as of yet unknown performance cost.

thunder/extend/__init__.py
IvanYashchuk: The current behavior as of https://github.com/Lightning-AI/lightning-thunder/pull/1780/commits/1fa6baf0503eeddc71862530402af5bba2241a1d is to exclude `update_aliases` from fused nvFuser region and then the output of `update_aliases` becomes the input to the fused region with nvFuser working for this small example:
```py
In [1]: import torch

In [2]: import thunder

In [3]: def foo(x):
   ...:    x.add_(1)
   ...:    return x
   ...: 

In [4]: jfoo = thunder.jit(foo, fusion_type="dataflow", skip_inplace_alias_updates=False, skip_inplace_functionalization=True)

In [5]: x = torch.randn(3, 3, device="cuda")

In [6]: jfoo(x);

In [7]: jfoo._lc_cs.last_traces[-1]
Out[7]: 
import torch

@torch.no_grad()
@no_autocast
def computation(x):
  # x: "cuda:0 f32[3, 3]"
  (t4,) = update_aliases((x,))
  del x
  [t1] = nvFusion0(t4)
    # t0 = prims.add(t4, 1.0)  # t0: "cuda:0 f32[3, 3]"
    # t1 = prims.copy_(t0, t4, grad_enabled=True)  # t1: "cuda:0 f32[3, 3]"
  del t4
  return (t1,)
```

thunder/core/update_aliases.py
```diff
@@ -0,0 +1,106 @@
...
+    # filter out view groups that don't have any tensors involved in inplace ops
+    view_groups = [group for group in view_groups if len(group.intersection(inplace_inputs)) != 0]
+    viewed = set(reduce(set.union, view_groups, set()))
+    encountered = set()
+
+    for bsym in computation_trace.bound_symbols:
+        if _is_inplace_op(bsym) or _is_view_creation_op(bsym) or _involves_viewed_args(bsym, viewed):
+            in_tensors = map(variableify, filter(lambda p: isinstance(p, TensorProxy), bsym.flat_proxy_args))
+            if _is_inplace_op(bsym):
+                in_tensor = list(in_tensors)[0]
```
IvanYashchuk, line 77: What would happen if `bsym.flat_proxy_args` is empty?

```suggestion
            in_tensors = list(map(variableify, filter(lambda p: isinstance(p, TensorProxy), bsym.flat_proxy_args)))
            if _is_inplace_op(bsym) and in_tensors:
                in_tensor = list(in_tensors)[0]
```

thunder/core/update_aliases.py
```diff
@@ -0,0 +1,106 @@
...
+                in_tensors = {in_tensor}
+            else:
+                in_tensors = set(in_tensors)
+            out_tensors = set(map(variableify, filter(lambda p: isinstance(p, TensorProxy), bsym.flat_proxy_outs)))
+            encountered.update(in_tensors)
+            group = set(reduce(set.union, filter(lambda g: any(g.intersection(in_tensors)), view_groups), set()))
+            if len(group) == 0:
```
IvanYashchuk, line 84: PEP8: "For sequences, (strings, lists, tuples), use the fact that empty sequences are false".
```suggestion
            if not group:
```

thunder/core/update_aliases.py
```diff
@@ -0,0 +1,106 @@
...
+                continue
+            views_encountered = group.intersection(encountered)
+            new_aliases = _get_new_aliases(views_encountered, computation_trace)
+
+            update_bsym, swap_map = _get_update_bsym(views_encountered, swap_map, new_aliases)
+            bsyms.append(update_bsym)
+            encountered.update(out_tensors)
+            new_bsym = bsym.from_bsym_swap_proxies(swap_map)
+            bsyms.append(new_bsym)
+            if _is_inplace_op(bsym) and len(out_tensors) == 1:
```
IvanYashchuk, line 96: Combined with the previous suggestion about `bsym.flat_proxy_args`.
```suggestion
            if _is_inplace_op(bsym) and len(out_tensors) == 1  and in_tensors:
```

thunder/core/update_aliases.py
```diff
@@ -0,0 +1,106 @@
...
+            else:
+                in_tensors = set(in_tensors)
+            out_tensors = set(map(variableify, filter(lambda p: isinstance(p, TensorProxy), bsym.flat_proxy_outs)))
+            encountered.update(in_tensors)
+            group = set(reduce(set.union, filter(lambda g: any(g.intersection(in_tensors)), view_groups), set()))
+            if len(group) == 0:
+                # This is a view creation with operands that are not involved in any inplace ops.
+                bsyms.append(bsym.from_bsym_swap_proxies(swap_map, skip_output=True))
+                continue
+            views_encountered = group.intersection(encountered)
```
IvanYashchuk, line 88: Should we take care of the case when `views_encountered` is empty?

```suggestion
            views_encountered = group.intersection(encountered)
            if not views_encountered:
                bsyms.append(bsym.from_bsym_swap_proxies(swap_map, skip_output=True))
                continue
```

thunder/core/update_aliases.py
beverlylytle: An inplace op with empty flat_proxy_args?  Then something else is surely wrong, but, of course, Safety First!

thunder/core/update_aliases.py
beverlylytle: done.

thunder/core/update_aliases.py
beverlylytle: Again, if views_encountered is empty, it would mean that a view creation op or an in-place op has empty flat_proxy_args, which does not align with my understanding of those types of ops.

mruberry (approved):
The merge conflict needs to be resolved, however

PR #1782, opened 2025-02-19
Remove unnecessary make_tensor implementation

### LINKED ISSUE #1774: [Developer Efficiency] Remove unnecessary make_tensor implementation
Since https://github.com/pytorch/pytorch/pull/96124 was merged into PyTorch, our own implementation of make_tensor is no longer necessary. We should try removing https://github.com/Lightning-AI/lightning-thunder/blob/main/thunder/tests/make_tensor.py wherever we still use it and use the upstream version, instead.

cc @borda @mruberry

lianakoleva:
Fixes #1774. 
- Possibly should move `make_tensor_like` to a different file and completely delete `thunder/tests/make_tensor.py`, but was not sure what location would be preferred. 
- Swapped `make((L, S), dtype=torch.bool, low=1, high=1, requires_grad=False).tril()` for `torch.ones((L, S), dtype=torch.bool, requires_grad=False).tril()` because torch `make_tensor` requires `high > low` while thunder `make_tensor` allowed `high == low`.


thunder/tests/opinfos.py
```diff
@@ -8269,7 +8269,7 @@ def scaled_dot_product_attention_sample_generator(op, device, dtype, requires_gr
 
     # mask cases
     q, k, v = make(N, n_head, L, E), make(N, n_head, S, E), make(N, n_head, S, Ev)
-    bool_attn_mask = make((L, S), dtype=torch.bool, low=1, high=1, requires_grad=False).tril()
+    bool_attn_mask = torch.ones((L, S), dtype=torch.bool, requires_grad=False).tril()
```
mruberry, line 8272: The test errors are real because this call to `torch.ones` doesn't pass a `device` argument like `make` does

mruberry:
Hey @lianakoleva, thanks for submitting this PR!



This is pretty cool; I was surprised to see the test failures, especially so many of the form:

```
FAILED thunder/tests/test_ops.py::test_core_vs_torch_consistency_index_put_nvfuser_cuda_thunder.dtypes.float64 - ValueError: `low` must be less than `high`, but got 0 >= 0
```

And I think your PR comment is correct that sometimes we are intentionally generating tensors that are full of just one value, and PyTorch's `make_tensor` complains when that's the case.

Maybe the simplest thing, then, is that we keep our own `make_tensor`, and it calls PyTorch's `make_tensor` whenever low != high, but when low == high it instead calls `torch.full`? That way we wouldn't have to maintain so much custom make_tensor logic, but we wouldn't require lots of rewriting from `make_tensor` to `full`. 

What do you think?

lianakoleva:
Alternatively, just do the rewrite to conform to PyTorch [convention](https://github.com/pytorch/pytorch/blob/main/torch/testing/_creation.py#L91-L92).

mruberry:
> That sounds reasonable. If I am reverting instances where thunder make_tensor to torch make_tensor, should I then standardize use of make_tensor to thunder across the codebase (currently it is mixed)?

Great question. I think it's OK to leave the mixed usage. We could clean it up, but I doubt it bothers anyone

thunder/tests/make_tensor.py
```diff
@@ -43,23 +18,9 @@ def make_tensor(
...
-    +---------------------------+------------+----------+
-    | unsigned integral type    | ``0``      | ``10``   |
-    +---------------------------+------------+----------+
-    | signed integral types     | ``-9``     | ``10``   |
-    +---------------------------+------------+----------+
-    | floating types            | ``-9``     | ``9``    |
-    +---------------------------+------------+----------+
-    | complex types             | ``-9``     | ``9``    |
-    +---------------------------+------------+----------+
+    Calls either torch.full or torch.testing.make_tensor to allow for low == high, whereas torch.testing.make_tensor
```
mruberry, line 21: Nice comment

mruberry (commented):
Looks good! Lots of cleanup. I suggested one tweak to merge more of the logic

thunder/tests/make_tensor.py
```diff
@@ -87,121 +48,27 @@ def make_tensor(
...
-    if dtype in _floating_types + _complex_types:
-        result.requires_grad = requires_grad
-
-    # NOTE This is a workaround. There are so many not supported operations that,
-    # even creating the test tensors is hard.
-    if dtype in _floating_8bit_types:
-        result = result.to(dtype)
-
-    return result
+    if low is not None and low == high:
```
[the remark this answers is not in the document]
lianakoleva: Good thought, thanks for the improvement and explanation

mruberry (approved):
Cool! Thanks @lianakoleva!

PR #1784, opened 2025-02-20
Split get_computation_and_inputs

t-vi:
Currently the function `get_computation_and_inputs` does three things:
- acquisition of the initial trace by running the jit/interpreter,
- application of the transforms and transform for execution etc. to get the prologue, computation, epilogue,
- running the prologue and extracting the inputs for compute.

It can be useful to have the first two split, so that is what we do.


t-vi:
Seems to work, I'll add a test.

lantiga (approved):
Looks great.

As a general consideration, it would be good to think about a higher level interface to do what is in `run_prologue`, as part of a follow-up PR.

For example, what moves together (cd, cs, ci) should stay together. It could be bundled in an object that allows to perform these operations without being exposed to internals.

For the future, but it would facilitate reasoning about model distribution etc.

t-vi:
Yeah, we definitely need to think about the cd, cs, ci.

PR #1785, opened 2025-02-20
remove redundant assert

kshitij12345:
This check above this assert does the same.

https://github.com/Lightning-AI/lightning-thunder/blob/7fd5686b571eda868f5b8ea02aea231f60518a5c/thunder/core/symbol.py#L343-L347


t-vi (approved):
Thank you @kshitij12345

PR #1787, opened 2025-02-21
Fix tensor.to

### LINKED ISSUE #1291: `ltorch.to` could return `self` even if `copy=True`
According to [the docs](https://pytorch.org/docs/stable/generated/torch.Tensor.to.html), `Tensor.to` should return a copy if `copy=True` even if none of device, dtype, and memory_format is specified.
 
Our dialect implementation of https://github.com/Lightning-AI/lightning-thunder/blob/30e4aa1e67005c58219d7f06b46836eedb74b27a/thunder/torch/__init__.py#L452-L465 apparently could return self

riccardofelluga:
#1775 wrongly uses `_copy_` while we need to register a clone op to have a copy of the tensor. Moved the tests to thunder's opinfos suite as per all the rest of the torch ops. 

This PR reverts #1775 and, together with #1292, fixes #1291

cc @mruberry

### DIFF
--- a/thunder/tests/test_core.py
+++ b/thunder/tests/test_core.py
@@ -1865,20 +1865,6 @@ def test_torch_tensor_to_memory_format(executor: TestExecutor, device: str, _):
         assert_close(torch_result, thunder_result, check_stride=True)
 
 
-def test_torch_tensor_to_return_type():
-    a = torch.randn(2, 4, 5, 3)
-
-    def torch_to(a):
-        return a.to(copy=True)
-
-    jfoo = thunder.jit(torch_to)
-    thunder_result = jfoo(a)
-    torch_result = torch_to(a)
-
-    assert thunder_result is not a
-    assert_close(torch_result, thunder_result)
-
-
 # TODO See issue "Add contiguous and clang.stride_order OpInfos that check stride
 # consistency with PyTorch"
 @instantiate(
--- a/thunder/torch/__init__.py
+++ b/thunder/torch/__init__.py
@@ -477,19 +477,7 @@ def to(
     )
 
     if copy:
-        if device is not None:
-            device = to_device(device)
-            a = prims.device_put(a, device)
-        if dtype is not None:
-            dtype = to_dtype(dtype)
-            a = prims.convert_element_type(a, dtype)
-        if memory_format is not None:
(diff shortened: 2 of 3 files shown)


t-vi:
@IvanYashchuk do you want to look at this before merging?

thunder/torch/__init__.py
```diff
@@ -472,31 +472,12 @@ def to(
     copy: bool = False,
     memory_format: None | torch.memory_format = None,
 ) -> TensorLike:
-
-    input_device = a.device
-    input_dtype = a.dtype
-    if copy and not _will_to_return_self(input_device, input_dtype, device, dtype, memory_format, copy):
-        b = prims.empty(a.shape, device=input_device, dtype=input_dtype)
-        return _copy_(b, a)
```
IvanYashchuk, line 480: empty + copy_ can certainly be replaced with a clone operation without the implications of an in-place operation. Nice change!

thunder/torch/__init__.py
```diff
@@ -472,31 +472,12 @@ def to(
...
-            dtype = to_dtype(dtype)
-            a = prims.convert_element_type(a, dtype)
-        if memory_format is not None:
-            # NOTE not sure if we need to handle torch.preserve_format explicitly
-            if memory_format == torch.channels_last:
-                a = prims.stride_order(a, (3, 0, 2, 1))
-            elif memory_format == torch.channels_last_3d:
-                a = prims.stride_order(a, (4, 0, 3, 2, 1))
-        return a
+        a = prims.clone(a)
```
IvanYashchuk, line 480: Nice code deduplication. All of the conversions are happening later in the function.

thunder/tests/test_core.py
```diff
@@ -1865,20 +1865,6 @@ def torch_to(a, memory_format):
         assert_close(torch_result, thunder_result, check_stride=True)
 
 
-def test_torch_tensor_to_return_type():
```
IvanYashchuk, line 1868: OpInfos is always the preferred way of testing functionality. Thanks for updating the test cases `to_opinfo `!

IvanYashchuk (approved):

t-vi (approved):
Supergood, thank you @riccardofelluga @IvanYashchuk

PR #1790, opened 2025-02-24
Optionally use shared memory pool in CUDAGraphs Transform

ali-alshaar7:
<details>

</details>

## What does this PR do?

- Adds an option to reuse memory pools in CUDAGraphs, as described [here](https://pytorch.org/docs/stable/notes/cuda.html#graph-memory-management)
- We note that using pools seems to have no effect on reserved memory increasing across captures

## PR review

## Did you have fun?

### DIFF
--- a/thunder/transforms/cudagraph.py
+++ b/thunder/transforms/cudagraph.py
@@ -59,11 +59,16 @@ class CUDAGraphRunner:
     change.
     """
 
-    def __init__(self):
+    def __init__(self, share_mem_pool: bool = False):
         self.cuda_graph_cache = {}  # cahce_key (.make_cache_key) -> (graph, static_inputs, static_outputs)
         self.python_callables = {}  # fn_name -> (callable. static_input_mask (or None))
         self.trace_symbols = {}  # fn_name -> (bsyms, inputs, outputs)
         self.name_counter = 1
+
+        if share_mem_pool:
+            self.global_mem_pool = torch.cuda.graph_pool_handle()
+        else:
+            self.global_mem_pool = None
 
     def get_static_buffer(self, x):
         if isinstance(x, torch.Tensor):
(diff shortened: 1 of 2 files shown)


thunder/transforms/cudagraph.py
```diff
@@ -59,12 +59,17 @@ class CUDAGraphRunner:
     change.
     """
 
-    def __init__(self):
+    def __init__(self, share_mem_pool: bool = False):
```
t-vi, line 62: Nit: Can we make this and the transform arg key-word-only, please?
```
    def __init__(self, *, share_mem_pool: bool = False):
```

thunder/transforms/cudagraph.py
```diff
@@ -59,12 +59,17 @@ class CUDAGraphRunner:
...
         self.cuda_graph_cache = {}  # cahce_key (.make_cache_key) -> (graph, static_inputs, static_outputs)
         self.python_callables = {}  # fn_name -> (callable. static_input_mask (or None))
         self.trace_symbols = {}  # fn_name -> (bsyms, inputs, outputs)
         self.name_counter = 1
 
+        if share_mem_pool:
+            self.global_mem_pool = torch.cuda.graph_pool_handle()
```
t-vi, line 69: Nit: I'd prefer if we could call it `mem_pool`, without global. It is specific to the transform/runner (which is good).

thunder/transforms/cudagraph.py
```diff
@@ -59,12 +59,17 @@ class CUDAGraphRunner:
...
+        else:
+            self.global_mem_pool = None
```
t-vi, line 71: Maybe add a comment that this means that a new pool is used for each graph.

t-vi (approved):
Great! Two nits, but let's get it merged.
The to my mind, the next step would be to check if we are using fw + bw and then which "saved for backward" tensors are outputs of the graph, and mark them static inputs in the backward.
The tricky bit will be that we see fw and bw sequentially, but thankfully in the right sequence.

riccardofelluga:
@ali-alshaar7  Is this new option covered by any test already?

ali-alshaar7:
> @ali-alshaar7 Is this new option covered by any test already?

So I haven't been able to get this option to do much memory saving unfortunately, but in theory it should. I'll create an issue with the repro script and more details instead so people can maybe correct it if I'm not testing properly.

riccardofelluga:
Oh thanks! @ali-alshaar7 I think it would be good for this PR to include a test that verifies that this new option works as expected regardless of the quantity of the memory savings

t-vi:
Thank you for adding the tests @ali-alshaar7

PR #1795, opened 2025-02-24
restore nvfuser take_along_axis support

jjsjann123:
Two things are done here:
1. adding back nvfuser take_along_axis support
2. adding an compile flag `


crcrpar:
Close as https://github.com/Lightning-AI/lightning-thunder/pull/2031 is in the main

PR #1797, opened 2025-02-25
Fix implementation of strides computation in cuDNN executor

IvanYashchuk:
There was a bug in the cuDNN executor code when computing the contiguous strides for a given shape. For example for a shape of (3, 4) the previous implementation was returning (3, 1) instead of correct (4, 1) strides.

The bug was benign because we were not executing/testing masked SDPA with the cuDNN executor (and the executor's checker was rightfully failing).

cc @mruberry, @vedaanta

### DIFF
--- a/thunder/executors/cudnnex.py
+++ b/thunder/executors/cudnnex.py
@@ -218,6 +218,7 @@ def torch_to_cudnn_dtype(lc_dtype: dtypes.dtype):
     }
     return _torch_to_cudnn_dtype_map[lc_dtype]
 
+
 def _compute_row_major_strides(shape):
     """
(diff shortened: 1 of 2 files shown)


thunder/executors/cudnnex.py
```diff
@@ -219,10 +220,17 @@ def torch_to_cudnn_dtype(lc_dtype: dtypes.dtype):
 
 
 def _compute_row_major_strides(shape):
-    strides = [1]
-    for dim in reversed(shape[:-1]):
```
IvanYashchuk, line 223: The bug is that it should have been `shape[1:]` instead of `shape[:-1]`.

t-vi (approved):
Thank you @IvanYashchuk

PR #1800, opened 2025-02-25
[ThunderFX report]: Improve the interfaces

kiya00:
<details>

</details>

## What does this PR do?

This PR
1. Reorganizes the `thunderfx_benchmark_reports`, so that users can use:
```
thunder_fxgraph_reports = get_thunder_fxgraph_reports(model, x)

# Frees the parameters and inputs to make room for the reports
del model
del x

thunderfx_benchmark_report_from_splits(thunder_fxgraph_reports, folder_path, compare_fusion=True)
```
(or if there's no OOM, we can use `thunderfx_benchmark_report` directly)
2. Fixes `run_forward_backward` so that it can take both function and module as input
3. Uses TorchInductor without TorchDynamo by default when benchmarking each FX graph to avoid  https://github.com/Lightning-AI/lightning-thunder/issues/1521

### DIFF
--- a/thunder/dynamo/benchmark_utils.py
+++ b/thunder/dynamo/benchmark_utils.py
@@ -53,7 +53,7 @@ class ThunderCompileSpecification(CompileSpecificationInterface):
         self.name = specification_name
         self.thunder_options: dict = kwargs
 
-    def compile(self, fn):
+    def compile(self, fn, *args):
         from thunder import jit
 
         return jit(fn, **self.thunder_options)
@@ -75,7 +75,7 @@ class TorchCompileSpecification(CompileSpecificationInterface):
         self.name = specification_name
         self.torch_compile_options: dict = kwargs
 
-    def compile(self, fn):
+    def compile(self, fn, *args):
         return torch.compile(fn, **self.torch_compile_options)
 
     def to_source(self, fn_name):
@@ -97,7 +97,7 @@ class TorchEagerSpecification(CompileSpecificationInterface):
     def __init__(self, specification_name="torcheager"):
         self.name = specification_name
 
-    def compile(self, fn):
+    def compile(self, fn, *args):
         return fn
 
     def to_source(self, fn_name):
@@ -111,9 +111,8 @@ class TorchInductorSpecification(CompileSpecificationInterface):
     https://github.com/Lightning-AI/lightning-thunder/issues/1521
     """
 
-    def __init__(self, inputs, specification_name="inductor_backend"):
+    def __init__(self, specification_name="inductor_backend"):
         self.name: str = specification_name
-        self.inputs: list = inputs
 
     @staticmethod
     def torch_inductor(fn, inputs):
@@ -123,8 +122,8 @@ class TorchInductorSpecification(CompileSpecificationInterface):
         fx_graph = symbolic_trace(fn)
         return inductor_compile(fx_graph, inputs)
 
-    def compile(self, fn):
-        return self.torch_inductor(fn, self.inputs)
+    def compile(self, fn, inputs):
+        return self.torch_inductor(fn, inputs)
 
     def to_source(self, fn_name):
         return f"TorchInductorSpecification.torch_inductor({fn_name}, inputs)"
--- a/thunder/dynamo/benchmark_utils.py
+++ b/thunder/dynamo/benchmark_utils.py
@@ -23,7 +23,7 @@ class CompileSpecificationInterface:
         - :class:`ThunderCompileSpecification`
     """
 
-    def compile(self, fn: Callable) -> Callable:
+    def compile(self, fn: Callable, **kwargs) -> Callable:
         """Compiles the given callable and returns the compiled version."""
         raise NotImplementedError("Subclasses should implement the 'compile' method if needed.")
 
@@ -53,7 +53,7 @@ class ThunderCompileSpecification(CompileSpecificationInterface):
         self.name = specification_name
         self.thunder_options: dict = kwargs
 
-    def compile(self, fn, *args):
+    def compile(self, fn, **kwargs):
         from thunder import jit
 
         return jit(fn, **self.thunder_options)
@@ -75,7 +75,7 @@ class TorchCompileSpecification(CompileSpecificationInterface):
         self.name = specification_name
         self.torch_compile_options: dict = kwargs
 
-    def compile(self, fn, *args):
+    def compile(self, fn, **kwargs):
         return torch.compile(fn, **self.torch_compile_options)
 
     def to_source(self, fn_name):
@@ -97,7 +97,7 @@ class TorchEagerSpecification(CompileSpecificationInterface):
     def __init__(self, specification_name="torcheager"):
         self.name = specification_name
 
-    def compile(self, fn, *args):
+    def compile(self, fn, **kwargs):
         return fn
 
     def to_source(self, fn_name):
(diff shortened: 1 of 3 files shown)


thunder/dynamo/benchmark_utils.py
```diff
@@ -123,8 +122,8 @@ def torch_inductor(fn, inputs):
         fx_graph = symbolic_trace(fn)
         return inductor_compile(fx_graph, inputs)
 
-    def compile(self, fn):
-        return self.torch_inductor(fn, self.inputs)
+    def compile(self, fn, inputs):
+        return self.torch_inductor(fn, inputs)
```
kiya00, line 126: when using the TorchInductor without Dynamo(torch._inductor.compile), it requires not only the callable, but also the inputs. Previously I made the inputs as the data member of the `TorchInductorSpecification` (so that in `run_benchmark` or `run_repro` the `compilespecification.compile(fn)` doesn't need to change), but it holds the inputs as extra memory. So here I modify the common call to be `compilespecification.compile(fn, inputs)` and the `compile` function of other compile specification classes  takes an unused `inputs` argument.

kiya00:
Hi @mruberry , it's ready to review~

thunder/dynamo/benchmark_utils.py
```diff
@@ -97,7 +97,7 @@ class TorchEagerSpecification(CompileSpecificationInterface):
     def __init__(self, specification_name="torcheager"):
         self.name = specification_name
 
-    def compile(self, fn):
+    def compile(self, fn, *args):
```
mruberry, line 100: I would suggest **kwargs instead of *args and then have the compile methods be:

```
def compile(self, fn, **kwargs):
```

and the `compile` method for `torch_inductor` would be:

```
def compile(self, fn, inputs, **kwargs)
```

or 

```
def compile(self, fn, *, inputs, **kwargs)
```

depending on whether you want to require that `inputs` be specified as a kwarg or not

thunder/dynamo/report.py
```diff
@@ -996,26 +1004,103 @@ def foo(x):
     return report
 
 
-def get_thunder_split_reports(fn: Callable, *args, thunder_compile_kwargs: dict = None, **kwargs):
+def check_torch_runnablility(fn: Callable, *args, **kwargs):
```
mruberry, line 1007: Comment describing this function and how to use it, please

mruberry: Would this more accurately be titlted `check_torch_compile_runnability`?

thunder/dynamo/report.py
```diff
@@ -996,26 +1004,103 @@ def foo(x):
...
+        print("The input callable can be successfully executed by torch.compile.")
+
+
+def get_thunder_fxgraph_reports(
+    fn: Callable, *args, thunder_compile_kwargs: dict = None, check_runnablility=True, **kwargs
+) -> list[ThunderFXGraphReport]:
+    """
+    Generates a list of :class:`ThunderFXGraphReport` objects for a given callable.
+
+    This function performs the following steps:
```
mruberry, line 1033: Would you extend this comment to describe what the arguments are used for, including `*args` and `**kwargs`? 

From a Python UX perspective, I understand why it's convenient to pass args and kwargs like this, but mixing the inputs to the callable and the options to this function can be a confusing user-experience. 

We don't have to change the signature in this PR, but it might make sense to make calling this function look like this:

```
get_thunder_fx_graph_reports(fn, thunder_compile_kwargs=..., check_runnability=...)(*args, **kwargs)
```

which would be done by returning a function or callable class from `get_thunder_fx_graph_reports`. 

fyi @IvanYashchuk as I think he and I discussed the pros and cons of this pattern previously

thunder/dynamo/report.py
```diff
@@ -996,26 +1004,103 @@ def foo(x):
...
+def thunderfx_benchmark_report_from_splits(
+    thunder_fxgraph_reports: list[ThunderFXGraphReport],
+    folder_path: str | PathLike,
+    thunder_compile_kwargs: dict = None,
+    compare_fusion: bool = False,
+    rtol=0.5,
+    atol=0.0,
+):
+    """
+    A utility function that analyzes the runnability and performance benchmarks of each Thunder-split FX graph and nvFusion regions(optional). It prints out the performance metrics and saves the benchmark script in `folder_path` if the difference exceeds the tolerance (`rtol`, `atol` in seconds).
```
mruberry, line 1068: This might create multiple files, right?

Would you elaborate on the directory structure and name of the files this creates? 

Not for this PR, but in the future should we think about the filenames including a short description of why the file was generated? Something like:

```
run09_graph13_speed_bench.py
```

?

thunder/dynamo/report.py
```diff
@@ -996,26 +1004,103 @@ def foo(x):
...
+    runnable_split_reports: list[ThunderSplitGraphReport] = []
+    for thunder_fxgraph_report in thunder_fxgraph_reports:
+        print(f"\n{thunder_fxgraph_report.graph_name}: {thunder_fxgraph_report.split_reason}\n")
+        graph_folder = folder_path / thunder_fxgraph_report.graph_name
+        graph_folder.mkdir()
+        for split_report in thunder_fxgraph_report.subgraph_reports:
+            try:
+                split_report.run_repro(thunderjit)
+            except Exception as e:
+                print(f"Failed to run {split_report.graph_name} using Thunder with exception: {e}\n")
```
mruberry, line 1086: In the future we might consider writing these to some kind of log.

Printing to stdout seems great for now

thunder/dynamo/report.py
```diff
@@ -1056,58 +1141,20 @@ def thunderfx_benchmark_report(
     Here is an example:
 
     ```python
-    split_reports = get_thunder_split_reports(model, x)
+    thunder_fxgraph_reports = get_thunder_fxgraph_reports(model, x)
```
mruberry, line 1144: Cool!

kiya00:
> It's very exciting to see this starting to be used!
> 
> I made a few comments for your review, @kiya00, mostly about elaborating on documentation.
> 
> What do you think you'd like to focus on next?

Thanks for the suggestions, I think I'll add the memory usage and the operator summary as you suggested, that's probably the most useful things I assume. Also I got some feedback from Voytek and setup the env to try to use it on NeMo, WDYT?

mruberry:
> > It's very exciting to see this starting to be used!
> > I made a few comments for your review, @kiya00, mostly about elaborating on documentation.
> > What do you think you'd like to focus on next?
> 
> Thanks for the suggestions, I think I'll add the memory usage and the operator summary as you suggested, that's probably the most useful things I assume. Also I got some feedback from Voytek and setup the env to try to use it on NeMo, WDYT?

Responding to customer feedback sounds like a great #1 priority.

Working on adding memory usage or an operator summary next sounds good. I'd pick something that won't take too long because I'd expect more feedback soon.

mruberry:
Let me know when you'd like this merged, @kiya00

thunder/dynamo/report.py
mruberry: Definitely we can't include the full reason -- let's not worry about it for now

kiya00:
> Responding to customer feedback sounds like a great https://github.com/Lightning-AI/lightning-thunder/pull/1 priority.
> 
> Working on adding memory usage or an operator summary next sounds good. I'd pick something that won't take too long because I'd expect more feedback soon.

Get it, I’ll prioritize responding to customer feedback. 
I updated the documentation, and it's ready to merge @mruberry

PR #1804, opened 2025-02-25
fix signature of polygamma_

beverlylytle:
The signature of `polygamma_` should reflect that of `torch.Tensor.polygamma_`, meaning the first input should be a tensor.  It currently matches the signature of `torch.special.polygamma`, which take the same inputs, but in reversed order.  This change aims to make `polygamma_` consistent with `torch.Tensor.polygamma_`.

As an added bonus, this renames `ltorch.type` to `ltorch.torch_type` to avoid overriding python's `type`.

### DIFF
--- a/thunder/torch/__init__.py
+++ b/thunder/torch/__init__.py
@@ -357,8 +357,9 @@ def _old_torch_typestring_to_devicetype_and_dtype(typestring: str) -> tuple[Devi
     return devicetype_str, dtype_str
 
 
-@torchsymbol(torch.Tensor.type, is_method=True)
-def type(a: TensorLike, /, dtype: None | str | dtypeLike = None, non_blocking: bool = False) -> str | TensorLike:
+# NOTE: Using name `torch_type` to avoid conflict with Python's `type`
+@torchsymbol(torch.Tensor.type, is_method=True, method_name="type", id="torch.Tensor.type")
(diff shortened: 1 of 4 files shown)


mruberry (approved):

thunder/torch/__init__.py
```diff
@@ -357,8 +357,9 @@ def _old_torch_typestring_to_devicetype_and_dtype(typestring: str) -> tuple[Devi
     return devicetype_str, dtype_str
 
 
-@torchsymbol(torch.Tensor.type, is_method=True)
-def type(a: TensorLike, /, dtype: None | str | dtypeLike = None, non_blocking: bool = False) -> str | TensorLike:
+# NOTE: Using name `torch_type` to avoid conflict with Python's `type`
```
IvanYashchuk, line 360: What problems does the previous name cause?

Yes, it overwrites the built-in `type`, but if any code inside this file needs it it's still available in `builtins.type`.

beverlylytle: I was trying to debug an issue involving types, and it was not at all obvious that `type` was overwritten.  I encountered mysterious errors when I attempted to use it.  Of course one can use `buitlins.type`, but one has to know to use it.  I thought that renaming this function might possibly save a future developer a headache.

PR #1808, opened 2025-02-26
Add softshrink

beverlylytle:
Add support for the softshrink operator.  

It wasn't appearing in the test_vjp_correctness, and I saw that it was being filtered out by `check_bsym_for_vjp` due to the subsymbol `sign`.   This adds the augmented forward and backward impls, too.


thunder/torch/__init__.py
```diff
@@ -1934,6 +1934,20 @@ def silu(a: TensorLike, /, inplace: bool = False) -> TensorLike:
...
+def softshrink(a: TensorProxy, /, lambd: float = 0.5) -> TensorLike:
+    utils.check(
+        not dtypes.is_complex_dtype(a.dtype),
+        lambda: f"softshrink not implemented for '{a.dtype}'",
+    )
+    utils.check(
+        lambd >= 0,
+        lambda: f"lambda must be greater or equal to 0, but found to be {lambd}'",
+    )
+    # `a * 0` has the correct behavior for NaNs
```
mruberry, line 1947: This comment could say a little more.

I'm guessing `a * 0` is our fastest pattern for constructing a tensor that is like `a` except full of zeros where `a` is not `NaN` and `NaN` where `a` is `NaN`?

mruberry:
The PR looks great, but seems like something is wrong in the `sign` grad formulas?

beverlylytle:
I think it's not so much that there's something wrong with the `sign` grad formula as much as numerical issues with the test.  `sign` is a piecewise constant function, meaning that its derivative is 0.  Or rather, 0 everywhere other than 0 where it is undefined--this point of discontinuity could add flakiness, but not consistent failure.  The grad formula added here returns 0 everywhere.  The vjp test compares Thunder's vjp to a vjp calculated with finite differences.  Thunder's result in the vjp test, the dot product of a random tensor and a zero tensor of the same size, is 0, exactly, as expected.  The finite differences method produces, not a zero tensor, but a tensor filled with a value on the order of 1e-15, tiny, but not zero.  The dot product of this epsilon tensor with a random tensor `t` is ~`t.numel * epsilon`.  When `t` has small enough size, this is ok, but when the tensor is large (eg (2^10, 2^10)), this is significant.  One option, is to force that samples given to `sign` are small.

thunder/torch/__init__.py
beverlylytle: Updated.

mruberry:
> I think it's not so much that there's something wrong with the `sign` grad formula as much as numerical issues with the test. `sign` is a piecewise constant function, meaning that its derivative is 0. Or rather, 0 everywhere other than 0 where it is undefined--this point of discontinuity could add flakiness, but not consistent failure. The grad formula added here returns 0 everywhere. The vjp test compares Thunder's vjp to a vjp calculated with finite differences. Thunder's result in the vjp test, the dot product of a random tensor and a zero tensor of the same size, is 0, exactly, as expected. The finite differences method produces, not a zero tensor, but a tensor filled with a value on the order of 1e-15, tiny, but not zero. The dot product of this epsilon tensor with a random tensor `t` is ~`t.numel * epsilon`. When `t` has small enough size, this is ok, but when the tensor is large (eg (2^10, 2^10)), this is significant. One option, is to force that samples given to `sign` are small.

Great analysis as always. We could also just not test sign's grad for now with a note about why those tests are skipped.

beverlylytle:
Ah, indeed, the vjp test for small samples passed on my machine, but not in CI.  Most recent commit has those tests skipped.

mruberry (approved):

PR #1811, opened 2025-02-28
Add hardtanh

beverlylytle:
Adds support for `torch.nn.functional.hardtanh`.


beverlylytle:
Hold up, I need to add `hardtanh_`.

PR #1813, opened 2025-02-28
[ThunderFX report] Adds TimerWithCUDAMemoryUsage to track CUDA memory usage alongside timing measurements

kiya00:
<details>

</details>

## What does this PR do?

- Adds TimerWithCUDAMemoryUsage to track CUDA memory usage alongside timing measurements
- Reorganizes the TimerSpecification classes
- Uses the WallTimeWithMemoryUsage as default and saves script in `path/graph_name/memory_issue/` folder when the memory usage exceeds `rtol,atol`
- Note that if using `thunderfx_benchmark_report` directly, i.e. the original model and inputs are not freed, the peak memory reported in the log (which gets from `run_benchmark`) will be larger than the number reported by running the saved script (there's only single graph in the script)

### DIFF
--- a/thunder/dynamo/benchmark_utils.py
+++ b/thunder/dynamo/benchmark_utils.py
@@ -444,8 +444,9 @@ def check_metrics(
         except Exception as e:
             msg = f"Benchmark {timer_fn.name} on {graph_name} using {compile_fn.name} failed with exception {e}, benchmark script failed_{filename} is saved"
             stream.write(msg)
+            failed_folder = folder_path / "failed"
             report.write_benchmark(
-                folder_path, compile_fn, timer_fn, file_name=f"failed_{filename1}", extra_comment_str=msg
+                failed_folder, compile_fn, timer_fn, file_name=f"failed_{filename1}", extra_comment_str=msg
             )
             return None
 
--- a/thunder/dynamo/benchmark_utils.py
+++ b/thunder/dynamo/benchmark_utils.py
@@ -241,7 +241,7 @@ class TimerWithCUDAMemoryUsage:
         Memory usage is recorded in megabytes (MB).
     """
 
-    def __init__(self, timer):
+    def __init__(self, timer=torch.utils.benchmark.utils.timer.timer):
         self.max_allocated_memory = 0.0
         self.timer = timer
 
@@ -262,8 +262,7 @@ class TimerInterface:
         - :class:`KernelTime`
     """
 
-    @staticmethod
-    def time(fn, *args, **kwargs):
+    def time(self, *args, **kwargs):
         """
         Measures the execution time of a given callable.
         Subclasses must implement this method to define a specific timing mechanism.
@@ -273,31 +272,32 @@ class TimerInterface:
         """
         raise NotImplementedError("Subclasses should implement the 'time' method if needed.")
 
-    @staticmethod
-    def to_source(*args, **kwargs) -> str:
+    def to_source(self, *args, **kwargs) -> str:
         """
         Converts the timing function into its source code representation.
         Subclasses can implement this method to define how the timing logic is represented as code.
         """
         raise NotImplementedError("Subclasses should implement the 'to_source' method if needed.")
 
-    @staticmethod
-    def import_str(*args, **kwargs):
+    def import_str(self, *args, **kwargs):
         """
         Returns the necessary import statements for the timing implementation.
         """
         return None
 
 
-class WallTime(TimerInterface):
+class TorchBenchmarkTimerSpecification(TimerInterface):
     """
-    A timing utility that measures execution time using `torch.utils.benchmark.Timer`.
-    This class implements `TimerInterface` and provides a method to measure wall-clock time
-    using PyTorch's benchmarking utilities.
+    A timing utility that measures execution time using :class:`torch.utils.benchmark.utils.timer.Timer`.
+    the inner timer is used as the custom timer in timeit.Timer to return the current time, default is :func:`torch.utils.benchmark.utils.timer.timer`.
+    See: :class:`torch.utils.benchmark.utils.timer.Timer` for more details.
     """
 
-    @staticmethod
-    def time(stmt="pass", setup="pass", globals=None, min_run_time: float = 0.2) -> Measurement:
+    def __init__(self, name, inner_timer=torch.utils.benchmark.utils.timer.timer):
+        self.inner_timer = inner_timer
+        self.name = name
+
+    def time(self, stmt="pass", setup="pass", globals=None, min_run_time: float = 0.2) -> Measurement:
         """
         Measures execution time using PyTorch's :func:`torch.utils.benchmark.Timer.blocked_autorange()`.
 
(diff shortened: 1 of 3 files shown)


kiya00:
Hi @mruberry , I added some draft code to record the maximum CUDA memory allocated.  
Should we also compare the memory usage, and if it exceeds a certain limit, save the script by default?

mruberry:
> Hi @mruberry , I added some draft code to record the maximum CUDA memory allocated. Should we also compare the memory usage, and if it exceeds a certain limit, save the script by default?

Sounds like a great idea, consistent with reviewing speed!

Both the speed comparison and the memory comparison should be optional.

thunder/dynamo/benchmark_utils.py
```diff
@@ -224,6 +224,34 @@ def cleanup(self):
             pass
 
 
+class TimerWithCUDAMemoryUsage:
```
mruberry, line 227: Cool! Great modularity that this is a separate timer. Even if the same fx graph shows a speed issue and a memory issue, it's probably simplest to create two scripts, one for each issue, and separate bugs.

kiya00:
Hi @mruberry , it's ready to review, I updated the changes in the PR description

thunder/dynamo/benchmark_utils.py
```diff
@@ -224,6 +227,33 @@ def cleanup(self):
...
+    Example usage:
+        t = TimerWithCUDAMemoryUsage(TimerInterface.time)
+        t0 = t()  # Records initial memory and time
+        # ... code to measure ...
+        t1 = t()  # Records final memory and time
+        duration = t1 - t0  # Get elapsed time
+        memory_mb = t.max_allocated_memory  # Get peak memory usage in B
+
+    Note:
+        The memory tracking adds some overhead to the timing measurements.
```
mruberry, line 243: In the near future, we should think how to incorporate `memory_reserved`

thunder/dynamo/benchmark_utils.py
```diff
@@ -224,6 +227,33 @@ def cleanup(self):
             pass
 
 
+class TimerWithCUDAMemoryUsage:
+    """
+    A timer wrapper that tracks CUDA memory usage alongside timing measurements.
```
mruberry, line 232: A link to https://pytorch.org/docs/stable/notes/cuda.html#memory-management might help explain how PyTorch measures memory usage

thunder/dynamo/benchmark_utils.py
```diff
@@ -282,40 +312,26 @@ def time(stmt="pass", setup="pass", globals=None, min_run_time: float = 0.2) ->
...
         """
-        t = TorchBenchmarkTimer(stmt=stmt, setup=setup, globals=globals)
-        return t.blocked_autorange(min_run_time=min_run_time)
-
-    @staticmethod
-    def import_str():
-        return ["from thunder.dynamo.benchmark_utils import WallTime"]
+        t = TorchBenchmarkTimer(stmt=stmt, setup=setup, globals=globals, timer=self.inner_timer)
+        measurement = t.blocked_autorange(min_run_time=min_run_time)
+        if hasattr(self.inner_timer, "max_allocated_memory"):
```
mruberry, line 317: This seems OK for now. A couple notes:

- The description of the timer should be clear that it calls `torch.cuda.reset_peak_memory_stats()`. Because if someone was separately tracking memory then using this timer would affect those, right? And that may be surprising.
- In the future we may want to refine this extensibility point. A `# TODO` comment could help us remember to do so

thunder/dynamo/benchmark_utils.py
```diff
@@ -370,83 +404,127 @@ def _check_tensor(x):
...
+def check_metrics(
     folder_path,
     report,
     compile_fn1: Callable,
     compile_fn2: Callable,
     timer_fn: Callable,
-    timer_name: str,
-    rtol=0.5,
-    atol=0.0,
+    perf_rtol=0.5,
```
mruberry, line 423: Nit: maybe `time_rtol` instead of `perf_rtol`. The word "performance" is often used to describe both speed and memory usage, so it can be helpful to be more precise and use words like "speed" or "time"

thunder/dynamo/benchmark_utils.py
```diff
@@ -370,83 +404,127 @@ def _check_tensor(x):
...
+    perf_record = False
+    memory_record = False
     log_strs = ""
     for m1, m2, name in zip(measure1, measure2, ("forward", "backward")):
-        if m1 is None:
-            assert m2 is None
-            continue
-        ret = check_threshold_log(
-            m1.median, m2.median, compile_fn1.name, compile_fn2.name, f"{graph_name} {name}", timer_name, rtol, atol
+        if timer_fn.name == "WallTimeWithMemoryUsage":
```
mruberry, line 463: This is OK for now, but in the future I don't think we want to rely on particular names to know whether to check memory usage or not.

One idea is to create a base class, like, `MaxAllocatedMemoryTimer`, that has an abstract method like `get_max_allocated()`, and then this check could become `if isinstance(timer, MaxAllocatedMemoryTimer)`. Then if someone wanted to implement a customer max allocated memory timer they would just inherit from this class. Similarly if we wanted to support reserve memory timers then there could be a `MaxReservedMemoryTimer` base class and timers could even inherit from both if they wanted to.

thunder/dynamo/benchmark_utils.py
mruberry: Below I suggest the extensibility point be refined by checking for a base class

thunder/dynamo/benchmark_utils.py
```diff
@@ -329,7 +345,27 @@ def get_pretty_time_str(a):
...
+    """
+    Converts memory size in bytes to human readable string with appropriate unit suffix.
+
+    Args:
+        value (float): Memory size in bytes
+
+    Returns:
+        str: Memory size with appropriate unit (B, KB, MB, GB, TB, PB)
+    """
+    suffixes = ["B", "KB", "MB", "GB", "TB", "PB"]
```
mruberry, line 358: Neat

thunder/dynamo/benchmark_utils.py
```diff
@@ -329,7 +345,27 @@ def get_pretty_time_str(a):
...
+    suffix_idx = 0
+
+    while suffix_idx + 1 < len(suffixes) and value >= 1024.0:
+        value /= 1024.0
```
mruberry, line 362: So here's the thing: is a kilobyte 1000 bytes or 1024 bytes?

See https://en.wikipedia.org/wiki/Byte#Multiple-byte_units for a discussion.

I'm not sure what convention PyTorch or NVIDIA uses (if any). To be as clear as possible I would use "KiB" (and "MiB", ...) to refer to kibibytes (mebibytes, ...) explicitly.

thunder/dynamo/benchmark_utils.py
```diff
@@ -370,83 +404,127 @@ def _check_tensor(x):
...
-    atol=0.0,
+    perf_rtol=0.5,
+    perf_atol=0.0,
+    memory_usage_rtol=0.5,
+    memory_usage_atol=0.0,
+    stream: TextIO = sys.stdout,
 ):
     """
-    Check the timing of graph report using two different compilation specifications with the provided timer configuration
+    Check the timing and memory usage(if using WallTimeWithMemoryUsage) of graph report using two different compilation specifications with the provided timer configuration
```
mruberry, line 430: "usage(" -> "usage ("

mruberry (approved):
I made some notes inline for your review, @kiya00. Looking forward to hearing your thoughts -- let me know when you'd like this merged

kiya00:
Hi @mruberry , thanks for the explanation, I added the comments and TODO in this PR, will start another one to make the memory measurement more extensible

mruberry:
@t-vi Looks like this is running afoul of the distributed test CI, so it will need your help to merge

PR #1828, opened 2025-03-03
Remove `FutureTensorLike` and use `FutureTensorProxy` thoroughly

crcrpar:
## What does this PR do?

Because it's purely an alias to `FutureTensorProxy` and it's not quite used.


t-vi (approved):
Thank you @crcrpar

Borda:
seems after this PR, we have a failing test on master...
cc: @t-vi

PR #1829, opened 2025-03-03
improve imports and type annotation of `thunder.executors.passes`

crcrpar:
## What does this PR do?

As per title, cleaning imports up and adding one type annotation.


t-vi (approved):
Thank you @crcrpar

PR #1831, opened 2025-03-04
`f.write(...)` <- `print(..., file=f)`

crcrpar:
## What does this PR do?

`code_str` seems to be guaranteed to be a string at https://github.com/Lightning-AI/lightning-thunder/blob/ff74bef8f31ae2624ae7a78a9cafa37755eb2cf2/thunder/dynamo/report.py#L470, thus it'd be reasonable to use `f.write` instead.

### DIFF
--- a/thunder/dynamo/report.py
+++ b/thunder/dynamo/report.py
@@ -472,7 +472,7 @@ class FXGraphReport:
         if file_name is None:
             file_name = f"{self.graph_name}.py"
         with open(folder / file_name, "w") as f:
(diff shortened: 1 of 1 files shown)


t-vi:
I don't have a preference for what dynamo/report.py does, but I don't really see the advantage here. 
To me, the print looks more intuitive. It's does not seem perf relevant. Also, if the new thing wants to be equivalent I think we're missing a newline.

t-vi (approved):
If you want (see comment)

PR #1834, opened 2025-03-04
Update vjp implementation to attach residuals to all outputs

IvanYashchuk:
This PR fixes the residuals/saved_for_backward saving and retrieving so that that backward function generation works correctly in the case when the first output of a bound symbol is unused.


IvanYashchuk:
@t-vi, could you please review this PR?

IvanYashchuk:
@t-vi, could you please review this PR?

t-vi:
Is this still  needed btw?

PR #1836, opened 2025-03-05
Update cuDNN Frontend version to 1.10.0

IvanYashchuk:
We were using a rather old cuDNN Frontend package, this PR updates to the latest version. 

cc @borda


IvanYashchuk:
@Borda, there's probably a problem with the Cache PIP step in CI. Could you please take a look?

crcrpar (approved):
I know the rule of thumb is one change per one pull request. We should also think of bumping cuda and pytorch to 12.4 and 2.6, respectively.

Borda (approved):

Borda:
updating dockers in https://dev.azure.com/Lightning-AI/lightning/_build/results?buildId=228304&view=logs&j=5590d1c6-4d73-5164-c2d0-7f7286b3996d

Issue #1768, opened 2025-02-17
Taking `requires_grad` propagation seriously in Thunder


IvanYashchuk:
## 🚀 Feature

Add correct `requires_grad` attribute propagation in Thunder with OpInfo-based tests.

Currently, Thunder queries the `requires_grad` attribute only for compiled function inputs (nn.Module parameters, buffers, and args) and uses this attribute to cache compiled artifacts and mask the output of the compiled backward function.

We could also faithfully model the propagation of this attribute, as in PyTorch Eager, or consistently, according to the rules that we need to develop. However, the exact benefits are unclear and need to be collected to consider whether it's worth the effort.

If we decide to do the `requires_grad` propagation we would be able to faithfully mark some outputs of the compiled function with `requires_grad=False` by disconnecting them from the Autograd graph (https://github.com/Lightning-AI/lightning-thunder/issues/1733). The performance benefits of doing so remain to be evaluated.

t-vi:
So my thinking is that globally:

- we have requires_grad on the inputs (easy!),
- we need to propagate automatically during the autodiff pass, so that is where we determine which outputs need grad,
- we need to keep it correct on outputs in further transformations.

In the autodiff pass:

- bsyms things that require grads on their outputs can things requiring grad as output by default (limited to floating point tensors),
- we only want to materialize `put_grad` for tensors that needing gradients and then "dce" things,
- the things that require grad are the `get_grad`s we have left, 
- there likely needs to be an escape hatch, but probably only for the grad_transform (joint forward / backward) style for things that have a requires_grad pattern that cannot be automatically inferred, so those need to set requires_grad as needed.

Probably one simple interesting case is `a * b` with various combinations of `requires_grad` on `a` and `b`.

PR #1838, opened 2025-03-05
TE: don't compute wgrad when weights are frozen

kshitij12345:
Problem - Currently, TE executor just assumes that we will be computing gradients for both input and weight. This is true for full finetuning, but in PEFT scenario the weights are frozen i.e. requires_grad=False. This leads to poor performance as the TE executor ends up doing unnecessary work of computing gradients for weights.

Solution - Based on the `requires_grad` on TensorProxy for weight and bias, enable/disable the relevant gradient computation.

NOTE - Thunder doesn't reliable propagate the requires_grad on intermediate values (see https://github.com/Lightning-AI/lightning-thunder/issues/1768) so we just assume that requires_grad on `inp` is True. However, it should be safe to check the `requires_grad` on Parameters as they are inputs to the jitted function.

### DIFF
--- a/thunder/distributed/prims.py
+++ b/thunder/distributed/prims.py
@@ -184,7 +184,14 @@ def synchronize_meta(
             # Assuming that the sharding is done on the first dimension
             # See [FSDP Sharding] in distributed/__init__.py
             unsharded_shape = a.shape[0] * group.size(), *a.shape[1:]
-            return TensorProxy(shape=unsharded_shape, like=a, distparallel_type=DistParallelType.REPLICATED)
+            # NOTE: We explicitly pass the `a.requires_grad` as we want to propogate the requires grad correctly on `nn.Parameter`'s
+            # as it maybe queried later.
+            return TensorProxy(
+                shape=unsharded_shape,
+                like=a,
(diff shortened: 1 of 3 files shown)


thunder/distributed/prims.py
```diff
@@ -184,7 +184,14 @@ def synchronize_meta(
...
             unsharded_shape = a.shape[0] * group.size(), *a.shape[1:]
-            return TensorProxy(shape=unsharded_shape, like=a, distparallel_type=DistParallelType.REPLICATED)
+            # NOTE: We explicitly pass the `a.requires_grad` as we want to propogate the requires grad correctly on `nn.Parameter`'s
+            # as it maybe queried later.
+            return TensorProxy(
+                shape=unsharded_shape,
+                like=a,
+                distparallel_type=DistParallelType.REPLICATED,
+                requires_grad=a.requires_grad,
+            )
```
kshitij12345, line 194: This change was required to get the TE FSDP tests working again.

Without this, the `requires_grad` on the sharded weight tensor is not propagated correctly.

kshitij12345:
Ping @t-vi for review and stamp

PR #1839, opened 2025-03-05
Add softmin op

beverlylytle:
Adds support for the `softmin` op, which is nearly identical to the `softmax` op.  However, unlike for `softmax` there is no `torch.softmin`  nor `torch.Tensor.softmin`.  

Also, it looks like `softmax` is supported on CPU for float16 now.

### DIFF
--- a/thunder/tests/opinfos.py
+++ b/thunder/tests/opinfos.py
@@ -8174,6 +8174,25 @@ softmax_opinfo = OpInfo(
 nn_ops.append(softmax_opinfo)
 
 
+softmin_opinfo = OpInfo(
+    ltorch.softmin,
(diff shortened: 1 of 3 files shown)


thunder/tests/opinfos.py
```diff
@@ -8161,38 +8161,34 @@ def softmax_sample_generator(op, device, dtype, requires_grad, **kwargs):
     sample_input_generator=softmax_sample_generator,
     torch_reference=torch.softmax,
     dtypes=(datatypes.floating,),
-    test_directives=(
```
mruberry, line 8164: Awesome!

mruberry (approved):

PR #1840, opened 2025-03-05
Decrease nvfuserex overhead by 10%.

csarofeen:
Found a couple minor changes make a pretty sizable difference to nvFuser latencies.

### DIFF
--- a/thunder/executors/nvfuserex_impl.py
+++ b/thunder/executors/nvfuserex_impl.py
@@ -464,9 +464,10 @@ class FusionDefinitionWrapper:
     disable_options: None | list[str] = None
 
     def __call__(self, *args):
-        fd = self.get_fd(self.to_descriptors(args))
-        self.last_used = fd
-
+        torch.cuda.nvtx.range_push("nvfuser_ex.__call__")
+        if self.last_used is None:
+            self.last_used = self.get_fd(self.to_descriptors(args))
+        fd = self.last_used
         if self.store_inputs:
             self.last_inputs = args
 
@@ -480,18 +481,18 @@ class FusionDefinitionWrapper:
         if hasattr(fd, "_selected_device"):
             kwargs["device"] = fd._selected_device
 
-        if nvfuser_version() >= LooseVersion("0.2.23"):
-            # nvFuser expects empty list instead of None values.
-            kwargs["_enable_options"] = self.enable_options if self.enable_options is not None else []
-            kwargs["_disable_options"] = self.disable_options if self.disable_options is not None else []
-
+        if NVFUSER_SUPPORTS_OPTIONS:
+            kwargs["_enable_options"] = self.enable_options or []
+            kwargs["_disable_options"] = self.disable_options or []
         elif self.enable_options or self.disable_options:
             warnings.warn(
                 f"nv_enable_options/nv_disable_options require nvFuser version 0.2.23 and above, found version {nvfuser_version()}. These options will be ignored."
             )
 
         with annotate_for_profile(self.name):
-            return fd.execute(args, **kwargs)
+            results = fd.execute(args, **kwargs)
+            torch.cuda.nvtx.range_pop()
+            return results
 
     def __repr__(self):
         return f"FusionDefinitionWrapper({self.name})"
@@ -2192,7 +2193,6 @@ def copy_(
 
 register_supported(PrimIDs.COPY_, copy_, _copy__check)
 
-
 # Removes excessive float casts, like those that occur when autocasting
 # NOTE This passes actually changes a program's semantics, because it will take a sequence like
 #   fp32 -> fp16 -> fp32 and remove all the operations, but casting fp32 values to fp16 can
@@ -2708,3 +2708,8 @@ def embedding(
 
 register_supported(PrimIDs.EMBEDDING, embedding, _embedding_check)
 register_supported(ltorch.embedding, embedding, _embedding_check)
+
+
+# At module/class level
+NVFUSER_SUPPORTS_OPTIONS = nvfuser_version() >= LooseVersion("0.2.23")
+


thunder/executors/nvfuserex_impl.py
```diff
@@ -464,9 +464,10 @@ class FusionDefinitionWrapper:
     disable_options: None | list[str] = None
 
     def __call__(self, *args):
-        fd = self.get_fd(self.to_descriptors(args))
-        self.last_used = fd
-
+        torch.cuda.nvtx.range_push("nvfuser_ex.__call__")
+        if self.last_used is None:
+            self.last_used = self.get_fd(self.to_descriptors(args))
+        fd = self.last_used
```
mruberry, line 470: IIRC, I think the reason the args are transformed to descriptors is that Thunder doesn't model properties of tensors, like strides, that nvFuser might use to pick among multiple file descriptors for the same fusion; maybe that's not the case anymore.

I'm not sure if we have any tests that check for calling the same fusion with tensors that pass Thunder's checks but have different strides. If not, we should probably add one

fyi @IvanYashchuk

thunder/executors/nvfuserex_impl.py
```diff
@@ -480,18 +481,18 @@ def __call__(self, *args):
         if hasattr(fd, "_selected_device"):
             kwargs["device"] = fd._selected_device
 
-        if nvfuser_version() >= LooseVersion("0.2.23"):
-            # nvFuser expects empty list instead of None values.
-            kwargs["_enable_options"] = self.enable_options if self.enable_options is not None else []
-            kwargs["_disable_options"] = self.disable_options if self.disable_options is not None else []
-
+        if NVFUSER_SUPPORTS_OPTIONS:
```
mruberry, line 484: If we really want to optimize this Python, the __call__ method can be generated differently depending on this option. Not performing the LooseVersion construction and the comparison in the function is a great improvement

thunder/executors/nvfuserex_impl.py
```diff
@@ -464,9 +464,10 @@ class FusionDefinitionWrapper:
     disable_options: None | list[str] = None
 
     def __call__(self, *args):
-        fd = self.get_fd(self.to_descriptors(args))
-        self.last_used = fd
-
+        torch.cuda.nvtx.range_push("nvfuser_ex.__call__")
```
IvanYashchuk, line 467: Adding NVTX markers in Python has a cost (less than 1 μs, more than 200 ns). This should be done only when profiling, not during regular execution. Other Python profiling tools, such as cProfile or py-spy, can be used to measure performance with per-function granularity with no additional markers in the codebase.

We have optional NVTX markers available through `annotate_for_profile`:
https://github.com/Lightning-AI/lightning-thunder/blob/69ae4d66e45be68115b260f5ad7af07a75dd8a5c/thunder/core/profile.py#L49-L51
Let's use it here instead of `range_push` + `range_pop`:
```py
@annotate_for_profile("nvfuser_ex.__call__")
def __call__(self, *args):
```
After this the markers will be added if the special environment variable is set to 1 `THUNDER_ANNOTATE_TRACES=1`:
https://github.com/Lightning-AI/lightning-thunder/blob/69ae4d66e45be68115b260f5ad7af07a75dd8a5c/thunder/core/profile.py#L10

thunder/executors/nvfuserex_impl.py
```diff
@@ -480,18 +481,18 @@ def __call__(self, *args):
...
         elif self.enable_options or self.disable_options:
             warnings.warn(
                 f"nv_enable_options/nv_disable_options require nvFuser version 0.2.23 and above, found version {nvfuser_version()}. These options will be ignored."
             )
 
         with annotate_for_profile(self.name):
-            return fd.execute(args, **kwargs)
+            results = fd.execute(args, **kwargs)
+            torch.cuda.nvtx.range_pop()
+            return results
```
IvanYashchuk, line 495: No code changes would be necessary here if the NVTX annotation is applied with a decorator
```py
@annotate_for_profile("nvfuser_ex.__call__")
def __call__(self, *args):
```

thunder/executors/nvfuserex_impl.py
IvanYashchuk: We can just drop the check and require nvFuser 0.2.23+

thunder/executors/nvfuserex_impl.py
IvanYashchuk: Unfortunately, as Mike says we can't generate the FusionDefinition once and use it for every later call. We need to be pessimistic here and expect that at runtime strides may change.
This type of execution should definitely be available as an option when we are sure strides will be the same for every invocation. It would also be great if nvFuser provided a way to directly execute the last compiled fusion bypassing all shape and stride checks.

csarofeen:
Thanks for your comments this makes sense. I wonder if we should change that. It should be on nvFuser to generate a new program if necessary on contiguity changes, not on Thunder. Am I missing something here @jacobhinkle @kevinstephano ?

thunder/executors/nvfuserex_impl.py
mruberry: If we do want to targets 10s of nanoseconds, there are probably some optimizations we could make by compiling more of the control-flow and nvtx decisions. If we assume the options are fixed when the fusion is constructed, their dictionary could also be prepopulated, which would avoid dictionary construction and lookups during call time. If we assume `self.store_inputs_meta` is fixed at compiletime, we don't need to branch on it. Etc.

Going a little farther, I think the operations in the fusion are the same regardless of file descriptor, so whether there's a "selected device" or not probably doesn't require a query for each file descriptor, and can probably happen at compile-time, too.

If we were to make all these changes, we'd probably save a couple hundred nanoseconds to a microsecond per fusion invocation

t-vi: To my mind we are at microseconds.

jacobhinkle:
> Thanks for your comments this makes sense. I wonder if we should change that. It should be on nvFuser to generate a new program if necessary on contiguity changes, not on Thunder. Am I missing something here @jacobhinkle @kevinstephano ?

We should be able to do that. Note that this will mean we would need to do concretization always, even when we have static shapes, since the contiguity could change. We would then compute as descriptor the stride order and contiguity of each input tensor and that would go into the concretization info which is a cache key for the top level of the FEC cache.

thunder/executors/nvfuserex_impl.py
IvanYashchuk: nvFuser 0.2.23+ is the new requirement added in https://github.com/Lightning-AI/lightning-thunder/pull/1840/commits/f11214adb9e77d4d8e5d0aa3f055cb9e92816179.

thunder/executors/nvfuserex_impl.py
IvanYashchuk: It turns out the NVTX markers were set by default and https://github.com/Lightning-AI/lightning-thunder/pull/1846 trims 1-2μs of CPU overhead from nvFuser regions execution.

csarofeen:
This is showing about a 6% overhead reduction. Thank you!

mruberry (approved):

t-vi (approved):
Thank you @csarofeen

PR #1841, opened 2025-03-06
refactoring `Proxy.__init__`'s logic to get `prefix`

crcrpar:
## What does this PR do?

As per title, by making it mandatory for each proxy class inheriting `ProxyInterface` to define a class variable of the default prefix and a property that returns the appropriate prefix of it.

cc @mruberry @lantiga


IvanYashchuk (approved):
Great refactoring and that allows for the future when out-of-core proxy objects exist without requiring changes on the Thunder side or always passing the prefix argument.

t-vi (approved):

PR #1843, opened 2025-03-06
Improve cache lookup overheads in FusionDefinitionWrapper

IvanYashchuk:
Inspired by the changes from https://github.com/Lightning-AI/lightning-thunder/pull/1096, the next step is to utilize only the necessary information to construct contiguity and stride order information from runtime tensors. Therefore, the shape and dtype are now always derived from ProxyTensors and this information is available at the trace construction time.

Here are the results of measuring CPU time for the `to_descriptors` function which is used to build the cache key:
image

Here's the breakdown for different components of the call method for the current PR:
image

"Other" is something that is not `to_descriptors`, `get_fd`, `fd.execute`. For example, version check, which is being removed in https://github.com/Lightning-AI/lightning-thunder/pull/1840.

cc @tfogal

### DIFF
--- a/thunder/executors/nvfuserex_impl.py
+++ b/thunder/executors/nvfuserex_impl.py
@@ -264,8 +264,9 @@ def create_fd(
                 lc_to_nv_map[x] = nv
             elif isinstance(x, TensorProxy):
                 utils.check_type(y, tuple)
-                symbolic_shape, contiguity, stride_order, dtype = y
-                nvdtype = lcdtype_to_nvdtype(dtypes.to_dtype(dtype))
+                contiguity, stride_order = y
+                symbolic_shape = compute_symbolic_shape(x._shape, x._shape)
+                nvdtype = lcdtype_to_nvdtype(x.dtype)
                 is_cpu = x.device == cpu
                 nv = fd.define_tensor(
                     shape=symbolic_shape, contiguity=contiguity, dtype=nvdtype, stride_order=stride_order, is_cpu=is_cpu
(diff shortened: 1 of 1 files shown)


csarofeen:
One thing I’d like to compare is how fast this is relative to:
https://github.com/NVIDIA/Fuser/blob/e4a93a5613b9b220d6e79c81257a8d585d581fb6/csrc/runtime/fusion_cache_utils.cpp#L279-L326

This code is what we’re using in nvFuser to cache sizes/strides internally in nvFuser. There’s unfortunately not an nvtx range on this so you can’t simply check without rebuilding nvFuser.

We use this to generate a cache id and set it in fusion executor cache: https://github.com/NVIDIA/Fuser/blob/e4a93a5613b9b220d6e79c81257a8d585d581fb6/csrc/runtime/fusion_executor_cache.cpp#L111

If caching takes a while we could actually generate this once and set it directly into FEC.

The only challenge is I don’t think this scales well to when contiguity is changing. As I don’t think contiguity changes are part of our concretization pass CC @jacobhinkle to check my knowledge here.

thunder/executors/nvfuserex_impl.py
```diff
@@ -400,48 +399,20 @@ def compute_contiguity(
...
-    return compute_symbolic_shape(proxy_shape, shape), *compute_contiguity(shape, stride)
-
-
-def to_descriptors(proxy_args, args) -> tuple:
-    def to_descriptor(proxy_arg, arg):
-        if isinstance(arg, Tensor):
-            return (*compute_tensor_descriptor(proxy_arg._shape, arg.shape, arg.stride()), arg.dtype)
-        elif isinstance(arg, Number):
-            return type(arg)
-        elif isinstance(arg, tuple):
```
mruberry, line 432: fyi @jjsjann123

jacobhinkle:
> The only challenge is I don’t think this scales well to when contiguity is changing. As I don’t think contiguity changes are part of our concretization pass CC @jacobhinkle to check my knowledge here.

That's right. Contiguity is currently encoded as part of the fusion definition and we do not modify it when we get new inputs. Instead there will be a separate FEC when you have inputs with different contiguity/stride order.

PR #1844, opened 2025-03-06
Remove `from torch import Tensor`

crcrpar:
## What does this PR do?

Remove `from torch import Tensor` as it's only used once and the other parts seem to use `torch.Tensor`.

### DIFF
--- a/thunder/executors/nvfuserex_impl.py
+++ b/thunder/executors/nvfuserex_impl.py
@@ -12,7 +12,6 @@ import warnings
 
 from looseversion import LooseVersion
 import torch
-from torch import Tensor
 
 import thunder.core.dtypes as dtypes
 import thunder.torch as ltorch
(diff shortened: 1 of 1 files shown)


thunder/executors/nvfuserex_impl.py
```diff
@@ -425,7 +424,7 @@ def compute_tensor_descriptor(
 
 def to_descriptors(proxy_args, args) -> tuple:
     def to_descriptor(proxy_arg, arg):
-        if isinstance(arg, Tensor):
+        if isinstance(arg, torch.Tensor):
```
IvanYashchuk, line 427: The reason to use `Tensor` directly compared to `torch.Tensor` is an attempt to have lower cost of this function. Let's take a look at what I think is about the average case of 16 inputs to the fusion:
```py
In [1]: import torch

In [2]: from torch import Tensor

In [3]: args = [torch.ones((2, 2)) for _ in range(16)]

In [4]: %%timeit
   ...: [isinstance(arg, Tensor) for arg in args]
   ...:
   ...:
1.15 µs ± 20.1 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
In [5]: %%timeit
   ...: [isinstance(arg, torch.Tensor) for arg in args]
   ...:
   ...:
1.66 µs ± 9.2 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

IvanYashchuk (commented):
In general, I agree that consistency in the codebase is important and it's probably better to use `torch.Tensor` for readability.
However, we shouldn't merge this change because it's on the CPU overhead critical path where any additional opcode counts. In Python using `torch.Tensor` is slower than just `Tensor` directly.

PR #1845, opened 2025-03-06
Add an option for skipping nvFuser wrapper cache

IvanYashchuk:
New option for bypassing Thunder level cache for nvFuser regions: `nv_skip_cache=True`.

We need an additional layer of caching to correct nvFuser fusion definitions in case the stride information of input changes at runtime. Thunder is designed to work with any stride without rebuilding the whole trace by default. This caching layer adds some amount of CPU overhead to each invocation of a fusion region in the Thunder-generated program and this option facilitates removing these overheads when the stride dynamism feature is not needed.

cc @tfogal


mruberry (approved):

csarofeen:
I don't think this is necessary with your other latency improvements. It does not appear to be a significant bottleneck anymore.

PR #1847, opened 2025-03-06
Inplace dtype mismatch

jjsjann123:
We could have a.add_(b), where a and b have different dtypes.

Type promotion would ended up having us trying to trigger inplace copy with mismatching dtypes.

### DIFF
--- a/thunder/torch/__init__.py
+++ b/thunder/torch/__init__.py
@@ -227,6 +227,7 @@ def register_function(torchfn, thunderfn_impl):
 
 def _copy_(a, b, /):
     cd = get_compile_data()
+    b = clang.maybe_convert_to_dtype(b, a.dtype)
     return prims.copy_(b, a, grad_enabled=cd.is_grad_enabled if cd is not None else False)
 
 
--- a/thunder/tests/test_inplace_copy.py
+++ b/thunder/tests/test_inplace_copy.py
@@ -191,3 +191,15 @@ def test_inplace_copy_of_leaf_requiring_grad_fails(executor, device, dtype):
     a = make_tensor((4, 4), device=device, dtype=tdtype, requires_grad=True)
     with pytest.raises(RuntimeError):
         jitted_fn(a)
+
+@instantiate(executors=(TorchExecutor,), dtypes=datatypes.float_math_dtypes)
+def test_inplace_copy_dtype_mismatch(executor, device, dtype):
+    def fn(x, y):
+        x.add_(y)
+
+    jitted_fn = executor.make_callable(fn)
+    tdtype = ttorch.to_torch_dtype(dtype)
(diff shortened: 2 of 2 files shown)


thunder/torch/__init__.py
```diff
@@ -227,6 +227,7 @@ def register_function(torchfn, thunderfn_impl):
 
 def _copy_(a, b, /):
     cd = get_compile_data()
+    b = clang.maybe_convert_to_dtype(b, a.dtype)
```
crcrpar, line 230: though it's not related to this pr but src and dst would be more readable than b and a

crcrpar (approved):

thunder/torch/__init__.py
beverlylytle: In [`copy__meta`](https://github.com/Lightning-AI/lightning-thunder/blob/56b922a4ba0eb87b3f5d1cb605648dbd571b6f96/thunder/core/prims.py#L4118) there are checks to enforce that `a` and `b` have the same shape, dtype and device.  Looking at the [torch documentation for `copy_`](https://pytorch.org/docs/stable/generated/torch.Tensor.copy_.html), none of those things are required to be true for Pytorch and empirical evidence suggests that `TensorProxy(like=copy_to)` conforms to Pytorch's behavior. @kiya00 As the author of `copy__meta`, do you remember why these checks were added?

On a side note, maybe it already did, but I noticed that `_copy_` confirms that Thunder passes the Bechdel test for software, so that's nice.

kiya00: IIRC(a bit vague though), there's no general inplace support back then, the copy is only added as an internal prim operator to support the batchnorm for the inplace update of stats (correspond to `fd.add_output(output, alias_input)`), so I didn't actually refer to the `torch.copy_` implementation and assumed it's the same type

kiya00 (approved):

thunder/torch/__init__.py
crcrpar: torch'a copy allow different dtype args for many ops but in thunder we usually cast tensors to their common dtype accordingly (see clang's elementwise binary op wrapper) so I think it makes the behavior consistent with the existing ops to have this type conversion here.

beverlylytle: Ok, but I guess there is still the shape issue:
```
def f(x, y):
    x.copy_(y)
    return x

a = torch.rand(10)
b = torch.ones(1)

f(a, b)  # this is ok

jf = thunder.jit(f)
jf(a, b)  # this is not ok
```

But I suppose this is out of scope.

beverlylytle (approved):

thunder/torch/__init__.py
mruberry: @beverlylytle Would you file an issue (or link an existing issue) tracking those?

PR #1848, opened 2025-03-07
add `std` op

khushi-411:
As per title.

### TODO
- [x] add grads in thunder/core/transforms.py


mruberry:
@beverlylytle Would you like to review this?

thunder/core/transforms.py
```diff
@@ -1761,6 +1761,24 @@ def var_backward(a, dim, correction, v, g):
...
+@register_backward(prims.PrimIDs.STD)
+def std_backward(a, dim, correction, s, g):
+    n_elem_reduced = a.numel() // s.numel() if a.numel() != 0 else 1
+    normalization_scalar = n_elem_reduced - correction
+    g = restore_reduced_dims(g, dim, a.shape)
+    if a.dtype != s.dtype:
+        a = prims.convert_element_type(a, s.dtype)
+    mean = prims.sum(a, dim) / n_elem_reduced
+    mean = restore_reduced_dims(mean, dim, a.shape)
+    return (g * (a - mean)) / (normalization_scalar * (s + 1e-8))
```
beverlylytle, line 1779: I know this is still a draft PR not up for review yet, but I wanted to give you a pointer on this return value that I think is causing the grad tests to fail.

I assume that this 1e-8 is to protect against dividing by zero.  If `s`, the standard deviation of `a`, is very small, then (practically) all values in `a` are going to be very close to its mean, meaning `a - mean` is filled mostly with very small values.   While `1e-8` is itself very small, in the case that `s` is also very small, the additional constant is going to have a huge effect on the returned value.  There are also some issues with the conventions around infinity and NaNs.  This [torch implementation](https://github.com/pytorch/pytorch/blob/00199acdb85a4355612bff28e1018b035e0e46b9/torch/csrc/autograd/FunctionsManual.cpp#L1849) gets around this problem by returning zero when `s` is zero.  Hopefully this should allow you to reenable the tests. 

Thank you for your efforts!

thunder/tests/opinfos.py
```diff
@@ -5800,6 +5800,63 @@ def glu_error_generator(op, device, **kwargs):
...
+            dtypes=(datatypes.float64,),
+        ),
+        # AssertionError: Tensor-likes are not close!
+        DecorateInfo(
+            pytest.mark.skip,
+            "test_phantom_grad_vs_torch_consistency",
+            dtypes=(datatypes.bfloat16, datatypes.float16, datatypes.float32, datatypes.float64),
+        ),
+    ),
+)
```
riccardofelluga, line 5856: @khushi-411 Interesting! What are the sources of these two errors?

khushi-411: @riccardofelluga, earlier, this was happening because my backward implementation was not correctly handling `nans`.

In the current implementation, there are still precision errors. I've added `custom_comparator` to handle it. I'll research more on this.

thunder/core/transforms.py
beverlylytle: Good, glad `masked_fill` helped with the NaNs, but I think there are still some issues here.  Standard deviation is the square root of variance.  Since variance is non-negative, any input given to `var` should also be valid for `std`.  Moreover, since square root is continuously differentiable on this domain, the vjp/grad for `std` shouldn't really have any numerical issues for any input that works for the vjp of `var`.  All this to say that I believe the sample generator and the test directives used for `std` should be the same as for `var`.  Since the tests are failing, there must be still some issue with this backward implementation.  Since the major difference between the backward for `var` and the backward for `std` is this `s`, I suspect that something must be off there.

khushi-411: Agree. I'm looking into this. will get back to you. Thank you!

thunder/tests/opinfos.py
khushi-411: You & @beverlylytle were so right!! The precision issues were because `dim != 0` was not correctly handled earlier. `unsqueezing` the `s` to `dim` dimension helped solve this issue.

thunder/core/transforms.py
```diff
@@ -1761,6 +1761,31 @@ def var_backward(a, dim, correction, v, g):
...
+    g = restore_reduced_dims(g, dim, a.shape)
+    if a.dtype != s.dtype:
+        a = prims.convert_element_type(a, s.dtype)
+    mean = prims.sum(a, dim) / n_elem_reduced
+    mean = restore_reduced_dims(mean, dim, a.shape)
+
+    if len(dim) == 1:
+        dim = dim[0]
+    if dim != 0:
+        s = s.unsqueeze(dim)
```
beverlylytle, line 1783: Looks like the tests are passing now.  Great!  But I do wonder: what's the difference between this and `s = restore_reduced_dims(s, dim, a.shape)`?

khushi-411: my bad.

fixed!

beverlylytle (approved):
Fantastic, thank you!

mruberry (approved):

khushi-411:
Thank you for all your insightful suggestions, @beverlylytle @riccardofelluga @mruberry!

PR #1853, opened 2025-03-07
[ThunderFX report] Uses fd in `run_benchmark` instead of `FusionDefinitionWrapper`

kiya00:
<details>

</details>

## What does this PR do?

The old `run_benchmark` for nvfusion symbol measures the callable `FusionDefinitionWrapper`, but saves the `fd.execute` in the benchmarking script. There's extra overhead of `FusionDefinitionWrapper`, so the wall time are different between `run_benchmark` and the saved script.
This PR makes their behavior uniform: both use `fd.execute` since it's hard to serialize `FusionDefinitionWrapper` into code string.

### DIFF
--- a/thunder/dynamo/benchmark_utils.py
+++ b/thunder/dynamo/benchmark_utils.py
@@ -140,7 +140,7 @@ class BoundSymbolNvfuserSpecification(CompileSpecificationInterface):
     # See the TODO in :class:`thunder.dynamo.report.ThunderFusionReport` for more details.
     def compile(self, nvfusion_bsym):
         fd = nvfusion_bsym._call_ctx[nvfusion_bsym.sym.name].last_used
-        return fd.execute
+        return lambda *args: fd.execute(args)
 
 
 class BoundSymbolTorchCompileSpecification(CompileSpecificationInterface):
--- a/thunder/dynamo/benchmark_utils.py
+++ b/thunder/dynamo/benchmark_utils.py
@@ -139,7 +139,8 @@ class BoundSymbolNvfuserSpecification(CompileSpecificationInterface):
     # Returns the nvFuser callable from the nvFuser bound symbol.
     # See the TODO in :class:`thunder.dynamo.report.ThunderFusionReport` for more details.
     def compile(self, nvfusion_bsym):
-        return nvfusion_bsym._call_ctx[nvfusion_bsym.sym.name]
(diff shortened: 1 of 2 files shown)


thunder/dynamo/report.py
```diff
@@ -796,6 +796,8 @@ def __str__(self):
     def run_benchmark(self, compile_fn: CompileSpecificationInterface, timer_fn: TimerInterface):
         compiled_fn = compile_fn.compile(self.nvfusion_bsym)
         inputs = self.make_example_inputs()
+        if compile_fn.name == "nvfuser":
```
mruberry, line 799: This is OK for now; in the future we probably want to let the compile specifications define a function that gets call on the inputs so that people can add their own specifications for nvfuser

thunder/dynamo/report.py
```diff
@@ -804,6 +806,8 @@ def run_repro(
     ):
         compiled_fn = compile_fn.compile(self.nvfusion_bsym)
         inputs = self.make_example_inputs()
+        if compile_fn.name == "nvfuser":
+            return compiled_fn(inputs)
         return compiled_fn(*inputs)
```
kiya00, line 811: here because `fd.execute` takes input as list, is different from the torch.compile

thunder/dynamo/benchmark_utils.py
```diff
@@ -139,7 +139,8 @@ def __init__(self, specification_name="nvfuser"):
     # Returns the nvFuser callable from the nvFuser bound symbol.
     # See the TODO in :class:`thunder.dynamo.report.ThunderFusionReport` for more details.
     def compile(self, nvfusion_bsym):
-        return nvfusion_bsym._call_ctx[nvfusion_bsym.sym.name]
+        fd = nvfusion_bsym._call_ctx[nvfusion_bsym.sym.name].last_used
+        return fd.execute
```
kshitij12345, line 143: This can probably be changed to 
```python
return lambda *args: fd.execute(args)
```

With this change, `if compile_fn.name == "nvfuser"` in `report.py` shouldn't be required.

Though a question would be that this will add a slight overhead of calling into the lambda first, not sure if that is acceptable.

mruberry (approved):
Cool!

fyi @t-vi , looks like CI is currently experiencing timeouts (unrelated to this PR)

kshitij12345 (approved):
LGTM, thanks @kiya00

Issue #1855, opened 2025-03-07
[reporting] Saved repro may have incorrect indentation.


kshitij12345:
Context - I am trying to get a minimal repro for failure running "microsoft/Phi-3-vision-128k-instruct".

Repro

```python
import torch
from transformers import AutoModelForCausalLM, AutoConfig
from transformers.models.phi3 import Phi3Config

model_id = "microsoft/Phi-3-vision-128k-instruct"

# Initialize the pre-trained model
cfg = AutoConfig.from_pretrained(model_id, trust_remote_code=True)
cfg.num_hidden_layers = 2

from thunder.dynamo import thunderfx
from thunder.dynamo.report import get_thunder_fxgraph_reports, fx_report, ThunderCompileSpecification

with torch.device("cuda"):
    model = AutoModelForCausalLM.from_config(cfg, trust_remote_code=True, torch_dtype=torch.bfloat16)

    input_ids = torch.randint(0, 200, (1, 512)) # torch.int64 cuda:0 torch.Size([1, 512]) tensor(1, device='cuda:0') tensor(32010, device='cuda:0')
    attention_mask = torch.zeros((1, 512), dtype=torch.int64)
    attention_mask[0, 426:] = 1
    # torch.uint8 cuda:0 torch.Size([1, 400, 356, 3]) tensor(0, device='cuda:0', dtype=torch.uint8) tensor(255, device='cuda:0', dtype=torch.uint8)
    pixel_values = torch.randint(0, 255, (1, 400, 356, 3), dtype=torch.uint8)
    labels = input_ids.clone().detach()

    reports = fx_report(model, **dict(input_ids=input_ids, attention_mask=attention_mask, pixel_values=pixel_values, labels=labels))
    for fx_graph in reports.fx_graph_reports:
        fx_graph.write_thunder_repro("Repros")


import os
import subprocess
import sys

# Specify the folder containing the Python files
folder_path = 'Repros'

# Change the current working directory to the specified folder
os.chdir(folder_path)

# Get a list of all Python files in the folder
python_files = [f for f in os.listdir() if f.endswith('.py')]

# Run each Python file and check its execution status
for file in python_files:
    try:
        # Run the Python file using subprocess
        result = subprocess.run([sys.executable, file], capture_output=True, text=True)

        # Check if the execution was successful
        if result.returncode == 0:
            print(f"{file} executed successfully.")
        else:
            print(f"{file} execution failed with return code {result.returncode}.")
[...]

Issue #1854, opened 2025-03-07
[reporting] Calling run_repro leads to jit error - Encountered exception AssertionError: attribute nodes of Graph object out of sync


kshitij12345:
Context - I am trying to get a minimal repro for failure running "microsoft/Phi-3-vision-128k-instruct" using `run_repro`.

Script
```python
import torch
from transformers import AutoModelForCausalLM, AutoConfig
from transformers.models.phi3 import Phi3Config

model_id = "microsoft/Phi-3-vision-128k-instruct"

# Initialize the pre-trained model
cfg = AutoConfig.from_pretrained(model_id, trust_remote_code=True)
cfg.num_hidden_layers = 2

from thunder.dynamo import thunderfx
from thunder.dynamo.report import get_thunder_fxgraph_reports, fx_report, ThunderCompileSpecification

with torch.device("cuda"):
    model = AutoModelForCausalLM.from_config(cfg, trust_remote_code=True, torch_dtype=torch.bfloat16)

    input_ids = torch.randint(0, 200, (1, 512)) # torch.int64 cuda:0 torch.Size([1, 512]) tensor(1, device='cuda:0') tensor(32010, device='cuda:0')
    attention_mask = torch.zeros((1, 512), dtype=torch.int64)
    attention_mask[0, 426:] = 1
    # torch.uint8 cuda:0 torch.Size([1, 400, 356, 3]) tensor(0, device='cuda:0', dtype=torch.uint8) tensor(255, device='cuda:0', dtype=torch.uint8)
    pixel_values = torch.randint(0, 255, (1, 400, 356, 3), dtype=torch.uint8)
    labels = input_ids.clone().detach()

    reports = fx_report(model, **dict(input_ids=input_ids, attention_mask=attention_mask, pixel_values=pixel_values, labels=labels))
    for fx_graph in reports.fx_graph_reports:
        try:
            # Mostly fails with 
            # Encountered exception AssertionError: attribute nodes of Graph object out of sync: <torch.fx.graph._node_list object at 0x7e1064d06360> vs. <torch.fx.graph._node_list object at 0x7e107bbd0230> while tracing GraphModule()
            fx_graph.run_repro(compile_fn=ThunderCompileSpecification())
        except Exception as e:
            print(e)
            # raise e

```

Most/All fx_graph repros fail with
Sample Error
```
[...]

PR #1865, opened 2025-03-11
Fixes higher order ops indentation in repro file

kiya00:
<details>

</details>

## What does this PR do?

Fixes the indentation error part in #1855， #1854

### DIFF
--- a/thunder/dynamo/report.py
+++ b/thunder/dynamo/report.py
@@ -9,7 +9,7 @@ from typing import TYPE_CHECKING
 import textwrap
 import copy
 from itertools import chain
-
+from looseversion import LooseVersion
 
 import torch
 from thunder.core.pytree import tree_flatten
@@ -24,6 +24,7 @@ from thunder.dynamo.utils import (
     CompilerType,
     example_input_meta_to_input,
     recompile_graph,
+    has_higher_order_operator,
 )
 
 from thunder.dynamo.repro_script_template import (
@@ -197,6 +198,14 @@ class FXGraphReport:
     """
 
     def __init__(self, graph: torch.fx.GraphModule, graph_name: str, example_input_meta: list[ExampleInputMetaData]):
+        if LooseVersion(torch.__version__) < LooseVersion("2.6.0"):
+            # NOTE: PyTorch 2.6 changes the structure of GraphModule for higher order ops.
+            # In newer torch version the higher order ops are nested as submodules within the module that uses them,
+            # but in the older version they were separate sibling modules.
+            if has_higher_order_operator(graph):
+                raise RuntimeError(
+                    "The Reporting Tool for Torch higher-order operators is supported only in PyTorch version 2.6 or later."
+                )
         self.graph = graph
         self.graph_name = graph_name
         self.example_input_meta = example_input_meta
--- a/thunder/dynamo/report.py
+++ b/thunder/dynamo/report.py
@@ -427,6 +427,7 @@ class FXGraphReport:
     ):
         torch._dynamo.reset()
         example_inputs = self.make_example_inputs()
+        # To avoid the AssertionError: attribute nodes of Graph object out of sync
         recompile_graph(self.graph)
         compiled_model = compile_fn.compile(self.graph, inputs=example_inputs)
         result = run_forward_backward(compiled_model, *example_inputs)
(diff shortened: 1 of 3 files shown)


thunder/dynamo/utils.py
```diff
@@ -696,6 +696,9 @@ def _readable(
         include_device=include_device,
     )
     module_code = verbose_python_code.src
+    submodule_names = [name for name, m in module.named_children() if hasattr(m, "graph")]
+    for submodule_name in submodule_names:
```
mruberry, line 700: A comment explaining what's happening here would be helpful

thunder/dynamo/report.py
```diff
@@ -491,6 +493,7 @@ def run_benchmark(self, compile_fn: CompileSpecificationInterface, time_fn: Time
         # to reset Dynamo's state *as if* you had started a fresh process invocation.
         torch._dynamo.reset()
         example_inputs = self.make_example_inputs()
+        recompile_graph(self.graph)
```
mruberry, line 496: Comments for why `recompile_graph` is called would make this more readable

mruberry (approved):
Nice! I left a couple suggestions for more comments but those can always be added in the future.

mruberry:
fyi @t-vi, CI failures appear unrelated

kiya00:
The failure appears to be related to differences in the implementation of higher-order operations between Torch 2.5.1 and the nightly version. so the test to pass on Torch nightly but fail on Torch 2.5.1. I'll see what happened

kiya00:
- Torch Module Structure Changes for Higher-Order Ops
In newer torch version the higher order ops are nested as submodules within the module that uses them, but in the older version they were separate sibling modules.
In torch2.5.1 the split module with higher order ops looks like:
```
GraphModule(
  (fwd_body_0): GraphModule()
  (bwd_body_0): GraphModule()
  (thunder_0): GraphModule()  # contains `torch.ops.higher_order.autograd_function_apply(fwd_body_0, bwd_body_0,...`
  (inductor_1): GraphModule()  
  (thunder_2): GraphModule()
)
```
in torch 2.6 or later:
```
GraphModule(
  (thunder_0): GraphModule(
    (fwd_body_0): GraphModule()
    (bwd_body_0): GraphModule()
  )
  (inductor_1): GraphModule()
  (thunder_2): GraphModule()
)
```

- Impact on Reproducing and Saving Scripts
In newer version it's automatically handled by Torch:
the code is like
```
def forward(self, l_x_ : torch.Tensor):
    fwd_body_0 = self.fwd_body_0
    bwd_body_0 = self.bwd_body_0
    cos = torch.cos(l_x_)
    autograd_function_apply = torch.ops.higher_order.autograd_function_apply(fwd_body_0, bwd_body_0, l_x_, args_tensor_mask = [True], non_differentiable_idx = []);  fwd_body_0 = bwd_body_0 = l_x_ = None
    y = cos + autograd_function_apply;  cos = autograd_function_apply = None
    return y
```
In the older version I need to support the callable type argument and also modify the code string generated by the graph module
```
def forward(self, l_x_ : torch.Tensor, fwd_body_0, bwd_body_0):
    cos = torch.cos(l_x_)
    autograd_function_apply = torch.ops.higher_order.autograd_function_apply(fwd_body_0, bwd_body_0, l_x_, args_tensor_mask = [True], non_differentiable_idx = []);  fwd_body_0 = bwd_body_0 = l_x_ = None
    y = cos + autograd_function_apply;  cos = autograd_function_apply = None
    return y
```
I’d like to ask if it’s worth supporting older Torch versions? Personally, I’m thinking of adding a version check to explicitly state that higher-order ops are only supported for Torch ≥ 2.6. WDYT? @mruberry @kshitij12345 
cc: @IvanYashchuk

mruberry:
> I’d like to ask if it’s worth supporting older Torch versions? Personally, I’m thinking of adding a version check to explicitly state that higher-order ops are only supported for Torch ≥ 2.6. WDYT? @mruberry @kshitij12345 cc: @IvanYashchuk

I was going to suggest the same thing!

If we need support for higher order operations on older Torch versions then we can add it later

kshitij12345:
> Personally, I’m thinking of adding a version check to explicitly state that higher-order ops are only supported for Torch ≥ 2.6. WDYT?

Makes sense to me as well. Also, based on #1872, it seems that thunder aims to support latest stable pytorch and nightlies, so it should be fine.

PR #1866, opened 2025-03-11
add softsign op

beverlylytle:
Adds support for the softsign op.

PR #1868, opened 2025-03-11
Skip rms_norm on CPU too for torch  < 2.7

riccardofelluga:
Thanks for the great catch @beverlylytle! From #1866


mruberry (approved):
Thanks @riccardofelluga and @beverlylytle !

fyi @t-vi, looks like the CI failure is unrelated

PR #1873, opened 2025-03-12
Upcast fp8 to fp32 in get_example_input_tensor_metadata for computing aminmax

crcrpar:
## What does this PR do?

While trying to identify the root cause of `thunder.jit(torchao_float8_linears)` backward fails to correctly convert matrices of `torch._scaled_mm` inputs, I found that `thunder.jit(..., nv_store_fusion_inputs=True, nv_store_fusion_inputs_meta=True)` when I tried it on a whim as `aminmax` does not support any fp8 dtypes.
However, the function this PR changes aims at getting max and min of `t: torch.Tensor`, it'd be fair to upcsat it to `torch.float32`.

### DIFF
--- a/thunder/dynamo/utils.py
+++ b/thunder/dynamo/utils.py
@@ -17,6 +17,7 @@ from thunder.torch.langctx import torchctx
 from thunder.core.utils import check
 
 if TYPE_CHECKING:
+    from numbers import Number
     from thunder.core.symbol import Symbol
     import os
     from typing import Any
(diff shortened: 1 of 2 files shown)


thunder/dynamo/utils.py
```diff
@@ -462,6 +462,8 @@ def _get_example_input_tensor_metadata(t: torch.Tensor) -> ExampleInputMetaData:
     min_val = None
     max_val = None
     if not isinstance(t, FakeTensor) and t.device.type != "meta" and t.numel() != 0:
+        if t.dtype in (torch.float8_e4m3fn, torch.float8_e4m3fnuz, torch.float8_e5m2, torch.float8_e5m2fnuz):
+            t = t.to(torch.flot32)
```
mruberry, line 466: `flot32` -> `float32`

mruberry:
@IvanYashchuk and/or @kshitij12345, would you like to review this PR?

Some thoughts before I excuse myself:

- Is there a test that we can add for this?
- Is there a more generic test for unsupported low precision datatypes than explicitly enumerating the actual dtypes?

IvanYashchuk:
There's a test for this function:
https://github.com/Lightning-AI/lightning-thunder/blob/d66fbbc3130565ef7fe109ca6bdba792feb530f3/thunder/tests/test_dynamo.py#L1038
but there's no test for unsupported cases.

thunder/dynamo/utils.py
```diff
@@ -462,6 +462,8 @@ def _get_example_input_tensor_metadata(t: torch.Tensor) -> ExampleInputMetaData:
     min_val = None
     max_val = None
     if not isinstance(t, FakeTensor) and t.device.type != "meta" and t.numel() != 0:
+        if t.dtype in (torch.float8_e4m3fn, torch.float8_e4m3fnuz, torch.float8_e5m2, torch.float8_e5m2fnuz):
```
mruberry, line 465: While it's OK to upcast `t` so that the min and maximum values can be computed, won't this change the dtype recorded for `t`?

Maybe an `aminmax` helper would be useful?

kiya00 (approved):

mruberry:
The CI failure is real but unrelated to this PR. I filed https://github.com/Lightning-AI/lightning-thunder/issues/1929 to track it.

fyi @t-vi this may need help getting merged

mruberry (approved):

PR #1877, opened 2025-03-13
add option of `_force_disable_execution_callback_file`

crcrpar:
## What does this PR do?

Add option `TraceCtx._force_disable_execution_callback_file` so that we do not need to see files of traces  of custom torch.autograd.Function and torch.ops.higher_order.autograd_function_apply.

### DIFF
--- a/thunder/core/jit_ext.py
+++ b/thunder/core/jit_ext.py
@@ -709,6 +709,7 @@ def _convert_pytorchfunc_to_thundertrace(
         return wrapped_func_result, None
 
     trace = TraceCtx()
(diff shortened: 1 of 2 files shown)


t-vi:
Personally, I'd rather do away with all this writing to file through context managers. Not sure what it is all about, we do have the compile_stats where you have all traces.

crcrpar:
`thunder.set_execution_callback_file` lets us modify all the traces translated into executable python in real-time. While I'd not say this feature has many users, so just removing this option itself sounds reasonable to me.

Issue #1857, opened 2025-03-07
Expected NumberType mismatch in pad meta


kshitij12345:
Found while working on https://github.com/Lightning-AI/lightning-thunder/pull/1850

```python
import torch

def forward(l_hidden_states_: "bf16[1, 512, 3072]", l_self_modules_qkv_proj_parameters_weight_: "bf16[9216, 3072]", l_position_ids_: "i64[1, 512]"):
    #     # File: /home/kkalambarkar/git/pytorch/torch/utils/_device.py:106 in __torch_function__, code: return func(*args, **kwargs)
    qkv: "bf16[1, 512, 9216]" = torch._C._nn.linear(l_hidden_states_, l_self_modules_qkv_proj_parameters_weight_, None);  l_hidden_states_ = l_self_modules_qkv_proj_parameters_weight_ = None

    # File: /home/kkalambarkar/git/pytorch/torch/utils/_device.py:106 in __torch_function__, code: return func(*args, **kwargs)
    getitem_3: "i64[1]" = l_position_ids_[(slice(None, None, None), -1)];  l_position_ids_ = None
    max_1: "i64[]" = getitem_3.max();  getitem_3 = None
    return (max_1,)

l_hidden_states_ = torch.randn([1, 512, 3072], dtype=torch.bfloat16, requires_grad=True)
l_self_modules_qkv_proj_parameters_weight_ = torch.randn([9216, 3072], dtype=torch.bfloat16)
l_position_ids_ = torch.randint(0, 10, size=(1, 512))

import thunder
thunder.jit(forward)(l_hidden_states_, l_self_modules_qkv_proj_parameters_weight_, l_position_ids_)

```

Error
```python
  File "/home/kkalambarkar/lightning-thunder/thunder/core/symbol.py", line 319, in __call__
    result = self.meta(*args, **kwargs)
  File "/home/kkalambarkar/lightning-thunder/thunder/core/langctxs.py", line 136, in _fn
    result = fn(*args, **kwargs)
  File "/home/kkalambarkar/lightning-thunder/thunder/core/prims.py", line 3228, in pad_meta
    utils.check_same_dtype(a, padding_value)
  File "/home/kkalambarkar/lightning-thunder/thunder/core/utils.py", line 254, in check_same_dtype
    check(
  File "/home/kkalambarkar/lightning-thunder/thunder/core/baseutils.py", line 146, in check
    raise exception_type(s())
[...]

PR #1878, opened 2025-03-13
Cast grad_chooser_backward output to match input

beverlylytle:
In #1857, an error was encountered when calculating the backward trace of a function that uses `max` with integer input.  The error was coming from a check in `pad_meta` where a float was found when an integer was expected.  The backward implementation of `max` involves a division operation, that when applied to integers results in a float.  This pr converts the result of that division to match the dtype of the input.

### DIFF
--- a/thunder/core/transforms.py
+++ b/thunder/core/transforms.py
@@ -1841,6 +1841,7 @@ def grad_chooser_backward(primal, x, x_shape, reduced_dims, g):
     argmax_locations = x == primal_repeated
     argmax_sum = keepdim_reduction(prims.sum, argmax_locations, reduced_dims)
     out = g_repeated * argmax_locations / argmax_sum
+    out = prims.convert_element_type(out, g.dtype)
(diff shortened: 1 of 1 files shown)


kshitij12345 (commented):
Fix makes sense to me, Thanks @beverlylytle.

I think we should add a small manual test (maybe like snippet below) to verify that this doesn't fail. 

```python
import torch

def forward(x, ids):
    x + x  # some operation on x is required to trigger the bug.
    return ids[0].max()

x = torch.randn([2, 2], dtype=torch.bfloat16, requires_grad=True)
ids = torch.randint(0, 10, size=(1, 512))

import thunder
thunder.jit(forward)(x, ids)
```

kshitij12345 (approved):
LGTM, thanks @beverlylytle

IvanYashchuk:
Waiting on code owner review from @lantiga, @mruberry, and/or @t-vi.

PR #1881, opened 2025-03-13
add `atleast_{1d, 2d, 3d}` ops & fix snippet_phantom_grad_vs_torch_consistency test

khushi-411:
`atleast_1d` operator returns `self` for one-dimensional tensor. It is a leaf tensor; however, this op is still differentiable, and the gradient value is `torch.tensor([1.])`. (The same goes for other similar operators like `atleast_2d`, `atleast_3d`, etc.)
```python
In [1]: import thunder, torch
In [2]: a = torch.tensor((4,), dtype=torch.float32, requires_grad=True)
In [3]: torch.atleast_1d(a).backward()
In [4]: a.grad
Out[4]: tensor([1.])
In [5]: a.reshape(-1)
Out[5]: tensor([4.], grad_fn=<ViewBackward0>)
In [6]: a.view(-1)
Out[6]: tensor([4.], grad_fn=<ViewBackward0>)
```

As in the `snippet_phantom_grad_vs_torch_consistency` test, it returns an `[]` empty list in such cases.
```python
E           AttributeError: 'function' object has no attribute 'name'

thunder/tests/test_grad.py:1348: AttributeError
------------------------------------- Captured stdout call -------------------------------------
tensor([1.2820], device='cuda:0', dtype=torch.float64, grad_fn=<ViewBackward0>)
[tensor([1.2820], device='cuda:0', dtype=torch.float64, grad_fn=<ViewBackward0>)]
tensor([-3.3215, -7.2744, -5.0842,  4.6119], device='cuda:0',
       dtype=torch.float64, requires_grad=True)
[]
```

This PR adds the support of the `atleast_{1d, 2d, 3d}` operator in thunder. And to make `atleast_{1d, 2d, 3d}` ops support gradient, this PR fixes this issue by returning grad values for cases where the operators return `self` and also support `grads`.

Credits & Thanks to @kshitij12345 for helping me find the gradient issue!! It was really helpful.


thunder/clang/__init__.py
```diff
@@ -2173,6 +2173,46 @@ def topk(
     return prims.topk(a, k, dim, bool(largest), bool(sorted), out=out)
 
 
+@clangop()
```
mruberry, line 2176: I would just define these as torch operators;  clang operators are a set of operations that are useful for building PyTorch (or NumPy) operations

mruberry (approved):
@beverlylytle, would you like to review the addition of these operators and the grad test changes?

@khushi-411 I just made one comment about how I don't think we need to add these ops to Thunder's core language

thunder/tests/test_grad.py
```diff
@@ -1320,7 +1320,12 @@ def is_output_differentiable(x):
         # torch.return_types.topk(
         # values=tensor([1., 1.]),
         # indices=tensor([0, 1]))
-        return x.grad_fn is not None
+        return x.grad_fn is not None or is_returning_self(x)
+
+    def is_returning_self(x):
```
beverlylytle, line 1325: Nice.

thunder/tests/test_grad.py
```diff
@@ -1355,7 +1360,7 @@ def filter_differentiable_outputs(outputs):
             grads = [torch.ones_like(torch_result)]
 
     torch_tensors_requiring_grad = tuple(f for f in torch_flats if isinstance(f, torch.Tensor) and f.requires_grad)
-    torch_grad_result = torch.autograd.grad(torch_result, torch_tensors_requiring_grad, grads)
+    torch_grad_result = torch.autograd.grad(torch_result, torch_tensors_requiring_grad, grads, allow_unused=True)
```
beverlylytle, line 1363: Why is this change needed?

thunder/tests/opinfos.py
```diff
@@ -6018,6 +6019,56 @@ def topk_error_generator(op, device, **kwargs):
...
+    yield SampleInput(make(()), make((2,)))
+    yield SampleInput(make((2,)), make((5, 5)))
+    yield SampleInput(make(()), make((2,)), make((4, 4)))
+    yield SampleInput(make(2, 3), make(4, 5), make(6, 6, 6), make(5, 5, 5, 5))
+
+
+atleast_1d_opinfo = OpInfo(
+    clang.atleast_1d,
+    supports_grad=True,
+    dtypes=(datatypes.floating, datatypes.exact, datatypes.complexfloating),
```
beverlylytle, line 6045: Isn't this the default value for dtypes? I think this line can be deleted, here and in the other opinfos.

beverlylytle (approved):
Looks good!

thunder/clang/__init__.py
khushi-411: I see your point. Quite informative, thank you!

thunder/tests/test_grad.py
khushi-411: I was getting something like `differentiated tensor is not used in the graph` earlier. I don't exactly remember, but it looks like I cannot reproduce this error anymore. Thanks for pointing it out; I've updated the PR.

thunder/torch/__init__.py
```diff
@@ -3005,6 +3005,52 @@ def topk(
     return clang.topk(a, k, dim, largest, sorted, out=out)
 
 
+@torchsymbol(torch.atleast_1d, is_method=True)
+def atleast_1d(
+    arg: Union[TensorLike, Sequence[TensorLike]], *args: TensorLike
+) -> Union[TensorLike, tuple[TensorLike, ...]]:
+    if not args and isinstance(arg, Sequence):
```
mruberry, line 3012: If the inputs is just `*tensors`, then I think this `if` statement can be removed, and the function would just be:

```
res = tuple(a if a.ndim >= 1 else unsqueeze(a, 0) for a in args)
return res if len(res) > 1 else res[0]
```

thunder/torch/__init__.py
```diff
@@ -3005,6 +3005,52 @@ def topk(
...
+def _unsqueeze_atleast(at_least_fn: Callable, dim: int, arg: TensorLike) -> TensorLike:
+    arg_ = at_least_fn(arg)
+    return unsqueeze(arg_, dim)
+
+
+@torchsymbol(torch.atleast_2d, is_method=True)
+def atleast_2d(
+    arg: Union[TensorLike, Sequence[TensorLike]], *args: TensorLike
+) -> Union[TensorLike, tuple[TensorLike, ...]]:
+    if not args and isinstance(arg, Sequence):
```
mruberry, line 3030: Like with the above, if the only argument is `*args` then I don't think the `if/else` is needed

thunder/torch/__init__.py
```diff
@@ -3005,6 +3005,52 @@ def topk(
...
+        args_ = arg
+    else:
+        assert not isinstance(arg, Sequence)
+        args_ = (arg,) + args
+    unsqueeze_atleast_1d = partial(_unsqueeze_atleast, atleast_1d, 0)
+    res = tuple(a if a.ndim >= 2 else unsqueeze_atleast_1d(a) for a in args_)
```
mruberry, line 3036: Instead of using `unsqueeze_atleast` as a helper, what would this look like if you defined a function within `atleast_2d` that handled the two unsqueezing cases?

```
if a.ndims == 0:
  return a.unsqueeze(0, 1)
if a.ndims == 1:
  return a.unsqueeze(0)
return a
```

?

I think this would be more readable and ensure the generated trace is as simple as possible

thunder/torch/__init__.py
```diff
@@ -3005,6 +3005,52 @@ def topk(
...
+    return res if len(res) > 1 else res[0]
+
+
+@torchsymbol(torch.atleast_3d, is_method=True)
+def atleast_3d(
+    arg: Union[TensorLike, Sequence[TensorLike]], *args: TensorLike
+) -> Union[TensorLike, tuple[TensorLike, ...]]:
+    if not args and isinstance(arg, Sequence):
```
mruberry, line 3044: Same comments here as for the above

thunder/torch/__init__.py
```diff
@@ -3005,6 +3005,52 @@ def topk(
     return clang.topk(a, k, dim, largest, sorted, out=out)
 
 
+@torchsymbol(torch.atleast_1d, is_method=True)
+def atleast_1d(
+    arg: Union[TensorLike, Sequence[TensorLike]], *args: TensorLike
```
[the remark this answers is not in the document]
khushi-411: Thank you!

Updated the code.

thunder/torch/__init__.py
khushi-411: Make sense! Definitely a better way. Thank you! I've updated the code accordingly.

mruberry:
Cool!

khushi-411:
Thank you very much, @mruberry, @beverlylytle, and @kshitij12345
 for all your insightful suggestions in this PR!!

PR #1885, opened 2025-03-14
simplify `match_fw_and_bw_saved_for_bw_proxies`

ali-alshaar7:
<details>

</details>

## What does this PR do?

Fixes # (issue).

## PR review

## Did you have fun?

### DIFF
--- a/thunder/core/rematerialization.py
+++ b/thunder/core/rematerialization.py
@@ -538,18 +538,15 @@ def match_fw_and_bw_saved_for_bw_proxies(
         new_required_for_bakward_fw_to_bw_map: Dict[str, Proxy]: mapping of bw names to forward proxies
     """
 
-    old_saved_for_backward_fw = (*fw_trace.bound_symbols[-1].args[1][0], *fw_trace.bound_symbols[-1].args[1][1])
-    old_saved_for_backward_bw = []
+    old_saved_for_backward_fw = [*fw_trace.output[1][0], *fw_trace.output[1][1]]
+    old_saved_for_backward_bw = [None] * len(old_saved_for_backward_fw)
     for bsym in bw_trace.bound_symbols:
-        if bsym.sym.id == PrimIDs.UNPACK_SEQUENCE:
-            flattened_args = tree_flatten(bw_trace.args[1])[0]
-            proxy_names = {y.name for y in flattened_args if isinstance(y, ProxyInterface)}
-            if all(
-                not isinstance(out, CollectionProxy) and out.name not in proxy_names
-                for out in bsym.flat_outs
-                if out is not None
-            ):
-                old_saved_for_backward_bw += bsym.flat_outs
+        if bsym.sym.id == PrimIDs.UNPACK_SEQUENCE and bsym.args[0].name == "C0":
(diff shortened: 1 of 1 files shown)


thunder/core/rematerialization.py
```diff
@@ -538,18 +538,15 @@ def match_fw_and_bw_saved_for_bw_proxies(
         new_required_for_bakward_fw_to_bw_map: Dict[str, Proxy]: mapping of bw names to forward proxies
     """
 
-    old_saved_for_backward_fw = (*fw_trace.bound_symbols[-1].args[1][0], *fw_trace.bound_symbols[-1].args[1][1])
-    old_saved_for_backward_bw = []
+    old_saved_for_backward_fw = [*fw_trace.output[1][0], *fw_trace.output[1][1]]
+    old_saved_for_backward_bw = [None] * len(old_saved_for_backward_fw)
```
IvanYashchuk, line 542: This line can be simple attribute access on TraceCtx.args instead of inspecting inputs of unpack sequence:
```suggestion
    old_saved_for_backward_bw = (*bw_trace.args[0][0], *bw_trace.args[0][1])
```
If proxies of bw_trace.args do not match the proxies in the unpack sequence operation it means there's a bug somewhere in the preceding transformations of the trace. They should be kept in sync.

IvanYashchuk (commented):
Great that you're continuing what was started in https://github.com/Lightning-AI/lightning-thunder/pull/1754! I hope you'll get all tests to pass.

Backward traces have a simple function signature with two arguments: "saved_for_backward" and "backward input". "saved_for_backward" is another tuple of two sequences: one for tensors to be able to free the memory and the second sequence for any other object saved from forward. Both should be accessible with `TraceCtx.args[0]`. This would simplify the code even further.

thunder/core/rematerialization.py
ali-alshaar7: Hey @IvanYashchuk, thanks for the pointers! 
```
If proxies of bw_trace.args do not match the proxies in the unpack sequence operation it means there's a bug somewhere in the preceding transformations
```
is exactly the issue here, the `TraceSubstitutionProcessor ` will update the names of proxies in the trace, and since the forward and backward are updated separately, they fall out of sync.

It seems the main issue is that the unpacks are a boundary for updating names, since it can't connect the unpacked proxy to its corresponding trace arg. Any ideas how this should be handled properly? Im betting there's something missing from the `TraceSubstitutionProcessor ` but not sure.

t-vi: Please do not rely on names for matching. They are not guaranteed to be the same across retracing.
There is a clear expectation that the saved for backwards returned by the forward are compatible with the ones consumed by the backward, but expecting name stability is asking for trouble.

k223kim (approved):

IvanYashchuk (approved):
Awesome! Would you update the pull request description with useful information like context on the change, and maybe pointing to the previous related changes? What issues does this PR resolve?

@t-vi, could you please review this PR as well?

t-vi (approved):
Thank you @ali-alshaar7 @IvanYashchuk @k223kim

PR #1887, opened 2025-03-17
Introduce plugins for the high level API

lantiga:
## What does this PR do?

Introduce `Plugins` in the high level API alongside `Recipes`, as a way to bundle lookasides, transforms and executors, leaving other model-specific to `Recipes`.

This way users can run

```python
with torch.device("cuda:0"):
    model = ViTForImageClassification.from_pretrained("google/vit-base-patch16-224")
    inp = torch.randn(128, 3, 224, 224)

thunder_model = thunder.compile(model, plugins=["fsdp", "reduce-overhead"])

thunder_out = thunder_model(inp)
```

The PR comes with a set of plugins that we can expand in the future.

Ideally plugins should be self standing and not require thinking from the user about ordering. In the future we will want to introduce a more sophisticated precedence/ordering strategy, that is not fleshed out in the current PR.

## PR review

## Did you have fun?


t-vi (approved):
Seems good.
I don't think the pre and post will solve our ordering problems (e.g. low precision/quant vs. distributed) but let's iterate in main.

The other thing I wonder is if we should have some CI for the examples (but we need to do it on smaller models, so it's not completely trivial), maybe have a "test" function (calling main after monkeypatching to smaller versions or so).

lantiga:
I agree re: pre vs post, ordering will need to be addressed in a follow up.
I'm thinking about a simple trait system but we'll need to figure it out.

PR #1888, opened 2025-03-18
bump GPU distributed timeout

t-vi:
...probably not the ultimate fix  (seems that something is unstable with the dev PT and distributed), but maybe for better error messages.


t-vi:
Merging this without review to stabilize CI.

PR #1896, opened 2025-03-19
add `frexp`, `isnan`, and `isinf`

khushi-411:
As per the title

#### TODO
- [x] check precision issue for test_vjp_correctness


khushi-411:
Oh no, so many errors in the CI! 😅 It passed on my local system, though. I'm looking into it and will ping you once it's ready.

mruberry:
@t-vi The CI failures appear to be due to download failures and code coverage not running properly. Do we ever use code coverage or is it just an extra source of failures?

thunder/core/prims.py
```diff
@@ -2216,6 +2217,20 @@ def _make_elementwise_unary_prim(
     output_dtype_kind=ELEMENTWISE_PRIM_OUTPUT_DTYPE_KIND.INT_FOR_NUMBER,
 )
 
+
+def frexp_meta(a: TensorProxy, *, out: None | TensorProxy = None) -> (TensorProxy, TensorProxy):
```
mruberry, line 2221: While it looks a like a few primitives do have an `out` keyword, they probably shouldn't. We should take `out` away from the primitive

thunder/clang/__init__.py
```diff
@@ -1602,6 +1602,16 @@ def floor(a: TensorLike | Number) -> TensorLike | Number:
     )
 
 
+@clangop()
+def frexp(a: TensorLike, *, out=None) -> tuple[TensorLike, TensorLike]:
```
mruberry, line 1606: Do clang ops typically have an `out` argument?

mruberry: Nit: `frexp(a: TensorLike, /) -> ...`

thunder/torch/__init__.py
```diff
@@ -1578,11 +1578,29 @@ def floor_(a):
...
+def frexp(a: TensorLike, *, out=None) -> (TensorLike, TensorLike):
+    return clang.frexp(a, out=out)
+
+
 @torchsymbol(torch.isfinite, is_method=True)
 def isfinite(a):
     return clang.isfinite(a)
 
 
+# TODO: add complex number support
```
mruberry, line 1591: Should this throw an error if given a complex input for now?

thunder/tests/opinfos.py
```diff
@@ -1175,6 +1175,35 @@ def _abs_torch(x: torch.Tensor | Number):
...
 elementwise_unary_ops.append(floor_opinfo)
 
+frexp_opinfo = OpInfo(
+    clang.frexp,
+    supports_grad=True,
+    dtypes=(datatypes.floating,),
+    sample_input_generator=elementwise_unary_generator,
+    torch_reference=_elementwise_unary_torch(torch.frexp),
+    test_directives=(
+        # AssertionError: Scalars are not close!
```
mruberry, line 1185: I wonder if widening the test tolerances would allow some of these tests to pass?

Maybe, as the comment suggests, a more numerically accurate fwd+bwd is needed

thunder/torch/__init__.py
mruberry: Or this could just add support for complex tensors. See the implementation here:

https://github.com/pytorch/pytorch/blob/32299e5f9a68602aa3b5ee77794703d9922bae96/torch/_refs/__init__.py#L717

mruberry:
@beverlylytle, would you please review this PR?

thunder/core/prims.py
khushi-411: Got it! Thank you for sharing the details!

I'll remove the `out` arg from here and submit another PR to remove it from the operators that support it.

thunder/clang/__init__.py
khushi-411: > Do clang ops typically have an out argument?

Yes, a few ops like `topk` and `sort`.

https://github.com/Lightning-AI/lightning-thunder/blob/a8c063676cbecad7c196dcfe4fbeacf85775c9c5/thunder/clang/__init__.py#L2165-L2184

mruberry: Like with the prims, we should probably not support `out` on clang ops

mruberry:
Cool start, @khushi-411! I made some comments inline for you review, and I asked @beverlylytle to take a look, too. I wonder what's going on with the grad computation for frexp?

thunder/torch/__init__.py
khushi-411: Sure, I'll update this once PR #1909 is merged. :-)

thunder/tests/opinfos.py
khushi-411: Hi, I was debugging the gradient issue and tried using `ldexp` and found it does not solve the issue—which makes sense, since it itself relies on the `exp2` operator. I stumbled upon a related PyTorch issue where `ldexp` was returning `inf` for `float64`. Here: https://github.com/pytorch/pytorch/issues/133267. Please correct me if I'm wrong, but I think this seems somewhat similar to our case. In that thread, they promoted the `exponent` values to `float32`, so I thought implementing something similar might help in our case, but unfortunately, it didn't help either. Here's my code:
```python
exponent_float = exponent.to(dtype=a.dtype)
two = ltorch.full_like(a, 2.0)
_pow = two.pow(-exponent_float)

a_grad = g_mantissa * _pow
a_grad = a_grad.masked_fill(~prims.isfinite(a_grad), 0.0)
```
Would love to hear your thoughts. Thank you!

thunder/torch/__init__.py
khushi-411: done! Thank you!

I'm sorry for the delay in updating the PR.

I had one doubt, could you please take a look? I've mentioned here, (since it was more relatable): https://github.com/Lightning-AI/lightning-thunder/pull/1896#discussion_r2046697737

thunder/tests/opinfos.py
mruberry: Interesting! Well, let's not worry about solving it in this PR.

mruberry (commented):
Just a couple small nits from me, @khushi-411! @beverlylytle would you like to take another look?

thunder/core/prims.py
```diff
@@ -2235,6 +2236,14 @@ def _exp2_number(a: Number) -> Number:
     output_dtype_kind=ELEMENTWISE_PRIM_OUTPUT_DTYPE_KIND.INT_FOR_NUMBER,
 )
 
+
+def frexp_meta(a: TensorProxy, /) -> (TensorProxy, TensorProxy):
+    utils.check_type(a, TensorProxy)
+    return TensorProxy(like=a), TensorProxy(like=a)
```
beverlylytle, line 2242: I don't think this is quite right.  The second element here shouldn't have a dtype like `a`, but rather an integer dtype.  I think that's what's causing the `RuntimeError:  INTERNAL ASSERT FAILED at "/workspace/Fuser/csrc/runtime/executor_utils.cpp":491`

mruberry: Great catch!

thunder/tests/opinfos.py
beverlylytle: I've been trying to figure out why the vjp tests are failing.  My first thought was that the derivative of the exponent component is zero except for a couple of places where it is undefined.  As we have often encountered, the finite differences approach to approximating the vjp is does not behave well for constant functions.  However, when I've tried to only compute the vjp for only the first component, the tests still fail.  My next thought was to push the sample inputs away from powers of 2 where there would be singularities for the mantissa, still no luck.  (Though still both of those things will probably be necessary to prevent flakey tests.). idk...

beverlylytle: Instead of skipping test_vjp_correctness entirely, you could instead have something like
```
sample_input_generator=partial(elementwise_unary_generator, small=True)
```

thunder/core/prims.py
```diff
@@ -2235,6 +2236,14 @@ def _exp2_number(a: Number) -> Number:
     output_dtype_kind=ELEMENTWISE_PRIM_OUTPUT_DTYPE_KIND.INT_FOR_NUMBER,
 )
 
+
+def frexp_meta(a: TensorProxy, /) -> (TensorProxy, TensorProxy):
+    utils.check_type(a, TensorProxy)
```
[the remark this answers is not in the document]
khushi-411: done!

thunder/core/prims.py
khushi-411: oops, my bad.

have updated the pr. Thanks!

thunder/tests/opinfos.py
```diff
@@ -1176,6 +1176,27 @@ def _abs_torch(x: torch.Tensor | Number):
 )
 elementwise_unary_ops.append(floor_opinfo)
 
+frexp_opinfo = OpInfo(
+    clang.frexp,
+    supports_grad=True,
+    dtypes=(datatypes.floating,),
+    sample_input_generator=partial(elementwise_unary_generator, small=True),
```
beverlylytle, line 1183: Could you make an issue about vjp tests failing for `frexp` for large samples and mention it in a TODO comment here?

thunder/tests/opinfos.py
```diff
@@ -1176,6 +1176,27 @@ def _abs_torch(x: torch.Tensor | Number):
...
+    torch_reference=_elementwise_unary_torch(torch.frexp),
+    test_directives=(
+        DecorateInfo(
+            pytest.mark.skip,
+            "test_phantom_grad_vs_torch_consistency",
+            executors=("nvfuser",),
```
beverlylytle, line 1189: What sort of errors are being produced here?  Is this a different problem than was addressed by the change to the meta function?

thunder/tests/opinfos.py
khushi-411: done! Issue: #1991

thunder/tests/opinfos.py
khushi-411: Oops, sorry, I forgot to remove it.

thunder/tests/opinfos.py
beverlylytle: Thank you!

beverlylytle (approved):
Looks good to me!

mruberry (approved):

PR #1906, opened 2025-03-25
Initial support for Python 3.13

t-vi:
Not much to see here, mostly opcode changes.
Most notable:
- The method / function is now `func, self_or_null` in a number of places,
- locals() now returns a new dict on every call,
- formatting got an update (e.g. values go through `__format__` usually),
- MAKE_FUNCTION is split up (but `__closure__` is still readonly when accessed from Python. Ugh.)

### DIFF
--- a/.github/workflows/ci-testing.yml
+++ b/.github/workflows/ci-testing.yml
@@ -33,6 +33,7 @@ jobs:
         include:
           - { os: "ubuntu-22.04", python-version: "3.11", requires: "latest" }
           - { os: "ubuntu-22.04", python-version: "3.12", requires: "latest" }
+          - { os: "ubuntu-22.04", python-version: "3.13", requires: "latest" }
         exclude:
           - { os: "windows-2022", suite: "ops" }
           - { os: "windows-2022", suite: "grads" }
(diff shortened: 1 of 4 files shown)


t-vi:
The alternative to the closure returning a new function object would be writing our own bytecode, something along the lines of

```python
import dis
import types
def set_closure(func, closure):
    return (closure, func)
co = set_closure.__code__.replace(co_code = set_closure.__code__.co_code[:4] + bytes((dis.SET_FUNCTION_ATTRIBUTE, 8)) + set_closure.__code__.co_code[6:])
set_closure = types.FunctionType(co, set_closure.__globals__, name=set_closure.__name__, argdefs=set_closure.__defaults__, closure=set_closure.__closure__, kwdefaults=set_closure.__kwdefaults__)

def make_g(x):
    def g():
        return x
    return g
g = make_g(2)

g3 = set_closure(g, (types.CellType(3),))

print(g3 is g)
print(g())
```

Of course, this is most faithful to the `SET_FUNCTION_ATTRIBUTE 0x8` opcode because it is just that.

Similarly, Python does not allow to create a new function without a (right number of items) closure by calling the FunctionType, but I imagine MAKE_FUNCTION would allow it.

lantiga (approved):
🚀

.github/workflows/ci-testing.yml
```diff
@@ -134,7 +135,7 @@ jobs:
               -n 4 --durations=250
 
       - name: Testing interpreter
-        if: matrix.python-version == '3.11' || matrix.python-version == '3.12'
+        if: matrix.python-version == '3.11' || matrix.python-version == '3.12' || matrix.python-version == '3.13'
```
Borda, line 138:
```suggestion
    if: matrix.python-version != '3.10'
```

PR #1908, opened 2025-03-26
Functional TE executor

### LINKED ISSUE #1056: TransformerEngine - Investigate to see if we can integrate with the newer Operation-based API
Currently, for the TransformerEngine (TE) integration, we rely on TE's internal functions and objects. This is leads to somewhat frequent breakages of the integration. See https://github.com/Lightning-AI/lightning-thunder/issues/1033 https://github.com/Lightning-AI/lightning-thunder/issues/117

TE introduced a new Operation based API - https://github.com/NVIDIA/TransformerEngine/pull/707 which we can potentially use. This API is expected to be more stable and provide the required functionality.

Filing this issue so that we can
* [x] Investigate how we can use this API
* [ ] Update our existing integration (if feasible)

riccardofelluga:
## What is this about?

This PR introduces a "_new_" executor aimed to replace the TransformerEngine executor. It makes use of the new TE's 2.0 functional API, hence the name. Fixes #1056

## Changes

This executor is designed to bring transparency and tune-ability to the FP8 computation. Therefore now the recipe and state creation are visible in the trace and can be manipulated*.

What will the functional TE executor trace look like:
``` python
import torch

@torch.no_grad()
@no_autocast
def computation(a, w):
  # a: "cuda:0 bf16[3, 768, 4096]"
  # w: "cuda:0 bf16[4096, 4096]"

  r3 = get_te_fp8_recipe_3()  # r3: "<class 'NoneType'>"
  s4 = get_te_fp8_state_4(r3, 'forward', 2)  # s4: "<class 'NoneType'>"
  (q6, q7) = get_te_fp8_quantizers_5(s4, 2)
  (t10, _, _) = te_functional_linear_fwd(a, w, None, q6, q7)
  del q6, q7

  (t13,) = te_fp8_amax_and_scale_update(r3, states=(s4,), tokens=(t10,))
  del r3, s4, t10
  return (t13,)
```

What the current TE executor trace looks like:
``` python
import torch
import transformer_engine.pytorch as transformer_engine

@torch.no_grad()
@no_autocast
def computation(a, w):
  # a: "cuda:0 bf16[3, 768, 4096]"
  # w: "cuda:0 bf16[4096, 4096]"

  (t4, _, _) = te_linear_1(a, w, None)
  return (t4,)
```

*still with some constraints.

## Test it out

<details>

<summary>To test this you can use this snippet:</summary>

``` python
import torch
import thunder
from thunder.executors.functional_teex import functional_te_ex
from thunder.transforms.functional_te_transform import TransformerEngineTransform
from transformer_engine.pytorch import Linear


te_linear1 = Linear(16, 16, bias=False)

w_c = te_linear1.weight.detach().clone()
w_c.requires_grad_(True)

w = te_linear1.weight.detach().clone()
w.requires_grad_(True)

with torch.no_grad():
    x = torch.randn(16, 16, device='cuda', requires_grad=True) * 1e-2
    x_c = x.detach().clone()
    x_c.requires_grad_(True)

def foo(a, w):
    return torch.nn.functional.linear(a, w)

jfoo = thunder.jit(foo, executors=(functional_te_ex, ), transforms=(TransformerEngineTransform(), ))

eager_optim = torch.optim.SGD(te_linear1.parameters())

target = torch.tensor(42., device="cuda")

n_iter = 5
for _ in range(n_iter):
    with te.fp8_autocast():
        o1 = te_linear1(x).sum()

    torch.nn.functional.mse_loss(o1, target).backward()

    eager_optim.step()
    eager_optim.zero_grad()


functional_optim = torch.optim.SGD([w,])

for _ in range(n_iter):
    o2 = jfoo(x_c, w).sum()

    torch.nn.functional.mse_loss(o2, target).backward()

    functional_optim.step()
    functional_optim.zero_grad()

torch.testing.assert_close(o1, o2)
torch.testing.assert_close(te_linear1.weight, w)

```

</details>

### StatefulExecutor

TransformerEngine recipes and states are visible in the trace and store in the trace's python context. This PR introduce the `StatefulExecutor` which can register stateful operators. Stateful means that they can store a state in an object and that object persists for as long as the trace lifetime. Therefore allowing states to be preserved when iterating multiple times over the same trace. 

### Transform

With the assumption of one recipe per compile function, the `TransformerEngineTransform` takes care of removing duplicate recipes and lays the grounds for more optimization opportunities that can be done post tracing. I will leave out the grouping of the amax and scale updates from this PR but will introduce it later on as it does not seem to impact performance.

### Supported Recipes

This executor supports `MXFP8` and `DelayedScaling` recipes that can be specified by using TransformerEngine `fp8_autocast` context manager like this:

```python
with te.fp8_autocast(enabled=True, fp8_recipe=fp8_recipe):
    out = thunder_compiled_fn(a, w)
```

In case no autocast is specified, Thunder relies on TE's default recipe for the device.

More info in the design document.

## TODO and misc.
- [x] Validate global executor state idea
- [x] Better linear check
- [x] Import check
- [x] Better comments
- [x] Improve bwd/fwd trace detection in transform
- [x] Add types
- [x] Tests

## Missing features

The functional TE executor is still missing the following features that will be introduced later on after this PR:

- Activation checkpointing #2027 
- DDP/FSDP #2081 

cc @mruberry

### DIFF
--- a/.azure/docker-build.yml
+++ b/.azure/docker-build.yml
@@ -37,9 +37,7 @@ schedules:
 jobs:
   - job: build_push
     strategy:
-      #maxParallel: "3"
       matrix:
-        # CUDA 12.1
         "cuda 12.6 | torch 2.7.0 | cudnn FE v1.10.0":
           { CUDA_VERSION: "12.6.3", TORCH_VERSION: "2.7.0", TRITON_VERSION: "3.3.0", CUDNN_FRONTEND_VERSION: "1.10.0" }
         "cuda 12.6 | torch nightly | cudnn FE v1.10.0":
@@ -49,7 +47,7 @@ jobs:
     cancelTimeoutInMinutes: "2"
     timeoutInMinutes: "95"
     variables:
-      UBUNTU_VERSION: "22.04"
+      UBUNTU_VERSION: "24.04"
       PYTHON_VERSION: "3.10"
       imageRepository: "pytorchlightning/lightning-thunder"
       imageTag: "ubuntu$(UBUNTU_VERSION)-cuda$(CUDA_VERSION)-cudnn-fe$(CUDNN_FRONTEND_VERSION)-py$(PYTHON_VERSION)-pt_${TORCH_VERSION/v/}"
--- a/.azure/gpu-tests.yml
+++ b/.azure/gpu-tests.yml
@@ -15,38 +15,29 @@ jobs:
   - job: testing
     strategy:
       matrix:
-        # CUDA 12.1
-        "ubuntu22.04 | cuda 12.1 | python 3.10 | torch 2.7.0 | main":
-          docker-image: "ubuntu22.04-cuda12.6.3-cudnn-fe1.10.0-py3.10-pt_2.7.0-dev"
-          CUDA_VERSION_MM: "121"
+        "main w/ torch 2.7.0":
+          docker-image: "ubuntu24.04-cuda12.6.3-cudnn-fe1.10.0-py3.10-pt_2.7.0-dev"
           testing: "main"
-        "ubuntu22.04 | cuda 12.1 | python 3.10 | torch 2.7.0 | ops":
-          docker-image: "ubuntu22.04-cuda12.6.3-cudnn-fe1.10.0-py3.10-pt_2.7.0-dev"
-          CUDA_VERSION_MM: "121"
+        "ops w/ torch 2.7.0":
+          docker-image: "ubuntu24.04-cuda12.6.3-cudnn-fe1.10.0-py3.10-pt_2.7.0-dev"
           testing: "ops"
-        "ubuntu22.04 | cuda 12.1 | python 3.10 | torch 2.7.0 | grads":
-          docker-image: "ubuntu22.04-cuda12.6.3-cudnn-fe1.10.0-py3.10-pt_2.7.0-dev"
-          CUDA_VERSION_MM: "121"
+        "grads w/ torch 2.7.0":
+          docker-image: "ubuntu24.04-cuda12.6.3-cudnn-fe1.10.0-py3.10-pt_2.7.0-dev"
           testing: "grads"
-        "ubuntu22.04 | cuda 12.1 | python 3.10 | torch 2.7.0 | distributed":
-          docker-image: "ubuntu22.04-cuda12.6.3-cudnn-fe1.10.0-py3.10-pt_2.7.0-dev"
-          CUDA_VERSION_MM: "121"
+        "distributed w/ torch 2.7.0":
+          docker-image: "ubuntu24.04-cuda12.6.3-cudnn-fe1.10.0-py3.10-pt_2.7.0-dev"
           testing: "distributed"
-        "ubuntu22.04 | cuda 12.1 | python 3.10 | torch-nightly | main":
-          docker-image: "ubuntu22.04-cuda12.6.3-cudnn-fe1.10.0-py3.10-pt_main-dev"
-          CUDA_VERSION_MM: "121"
+        "main w/ torch-nightly":
+          docker-image: "ubuntu24.04-cuda12.6.3-cudnn-fe1.10.0-py3.10-pt_main-dev"
           testing: "main"
-        "ubuntu22.04 | cuda 12.1 | python 3.10 | torch-nightly | ops":
-          docker-image: "ubuntu22.04-cuda12.6.3-cudnn-fe1.10.0-py3.10-pt_main-dev"
-          CUDA_VERSION_MM: "121"
+        "ops w/ torch-nightly":
+          docker-image: "ubuntu24.04-cuda12.6.3-cudnn-fe1.10.0-py3.10-pt_main-dev"
           testing: "ops"
-        "ubuntu22.04 | cuda 12.1 | python 3.10 | torch-nightly | grads":
-          docker-image: "ubuntu22.04-cuda12.6.3-cudnn-fe1.10.0-py3.10-pt_main-dev"
-          CUDA_VERSION_MM: "121"
+        "grads w/ torch-nightly":
+          docker-image: "ubuntu24.04-cuda12.6.3-cudnn-fe1.10.0-py3.10-pt_main-dev"
           testing: "grads"
-        "ubuntu22.04 | cuda 12.1 | python 3.10 | torch-nightly | distributed":
-          docker-image: "ubuntu22.04-cuda12.6.3-cudnn-fe1.10.0-py3.10-pt_main-dev"
-          CUDA_VERSION_MM: "121"
+        "distributed w/ torch-nightly":
+          docker-image: "ubuntu24.04-cuda12.6.3-cudnn-fe1.10.0-py3.10-pt_main-dev"
           testing: "distributed"
     # how much time to give 'run always even if cancelled tasks' before stopping them
     cancelTimeoutInMinutes: "2"
@@ -65,7 +56,6 @@ jobs:
     steps:
       - bash: |
           echo $(DEVICES)
-          echo "CUDA_VERSION_MM=$CUDA_VERSION_MM"
           lspci | egrep 'VGA|3D'
           whereis nvidia
           nvidia-smi
@@ -140,7 +130,6 @@ jobs:
 
       - bash: |
           set -ex
-          export CUDA_LAUNCH_BLOCKING=1
           coverage run --source thunder -m \
             pytest thunder/tests/test_ops.py \
               -m "not standalone" \
@@ -156,12 +145,13 @@ jobs:
           ./codecov --token=$(CODECOV_TOKEN) --commit=$(Build.SourceVersion) \
             --flags=gpu,pytest,regular --name="GPU-coverage" --env=linux,azure
         condition: eq(variables['testing'], 'ops')
+        env:
+          CUDA_LAUNCH_BLOCKING: "1"
         timeoutInMinutes: "40"
         displayName: "Testing: ops"
 
       - bash: |
           set -ex
-          export CUDA_LAUNCH_BLOCKING=1
           coverage run --source thunder -m \
             pytest thunder/tests/test_grad.py \
               -m "not standalone" \
@@ -177,6 +167,8 @@ jobs:
           ./codecov --token=$(CODECOV_TOKEN) --commit=$(Build.SourceVersion) \
             --flags=gpu,pytest,regular --name="GPU-coverage" --env=linux,azure
         condition: eq(variables['testing'], 'grads')
+        env:
+          CUDA_LAUNCH_BLOCKING: "1"
         timeoutInMinutes: "40"
         displayName: "Testing: grads"
 
--- a/.azure/notebook-runs.yml
+++ b/.azure/notebook-runs.yml
@@ -15,12 +15,10 @@ jobs:
   - job: jupyter
     strategy:
       matrix:
-        "ubuntu22.04 | cuda 12.7 | torch 2.7":
-          docker-image: "ubuntu22.04-cuda12.6.3-cudnn-fe1.10.0-py3.10-pt_2.7.0-dev"
-          CUDA_VERSION_MM: "126"
-        "ubuntu22.04 | cuda 12.6 | torch-nightly":
-          docker-image: "ubuntu22.04-cuda12.6.3-cudnn-fe1.10.0-py3.10-pt_main-dev"
-          CUDA_VERSION_MM: "126"
+        "notebooks w/ torch 2.7":
+          docker-image: "ubuntu24.04-cuda12.6.3-cudnn-fe1.10.0-py3.10-pt_2.7.0-dev"
+        "notebooks w/ torch-nightly":
+          docker-image: "ubuntu24.04-cuda12.6.3-cudnn-fe1.10.0-py3.10-pt_main-dev"
     # how long to run the job before automatically cancelling
     timeoutInMinutes: "45"
     # how much time to give 'run always even if cancelled tasks' before stopping them
@@ -38,7 +36,6 @@ jobs:
     steps:
       - bash: |
           echo $(DEVICES)
-          echo "CUDA_VERSION_MM=$CUDA_VERSION_MM"
           lspci | egrep 'VGA|3D'
           whereis nvidia
           nvidia-smi
--- a/.github/run-benchmark-as-lit-jobs.py
+++ b/.github/run-benchmark-as-lit-jobs.py
@@ -1,13 +1,16 @@
 import glob
 import json
+import sys
 from datetime import datetime
 
 from lightning_sdk import Studio, Job, Machine, Status
 
 
-def main():
+def main(gh_run_id: str = ""):
+    if not gh_run_id:
+        gh_run_id = datetime.now().strftime("%Y-%m-%d|%H:%M:%S")
     print("Creating studio...")
-    s = Studio("thunder-benchmark", "oss-thunder", org="lightning-ai", create_ok=True)
+    s = Studio(f"thunder-benchmark-run{gh_run_id}", "oss-thunder", org="lightning-ai", create_ok=True)
 
     print("Uploading package and benchmark script...")
     s.upload_folder("dist", remote_path="dist")
@@ -20,9 +23,8 @@ def main():
     s.run(f"pip install {pkg_path} -U transformers 'numpy<2.0' 'nvfuser_cu128_torch27==0.2.27.dev20250501'")
 
     print("Running HF benchmark script...")
-    timestamp = datetime.now().strftime("%Y-%m-%d|%H:%M:%S")
     job = Job.run(
-        name=f"benchmark-hf-{timestamp}",
+        name=f"benchmark-run{gh_run_id}",
         command="pip list && python benchmarks/benchmark_hf.py",
         studio=s,
         machine=Machine.L40S,
@@ -30,7 +32,6 @@ def main():
     )
 
     print("Stopping studio...")
-    s.run("rm -rf dist/")
     s.stop()
 
     print("Waiting for job to finish...")
@@ -49,7 +50,12 @@ def main():
         print(job.logs)
         print("=" * 80)
         raise RuntimeError("Benchmark HF job failed")
+    else:  # clean up
+        job.delete()
+        s.delete()
 
 
 if __name__ == "__main__":
-    main()
+    # parse command line arguments
+    args = sys.argv[1:]
+    main(*args)
--- a/.github/run-quickstart-as-lit-jobs.py
+++ b/.github/run-quickstart-as-lit-jobs.py
@@ -1,4 +1,5 @@
 import json
+import sys
 from datetime import datetime
 import glob
 import os.path
@@ -6,9 +7,11 @@ import os.path
 from lightning_sdk import Studio, Job, Machine, Status
 
 
-def main():
+def main(gh_run_id: str = ""):
+    if not gh_run_id:
+        gh_run_id = datetime.now().strftime("%Y-%m-%d|%H:%M:%S")
     print("Creating studio...")
-    s = Studio("thunder-quickstarts", "oss-thunder", org="lightning-ai", create_ok=True)
+    s = Studio(f"thunder-quickstarts-run{gh_run_id}", "oss-thunder", org="lightning-ai", create_ok=True)
     print("Uploading package and scripts...")
     s.upload_folder("dist", remote_path="dist")
     pkg_path = glob.glob("dist/*.whl")[0]
@@ -23,10 +26,9 @@ def main():
     print("Found quickstart scripts:", ls_quickstart)
 
     print("Running quickstart scripts...")
-    timestamp = datetime.now().strftime("%Y-%m-%d|%H:%M:%S")
     jobs = {
         os.path.basename(script): Job.run(
-            name=f"ci-{timestamp}_{script}",
+            name=f"ci-run{gh_run_id}_{script}",
             command=f"pip list && python quickstart/{os.path.basename(script)}",
             studio=s,
             machine=Machine.L40S,
@@ -36,7 +38,6 @@ def main():
     }
 
     print("Stopping studio...")
-    s.run("rm -rf dist/")
     s.stop()
 
     print("Waiting for jobs to finish...")
@@ -47,6 +48,8 @@ def main():
         report[name] = str(job.status).lower()
         if job.status != Status.Completed:
             failures[name] = job.logs
+        else:  # clean up successful jobs
+            job.delete()
 
     with open("quickstart_report.json", "w") as fp:
         json.dump(report, fp, indent=4)
@@ -61,9 +64,11 @@ def main():
 
     assert not failures
 
-    # print("Cleaning up...")
-    # s.delete()
+    print("Cleaning up...")
+    s.delete()
 
 
 if __name__ == "__main__":
-    main()
+    # parse command line arguments
+    args = sys.argv[1:]
+    main(*args)
--- a/.github/workflows/ci-benchmark.yml
+++ b/.github/workflows/ci-benchmark.yml
@@ -6,6 +6,8 @@ on:
     paths:
       - ".github/workflows/benchmark-hf.yml"
       - ".github/run-benchmark-as-lit-jobs.py"
+  push: # enable tracing which commit to master broke it...
+    branches: [main]
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
@@ -38,7 +40,7 @@ jobs:
           LIGHTNING_API_KEY: ${{ secrets.LIGHTNING_API_KEY }}
         run: |
           pip install lightning_sdk -U -q
-          python .github/run-benchmark-as-lit-jobs.py
+          python .github/run-benchmark-as-lit-jobs.py ${{ github.run_id }}
 
       - name: Post Slack Notification
         if: always() && github.event_name != 'pull_request'
--- a/.github/workflows/ci-quickstart.yml
+++ b/.github/workflows/ci-quickstart.yml
@@ -43,7 +43,7 @@ jobs:
           LIGHTNING_API_KEY: ${{ secrets.LIGHTNING_API_KEY }}
         run: |
           pip install lightning_sdk -U -q
-          python .github/run-quickstart-as-lit-jobs.py
+          python .github/run-quickstart-as-lit-jobs.py ${{ github.run_id }}
 
       - name: Load report
         if: always()
--- a/dockers/ubuntu-cuda/Dockerfile
+++ b/dockers/ubuntu-cuda/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-ARG UBUNTU_VERSION="22.04"
+ARG UBUNTU_VERSION="24.04"
 ARG CUDA_VERSION="12.6.3"
 # select devel | runtime
 ARG IMAGE_TYPE="devel"
@@ -108,7 +108,7 @@ RUN \
         git checkout "${TORCH_VERSION}" && \
         git submodule sync && \
         git submodule update --init --recursive && \
-        pip install cmake==3.31.6 && \
+        pip install "cmake==3.31.6" -r requirements.txt && \
         pip install . && \
         pip install "pytorch-triton==$(cat .ci/docker/triton_version.txt)" --index-url="https://download.pytorch.org/whl/nightly/" && \
         cd .. && \
@@ -130,15 +130,7 @@ ENV NVFUSER_BUILD_NO_TEST=1
 ENV NVFUSER_BUILD_NO_BENCHMARK=1
 
 RUN \
-    if [ "${TORCH_INSTALL}" == "source" ]; then \
-        # building nvFuser from source
-        git clone --recursive https://github.com/NVIDIA/Fuser.git && \
-        cd Fuser && \
-        pip install -r requirements.txt && \
-        pip install -v --no-build-isolation python/ && \
-        cd .. && \
-        rm -rf Fuser ; \
-    elif [ "${TORCH_INSTALL}" == "test" ]; then \
+    if [ "${TORCH_INSTALL}" == "source" ] || [ "${TORCH_INSTALL}" == "test" ]; then \
         # building nvFuser from source
         git clone --recursive https://github.com/NVIDIA/Fuser.git && \
         cd Fuser && \
--- a/examples/quickstart/requirements.txt
+++ b/examples/quickstart/requirements.txt
@@ -1,1 +1,3 @@
 transformers
+accelerate
+nvfuser-cu128-torch27
--- a/thunder/__init__.py
+++ b/thunder/__init__.py
@@ -294,10 +294,7 @@ def compile(
             plugins_.append(plugin)
     plugins = plugins_
 
-    if recipe is None and not plugins:
-        return thunder.jit(fn)
-
-    if recipe is None and plugins:
+    if recipe is None:
         recipe = thunder.recipes.BaseRecipe()
 
     if recipe == "auto":
--- a/thunder/benchmarks/benchmark_hf.py
+++ b/thunder/benchmarks/benchmark_hf.py
@@ -166,7 +166,7 @@ if __name__ == "__main__":
     torch.backends.cuda.matmul.allow_tf32 = True
     root.mkdir(parents=True, exist_ok=True)
 
-    # inference_gen()
-    # inference_fwd()
+    inference_gen()
+    inference_fwd()
     training_fwd()
     training_fwd_bwd()
--- a/thunder/core/recipe.py
+++ b/thunder/core/recipe.py
@@ -6,7 +6,7 @@ import warnings
 import torch
 
 from thunder.core.transform_common import Transform
-from thunder.extend import Executor, get_default_executors
+from thunder.extend import Executor
 
 
 @contextmanager
@@ -80,7 +80,7 @@ class Recipe:
         return None
 
     def setup_executors(self) -> list[Executor]:
-        return list(get_default_executors())
+        return []
 
     def setup_config(self) -> dict[str, Any]:
         return {}
--- a/thunder/core/transforms.py
+++ b/thunder/core/transforms.py
@@ -1848,8 +1848,9 @@ def pad_backward(a, padding_config, g):
 
     # Un-pad by padding with zero values
     zero_padding_config = [(-lo, -hi, 0) for lo, hi, _ in padding_config]
+    zero_value = dtypes.dtype_to_numbertype(g.dtype)(0)
 
-    g = prims.pad(g, 0.0, zero_padding_config)
+    g = prims.pad(g, zero_value, zero_padding_config)
 
     # Un-slice by slicing with a stride of value (dilation + 1)
     for dim, (_, _, d) in enumerate(padding_config):
--- a/thunder/dynamo/benchmark_utils.py
+++ b/thunder/dynamo/benchmark_utils.py
@@ -73,6 +73,19 @@ class ThunderCompileSpecification(CompileSpecificationInterface):
         return ["import thunder"]
 
 
+class ThunderCompilerOnGraphModuleSpecification(CompileSpecificationInterface):
+    def __init__(self, specification_name="thunderfx", **kwargs):
+        self.name = specification_name
+        self.thunder_options: dict = kwargs
+
+    def compile(self, gm, **kwargs):
+        from thunder.dynamo import ThunderCompiler
+
+        thunder_compiler = ThunderCompiler(**kwargs)
+        split_gm = thunder_compiler(gm, sample_args=None)
+        return split_gm
+
+
 class TorchCompileSpecification(CompileSpecificationInterface):
     """
     A compile specification for :func:`torch.compile`.
@@ -118,20 +131,24 @@ class TorchInductorSpecification(CompileSpecificationInterface):
     https://github.com/Lightning-AI/lightning-thunder/issues/1521
     """
 
-    def __init__(self, specification_name="inductor_backend"):
+    def __init__(self, specification_name="inductor_backend", *, skip_symbolic_trace=True):
         self.name: str = specification_name
+        # self.skip_symbolic_trace decides whether to skip symbolic trace for self.compile
+        self.skip_symbolic_trace = skip_symbolic_trace
 
     @staticmethod
-    def torch_inductor(fn, inputs):
+    def torch_inductor(fn, inputs, *, skip_symbolic_trace=False):
         from torch._inductor import compile as inductor_compile
         from torch.fx import symbolic_trace
 
-        fx_graph = symbolic_trace(fn)
-        return inductor_compile(fx_graph, inputs)
+        if not skip_symbolic_trace:
+            fn = symbolic_trace(fn)
+        return inductor_compile(fn, inputs)
 
     def compile(self, fn, *, inputs, **kwargs):
-        return self.torch_inductor(fn, inputs)
+        return self.torch_inductor(fn, inputs, skip_symbolic_trace=self.skip_symbolic_trace)
 
+    # to_source will always use symbolic trace
     def to_source(self, fn_name):
         return f"TorchInductorSpecification.torch_inductor({fn_name}, inputs)"
 
@@ -495,8 +512,8 @@ def check_metrics(
             )
             return None
 
-    measure1 = try_and_log_benchmark(compile_fn1, filename1)
-    measure2 = try_and_log_benchmark(compile_fn2, filename2)
+    _, *measure1 = try_and_log_benchmark(compile_fn1, filename1)
+    _, *measure2 = try_and_log_benchmark(compile_fn2, filename2)
 
     if measure1 is None or measure2 is None:
         return measure1, measure2
--- a/thunder/dynamo/report.py
+++ b/thunder/dynamo/report.py
@@ -69,13 +69,15 @@ if TYPE_CHECKING:
     from thunder.dynamo.benchmark_utils import CompileSpecificationInterface, TimerInterface
 
 
-def run_forward_backward(fn, *args, **kwargs):
+def run_forward_backward(fn, *args, benchmark=False, **kwargs):
     result = fn(*args, **kwargs)
     result = sequencify(result)
 
     differentiable_tensor_result = list(filter(lambda x: isinstance(x, torch.Tensor) and x.requires_grad, result))
 
     if not differentiable_tensor_result:
+        if benchmark:
+            return
         return result, None
 
     forward_inputs = tree_flatten((args, kwargs))[0]
@@ -94,6 +96,10 @@ def run_forward_backward(fn, *args, **kwargs):
         i.grad = None
 
     torch.autograd.backward(differentiable_tensor_result, output_grads, inputs=inputs_requires_grad)
+    if benchmark:
+        for i in inputs_requires_grad:
+            i.grad = None
+        return
     return result, [t.grad for t in inputs_requires_grad]
 
 
@@ -451,13 +457,16 @@ class FXGraphReport:
     ):
         torch._dynamo.reset()
         example_inputs = self.make_example_inputs()
+        # ref: https://github.com/pytorch/pytorch/blob/0ef5ba43a6e7fe806ea9f27929bf4328ffd1ebf4/torch/_inductor/compile_fx.py#L1921-L1922
+        # The compile_fn may mutate the GraphModule, so we need to deepcopy it
+        graph = copy.deepcopy(self.graph)
         # To avoid the AssertionError: attribute nodes of Graph object out of sync
-        recompile_graph(self.graph)
-        compiled_model = compile_fn.compile(self.graph, inputs=example_inputs)
+        recompile_graph(graph)
+        compiled_model = compile_fn.compile(graph, inputs=example_inputs)
         result = run_forward_backward(compiled_model, *example_inputs)
 
         if check_consistency:
-            eager_result = run_forward_backward(self.graph, *example_inputs)
+            eager_result = run_forward_backward(graph, *example_inputs)
             torch.testing.assert_close(result, eager_result)
         return result
 
@@ -516,7 +525,13 @@ class FXGraphReport:
         format_python_file(folder / file_name)
 
     def run_benchmark(
-        self, compile_fn: CompileSpecificationInterface, time_fn: TimerInterface, *, reset_torch_dynamo=True
+        self,
+        compile_fn: CompileSpecificationInterface,
+        time_fn: TimerInterface,
+        *,
+        reset_torch_dynamo=True,
+        example_inputs=None,
+        measure_fwd_bwd_together=False,
     ):
         # From torch.compile docs - https://pytorch.org/docs/stable/generated/torch.compile.html
         # > Multiple compiled results can be associated with a frame up to torch._dynamo.config.cache_size_limit, which defaults to 8; at which point we will fall back to eager.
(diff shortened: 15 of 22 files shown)


thunder/executors/functional_teex.py
```diff
@@ -0,0 +1,236 @@
...
+from transformer_engine.pytorch.fp8 import (
+    _amax_and_scale_update,
+    get_fp8_max,
+    get_default_fp8_recipe,
+    Recipe,
+    RecipeState,
+)
+
+
+class StatefulExecutor(OperatorExecutor):
```
kshitij12345, line 23: This seems to be a useful class. Should this be moved to `thunder/extend/__init__.py` so that it is available for other stateful executors in future?

thunder/executors/functional_teex.py
```diff
@@ -0,0 +1,236 @@
...
+functional_te_ex = StatefulExecutor("functional_te")
+register_executor(functional_te_ex)
+
+
+# TODO
+def _functional_te_checker(a, w, /, bias):
+    # TE 2.0 functional API does not yet support bias=True
+    # https://github.com/NVIDIA/TransformerEngine/blob/1321b9b5dc96d67d20d6682c52116a3657f293d3/transformer_engine/pytorch/ops/basic/basic_linear.py#L46-L47
+    # https://github.com/NVIDIA/TransformerEngine/blob/1321b9b5dc96d67d20d6682c52116a3657f293d3/transformer_engine/pytorch/ops/basic/basic_linear.py#L505
+    if bias is not None:
```
kshitij12345, line 50: I am a bit confused. It seems like `_functional_forward` accepts `bias` argument.

https://github.com/NVIDIA/TransformerEngine/blob/1321b9b5dc96d67d20d6682c52116a3657f293d3/transformer_engine/pytorch/ops/basic/basic_linear.py#L334-L342

which is also passed to `general_gemm`
https://github.com/NVIDIA/TransformerEngine/blob/1321b9b5dc96d67d20d6682c52116a3657f293d3/transformer_engine/pytorch/ops/basic/basic_linear.py#L505-L515

And I see that `_functional_backward` doesn't return a grad for `bias` but I think it should be straight forward to do it with a grad rule.

https://github.com/Lightning-AI/lightning-thunder/blob/bff81826518f6c95150fc89a4fe9c11a9ccce9d7/thunder/core/transforms.py#L1311-L1316

NOTE: This could be addressed in a follow-up PR.

thunder/executors/functional_teex.py
```diff
@@ -0,0 +1,236 @@
...
+        return False
+    return _linear_checker(a, w, bias)
+
+
+# Assuming one recipe per machine since it's based on hw capabilities.
+functional_te_recipe = get_default_fp8_recipe()
```
kshitij12345, line 56: Technically you can have both `delayed` and `block` scaling recipe on Blackwell. So, we should allow user to choose the recipe as well.

thunder/executors/functional_teex.py
```diff
@@ -0,0 +1,236 @@
...
+def _te_fp8_quantizers_meta(recipe_state: RecipeState, num_quantizers: int):
+    # giving num quantizers so there is no need to allocate extra things
+    return (*(AnyProxy(num_quantizers, prefix="q") for _ in range(num_quantizers)),)
+
+
+class TEQuantizerState:
+    def __init__(self):
+        self.quantizers = None
+
+    def __call__(self, recipe_state: RecipeState, num_quantizers: int):
```
kshitij12345, line 79: `num_quantizers` is unused here.

thunder/executors/functional_teex.py
```diff
@@ -0,0 +1,236 @@
...
+
+
+_te_fp8_quantizers = functional_te_ex.register_stateful_operator(
+    "te_fp8_quantizers", TEQuantizerState, meta=_te_fp8_quantizers_meta
+)
+
+
+def _te_fp8_state_meta(recipe, mode: str, num_quantizers: int, /):
+    # mode just to give the proxy something wihout allocating extra stuff
+    return AnyProxy(mode, prefix="s")
```
kshitij12345, line 96: Using `AnyProxy(None, prefix="s")` could also be an option.

thunder/executors/functional_teex.py
```diff
@@ -0,0 +1,236 @@
...
+
+
+class TERecipeState:
+    def __init__(self):
+        self.state = None
+
+    def __call__(self, recipe: Recipe, mode, num_quantizers):
```
kshitij12345, line 103: It'd be nice to mention what `mode` is as a comment. 

Had to look at TE to understand: https://github.com/NVIDIA/TransformerEngine/blob/1321b9b5dc96d67d20d6682c52116a3657f293d3/transformer_engine/pytorch/fp8.py#L913-L921

thunder/executors/functional_teex.py
```diff
@@ -0,0 +1,236 @@
...
+    w: TensorProxy,
+    bias: TensorProxy | None,
+    input_quantizer: Float8Quantizer,
+    weight_quantizer: Float8Quantizer,
+):
+    return TensorProxy(like=a), TensorProxy(like=a), TensorProxy(like=w)
+
+
+def _linear_fwd_impl(a, w, bias, input_quantizer, weight_quantizer):
+    out, gemm_a, gemm_w = BasicLinear._functional_forward(
```
kshitij12345, line 133: Wdyt about renaming `gemm_a` -> `quantized_a`?

thunder/executors/functional_teex.py
```diff
@@ -0,0 +1,236 @@
...
+# Assuming one recipe per machine since it's based on hw capabilities.
+functional_te_recipe = get_default_fp8_recipe()
+
+
+def _te_fp8_recipe_meta(recipe_name: str):
+    return AnyProxy(recipe_name, prefix="r")
+
+
+def _te_fp8_recipe_impl(recipe_name: str):
+    return functional_te_recipe
```
kshitij12345, line 64: This just ignore the `recipe_name`. Is it just to have the name in the trace?

thunder/executors/functional_teex.py
```diff
@@ -0,0 +1,236 @@
...
+    return out, gemm_a, gemm_w
+
+
+_te_linear_fwd = functional_te_ex.register_operator(
+    "te_functional_linear_fwd", meta=_linear_fwd_meta, fn=_linear_fwd_impl
+)
+
+
+def _te_linear_execution_transform(a, w, /, bias):
+    recipe = _te_fp8_recipe("delayed")
```
kshitij12345, line 150: On Blackwell, default recipe is `MXFP8BlockScaling`. So the trace would say `delayed` but we would be using `MXFP8BlockScaling` under the hood.

thunder/transforms/functional_te_sync.py
```diff
@@ -0,0 +1,66 @@
...
+
+        sync_states = []
+        last_fp8_linear_idx = 0
+        new_trace = from_trace(computation_trace)
+        new_bsyms = []
+
+        swap_map: dict[VariableInterface, TensorProxy] = {}
+
+        for bsym in computation_trace.bound_symbols:
+            if bsym.sym.name == "te_fp8_recipe":
```
kshitij12345, line 32: It would be nice to have comment for what this for-loop does. As far as I understood, it seems to do
1. Replace the output proxy of all `te_fp8_recipe` with the output of first `te_fp8_recipe`.
2. Gathers state to be passed to `_te_fp8_synchronization`.

Is that correct?

thunder/executors/functional_teex.py
riccardofelluga: Fair enough! It can totally be moved an reused. I was thinking about this from the start but I didn't know if it would end up being "generalized" enough. Will move it there!

thunder/executors/functional_teex.py
riccardofelluga: Thanks for the input! That is actually very good news, I've tried a couple of runs with the bias but the backward transformation was what stopped me. It didn't occur to me that I can use a grad rule. Will implement this asap

thunder/executors/functional_teex.py
riccardofelluga: Yess indeed, as of now I've just been working with delayed, but I'll add support for mxfp8 soon

thunder/executors/functional_teex.py
riccardofelluga: I am also considering what would be the benefits of creating an `FP8StateProxy`. So far it doesn't seem it's needed

thunder/executors/functional_teex.py
riccardofelluga: this outputs are not always quantized so I need to study the API a bit more to understand the situations where these tensors are quantized

thunder/executors/functional_teex.py
riccardofelluga: Yess it's not nice, it's more about the groundwork for supporting the two different recipes(when removing get_default_recipe above)

thunder/transforms/functional_te_sync.py
riccardofelluga: Yes exactly! But you're right, let me add a comment explaining the transform.

riccardofelluga:
While I am still working on this I just want to add that by using this patch you can try this executor on the litgpt benchmark script:

<details>
<summary>benchmarking_litgpt.py diff</summary>

```diff
diff --git a/thunder/benchmarks/benchmark_litgpt.py b/thunder/benchmarks/benchmark_litgpt.py
index 952426b3..a22a5574 100644
--- a/thunder/benchmarks/benchmark_litgpt.py
+++ b/thunder/benchmarks/benchmark_litgpt.py
@@ -23,6 +23,8 @@ from torch.distributed.algorithms._checkpoint.checkpoint_wrapper import (
 import thunder
 from thunder.dynamo import ThunderCompiler

+import thunder.executors
+import thunder.executors.functional_teex
 from thunder.tests.litgpt_model import Config, GPT, Block
 from lightning.fabric.utilities.throughput import measure_flops
 from lightning.fabric.utilities import Throughput
@@ -580,6 +582,14 @@ class Benchmark_litGPT:
             model = torch.compile(model)
         elif "thunder" in self.compile:
             executors = list(thunder.get_default_executors())
+            transforms = []
+
+            if "functional_te" in self.compile:
+                from thunder.executors.functional_teex import functional_te_ex
+                from thunder.transforms import TESynchronizationTransform
+                executors.insert(0, functional_te_ex)
+                transforms.insert(0, TESynchronizationTransform())
+
             if "inductor_cat" in self.compile:
                 from thunder.executors.torch_compile import torch_compile_cat_ex as torch_compile_ex

@@ -601,7 +611,7 @@ class Benchmark_litGPT:

                     dynamo_config.cache_size_limit = 64

-                self.backend = ThunderCompiler(executors=executors)
+                self.backend = ThunderCompiler(executors=executors, transforms=transforms)
                 # Because Lightning Fabric is imported in this script it monkey patches the torch.compile function
                 # https://github.com/Lightning-AI/pytorch-lightning/blob/828fd998961f6a60f92c35254bb94d6e049ad069/src/lightning/fabric/wrappers.py#L421
                 # using __wrapped__ to access the original torch.compile function did not work
@@ -610,7 +620,8 @@ class Benchmark_litGPT:
             else:
                 jit_options = {}
                 jit_options["fp8_shard_intermediate_activation"] = self.fp8_shard_intermediate_activation
-                model = thunder.jit(model, executors=executors, **jit_options)
+                jit_options["auto_recompute_intermediates"] = False
+                model = thunder.jit(model, executors=executors, transforms=transforms, **jit_options)

         elif self.compile != "eager":
             raise ValueError(f"Invalid compile option: {self.compile}")
```
</details>

To run use the following command:
```shell
python thunder/benchmarks/benchmark_litgpt.py --max_iters 20 --warmup_iters 5 --model_name meta-llama/Llama-2-7b-hf --compile thunder+functional_te+dynamo --checkpoint_activations False
```

thunder/executors/functional_teex.py
timmoon10: TE has recently added recipes for FP8 current scaling and DeepSeek-style FP8. Making this configurable could be useful. It could check a global variable or envvar, or perhaps there could be a way to configure specific GEMMs to run with different recipes?

riccardofelluga:
This is still WIP as I am currently investigating the numerical mismatch of this executor compared to the TransformerEngine library, it looks like that after a couple of iterations the scale and amax values for the backward state start differing significantly from the reference implementation

thunder/transforms/functional_te_transform.py
IvanYashchuk, line 1: This transform code should be next to the executor because they are meant to be used together, not independently. If there's no such thing now, then if you really need a whole trace transform after applying "transform for operator execution", there should be a new attribute on OperatorExecutors with defined transforms. If the OperatorExecutor transform is defined like this, then it should be used and applied right after replacing operators with operators from the given OperatorExecutor.

thunder/transforms/functional_te_transform.py
```diff
@@ -0,0 +1,110 @@
...
+
+        sync_trace = del_last_used(new_trace)
+
+        end_time_ns = time.perf_counter_ns()
+        elapsed_time_ns = end_time_ns - start_time_ns
+        elapsed_time_millis = elapsed_time_ns // 1000000
+
+        sync_trace.set_provenance(
+            TraceProvenance(f"TransformerEngine Synchronization transform (took {elapsed_time_millis} milliseconds)")
+        )
```
IvanYashchuk, line 108: This block of code could be a decorator.

thunder/executors/functional_teex.py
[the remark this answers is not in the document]
riccardofelluga: Functional as it uses the functional API, tho yeah it's just to differentiate it from the other transformer engine executor till it replaces it. Can totally be named "transformer_engine_v2"!

thunder/transforms/functional_te_transform.py
riccardofelluga: Yess that's a good idea but I'd add the "default transform" for the executors in another PR. In this one I'll move the transform into the executor #2039

thunder/extend/__init__.py
```diff
@@ -273,6 +274,34 @@ def register_implementation(
...
+        super().__init__(name, version=version)
+        self.state_count: int = 0
+
+    def register_stateful_operator(self, base_name: str, state_class, *, meta):
+        def register_state(*args, **kwargs):
+            state_id = self.state_count
+            name = f"{base_name}_{state_id}"
+
+            def bind_state(bsym):
+                bsym._call_ctx = {name: state_class()}
```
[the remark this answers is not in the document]
riccardofelluga: In the other executors those are needed in order to add the TE context manager to the trace. This executor does not do that so there is no need to store recipe and an import ctx key. 

Also they are blended into one when the trace is executed https://github.com/Lightning-AI/lightning-thunder/blob/bbf85159a4e54174157e88483b6fadf9153cfdf0/thunder/core/trace.py#L348-L361

thunder/executors/functional_teex.py
```diff
@@ -0,0 +1,264 @@
...
+    return AnyProxy(None, prefix="r")
+
+
+class TERecipe:
+    def __init__(self):
+        self.fp8_recipe = None
+
+    def __call__(self) -> list[Recipe]:
+        # If the Thunder trace is not ran inside an fp8_autocast region then
+        # use the default recipe.
```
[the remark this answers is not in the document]
riccardofelluga: The fp8_autocast context manager sets the current recipe in`FP8GlobalStateManager.FP8_RECIPE`.

The only way to change recipe is by using the `fp8_autocast` context manager, so every time this is called, this check will pick the correct recipe since it checks the variable set by that context manager. 

The other option is that no fp8_autocast is used, in that case the recipe won't change so this stores the default one.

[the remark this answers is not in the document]
riccardofelluga: Tho it is true that I still need to accommodate for changes in the recipe as the states are created based on that. So I will add that in this PR

thunder/extend/__init__.py
```diff
@@ -273,6 +274,34 @@ def register_implementation(
...
 
 
+class StatefulExecutor(OperatorExecutor):
+    def __init__(self, name, *, version=None):
+        super().__init__(name, version=version)
+        self.state_count: int = 0
+
+    def register_stateful_operator(self, base_name: str, state_class, *, meta):
+        def register_state(*args, **kwargs):
+            state_id = self.state_count
```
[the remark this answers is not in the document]
riccardofelluga: I've settled on count by using the same convention used in the temporary executor class

thunder/executors/transformer_engine_v2ex.py
```diff
@@ -0,0 +1,19 @@
...
+
+from thunder import Transform
+from thunder.extend import StatefulExecutor
+
+__all__ = ["transformer_engine_v2_ex", "TransformerEngineTransform"]
+
+transformer_engine_v2_ex: None | StatefulExecutor = None
+TransformerEngineTransform: None | Transform = None
+
+if importlib.util.find_spec("transformer_engine"):
```
crcrpar, line 12: `package_available` could be used as well
```python
from lightning_utilities.core.imports import package_available

if package_available("transformer_engine"):
```

thunder/extend/__init__.py
```diff
@@ -273,6 +274,34 @@ def register_implementation(
         self.implmap[_id] = impl
 
 
+class StatefulExecutor(OperatorExecutor):
```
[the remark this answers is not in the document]
crcrpar: Could you add this class to https://github.com/Lightning-AI/lightning-thunder/blob/dea182c6e1804f5a52b26b0cb3bde8e67c031973/docs/source/reference/extend/index.rst?

thunder/extend/__init__.py
```diff
@@ -273,6 +274,78 @@ def register_implementation(
...
+        ...     StateClass,
+        ...     [...same arguments that would have been passed to register_operator...]
+        ... )
+
+    For more examples check out transformer engine executor.
+    """
+
+    def __init__(self, name: Hashable, *, version: Any | None = None):
+        super().__init__(name, version=version)
+        self.counter: int = 0
```
crcrpar, line 302: I just might be paranoid but shouldn't we have this as a "private" member variable and expose this through `@property`? I mean,
```python
...
        self._counter: int = 0

    @property
    def property(self) -> int: return self._property
```

thunder/transforms/__init__.py
```diff
@@ -4,7 +4,6 @@
 from .prune_prologue_checks import PrunePrologueChecks
 from .extraction_only_prologue_transform import ExtractionOnlyPrologueTransform
 
-
```
crcrpar, line 7: This diff does not look expected to me

thunder/tests/test_transformer_engine_v2_executor.py
```diff
@@ -0,0 +1,425 @@
...
+    assert "get_te_fp8_state" in bwd_trace.bound_symbols[14].sym.name
+    assert bwd_trace.bound_symbols[14].sym.name in bwd_trace_pyctx.keys()
+
+    assert "get_te_fp8_quantizers" in bwd_trace.bound_symbols[15].sym.name
+    assert bwd_trace.bound_symbols[15].sym.name in bwd_trace_pyctx.keys()
+
+    assert "te_functional_linear_bwd" in bwd_trace.bound_symbols[16].sym.name
+    assert bwd_trace.bound_symbols[16].sym.name in bwd_trace_pyctx.keys()
+
+    assert "te_fp8_amax_and_scale_update" in bwd_trace.bound_symbols[18].sym.name
```
crcrpar, line 425: I'm fine with these but just wondering if we could avoid hardcoding here.

thunder/tests/test_transformer_engine_v2_executor.py
```diff
@@ -0,0 +1,425 @@
...
+    assert_close(x.grad, x_te.grad)
+    assert_close(w1.grad, te_linear1.weight.grad)
+    assert_close(w2.grad, te_linear2.weight.grad)
+
+    # Verifies te_linear was called
+    forward_trace = thunder.last_traces(cfn)
+    backward_trace = thunder.last_backward_traces(cfn)
+
+    assert any(bsym.sym.name.startswith("te_functional_linear") for bsym in forward_trace[-1].bound_symbols)
+    assert any(bsym.sym.name.startswith("te_functional_linear_bwd") for bsym in backward_trace[-1].bound_symbols)
```
crcrpar, line 88: how about check the number of bsyms satisfying the conditions as well?
e.g.

```python
assert 2 == len(filter(
    lambda bsym: bsym.sym.name.startswith("te_functional_linear"),
    forward_trace[-1].bound_symbols,
))
```

thunder/tests/test_transformer_engine_v2_executor.py
```diff
@@ -0,0 +1,425 @@
...
+    w1 = te_linear1.weight.detach().clone()
+    w1.requires_grad_(True)
+    w2 = te_linear2.weight.detach().clone()
+    w2.requires_grad_(True)
+
+    def fn(x, w1, w2):
+        o = torch.nn.functional.linear(x, w1)
+        return torch.nn.functional.linear(o + x, w2)
+
+    cfn = thunder.jit(fn, executors=[transformer_engine_v2_ex], transforms=[TransformerEngineTransform()])
```
crcrpar, line 61: out of curiosity
1. what would happen if `executors` accidentally has both of the two transformer engine executors?
2. what would happen if I forget to specify the transform class?

crcrpar (commented):
generally looks nice. I just have some questions and some minor comments

thunder/executors/transformer_engine_v2ex.py
riccardofelluga: Yess thanks for the hint! it's indeed way more readable than `importlib.util.find_spec`

thunder/extend/__init__.py
riccardofelluga: I agree on the making private with the underscore, but I think that there is no need to expose it to discourage devs from using it for other things outside the class

thunder/tests/test_transformer_engine_v2_executor.py
riccardofelluga: Yess updated!

thunder/tests/test_transformer_engine_v2_executor.py
riccardofelluga: On the first question what happens depends on the order the executors are specified, the first in list will take over the linear layers so the second transformer engine executor wont be used.(In any case this would be a momentary problem till the old TE executor is replaced)

Regarding the second question, nothing bad will happen but there will be numerical inconstistencies between Thunder  and the same code using just transformer engine library. For this the ideal solution would be to specify a "default transform" for the executor which is something we are thinking about in #2039

thunder/executors/transformer_engine_v2ex_impl.py
```diff
@@ -0,0 +1,390 @@
...
+
+
+class TERecipe:
+    def __init__(self):
+        self.fp8_recipe = None
+
+    def __call__(self) -> list[Recipe]:
+        # If the Thunder trace is not ran inside an fp8_autocast region then
+        # use the default recipe.
+        if not FP8GlobalStateManager.FP8_RECIPE:
```
kshitij12345, line 74: With the newer Operation based (functional API) , linear doesn't need to be run under `fp8_autocast` manager. It determines if casting should be applied based on the quantizers passed.

However, with the current approach here, the user needs to call the thunder function under FP8 autocast. Do we want that? IIUC, to use this executor user needs to pass the executor, pass the transform and call the thunder function under `fp8_autocast`.

I think, it would be nicer if the user passes the recipe as a compile option or we provide `get_te_executor(recipe=None)` function which will get the executor which will apply the correct recipe (and get default if `recipe=None`).

Wdyt?

thunder/executors/transformer_engine_v2ex_impl.py
```diff
@@ -0,0 +1,390 @@
...
+    return AnyProxy(None, prefix="s")
+
+
+class TERecipeState:
+    def __init__(self):
+        self.parent_recipe_type: None | RecipeType = None
+        self.state = None
+
+    def __call__(self, recipe: Recipe, mode: str, num_quantizers: int) -> RecipeState:
+        recipe_type = get_fp8_recipe_type(recipe)
```
kshitij12345, line 125: Do we need this function `get_fp8_recipe_type` and the enum `RecipeType` only to verify if current recipe matches the previous? Can't we just compare the recipes directly?

thunder/executors/transformer_engine_v2ex_impl.py
```diff
@@ -0,0 +1,390 @@
...
+    # Therefore it's needed here to inform the meta.
+    return (*(AnyProxy(None, prefix="q") for _ in range(num_quantizers)),)
+
+
+class TEQuantizerState:
+    def __init__(self):
+        self.quantizers = None
+        self.parent_recipe_state: None | RecipeState = None
+
+    def __call__(self, recipe_state: RecipeState, num_quantizers: int) -> list[Float8Quantizer]:
```
kshitij12345, line 99: Depending on the recipe it could be other quantizer as well.

https://github.com/NVIDIA/TransformerEngine/blob/0e45e138c08af8f3b38e46eea58e2e9dbe628d42/transformer_engine/pytorch/tensor/mxfp8_tensor.py#L24

https://github.com/NVIDIA/TransformerEngine/blob/0e45e138c08af8f3b38e46eea58e2e9dbe628d42/transformer_engine/pytorch/tensor/float8_blockwise_tensor.py#L21

I think annotating with their base class `Quantizer` would be appropriate.

thunder/executors/transformer_engine_v2ex_impl.py
```diff
@@ -0,0 +1,390 @@
...
+    w: TensorProxy,
+    bias: TensorProxy | None,
+    input_quantizer: AnyProxy,
+    weight_quantizer: AnyProxy,
+):
+    out_shape = (*a.shape[:-1], w.shape[0])
+    return TensorProxy(like=a, shape=out_shape), TensorProxy(like=a), TensorProxy(like=w)
+
+
+def _linear_fwd_impl(a, w, bias, input_quantizer: Float8Quantizer, weight_quantizer: Float8Quantizer):
```
kshitij12345, line 160: Float8Quantizer -> Quantizer

thunder/executors/transformer_engine_v2ex_impl.py
```diff
@@ -0,0 +1,390 @@
...
+            if bsym.sym.id == prims.PrimIDs.DEL:
+                continue
+
+            # Save the first occurrence of a recipe symbol and map any later ones in the redundant_map
+            if "get_te_fp8_recipe" in bsym.sym.name:
+                # Store the first occurrence
+                if not self.fp8_recipe:
+                    self.fp8_recipe = bsym
+                else:
+                    vsrc = variableify(bsym.output)
```
kshitij12345, line 348: We should assert in `else` that the recipes match.

thunder/tests/test_transformer_engine_v2_executor.py
```diff
@@ -0,0 +1,443 @@
...
+
+hybrid_fp8_delayed_scaling_recipe = recipe.DelayedScaling()
+mxfp8_e4m3_recipe = recipe.MXFP8BlockScaling()
+
+# `None` is used to test the default recipe.
+recipes = (None, hybrid_fp8_delayed_scaling_recipe, mxfp8_e4m3_recipe)
+recipe_ids = ("default", "delayed_scaling", "mxfp8_e4m3")
+
+
+@requiresCUDA
```
kshitij12345, line 32: The tests here are very similar to the existing `test_transformer_engine_executor`. Would it be possible to update the existing tests to work with both executors?

kshitij12345 (commented):
Overall the PR looks good, great work! Just have a few comments and questions.

Also, regarding the trace representation (from the PR description),

```python
import torch

@torch.no_grad()
@no_autocast
def computation(a, w):
  # a: "cuda:0 bf16[3, 768, 4096]"
  # w: "cuda:0 bf16[4096, 4096]"

  r3 = get_te_fp8_recipe_3()  # r3: "<class 'NoneType'>"
  s4 = get_te_fp8_state_4(r3, 'forward', 2)  # s4: "<class 'NoneType'>"
  (q6, q7) = get_te_fp8_quantizers_5(s4, 2)
  (t10, _, _) = te_functional_linear_fwd(a, w, None, q6, q7)
  del q6, q7

  (t13,) = te_fp8_amax_and_scale_update(r3, states=(s4,), tokens=(t10,))
  del r3, s4, t10
  return (t13,)
```

In the trace, we see `del r3, s4, t10` but IIUC, `s4` is not really deleted (as for delayed scaling, we need to preserve it across iteration). Is that correct? Is there a way to tag s4 so that it is not in `del` in the trace?

thunder/executors/transformer_engine_v2ex_impl.py
riccardofelluga: Not quite, the Thunder function does not need to be called exclusively under an `fp8_autocast` context. When the trace is ran, that `__call__` function that you highlighted will check weather there is an active autocast region or not by looking into TE's `FP8GlobalStateManager`. In the case that there is no active `fp8_autocast` region or the `fp8_autocast` region has been called with a None recipe, Thunder will pick the default recipe for the device onto which the trace is ran.

Also in the same spirit, since this is done at runtime, no recipe nor `fp8_autocast` region is needed to create executor or transform

thunder/tests/test_transformer_engine_v2_executor.py
riccardofelluga: Yes they can be parametrized by executor too, i kept it separate because I am thinking that it might be when we switch the old te with the new executor

thunder/executors/transformer_engine_v2ex_impl.py
kshitij12345: If user wanted to specify a different recipe than default then they would need something like the following, right?

```python
tfn = thunder.jit(fn, executors=(functional_te_ex, ), transforms=(TransformerEngineTransform(), )
with fp8_autocast(user_recipe):
    tfn(x)

tfn(x)  # Default recipe for low precision GEMM
```

If the user is familiar with TE API, this may seem a bit confusing. Because for TE API without `fp8_autocast`, computation happens in higher precision.
```python
m = te.Linear(16, 16)

with fp8_autocast(user_recipe):
      m(x)  # User recipe is used.

m(x)  # High-precision is used for computation
```

thunder/tests/test_transformer_engine_v2_executor.py
kshitij12345: I think it'd better to have only a single set of tests. We can remove the parametrization for the older executor when we remove it.

thunder/executors/transformer_engine_v2ex_impl.py
riccardofelluga: Yess updated to check for ids and not using type

thunder/executors/transformer_engine_v2ex_impl.py
riccardofelluga: Yess exaclty, because `with_quantized_compute` is hardcoded to True. Otherwise it is possible to check `FP8GlobalStateManager.FP8_ENABLED` at runtime and match TE behavior

riccardofelluga: However, I don't think we should match the behavior in this case. Because yes we use the context manager to provide the recipe, but my assumption is that if you are using te executor you want fp8 to be enabled(like the current te executor)

kshitij12345: Right I agree that passing the executor implies wanting to use FP8 when the executor is passed.

I have two questions regarding `fp8_autocast` usage to detect recipe:
1. Not using `fp8_autocast` will have different behaviour when used with te.Linear (with PyTorch eager) and when used with thunder. 

2. To get recipe from `fp8_autocast`, we need to rely on `FP8GlobalStateManager` which is undocumented, so I am not sure if it is ok to rely on it (as that is one of the reasons we want to move from current executor).

That is why I think it would be better if the recipe is passed as option to `thunder.jit` or to the executor.

thunder/tests/test_transformer_engine_v2_executor.py
kshitij12345: Got it.

IvanYashchuk:
TransformerEngine is not installed in CI, so the green status may be false (https://github.com/Lightning-AI/lightning-thunder/issues/196).

At the moment I see one failing test:
```
FAILED tests/test_transformer_engine_v2_executor.py::test_te_linear_forward_backward_multiple_iteration[default] - AssertionError: Tensor-likes are not close!

Mismatched elements: 16777216 / 16777216 (100.0%)
Greatest absolute difference: 8519680.0 at index (0, 3162) (up to 1e-05 allowed)
Greatest relative difference: 1.0078125 at index (0, 0) (up to 0.016 allowed)
```

IvanYashchuk (approved):
Could you please create an issue about the next steps for testing DPP and FSDP?

What's the performance of Llama 3 on a single GPU with this and the previous version of the TransformerEngine executor?

thunder/executors/transformer_engine_v2ex_impl.py
```diff
@@ -0,0 +1,371 @@
...
+def _get_te_fp8_quantizers_meta(recipe_state: RecipeState, num_quantizers: int):
+    # num_quantizers can be taken from recipe state, but not at trace time.
+    # Therefore it's needed here to inform the meta.
+    return (*(AnyProxy(None, prefix="q") for _ in range(num_quantizers)),)
+
+
+class TEQuantizerState:
+    def __init__(self):
+        self.quantizers: None | list[Quantizer] = None
+        self.parent_recipe_state: None | RecipeState = None
```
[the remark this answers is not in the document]
riccardofelluga: Parent as in the recipe state that was used to create the quantizers

thunder/executors/transformer_engine_v2ex.py
```diff
@@ -0,0 +1,20 @@
...
+
+__all__ = ["transformer_engine_v2_ex", "TransformerEngineTransform"]
+
+transformer_engine_v2_ex: None | StatefulExecutor = None
+TransformerEngineTransform: None | Transform = None
+
+if package_available("transformer_engine"):
+    import thunder.executors.transformer_engine_v2ex_impl as impl
+
+    transformer_engine_v2_ex = impl.functional_te_ex
```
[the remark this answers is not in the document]
riccardofelluga: Oversight! 😬 Updated the name

thunder/tests/test_transformer_engine_v2_executor.py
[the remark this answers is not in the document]
riccardofelluga: v1 tests do not use the autocast for the thunder runs in the tests and v2 has different symbol names to check in the trace. As of writing all the tests have been ported in the v2 file, if any will be added later to the v1 testing suite I will port them and unify the tests accordingly

riccardofelluga:
> TransformerEngine is not installed in CI, so the green status may be false (#196).
> 
> At the moment I see one failing test:
> 
> ```
> FAILED tests/test_transformer_engine_v2_executor.py::test_te_linear_forward_backward_multiple_iteration[default] - AssertionError: Tensor-likes are not close!
> 
> Mismatched elements: 16777216 / 16777216 (100.0%)
> Greatest absolute difference: 8519680.0 at index (0, 3162) (up to 1e-05 allowed)
> Greatest relative difference: 1.0078125 at index (0, 0) (up to 0.016 allowed)
> ```

Actually this is a "feature" 😁 when the recipe is None, then TE creates a new recipe under the hood for every time we enter the `fp8_autocast` context. So numbers are not comparable because the recipes are always different(i've added the skip)

thunder/executors/transformer_engine_v2ex_impl.py
riccardofelluga: Marking as completed, moving the conversation to #2080

kshitij12345:
> Actually this is a "feature" 😁 when the recipe is None, then TE creates a new recipe under the hood for every time we enter the fp8_autocast context. So numbers are not comparable because the recipes are always different(i've added the skip)

I think this could lead to problem with convergence. 

TE v2: `python thunder/benchmarks/benchmark_litgpt.py --max_iters 20 --warmup_iters 5 --model_name meta-llama/Llama-3.2-1B --n_layers 5 --compile thunder+transformerengine_v2+dynamo --checkpoint_activations False --block_size 4096`

```
iter 0: loss 11.9375, iter time: 13155.99ms, t: 4096
iter 1: loss 11.8750, iter time: 111.68ms, t: 4096
iter 2: loss 11.8125, iter time: 106.28ms, t: 4096
iter 3: loss 11.7500, iter time: 112.01ms, t: 4096
iter 4: loss 11.6875, iter time: 105.83ms, t: 4096
iter 5: loss 11.6250, iter time: 108.55ms, t: 4096
iter 6: loss 11.5000, iter time: 108.41ms, t: 4096
iter 7: loss 11.4375, iter time: 106.66ms, t: 4096
iter 8: loss 11.3750, iter time: 112.77ms, t: 4096
iter 9: loss 11.3125, iter time: 106.67ms, t: 4096
iter 10: loss 11.2500, iter time: 114.51ms, t: 4096
iter 11: loss 11.1875, iter time: 109.91ms, t: 4096
iter 12: loss 11.1250, iter time: 108.99ms, t: 4096
iter 13: loss 11.0625, iter time: 107.20ms, t: 4096
iter 14: loss 11.0000, iter time: 114.36ms, t: 4096
iter 15: loss 10.8750, iter time: 108.55ms, t: 4096
iter 16: loss 10.8750, iter time: 109.44ms, t: 4096
iter 17: loss 10.7500, iter time: 106.26ms, t: 4096
iter 18: loss 10.6250, iter time: 113.68ms, t: 4096
iter 19: loss 10.5000, iter time: 107.46ms, t: 4096
Model name: meta-llama/Llama-3.2-1B
Seq Length: 4096
Micro BS: 1
Global BS: 1
Number of Layers: 5
Number of parameters: 0.83B
Distributed Mode: none
Compiler: thunder+transformerengine_v2+dynamo
Low Precision Mode: none
Average iter time: 109.58 ms
Memory used: 12.11 GB
Saved for backward size: 989.54 MiB
Saved for backward number of tensors: 25
Tokens/s: 37352.76
Tokens/s/GPU: 37352.76
TFLOP/s: 145.82
```

TE v1: `python thunder/benchmarks/benchmark_litgpt.py --max_iters 20 --warmup_iters 5 --model_name meta-llama/Llama-3.2-1B --n_layers 5 --compile thunder+transformerengine+dynamo --checkpoint_activations False --block_size 4096`

```
iter 0: loss 11.9375, iter time: 12751.10ms, t: 4096
iter 1: loss 11.8750, iter time: 117.55ms, t: 4096
iter 2: loss 11.0000, iter time: 126.23ms, t: 4096
iter 3: loss 8.3750, iter time: 131.56ms, t: 4096
iter 4: loss 6.8125, iter time: 129.35ms, t: 4096
iter 5: loss 5.6875, iter time: 135.55ms, t: 4096
iter 6: loss 5.0000, iter time: 129.78ms, t: 4096
iter 7: loss 4.7500, iter time: 126.43ms, t: 4096
iter 8: loss 4.6562, iter time: 125.65ms, t: 4096
iter 9: loss 4.6562, iter time: 132.58ms, t: 4096
iter 10: loss 4.6250, iter time: 113.54ms, t: 4096
iter 11: loss 4.6562, iter time: 126.40ms, t: 4096
iter 12: loss 4.6250, iter time: 123.81ms, t: 4096
iter 13: loss 4.6250, iter time: 128.72ms, t: 4096
iter 14: loss 4.6250, iter time: 132.81ms, t: 4096
iter 15: loss 4.6250, iter time: 119.90ms, t: 4096
iter 16: loss 4.6250, iter time: 126.24ms, t: 4096
iter 17: loss 4.6250, iter time: 124.86ms, t: 4096
iter 18: loss 4.6562, iter time: 121.86ms, t: 4096
iter 19: loss 4.6250, iter time: 112.24ms, t: 4096
Model name: meta-llama/Llama-3.2-1B
Seq Length: 4096
Micro BS: 1
Global BS: 1
Number of Layers: 5
Number of parameters: 0.83B
Distributed Mode: none
Compiler: thunder+transformerengine+dynamo
Low Precision Mode: none
Average iter time: 125.38 ms
Memory used: 12.11 GB
Saved for backward size: 1594.54 MiB
Saved for backward number of tensors: 40
Tokens/s: 32859.21
Tokens/s/GPU: 32859.21
TFLOP/s: 128.28
```

riccardofelluga:
> > Actually this is a "feature" 😁 when the recipe is None, then TE creates a new recipe under the hood for every time we enter the fp8_autocast context. So numbers are not comparable because the recipes are always different(i've added the skip)
> 
> I think this could lead to problem with convergence.


@kshitij12345 I've saved the full log of my runs and I did not observe any problem in convergence(note the following is from the output of my previous perf comment)

```shell
python thunder/benchmarks/benchmark_litgpt.py --max_iters 20 --warmup_iters 5 --model_name meta-llama/Llama-3.2-1B --n_layers 1 --compile thunder+transformerengine+dynamo --checkpoint_activations False
Loading model with {'name': 'Llama-3.2-1B', 'hf_config': {'org': 'meta-llama', 'name': 'Llama-3.2-1B'}, 'block_size': 131072, 'n_layer': 1, 'n_embd': 2048, 'vocab_size': 128000, 'padding_multiple': 512, 'padded_vocab_size': 128256, 'norm_class_name': 'RMSNorm', 'norm_eps': 1e-05, 'norm_qk': False, 'post_attention_norm': False, 'post_mlp_norm': False, 'parallel_residual': False, 'shared_attention_norm': False, 'n_head': 32, 'head_size': 64, 'n_query_groups': 8, 'attn_bias': False, 'attention_scores_scalar': None, 'sliding_window_size': None, 'sliding_window_indices': None, 'attention_logit_softcapping': None, 'rope_base': 500000, 'rotary_percentage': 1.0, 'rope_condense_ratio': 1, 'rope_adjustments': {'factor': 32.0, 'low_freq_factor': 1.0, 'high_freq_factor': 4.0, 'original_max_seq_len': 8192}, 'intermediate_size': 8192, 'bias': False, 'mlp_class_name': 'LLaMAMLP', 'gelu_approximate': 'none', 'n_expert': 0, 'n_expert_per_token': 0, 'scale_embeddings': False, 'lm_head_bias': False, 'final_logit_softcapping': None, 'rope_local_base_freq': None, 'rope_indices': None, 'rope_n_elem': 64}
Time to instantiate model: 0.26 seconds.
/opt/pytorch/lightning-thunder/thunder/dynamo/utils.py:494: Converting a tensor with requires_grad=True to a scalar may lead to unexpected behavior.
Consider using tensor.detach() first. (Triggered internally at /opt/pytorch/pytorch/aten/src/ATen/native/Scalar.cpp:22.)
iter 0: loss 11.9375, iter time: 8263.78ms, t: 131072
iter 1: loss 11.3750, iter time: 893.59ms, t: 131072
iter 2: loss 9.8750, iter time: 885.15ms, t: 131072
iter 3: loss 7.5000, iter time: 859.14ms, t: 131072
iter 4: loss 5.5938, iter time: 866.76ms, t: 131072
iter 5: loss 4.9062, iter time: 882.27ms, t: 131072
iter 6: loss 4.6875, iter time: 856.15ms, t: 131072
iter 7: loss 4.6250, iter time: 884.59ms, t: 131072
iter 8: loss 4.6562, iter time: 845.54ms, t: 131072
iter 9: loss 4.6562, iter time: 872.12ms, t: 131072
iter 10: loss 4.6250, iter time: 847.81ms, t: 131072
iter 11: loss 4.6250, iter time: 847.33ms, t: 131072
iter 12: loss 4.6250, iter time: 849.47ms, t: 131072
iter 13: loss 4.6250, iter time: 843.24ms, t: 131072
iter 14: loss 4.6250, iter time: 841.88ms, t: 131072
iter 15: loss 4.6250, iter time: 857.12ms, t: 131072
iter 16: loss 4.6250, iter time: 849.65ms, t: 131072
iter 17: loss 4.6250, iter time: 853.72ms, t: 131072
iter 18: loss 4.6250, iter time: 846.88ms, t: 131072
iter 19: loss 4.6250, iter time: 854.12ms, t: 131072
Model name: meta-llama/Llama-3.2-1B
Seq Length: 131072
Micro BS: 1
Global BS: 1
Number of Layers: 1
Number of parameters: 0.59B
Distributed Mode: none
Compiler: thunder+transformerengine+dynamo
Low Precision Mode: none
Average iter time: 855.53 ms
Memory used: 149.84 GB
Saved for backward size: 13515.01 MiB
Saved for backward number of tensors: 40
Tokens/s: 153549.29
Tokens/s/GPU: 153549.29
TFLOP/s: 792.64

python thunder/benchmarks/benchmark_litgpt.py --max_iters 20 --warmup_iters 5 --model_name meta-llama/Llama-3.2-1B --n_layers 1 --compile thunder+transformerengine_v2+dynamo --checkpoint_activations False
Loading model with {'name': 'Llama-3.2-1B', 'hf_config': {'org': 'meta-llama', 'name': 'Llama-3.2-1B'}, 'block_size': 131072, 'n_layer': 1, 'n_embd': 2048, 'vocab_size': 128000, 'padding_multiple': 512, 'padded_vocab_size': 128256, 'norm_class_name': 'RMSNorm', 'norm_eps': 1e-05, 'norm_qk': False, 'post_attention_norm': False, 'post_mlp_norm': False, 'parallel_residual': False, 'shared_attention_norm': False, 'n_head': 32, 'head_size': 64, 'n_query_groups': 8, 'attn_bias': False, 'attention_scores_scalar': None, 'sliding_window_size': None, 'sliding_window_indices': None, 'attention_logit_softcapping': None, 'rope_base': 500000, 'rotary_percentage': 1.0, 'rope_condense_ratio': 1, 'rope_adjustments': {'factor': 32.0, 'low_freq_factor': 1.0, 'high_freq_factor': 4.0, 'original_max_seq_len': 8192}, 'intermediate_size': 8192, 'bias': False, 'mlp_class_name': 'LLaMAMLP', 'gelu_approximate': 'none', 'n_expert': 0, 'n_expert_per_token': 0, 'scale_embeddings': False, 'lm_head_bias': False, 'final_logit_softcapping': None, 'rope_local_base_freq': None, 'rope_indices': None, 'rope_n_elem': 64}
Time to instantiate model: 0.25 seconds.
/opt/pytorch/lightning-thunder/thunder/dynamo/utils.py:494: Converting a tensor with requires_grad=True to a scalar may lead to unexpected behavior.
Consider using tensor.detach() first. (Triggered internally at /opt/pytorch/pytorch/aten/src/ATen/native/Scalar.cpp:22.)
iter 0: loss 11.9375, iter time: 8296.45ms, t: 131072
iter 1: loss 11.3750, iter time: 872.37ms, t: 131072
iter 2: loss 9.8750, iter time: 858.54ms, t: 131072
iter 3: loss 7.5000, iter time: 859.69ms, t: 131072
iter 4: loss 5.5938, iter time: 860.72ms, t: 131072
iter 5: loss 4.9062, iter time: 874.81ms, t: 131072
iter 6: loss 4.6875, iter time: 876.73ms, t: 131072
iter 7: loss 4.6250, iter time: 873.63ms, t: 131072
iter 8: loss 4.6562, iter time: 844.99ms, t: 131072
iter 9: loss 4.6562, iter time: 843.06ms, t: 131072
iter 10: loss 4.6250, iter time: 845.17ms, t: 131072
iter 11: loss 4.6250, iter time: 842.99ms, t: 131072
iter 12: loss 4.6250, iter time: 849.01ms, t: 131072
iter 13: loss 4.6250, iter time: 841.71ms, t: 131072
iter 14: loss 4.6250, iter time: 845.71ms, t: 131072
iter 15: loss 4.6250, iter time: 850.40ms, t: 131072
iter 16: loss 4.6250, iter time: 845.76ms, t: 131072
iter 17: loss 4.6250, iter time: 852.72ms, t: 131072
iter 18: loss 4.6250, iter time: 847.25ms, t: 131072
iter 19: loss 4.6250, iter time: 847.31ms, t: 131072
Model name: meta-llama/Llama-3.2-1B
Seq Length: 131072
Micro BS: 1
Global BS: 1
Number of Layers: 1
Number of parameters: 0.59B
Distributed Mode: none
Compiler: thunder+transformerengine_v2+dynamo
Low Precision Mode: none
Average iter time: 852.15 ms
Memory used: 149.84 GB
Saved for backward size: 12910.01 MiB
Saved for backward number of tensors: 25
Tokens/s: 154106.03
Tokens/s/GPU: 154106.03
TFLOP/s: 795.52
```

kshitij12345:
> @kshitij12345 I've saved the full log of my runs and I did not observe any problem in convergence(note the following is from the output of my previous perf comment). Could you share more details on your setup?

Sure, I ran it on RTX6000 (probably will also happen on H100). I guess this would happen when default recipe is DelayedScaling which requires information from previous iteration to be carried over.

For the above logs, I guess you ran them on B200 where default recipe doesn't carry state across iterations.

riccardofelluga:
I think everyone had opportunity to review this PR so let me push it forward by asking for a stamp 🙏 @t-vi @mruberry

thunder/executors/transformer_engine_v2ex_impl.py
```diff
@@ -0,0 +1,371 @@
...
+        Finds and replaces TE executor recipe calls and replaces them with one.
+
+        This function may be called twice, once with the forward trace and once with the backward trace.
+        It will save the first occurance of a recipe from the trace and use it to replce all the others.
+
+        Args:
+            computation_trace: Trace to perform the replacement on.
+        """
+
+        if "transformer_engine_v2_ex" not in map(lambda x: x.name, kwargs["executors_list"]):
```
riccardofelluga, line 310:
```suggestion
        if "transformer_engine_v2" not in map(lambda x: x.name, kwargs["executors_list"]):
```

t-vi (approved):
I'm not sure about the new executor class, but let's try this.

PR #1909, opened 2025-03-27
add `imag` op

khushi-411:
As per the title.

referred to implementation of the `real` op.


thunder/clang/__init__.py
```diff
@@ -1791,6 +1791,20 @@ def real(a: TensorProxy | Number):
     )
 
 
+@clangop(method_name="imag")
+def imag(a: TensorProxy | Number) -> TensorLike:
```
mruberry, line 1795: Nit: `(a: ..., /) -> TensorLike:`

thunder/core/prims.py
```diff
@@ -2375,6 +2376,29 @@ def real_meta(a: complex | TensorProxy) -> float | TensorProxy:
...
+
+def imag_meta(a: complex | TensorProxy) -> float | TensorProxy:
+    utils.check_type(a, (TensorProxy, complex))
+    dtyp = dtypes.to_dtype(a, true_dtype=True)
+    utils.check(
+        dtyp, lambda: f"imag expected a complex tensor or number, but receive a tensor or number with dtype {dtyp}"
+    )
+    output_dtype = dtypes.corresponding_real_dtype(dtyp)
+
+    if isinstance(a, complex):
```
mruberry, line 2388: Nice handling of Python numbers and tensorlikes

thunder/tests/opinfos.py
```diff
@@ -2002,6 +2002,34 @@ def hardtanh_singularity_fn_producer(sample):
...
+    for dtype in dtypes:
+        make = partial(make_tensor, device=device, dtype=dtype)
+        for shape in cases:
+            yield (SampleInput(make(shape)), RuntimeError, err_msg)
+
+
+imag_opinfo = OpInfo(
+    clang.imag,
+    dtypes=(datatypes.complexfloating,),
+    sample_input_generator=elementwise_unary_generator,
```
mruberry, line 2025: Nice generator reuse

thunder/tests/opinfos.py
```diff
@@ -2002,6 +2002,34 @@ def hardtanh_singularity_fn_producer(sample):
 elementwise_unary_ops.append(real_opinfo)
 
 
+def imag_error_generator(op, device, **kwargs):
```
mruberry, line 2005: Cool error generator

thunder/tests/opinfos.py
```diff
@@ -2002,6 +2002,34 @@ def hardtanh_singularity_fn_producer(sample):
...
+    dtypes = [torch.float32, torch.float64, torch.bfloat16, torch.int64]
+    cases = (
+        (),
+        (5),
+        (4, 4),
+        (4, 1, 4),
```
mruberry, line 2011: The 3 and 4d cases can probably be removed (just to make the test run faster)

thunder/tests/opinfos.py
```diff
@@ -2002,6 +2002,34 @@ def hardtanh_singularity_fn_producer(sample):
...
+        (1, 2, 3, 4),
+    )
+
+    err_msg = "imag is not implemented for tensors with non-complex dtypes"
+    for dtype in dtypes:
```
mruberry, line 2016: using itertools.product might make this easier to read:

```
for dtype, shape in product(dtypes, cases):
```

thunder/tests/opinfos.py
```diff
@@ -2002,6 +2002,34 @@ def hardtanh_singularity_fn_producer(sample):
 elementwise_unary_ops.append(real_opinfo)
 
 
+def imag_error_generator(op, device, **kwargs):
+    dtypes = [torch.float32, torch.float64, torch.bfloat16, torch.int64]
```
mruberry, line 2006: Just to reduce test time, it's probably OK to test only float32 and int64

thunder/torch/__init__.py
```diff
@@ -1766,6 +1766,11 @@ def real(a):
     return clang.real(a)
 
 
+@torchsymbol(torch.imag, is_method=False)
```
mruberry, line 1769: Isn't `imag` also a method? 

https://pytorch.org/docs/stable/generated/torch.Tensor.imag.html

mruberry (commented):
Cool! Lots of careful details to consider with `imag`. I left some comments inline for your review, @khushi-411.

@beverlylytle, would you like to review this?

thunder/core/prims.py
```diff
@@ -2375,6 +2376,29 @@ def real_meta(a: complex | TensorProxy) -> float | TensorProxy:
     meta=real_meta,
 )
 
+
+def imag_meta(a: complex | TensorProxy) -> float | TensorProxy:
+    utils.check_type(a, (TensorProxy, complex))
+    dtyp = dtypes.to_dtype(a, true_dtype=True)
+    utils.check(
+        dtyp, lambda: f"imag expected a complex tensor or number, but receive a tensor or number with dtype {dtyp}"
+    )
```
beverlylytle, line 2385: I know this is based on the meta for `real`, but I don't understand this check either here or there.  It asserts that `dtyp` isn't falsey.  The output of `to_dtype` is either a dtype (never falsey, right?) or `None`.  But it can never be `None` because the the `check_type` in the first line of the function.  Based on the error message, should this instead be `utils.check(is_complex_dtype(dtyp), ...`?

beverlylytle (commented):
Thank you for the PR!  

It should be the case that the way `imag` is supported should mirror the way `real` is supported.  Maybe I'm just being slow on a Monday morning, but somethings about the way `real` is supported don't make sense to me.  If you agree that something's fishy with `real`, then let's do it "right" in this PR and file a bug with the issues for `real`.

Would you be interested in also adding the implementation for nvfuser's executor?  It should be very similar to the [one](https://github.com/Lightning-AI/lightning-thunder/blob/1fc0cc08308fcda0302acf0a964d9c33a57aa1a7/thunder/executors/nvfuserex_impl.py#L1617) for `real`?

thunder/core/prims.py
mruberry: Yes! In general we shouldn't rely on implicit conversions to bool, anyway, because it can be confusing

thunder/torch/__init__.py
khushi-411: Thank you! After making the change in my latest commit, this works:
```python

In [1]: import torch
   ...: import thunder
   ...: 
   ...: def f(x):
   ...:     z = x.real
   ...:     return z
   ...: 
   ...: def g(x):
   ...:     z = torch.real(x)
   ...:     return z
   ...: 
   ...: jf = thunder.jit(f)
   ...: jg = thunder.jit(g)
   ...: 
   ...: x = torch.rand((2, 2), dtype=torch.complex128)
   ...: 
   ...: f(x)            # OK  
   ...: g(x)           # OK  
   ...: 
   ...: jf(x)           # OK
   ...: jg(x)
Out[1]: 
tensor([[0.5807, 0.8727],
        [0.1300, 0.8024]], dtype=torch.float64)
```

mruberry (commented):
This look good, although the CI failure is unexpected:

```
        torch.testing.assert_close(actual_grad, expect_grad)
    
        from torch.autograd.gradcheck import GradcheckError
        from torch.testing._internal.common_utils import gradcheck
    
        with pytest.raises(GradcheckError):
>           gradcheck(jitted, (x,))

thunder/tests/test_jit_general.py:1269: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/opt/hostedtoolcache/Python/3.13.2/x64/lib/python3.13/site-packages/torch/testing/_internal/common_utils.py:4914: in gradcheck
    return torch.autograd.gradcheck(fn, inputs, **kwargs)
/opt/hostedtoolcache/Python/3.13.2/x64/lib/python3.13/site-packages/torch/autograd/gradcheck.py:2055: in gradcheck
    return _gradcheck_helper(**args)
/opt/hostedtoolcache/Python/3.13.2/x64/lib/python3.13/site-packages/torch/autograd/gradcheck.py:2109: in _gradcheck_helper
    _test_batched_grad(tupled_inputs, o, i)
/opt/hostedtoolcache/Python/3.13.2/x64/lib/python3.13/site-packages/torch/autograd/gradcheck.py:1156: in _test_batched_grad
    result = vmap(vjp)(torch.stack(grad_outputs))
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

args = (<function _test_batched_grad.<locals>.vjp at 0x7f716ea3aa20>,)
kwargs = {}

    @functools.wraps(arg)
    def wrapper(*args, **kwargs):
>       warn(msg, category=category, stacklevel=stacklevel + 1)
E       FutureWarning: Please use `torch.vmap` instead of `torch._vmap_internals.vmap`.

/opt/hostedtoolcache/Python/3.13.2/x64/lib/python3.13/warnings.py:636: FutureWarning
```

It looks like this error originates in PyTorch (the `gradcheck.py` file has a few uses of `_vmap`). I'm guessing we weren't hitting it before because we weren't testing any complex autodifferentiation. Maybe @beverlylytle can advise us, but I think we need to suppress this warning when calling `gradcheck`, maybe like this:

```
@pytest.mark.filterwarnings("ignore:Please use `torch.vmap`")
```

Maybe we should just wrap the `gradcheck` function in our own helper that suppresses this warning?

beverlylytle:
Hmph, that warning is interesting.  There's an attempt to filter out such warnings at the [place](https://github.com/pytorch/pytorch/blob/f74d5d576aedf053b7574f3eb06d12417d80625a/torch/autograd/gradcheck.py#L1151) where `vmap` is called.  I imagine that pytest is interfering with `warnings`, so adding a filter makes sense.  It's even done for another test in that file.

khushi-411:
Thank you for the insights! I've implemented a helper function to wrap the `gradcheck` function, similar to the one implemented in PyTorch. Please let me know what do you think about it.

beverlylytle:
Hmm... there seems to be some issues with the CI... "Started 12 hours ago..." is not a usual thing to see.  And it looks like other PRs are affected.  In fact I think the vmap issue isn't related to this PR at all, and may be address in another PR.  Hopefully this can soon be sorted out, this branch can be updated and merged.

thunder/tests/test_jit_general.py
```diff
@@ -1205,6 +1206,13 @@ def forward(self, x):
     assert jitted.l1.weight.grad is not None
 
 
+@contextmanager
+def suppress_warnings():
+    with warnings.catch_warnings():
+        warnings.filterwarnings("ignore", message="Please use `torch.vmap`")
+        yield
+
+
```
beverlylytle, line 1215: I think the same thing could be accomplished more succinctly with the annotation `@pytest.mark.filterwarnings("ignore:Please use `torch.vmap`")`.

khushi-411: Yeah, makes sense. I just added these because Mike said:

> Maybe we should just wrap the gradcheck function in our own helper that suppresses this warning?

But I'll fix this with the decorator. Thank you!

beverlylytle (approved):
Looks great, especially with the additional improvement to `real`.

thunder/executors/nvfuserex_impl.py
```diff
@@ -1615,6 +1615,15 @@ def real(a: TensorProxy | Number, *, fd: FusionDefinition, lc_to_nv_map: dict) -
 register_supported(PrimIDs.REAL, real, _elementwise_unary_check)
 
 
+def imag(a: TensorProxy | Number, *, fd: FusionDefinition, lc_to_nv_map: dict) -> Any:
+    nva = getnv(a, fd, lc_to_nv_map)
+
+    return fd.ops.real(nva)
```
mruberry, line 1621: This looks like a typo. How is nvFuser passing the tests if it's taking the real part and not the imaginary part from the tensor? Is it not being tested properly?

beverlylytle: It's not passing exactly, operator consistency tests with complex input are skipped: https://github.com/Lightning-AI/lightning-thunder/blob/858a54d331159a266c586b626a3c78bc382f3ea8/thunder/tests/test_ops.py#L56

Fully enabling them would increase the CI run time substantially.  Should there be some sort of flag in OpInfo to opt in to testing with complex dtypes?

mruberry: That's right! Thank you for reminding me.

@beverlylytle, Would you file an issue to allow passing an option to run the complex tests, and a separate issue to allow a test to specify that its complex variants should not be skipped?

mruberry (approved):

PR #1914, opened 2025-03-28
Simplify quickstarts with transformers

lantiga:
## What does this PR do?

With the latest `transformers` models can be traced through without a recipe.

Also on latest generation GPUs we get a good speed up without reduce overhead, so I moved it to an extra step.

As a note, I'm getting an memory access error using CUDAGraphs on the llm quickstart with the latest nvfuser, so I'm not adding the reduce overhead plugin to that particular quickstart for now.


t-vi (approved):
Supergood!

PR #1915, opened 2025-03-28
add `ldexp` op

khushi-411:
As per the title.

~wip~


thunder/clang/__init__.py
```diff
@@ -2017,6 +2017,12 @@ def logical_xor(a: TensorLike, b: TensorLike, /) -> TensorLike:
     return a ^ b
 
 
+@clangop()
+def ldexp(a: TensorLike, b: TensorLike, /) -> TensorLike:
+    a, b = maybe_broadcast(a, b)
+    return prims.mul(a, prims.exp2(b))
```
mruberry, line 2023: Two things here:

1. I think we can just define ldexp in the torch namespace, not the core language namespace
2. Core language operators should prefer calling other core language operations when possible (so `a * exp2(b)` instead of directly calling primitives); similarly torch language operations should prefer calling torch language operations when possible, then core language operations, then primitives

thunder/tests/opinfos.py
```diff
@@ -1104,6 +1104,7 @@ def _abs_torch(x: torch.Tensor | Number):
 
 exp2_opinfo = OpInfo(
     clang.exp2,
+    supports_grad=True,
```
mruberry, line 1107: Cool!

thunder/clang/__init__.py
khushi-411: Interesting details! I'll keep this in my mind. Thank you so much for this.

thunder/tests/opinfos.py
```diff
@@ -2288,6 +2289,38 @@ def fmod_sample_input_generator(op, device, dtype, requires_grad, **kwargs):
...
 
+
+def ldexp_sample_generator(op, device, dtype, requires_grad, **kwargs):
+    make = partial(make_tensor, device=device, dtype=dtype, requires_grad=requires_grad)
+    a = make((4, 4), **kwargs)
+    b = make((4, 4), **kwargs)
+    c = make((4, 1), **kwargs)
+
+    yield SampleInput(a, b)
+    yield SampleInput(a, c)
```
khushi-411, line 2300: I had to create this function seperately instead of using `elementwise_binary_generator` because it includes tests for cases where the input tensors are in different device, and these are error cases in our case here in `ldexp`.

https://github.com/Lightning-AI/lightning-thunder/blob/2d18b7a07fb5e83d138707b7bf97bd2f82da4f2c/thunder/tests/opinfos.py#L2038-L2046

thunder/tests/opinfos.py
```diff
@@ -2288,6 +2289,38 @@ def fmod_sample_input_generator(op, device, dtype, requires_grad, **kwargs):
...
+        b = make((4, 4), device="cpu")
+
+        err_msg = "Expected all tensors to be on the same device, but found at least two devices, cuda and cpu"
+        yield (SampleInput(a, b), RuntimeError, err_msg)
+
+
+ldexp_opinfo = OpInfo(
+    ltorch.ldexp,
+    supports_grad=True,
+    dtypes=(datatypes.float32, datatypes.float64),
```
mruberry, line 2317: Can this op be defined for all floating types or all inexact types instead of just fp32 and fp64?

mruberry (approved):
Cool! I have one last question about the operator's supported dtypes, but this looks great!

thunder/tests/opinfos.py
khushi-411: Nice catch. updated the tests to `datatypes.all_dtypes`.

Thank you!

khushi-411:
Thank you for all your insightful suggestions, @mruberry! :-)

PR #1921, opened 2025-03-31
remove out arg from prim and clang operators.

khushi-411:
As per the title.

Originally discussed here: https://github.com/Lightning-AI/lightning-thunder/pull/1896#discussion_r2014469339

### DIFF
--- a/docs/source/advanced/contrib_thunder.rst
+++ b/docs/source/advanced/contrib_thunder.rst
@@ -383,7 +383,7 @@ lines
 
   @torchctx
   def _topk_prim_grad(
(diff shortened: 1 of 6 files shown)


beverlylytle (approved):
My favorite PRs are the PRs that delete code.  🔥 

That being said, while I see that Thunder didn't really support `out=`, that to do so would be akin to support in-place ops and that they seem only to [exist in PyTorch for backward compatibility reasons](https://github.com/pytorch/pytorch/issues/126016#issuecomment-2110851555), they do still exist in PyTorch.  @mruberry has a deliberate decision been made to never support them? Or is the suggestion to do this only tidying up for the moment?

mruberry (approved):

Issue #1936, opened 2025-04-03
Add decomposing executor for testing


beverlylytle:
The TorchTestExecutor is not decomposing ops.  This means that when checking the consistency of ltorch ops to pytorch for cpu input in, eg, `test_core_vs_torch_consistency` we are checking the results of an op jitted to something of the form
```
import torch
import torch.nn.functional
from thunder.executors.torchex import no_autocast

@torch.no_grad()
@no_autocast
def computation(t_0):
  # t_0: "cpu f32[4, 4]"
  t7 = torch.nn.functional.threshold(t_0, 0.0, 5.0, False)  # t7: "cpu f32[4, 4]"
    # t7 = ltorch.threshold(t_0, 0.0, 5.0, False)  # t7: "cpu f32[4, 4]"
      # t6 = ltorch.gt(t_0, 0.0)  # t6: "cpu b8[4, 4]"
        # t6 = prims.gt(t_0, 0.0)  # t6: "cpu b8[4, 4]"
      # t7 = ltorch.where(t6, 5.0, t_0)  # t7: "cpu f32[4, 4]"
        # t7 = prims.where(t6, 5.0, t_0)  # t7: "cpu f32[4, 4]"
  return (t7,)
```
rather than something that has been lowered to the subsymbols, something like (or perhaps lower)
```
import ltorch
from thunder.executors.torchex import no_autocast

@torch.no_grad()
@no_autocast
def computation(x):
  # x: "cpu f32[11]"

  t7 = ltorch.threshold(t_0, 0.0, 5.0, False)  # t7: "cpu f32[4, 4]"
    # t6 = ltorch.gt(t_0, 0.0)  # t6: "cpu b8[4, 4]"
      # t6 = prims.gt(t_0, 0.0)  # t6: "cpu b8[4, 4]"
    # t7 = ltorch.where(t6, 5.0, t_0)  # t7: "cpu f32[4, 4]"
      # t7 = prims.where(t6, 5.0, t_0)  # t7: "cpu f32[4, 4]"
  return (t7,)
```
Essentially, the majority of consistency tests are comparing pytorch results to pytorch results.  The exceptions are the tests run with the nvfuser executor.  There are no tests for the consistency of  implementations in `thunder.torch.__init__.py` on cpu input.

t-vi:
This is totally on purpose. We want things to also be consistent when they are not decomposed. And we don't want decomposition when not fusing ops.

beverlylytle:
[...]

PR #1938, opened 2025-04-03
add the threshold op

beverlylytle:
Add support for `torch.nn.functional.threshold`.

### DIFF
--- a/thunder/torch/__init__.py
+++ b/thunder/torch/__init__.py
@@ -2021,12 +2021,12 @@ def threshold(a: TensorProxy, /, threshold: float, value: float, inplace: bool =
 _inplace_to_out_of_place[threshold] = threshold, 3
 
 # alias to avoid conflict with keyword argument `threshold` in threshold_
-threshold_fn = threshold
+_threshold = threshold
 
 
 @torchsymbol(torch.nn.functional.threshold_, is_method=False, tags=(prims.OpTags.IN_PLACE,))
 def threshold_(a: TensorProxy, /, threshold: float, value: float) -> TensorLike:
-    return _copy_(a, threshold_fn(a, threshold, value, False))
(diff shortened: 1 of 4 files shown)


mruberry:
@beverlylytle Would you like to review this?

thunder/torch/__init__.py
```diff
@@ -2009,6 +2009,26 @@ def tanhshrink(a: TensorLike, /) -> TensorLike:
 
 _inplace_to_out_of_place[tanhshrink] = tanhshrink, -1
 
+
+@torchsymbol(torch.threshold, torch.nn.functional.threshold, id="torch.threshold", is_method=False)
+def threshold(a: TensorProxy, /, threshold: float, value: float, inplace: bool = False) -> TensorLike:
```
mruberry, line 2014: PyTorch reference implementation for comparison:

https://github.com/pytorch/pytorch/blob/781d28e2655f88ae2fef827ed110f22ed553a0ab/torch/_refs/nn/functional/__init__.py#L917

The PyTorch reference implementation uses a different comparison -- is the NaN propagation different between the two?

beverlylytle:
image

I'd be happy to review it.

mruberry:
> image
> 
> I'd be happy to review it.

haha, my mistake; looking at too many PRs at once

thunder/torch/__init__.py
beverlylytle: Indeed it does not propagate NaNs correctly.  Changing.

beverlylytle: I was foolishly checking this on cpu tensors and bumping into https://github.com/Lightning-AI/lightning-thunder/issues/1936.

mruberry (approved):

PR #1943, opened 2025-04-04
add rrelu support

beverlylytle:
Add support for `torch.nn.functional.rrelu`.


thunder/torch/__init__.py
```diff
@@ -1867,6 +1867,31 @@ def relu6(a: TensorProxy, /, inplace: bool = False) -> TensorLike:
 _inplace_to_out_of_place[relu6] = relu6, 1
 
 
+@torchsymbol(torch.rrelu, torch.nn.functional.rrelu, id="torch.rrelu", is_method=False)
+def rrelu(
+    a: TensorProxy, /, lower: float = 0.125, upper: float = 1.0 / 3, training: bool = False, inplace: bool = False
```
mruberry, line 1872: It's funny to see a constant like `1.0 / 3` as a default. It is completely correct to do this, and it will only be computed once when the function is defined, it's just an unusual pattern.

thunder/torch/__init__.py
```diff
@@ -1867,6 +1867,31 @@ def relu6(a: TensorProxy, /, inplace: bool = False) -> TensorLike:
...
+) -> TensorLike:
+    if training:
+        noise = uniform_like(a, minval=lower, maxval=upper, dtype=a.dtype, device=a.device)
+        out = where(a <= 0, a * noise, a)
+    else:
+        out = where(a <= 0, a * (upper + lower) / 2, a)
+    if inplace:
+        return _copy_(a, out)
```
mruberry, line 1880: Not for this PR, but it's interesting to think if the pattern for `inplace` should be to include the copy, or to call `rrelu_` here (which would then call `rrelu and `_copy_`), or to translate `inplace=True` calls to their inplace variants for consistency

Use of `inplace` is very uncommon so this isn't a pressing issue

mruberry (approved):

thunder/torch/__init__.py
beverlylytle: I have wondered this myself, but defaulted to the existing pattern.

PR #1944, opened 2025-04-04
add normal op

### LINKED ISSUE #1934: implement torch.normal
(torch.randn + scale + shift)

khushi-411:
Fixes #1934

wip

#### TODO
- [ ] fix for std < 0

### DIFF
--- a/thunder/tests/opinfos.py
+++ b/thunder/tests/opinfos.py
@@ -6442,6 +6442,51 @@ zeros_like_opinfo = OpInfo(
 tensor_creation_ops.append(zeros_like_opinfo)
 
 
+def torch_normal_and_zero(*args, **kwargs):
+    return ltorch.full_like(ltorch.normal(*args, **kwargs), 0)
+
+
+def normal_sample_generator(op, device, dtype, requires_grad=False, **kwargs):
+    make = partial(make_tensor, device=device, dtype=dtype, requires_grad=False)
+
+    cases = (
+        (4,),
+        (4, 5),
+        (6, 5, 4),
+    )
+
+    for shape in cases:
+        mean = make(shape)
+        std = torch.abs(make(shape))
+        yield SampleInput(mean, std)
+
+    # to ensure broadcasting
+    yield SampleInput(make((3, 1)), torch.abs(make((1, 3))))
+
+
+def normal_error_generator(op, device, dtype=torch.float32, requires_grad=False, **kwargs):
+    make = partial(make_tensor, device=device, dtype=dtype, requires_grad=requires_grad)
+
+    # mean = make((4, 4))
+    # std = -1
+    # err_msg = f"normal expects std >= 0.0, but found std -1"
+    # yield (SampleInput(mean, std), RuntimeError, err_msg)
+
+    err_msg = "generator is not None which"
+    yield (SampleInput(1, 2, generator=torch.Generator()), NotImplementedError, err_msg)
+
+
+normal_opinfo = OpInfo(
+    name="normal",
+    op=torch_normal_and_zero,
+    sample_input_generator=normal_sample_generator,
+    error_input_generator=normal_error_generator,
+    torch_reference=lambda *args, **kwargs: torch.normal(*args, **kwargs).fill_(0),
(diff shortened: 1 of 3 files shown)


thunder/core/proxies.py
```diff
@@ -1690,7 +1690,7 @@ def __int__(self):
         raise NotImplementedError
 
     def __bool__(self):
-        raise NotImplementedError
+        return bool(self.shape)
```
mruberry, line 1693: I don't think this is the correct conversion to bool for tensors, for example:

```
a = torch.randn((2, 2))
bool(a)
# RuntimeError: Boolean value of Tensor with more than one value is ambiguous
```

khushi-411: Ah, right, that makes sense. Thanks for pointing it out! :-) I've updated the PR to align with PyTorch's behaviour now.
```python
>>> import thunder, torch
>>> a = torch.randn((2, 2))  # 2d tensor
>>> b = torch.randn(())  # scalar tensor
>>> c = torch.tensor([])  # empty tensor
>>> def foo(a):
...     return bool(a)
... 
>>> jfoo = thunder.jit(foo)
>>> jfoo(a)
# RuntimeError: Boolean value of Tensor with more than one value is ambiguous
>>> jfoo(b)
True
>>> jfoo(c)
# RuntimeError: Boolean value of Tensor with no values is ambiguous
>>> foo(a)
# RuntimeError: Boolean value of Tensor with more than one value is ambiguous
>>> foo(b)
True
>>> foo(c)
# RuntimeError: Boolean value of Tensor with no values is ambiguous
```

beverlylytle: I think there are a few edge cases not addressed here:
```
import torch

a = torch.zeros(())
b = torch.zeros((1,))

assert bool(a) == False
assert bool(b) == False
```

thunder/torch/__init__.py
```diff
@@ -802,6 +802,26 @@ def randn_like(
     return randn(a.shape, dtype=dtype, device=device)
 
 
+@torchsymbol(is_method=False, id="torch.normal")
+def normal(
+    mean: TensorLike,
```
beverlylytle, line 807: Looking at the [PyTorch documentation](https://pytorch.org/docs/stable/generated/torch.normal.html), `mean` could also be a `Number`, and in that case the output would have the shape of `std`.  It would be good to ensure all variations of shapes of `mean` and `std` are included in the test cases.

thunder/tests/opinfos.py
```diff
@@ -6504,6 +6504,57 @@ def varargs_tensor_creation_op_sample_generator(*args, **kwargs):
...
+    # Failed: DID NOT RAISE <class 'RuntimeError'>
+    # std = -torch.abs(make((4, 4)))
+    # err_msg = "normal expects all elements of std >= 0.0"
+    # yield (SampleInput(mean, std), RuntimeError, err_msg)
+
+    err_msg = "generator is not None which"
+    yield (SampleInput(1, 2, generator=torch.Generator()), NotImplementedError, err_msg)
+
+
+normal_opinfo = OpInfo(
```
beverlylytle, line 6547: Could you please add a comment either like or referencing the comment for `randn` to explain why zeros are being compared to zeros for this test?

thunder/tests/opinfos.py
```diff
@@ -6504,6 +6504,57 @@ def varargs_tensor_creation_op_sample_generator(*args, **kwargs):
...
+
+    cases = (
+        (4,),
+        (4, 5),
+        (6, 5, 4),
+    )
+
+    for shape in cases:
+        mean = make(shape)
+        std = torch.abs(make(shape))
```
beverlylytle, line 6522: FYI `make_tensor` takes a [`low` keyword arg](https://github.com/Lightning-AI/lightning-thunder/blob/e6e5c26f36e93472aafd9224208fcc62d43e7044/thunder/tests/make_tensor.py#L5). So there could be separate `make_mean` and `make_std` functions without the use of `torch.abs`.

beverlylytle (commented):
Thank you!  This is a good start, but the devil is in the edge cases here.

PR #1949, opened 2025-04-09
Remove the `torchcompile_cat` executor from the default executor set

kevinstephano:
## What does this PR do?

This PR is removing the `torchcompile_cat` executor from Thunder's default executor list in favor of nvFuser's implementation of RoPE and the surrounding logic for RoPE.

nvFuser significantly upgraded it's RoPE and surrounding logic performance with [nvFuser PR 3848](https://github.com/NVIDIA/Fuser/pull/3848).  Regression testing on a large set of nemo based HuggingFace models showed 1-9% speedup end-2-end with many models not showing a difference.  However, most importantly, there were no slowdowns.  Therefore, this should be only a performance bonus over current expectations.

cc @mruberry @tfogal


lantiga:
Thank you @kevinstephano. A couple of clarifications:

- what is the performance impact on other uses of `cat`?
- what are the requirements on the specific way RoPE is implemented in order for it to be optimized by nvFuser?

IvanYashchuk (approved):
🚀

kevinstephano:
> what are the requirements on the specific way RoPE is implemented in order for it to be optimized by nvFuser?

There should be no specific requirements as nvFuser was augmented to generally execute all the operations between the Input Linears of Multihead Attention and the first matmul of `scaled_dot_product_attention`.

kevinstephano:
This is the breadth of models and parameters that were tested across H100 and B200 NVIDIA gpus.   The geomean was `4%` speedup.
image

kevinstephano:
>what is the performance impact on other uses of cat?
I will look into this.  I thought the primary `cat` usage was RoPE in transformer models, so I will see where else I can find it show up.  `cat` might have been used in the backward pass of multihead attention related to transformer implementations that use 3 separate linears for the Input Linears in MHA or some non-fused version of `scaled_dot_product_attention`.

lantiga (approved):

PR #1954, opened 2025-04-10
Prototype of split the thunderFX compilation and execution

kiya00:
<details>

</details>

## What does this PR do?

Thanks to @mruberry 's idea (The design doc https://docs.google.com/document/d/16BNLm95dwHBRxHMNV4YFtibS_UdXHFSIPYBfA5KajE8/edit?tab=t.0#heading=h.4gbt7pm6ctc0)

This PR prototypes split the thunderFX compilation and execution, provide extension point for customer optimize and filter function

usage:
```
import torch
from thunder.dynamo import thunder_profile, thunder_optimize
def foo(x):
    x = x.exp()
    y = torch.sin(x) + torch.cos(x)
    return y + 1

x = torch.ones(2, 2, device="cuda", requires_grad=True)
y = torch.ones(4, 4, device="cuda", requires_grad=True)
z = torch.ones(8, 4, device="cuda", requires_grad=True)
# pfoo is the profiling callable that wraps the torch.compiled function and profile stats
pfoo = thunder_profile(foo)
# runs the torch.compiled function with given input x and collect the profile stats
pfoo(x)
# dynamo generates new dynamic gm
pfoo(y)
# torch uses the dynamic graphs instead of the static ones
pfoo(x)
pfoo(z)
pfoo(x)
pfoo(y)

# pfoo._tao.id_to_profile_stats records the gm, a dict with input metadata as key and called times as value
# the stats here shows 2 gms, one static and one dynamic
# input x, y are called 2 times, z is called 1 time for the dynamic graph
# input x is called 1 time for the static graph

# For now the default filter filters out the dynamic graphs, only the static ones are optimized.
# The default optimize function first splits the gm into thunder supported/unsupported subgraphs. For Thunder-supported subgraph,  the compiler (Thunder or Inductor) is selected based on benchmark results.
optfoo = thunder_optimize(pfoo)
# fallbacks to eager for the dynamic graph
optfoo(x)

# not allowed to profile after execution, raises "No longer profiling"
pfoo(x)
```


kshitij12345:
Regarding the API (Not necessarily recommending the names but this pattern)
```python
# This returns an object which will apply the compile strategy over 
cfn = thunder.compile_with_strategy(fn, compile_strategy)

# For profiling
cfn.profile(x)

# For execution
cfn.execute(x) or cfn(x)
```

I feel this is more intuitive and easy to read. Wdyt @kiya00?

thunder/dynamo/report.py
```diff
@@ -515,13 +515,16 @@ def write_repro(
...
         format_python_file(folder / file_name)
 
-    def run_benchmark(self, compile_fn: CompileSpecificationInterface, time_fn: TimerInterface):
+    def run_benchmark(
+        self, compile_fn: CompileSpecificationInterface, time_fn: TimerInterface, reset_torch_dynamo=True
+    ):
         # From torch.compile docs - https://pytorch.org/docs/stable/generated/torch.compile.html
         # > Multiple compiled results can be associated with a frame up to torch._dynamo.config.cache_size_limit, which defaults to 8; at which point we will fall back to eager.
         # Ref: https://github.com/pytorch/pytorch/blob/34d726011f482b716d879bf665aef100a7c08a8d/torch/_dynamo/__init__.py#L97
         # > reset function clears all compile caches and restore initial state.  This function is intended
```
mruberry, line 524: This comment will have to be updated because of the new `reset_torch_dynamo` option

thunder/dynamo/report.py
```diff
@@ -515,13 +515,16 @@ def write_repro(
             print(code_str, file=f)
         format_python_file(folder / file_name)
 
-    def run_benchmark(self, compile_fn: CompileSpecificationInterface, time_fn: TimerInterface):
+    def run_benchmark(
+        self, compile_fn: CompileSpecificationInterface, time_fn: TimerInterface, reset_torch_dynamo=True
```
mruberry, line 519: Nit: `*, reset_torch_dynamo: bool = True` -- might as well make it kwarg-only for readability

kiya00:
Hi @mruberry , I think it's ready to review

thunder/dynamo/compiler.py
```diff
@@ -284,3 +289,107 @@ def last_backward_traces(self) -> list[Trace]:
 
     c = CompiledObject(backend, compiled)
     return c
+
+
+def thunder_profile(fn):
+    """
+    This function returns a profiling callable that wraps the torch compiled function and profiling
```
mruberry, line 296: Nice comment

thunder/dynamo/compiler.py
```diff
@@ -284,3 +289,107 @@ def last_backward_traces(self) -> list[Trace]:
...
+
+    For an example, see :func:`thunder.dynamo.compiler.thunder_optimize`
+    """
+    torch.compiler.reset()
+    tao: ThunderAoTOptimizer = ThunderAoTOptimizer()
+
+    def dispatching_backend(gm, *args):
+        if tao.is_profiling:
+            idx: int = id(gm)
+            # The gm needs to be recorded during Torch JIT compilation,
```
mruberry, line 308: Great comment

thunder/dynamo/compiler.py
```diff
@@ -284,3 +289,107 @@ def last_backward_traces(self) -> list[Trace]:
...
+
+
+def thunder_optimize(
+    profiling_callable,
+    *,
+    gm_filter=default_filter,
+    optimizer=default_optimizer,
+):
+    """
+    Optimizes the given profiling callable based on the statistics collected during profiling.
```
mruberry, line 352: Create comment

thunder/dynamo/compiler.py
```diff
@@ -284,3 +289,107 @@ def last_backward_traces(self) -> list[Trace]:
...
+
+    Args:
+        profiling_callable: The callable returned by :func:`thunder.dynamo.compiler.thunder_profile`
+        gm_filter: Function that selects which graph modules to optimize
+        optimizer: Function that performs the actual optimization on each graph module
+
+    Returns:
+        The torch compiled function
+
+    Example:
```
mruberry, line 362: Nice code example

thunder/dynamo/utils.py
```diff
@@ -831,3 +860,137 @@ def get_torch_compile_kwargs(**kwargs) -> dict:
     torch.compile = inspect.unwrap(torch.compile)
     torch_compile_kwarg_names = inspect.getfullargspec(torch.compile).kwonlyargs
     return {k: v for k, v in kwargs.items() if k in torch_compile_kwarg_names}
+
+
+def default_compile_strategy(gm, example_inputs) -> CompilerFn:
+    """
+    Default compile strategy for each Thunder-supported subgraph that selects the best compiler
```
mruberry, line 867: Great comment

thunder/dynamo/utils.py
```diff
@@ -831,3 +860,137 @@ def get_torch_compile_kwargs(**kwargs) -> dict:
...
+    (Thunder or Inductor) based on benchmark results. This function benchmarks both compilers
+    on the given graph module and returns the compiler function that demonstrated better performance.
+    """
+    from thunder.dynamo.report import FXGraphReport
+    from thunder.dynamo.benchmark_utils import ThunderCompileSpecification, TorchInductorSpecification, WallTime
+
+    example_input_metadata = input_to_example_input_meta(example_inputs)
+    report = FXGraphReport(gm, "gm", example_input_metadata)
+    thunderjit = ThunderCompileSpecification()
+    torchinductor = TorchInductorSpecification()
```
mruberry, line 877: What does this do if there are multiple inputs? Does it run a benchmark for each observed input?

Would it make more sense, for now, to not optimize gms that have multiple distinct inputs? This could throw a `NotImplementedError` instead

thunder/dynamo/compiler.py
```diff
@@ -284,3 +289,107 @@ def last_backward_traces(self) -> list[Trace]:
...
+    # 2) Replaces the FX graphs that are to be optimized with the optimized
+    #      callables, and other graphs with the graph modules
+    tao = profiling_callable._tao
+    dispatch_map = tao.dispatch_map
+    id_to_gm_map = tao.id_to_gm_map
+    for idx, call in dispatch_map.items():
+        if idx in indices_to_optimize:
+            gm = id_to_gm_map[idx]
+            profile_stats = tao.id_to_profile_stats[idx]
+            dispatch_map[idx] = optimizer(gm, profile_stats)
```
mruberry, line 387: If we let `optimizer` throw a `NotImplementedError` (see below), then a slight expansion of this might be:

```
try: 
  dispatch_map[idx] = optimizer(gm, profile_stats)
catch NotImplementedError:
  # Executes the gm eagerly if the optimizer can't optimize it
  dispatch_map[idx] = gm
```

thunder/dynamo/utils.py
```diff
@@ -831,3 +860,137 @@ def get_torch_compile_kwargs(**kwargs) -> dict:
...
+    m1 = report.run_benchmark(thunderjit, WallTime, reset_torch_dynamo=False)
+    m2 = report.run_benchmark(torchinductor, WallTime, reset_torch_dynamo=False)
+    if sum(m.median for m in m1 if m is not None) > sum(m.median for m in m2 if m is not None):
+        return _torch_inductor_compile
+    else:
+        from thunder import jit
+
+        return lambda *args: (jit(args[0]), "thunder")
```
mruberry, line 885: This is really interesting -- what's happening in this pattern? This probably needs a comment

thunder/dynamo/utils.py
```diff
@@ -831,3 +860,137 @@ def get_torch_compile_kwargs(**kwargs) -> dict:
...
+
+        def forward(self, *args):
+            return self.optimized_callable(*args)
+
+    return InductorModuleWrapper(torch._inductor.compile(gm, example_inputs)), "inductor"
+
+
+class ThunderAoTOptimizer:
+    """
+    Helper class that keeps track of profiling data used by the Ahead-of-Time (AoT) optimization process.
```
mruberry, line 902: Another great comment

thunder/dynamo/utils.py
```diff
@@ -831,3 +860,137 @@ def get_torch_compile_kwargs(**kwargs) -> dict:
...
+    4. Returns the optimized composite GraphModule with all partitions compiled
+
+    Args:
+        gm: The GraphModule to optimize
+        stats: ProfileStats object containing profiling information for the GraphModule
+
+    Returns:
+        The optimized GraphModule
+    """
+    from thunder import jit
```
mruberry, line 970: This is where we could add the check that the graph is static vs dynamic, I think

thunder/dynamo/utils.py
```diff
@@ -831,3 +860,137 @@ def get_torch_compile_kwargs(**kwargs) -> dict:
...
+        if not dynamic:
+            choosen.add(idx)
+    return choosen
+
+
+def default_optimizer(gm: torch.fx.GraphModule, stats: ProfileStats) -> Callable:
+    """
+    Default optimizer function that optimizes a GraphModule based on profiling statistics.
+
+    This function:
```
mruberry, line 957: Great description

thunder/dynamo/utils.py
```diff
@@ -831,3 +860,137 @@ def get_torch_compile_kwargs(**kwargs) -> dict:
...
+    1. Splits the GraphModule into Thunder-supported and unsupported partitions
+    2. Compiles unsupported partitions using PyTorch's Inductor backend
+    3. For supported partitions, selects the best compiler (Thunder or Inductor) based on benchmark results
+    4. Returns the optimized composite GraphModule with all partitions compiled
```
mruberry, line 961: Super cool

mruberry (commented):
Exciting stuff! Things overall look really good. One last tweak I mention in a few comments is shifting the dynamic input check into the optimizer and switching the filter to just use statistics about how often a graph is called to decide whether to filter it or not.

This is a change from what the technical proposal asks for, and is my mistake in the original design. But I think it makes more sense now as we work through the code together.

Great work, @kiya00!

thunder/dynamo/utils.py
kiya00: I modified to throw error for dynamic graphs. For the static graph, gm has the input data inside the node.meta (it's deepcopied from the Thunder jit compile phrase to keep the input data range), and the splitter needs that to get the subgraphs, the inferring of the input meta of the subgraph is kind of taken care by Torch, so this is how I get the example input to run the benchmark on each subgraph. For the dynamic graph I haven't thought of how to propagate the multiple input meta of the gm to the input meta of each subgraph

kiya00:
Hi @mruberry , thanks for the detailed suggestions, I fixed it: 

moves the symbolic input check to optimizer; filters out graphs that
 called less than cutoff times; add eager benchmark

mruberry (approved):
I think the next step is to look at some e2e benchmarks and see how this performs!

PR #1956, opened 2025-04-10
autodiff processor

t-vi:
Will clean up add comments etc. to get a discussion basis.

Goals
- model after transform for execution, in particular use TraceSubstitutionProcessor and to the decomposition the same way (by having a "queue" of  bound symbols to process),
- no use of "thunder.common.trace", use push scope instead,
- no conversion between functions and traces,
- (builds the joint trace instead to integrate with @beverlylytle 's work) generate the backward immediately with the augmented forward,

- [TODO] want to integrate more into the transform system,
- [TODO] propagate requires_grad in the aug forward pass,
- [TODO, probably some things vanish on their own now] clear terminology. Not calling bound symbols "symbol", no "VJPDual" when there is no duality involved,
- (seems to be better achieved by joint traces) clearer contract between forward and backward (it seems bad that one could not deduplicate saved for backwards),
- TODO/need to verify: better robustness (e.g. there is no error message if you use get_grad on a non-output currently, and that seemed hard to debug),

It still is somewhat tricky because the backward generation does not fit the tooling well, but it seems like a start for the decomposition and dropping `thunder.common.trace`.

I didn't fix rematerialize_forward_and_backward yet. My understanding is that @beverlylytle 's work will obsolete that (please correct me if I'm wrong.).

### DIFF
--- a/thunder/core/symbol.py
+++ b/thunder/core/symbol.py
@@ -96,15 +96,6 @@ def default_python_printer(
 
     s = f"{result_str}{bsym.name_with_module()}({arg_str}{', ' if (len(arg_str) > 0 and len(kwarg_str) > 0) else ''}{kwarg_str}){comment_str}"
 
-    if bsym.header:
-        header_lines = (
-            bsym.header
-            if isinstance(bsym.header, Sequence) and not isinstance(bsym.header, str)
-            else bsym.header.splitlines()
-        )
-        header_lines = (f"# {line}" for line in header_lines)
-        return chain(header_lines, [s])
-
     return s
 
 
@@ -673,6 +664,18 @@ class BoundSymbol(BoundSymbolInterface):
         lines = []
 
         s = self.sym.python_printer(self, self._out_printables, self._arg_printables, self._kwarg_printables)
+
+        if self.header:
+            if isinstance(s, str):
+                s = [s]
+
+            header_lines = (
+                self.header
+                if isinstance(self.header, Sequence) and not isinstance(self.header, str)
+                else self.header.splitlines()
+            )
+            header_lines = (f"# {line}" for line in header_lines)
+            s = chain(header_lines, s)
 
         comment = "# " if commented else ""
 
--- a/thunder/core/symbol.py
+++ b/thunder/core/symbol.py
@@ -330,9 +330,8 @@ class Symbol:
             # vjp transform (applied later).
             def tag_tensorproxy_output_as_detached(proxy):
                 if isinstance(proxy, TensorProxy):
-                    # We need to remove name from trace, otherwise replace will return a proxy with new name.
-                    trace.names.remove(proxy.name)
-                    return proxy.replace(tags=(ProxyTag.DETACHED_AUTOGRAD_GRAPH,))
+                    proxy.tags.add(ProxyTag.DETACHED_AUTOGRAD_GRAPH)
+
                 return proxy
 
             result = tree_map(tag_tensorproxy_output_as_detached, result)
--- a/thunder/tests/test_grad.py
+++ b/thunder/tests/test_grad.py
@@ -1967,3 +1967,25 @@ def test_benchmark_grad():
     backward_fn, backward_setup = backward_only(tfunc, x)
     backward_args = backward_setup()
     backward_fn(*backward_args)
+
+
+def test_disambiguate_grad_names():
+    def fn(a, grad_for_a):
+        return a * grad_for_a
+
+    jfn = thunder.jit(fn)
+
+    a = torch.randn(2, 2, requires_grad=True)
+    b = torch.randn(2, 2, requires_grad=True)
+
+    # this should work
+    res = jfn(a, b)
+    ref = fn(a, b)
+
+    go = torch.randn_like(res)
+
+    res_grads = torch.autograd.grad(res, (a, b), go)
+    ref_grads = torch.autograd.grad(ref, (a, b), go)
+
+    assert_close(res, ref)
+    assert_close(res_grads, ref_grads)
--- a/thunder/tests/test_grad.py
+++ b/thunder/tests/test_grad.py
@@ -1856,7 +1856,7 @@ def test_inconsistent_output_length_grad_transform():
 
     with pytest.raises(
         RuntimeError,
-        match="number of outputs of the original forward function must be the same as the number of primal outputs",
+        match="The number of outputs of the gradient transform function must be the same as the nunmber of outputs of the original forward function.",
     ):
         _ = jf(a)
 
@@ -1942,35 +1942,6 @@ def test_adhoc_executor_grad(executor, device, _):
 
     torch.testing.assert_close(actual, expected)
     torch.testing.assert_close(actual_gr, expected_gr)
-
-
-@pytest.mark.parametrize("device", ("cuda", "cpu"))
-def test_backward_recomputation_decomposed_ops(device):
-    if device == "cuda" and not torch.cuda.is_available():
-        pytest.skip("CUDA is not available")
-
-    def fn(a):
-        return torch.nn.functional.gelu(a)
-
-    # rematerialization will also trigger recomputation here.
-    jfn = thunder.jit(fn, executors=())
-    jfn2 = thunder.jit(fn, auto_recompute_intermediates=True)
-    a = torch.randn(2, 2, device=device, requires_grad=True)
-    res = jfn(a)
-    res2 = jfn2(a)
-    assert len(res.grad_fn.saved_tensors) == 3  # should be decomposed
-    assert len(res2.grad_fn.saved_tensors) == 1
-
-    if NVFUSER_AVAILABLE and device == "cuda":
-        # check everything is fused
-        assert {bsym.sym.name for bsym in thunder.last_backward_traces(jfn2)[-1].bound_symbols} == {
-            "nvFusion0",
-            "clear_mutable_collection",
-            "python_return",
-            "python_del",
-            "unpack_sequence",
-            "unpack_trivial",
-        }
 
 
 @requiresCUDA
--- a/thunder/tests/test_grad.py
+++ b/thunder/tests/test_grad.py
@@ -1856,7 +1856,7 @@ def test_inconsistent_output_length_grad_transform():
 
     with pytest.raises(
         RuntimeError,
-        match="The number of outputs of the gradient transform function must be the same as the nunmber of outputs of the original forward function.",
+        match="The number of outputs of the gradient transform function must be the same as the number of outputs of the original forward function.",
     ):
         _ = jf(a)
 
--- a/thunder/transforms/autodiff.py
+++ b/thunder/transforms/autodiff.py
@@ -1,0 +1,408 @@
+import thunder.core.transforms
+from thunder.core.transforms import ForwardBackwardTraces
+
+from thunder.core import prims, utils
+from thunder.core.transforms import (
+    is_constant_for_vjp,
+    _get_gradfn_and_executor,
+    augmented_forward_impls,
+    backward_impls,
+    recompute_saved_for_backward,
+)
+from thunder.core.proxies import ProxyTag
+from thunder.core.symbol import BoundSymbol
+from thunder.core.vjp_utils import make_aug_forward_and_backward
+from thunder.core.pytree import tree_map
+import thunder
+import time
+
+
+def _should_recompute_bsym_in_backward(bsym):
+    return any((ProxyTag.RECOMPUTE_IN_BACKWARD in o.tags) for o in bsym.flat_proxy_outs)
+
+
+# Transforms a trace by determining which grad transforms to call given the list of executors in priority order
+# This pass tries to preserve the original trace and proxies.
+def grad_transform_on_trace(trace, /, *args, **kwargs):
+    # This processes the bsyms to map symbols to operator executors:
+    # - in the order of the executor list
+    #   - if the executor defines a grad transform, call that to
+    #     get fw + bw,
+    # - if there is a augmented forward registered, use that and the backwared registered,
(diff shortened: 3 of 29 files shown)


thunder/transforms/autodiff.py
```diff
@@ -0,0 +1,281 @@
...
+from thunder.core.utils import check
+from thunder.core.vjp_utils import make_aug_forward_and_backward
+from thunder.core.pytree import tree_map
+import thunder
+import time
+
+
+# Transforms a trace by determining which grad transforms to call given the list of executors in priority order
+# This pass tries to preserve the original trace and proxies.
+def grad_transform_on_trace(trace, /, *args, **kwargs):
```
IvanYashchuk, line 20: That's cool that it works, but these more than 100 lines of code can be condensed into <10 lines of code. In the current prototype, every operation in the input trace is reevaluated/reinterpreted, redirecting all calls to the augmented forward and backward functions. Here's the condensed version of the same logic:
```py
from thunder.prims import get_grad
from thunder.core.transforms import put_grads, augmented_forward_pass, backward_pass

def grad_transform_on_trace(trace, /, *args, **kwargs):
    def joint_forward_and_backward(*args, **kwargs):
        result, all_intermediates = augmented_forward_pass(*args, trace=trace, **kwargs)
        output_grad = get_grad(result)
        backward_result = backward_pass(all_intermediates, trace, cotangents)
        return {"output": result, "grad_flat_args": backward_result}
    return trace(use_dce=False)(joint_forward_and_backward, *args, **kwargs)
```

t-vi:
Aside form not computing the gradient of process groups, there is an error in distributed stemming from an incomplete substitution of proxies in the subsymbols following synchronize. This is an example from an input trace in test_fsdp (`t_net2_weight` should have been replaced with `t22` also in the `prims.linear`):
```
  # /home/zeus/miniconda3/envs/cloudspace/lib/python3.12/site-packages/torch/nn/modules/linear.py:125:          return F.linear(input, self.weight, self.bias)
  t18 = ltorch.linear(input, t22, t21)  # t18: "cuda:1 f32[2, 2]"
    # t18 = prims.linear(input, t_net2_weight, t_net2_bias)  # t18: "cuda:1 f32[2, 2]"
```

I'll look at fixing the FSDP transform.

thunder/transforms/autodiff.py
t-vi: Yeah, but how many LOC is augmented forward pass and backward_pass? How many LOC is trace?
All these have to go, and this is showing them out.

t-vi: Not to forget that there is an entire caching mechanism we try to drop here.

thunder/transforms/autodiff.py
```diff
@@ -0,0 +1,301 @@
...
+
+        backward_part_bsyms.insert(0, bsym.from_bsym())
+
+    # collect needed computation
+    saved_for_backward = {}
+    for bsym in backward_part_bsyms:
+        saved_for_backward.update(
+            (a.name, a)
+            for a in bsym.flat_proxy_args
+            if a.name in forward_proxy_names and a.name not in backward_part_bsyms
```
beverlylytle, line 242: `a.name not in backward_part_bsyms` is always true. Did you mean something else here?

t-vi: Indeed. Thank you for catching.

thunder/transforms/autodiff.py
```diff
@@ -0,0 +1,311 @@
...
+                aug_fwd_impl = augmented_forward_impls.get(bsym.sym.id)
+                if aug_fwd_impl is not None:
+                    bwd_impl = backward_impls.get(bsym.sym.id)
+
+                    def joint_forward_backward(*args, **kwargs):
+                        aug_fwd_res = aug_fwd_impl(*bsym.args, **bsym.kwargs)
+                        if isinstance(aug_fwd_res, thunder.core.transforms.VJPDual):
+                            res, saved_for_backward = aug_fwd_res
+                        else:
+                            res, saved_for_backward = aug_fwd_res
```
beverlylytle, line 151: This looks a little funny.

t-vi: I need to figure out what the return convention for aug forward actually is.

Also for backward:

https://github.com/Lightning-AI/lightning-thunder/blob/150d7e0bebda9c2f27bddcf7e21206d0c7d37ee9/thunder/core/transforms.py#L2274-L2280

why is it not `return None, g_weight`

And not to mention that we also have this:

https://github.com/Lightning-AI/lightning-thunder/blob/150d7e0bebda9c2f27bddcf7e21206d0c7d37ee9/thunder/core/transforms.py#L1360-L1377

which looks fishy to me (shouldn't the `put_grad` be on `weight` and `a` is the index?).

t-vi:
One failure is from the assumption that augmented forward and backward implementations would have flat inputs and outputs, an assumption that seems to be violated (only) by the automatically registered `multi_dot`.
I'd be inclined to not fix this but to either move auto registration to the grad_transform scheme or manually provide multi_dot.
I would likely be interesting to actually write a decomposition for multi_dot, as it reduces to a sequence of matmuls.

t-vi:
Mostly for my own reference, these seem to be the remaining CPU test failures (there are more on GPU, in particular memory and distributed). 
- [ ]  `FAILED thunder/tests/test_grad.py::test_torch_checkpoint - assert 1 == 2`
- [ ] `FAILED thunder/tests/test_grad.py::test_grad_split_unused_output[cpu] - RuntimeError: grad of non-output and non-intermediate t4 requested`
- [ ] `FAILED thunder/tests/test_grad.py::test_backward_recomputation_decomposed_ops[cpu] - assert 3 == 1`
- [ ] `FAILED thunder/tests/test_grad.py::test_inconsistent_output_length_grad_transform - AssertionError: argument layout mismatch: [<TensorProxy(name="t0", dtype=thunder.dtypes.float32, shape=(2, 2))>, <AnyProxy '<object object at 0x...`
- [ ] `FAILED thunder/tests/test_grad.py::test_too_few_results_from_backward - AssertionError: Regex pattern did not match.`

riccardofelluga:
All pretty interesting stuff! As far as architecture goes, is the idea of trying to keep one single trace for as long as possible inside thunder and then split it only at the very end to pipe it in the torch autograd infrastructure? 

Maybe not in this PR but just asking to get this out to understand the direction. I think it would be pretty beneficial to do all the compiler passes with one trace, as it allows to reason with a wider scope(and later transformations can take advantage of this one single trace)

t-vi:
> is the idea of trying to keep one single trace for as long as possible inside thunder and then split it only at the very end to pipe it in the torch autograd infrastructure

So (quite) a few of us @beverlylytle @IvanYashchuk @lantiga chatted about this.

There are main two parts we want to evolve the autodiff:
- this PR produces a joint trace (and changes the way we produce it), but for the moment splits it very soon (as before),
- @beverlylytle is working on keeping a single trace until after transform for execution.

t-vi:
I'll split out some of the stand-alone features.

t-vi:
I am factoring out fixes, so I guess this will need to merge after them (and after merging main).

thunder/core/symbol.py
```diff
@@ -321,9 +321,8 @@ def __call__(self, *args, **kwargs):
             # vjp transform (applied later).
             def tag_tensorproxy_output_as_detached(proxy):
                 if isinstance(proxy, TensorProxy):
-                    # We need to remove name from trace, otherwise replace will return a proxy with new name.
-                    trace.names.remove(proxy.name)
-                    return proxy.replace(tags=(ProxyTag.DETACHED_AUTOGRAD_GRAPH,))
+                    proxy.tags.add(ProxyTag.DETACHED_AUTOGRAD_GRAPH)
```
t-vi, line 324: This is a bit of a funny change, I am seeing the traces for autograd functions having two versions of the proxy, but have not managed to isolate the problem in the old autodiff setting. Apparently it re-executes things more so it does not hit the mismatch in any of the checked traces.

I am relatively confident though, that the trick here is problematic in that it introduces a collision - two distinct proxies with the same name.

thunder/tests/test_grad.py
```diff
@@ -1944,35 +1944,6 @@ def func(x):
     torch.testing.assert_close(actual_gr, expected_gr)
 
 
-@pytest.mark.parametrize("device", ("cuda", "cpu"))
-def test_backward_recomputation_decomposed_ops(device):
```
t-vi, line 1948: This test is for an old experimental feature `auto_recompute_intermediates` that has not caught on, so I'm removing the test here.

t-vi (commented):
There are two details here in the surroundings of the autodiff itself.
The other changes outside the thunder.transforms.autodiff are benign updates of tests.

thunder/transforms/autodiff.py
t-vi: fixed now.

lantiga (approved):
Epic 🚀, just a few nits here and there

thunder/transforms/autodiff.py
```diff
@@ -0,0 +1,477 @@
...
+                                    new_grad = self.add_bsyms_from_function(thunder.torch.add, current_grad, new_grad)
+
+                                grad_proxy_map[p.name] = new_grad
+                                self.write(new_grad, new_grad)
+                            elif nbsym.sym == prims.get_grad:
+                                # get grad reads a gradient, this could be
+                                # - the gradient of an intermediate (possibly combined with an output) form put_grad above
+                                #   note that all put_grads are happen before all get_grads from the dataflow of the forward,
+                                #   in this case we use the gradient from the put_grad(s)
+                                # - an output of the forward function (that has not been combined with an intermediate in pu grd),
```
beverlylytle, line 97: "put grad"?

thunder/transforms/autodiff.py
```diff
@@ -0,0 +1,477 @@
...
+                                    self.add_processed_bsyms([new_bsym])
+                                    grad_proxy_map[name] = new_bsym.output
+                                    self.write(nbsym.output, new_bsym.output)
+                            else:
+                                # all other symbols are just computation
+                                self.add_processed_bsyms([nbsym.from_bsym()])
+                self.collected_bw_part_bsyms.clear()
+
+                # we collect the gradients fo the flat args in the output dictionary's 'grad_flat_args'
+                # this means that the joint trace as them as outputs, which is good.
```
beverlylytle, line 133:
```suggestion
                # this means that the joint trace has them as outputs, which is good.
```

thunder/transforms/autodiff.py
```diff
@@ -0,0 +1,477 @@
...
+    # we also have the inputs available, so we add flat_args.
+    # for inplace, we need to update this (or have flat args be the right thing?...)
+    forward_proxy_names.update(a.name for a in return_bsym.args[0]["flat_args"] if isinstance(a, thunder.Proxy))
+
+    # We keep track of the names of proxies we recompute in the backward as those will not need to be part of the
+    # ones saved in the forward for the backward
+    backward_recomputed_proxy_names = set()
+
+    # here we go through the bound symbols to assign to the three categories, keeping track of the forward names, the backward needed
+    # and the backward available names.
```
beverlylytle, line 350: I think this comment is stale, as there is now only `backward_recomputed_proxy_names`.

thunder/transforms/autodiff.py
```diff
@@ -0,0 +1,477 @@
...
+    saved_for_backward_tensors = [p for p in saved_for_backward.values() if isinstance(p, thunder.TensorProxy)]
+    saved_for_backward_other = [p for p in saved_for_backward.values() if not isinstance(p, thunder.TensorProxy)]
+
+    # we build the forward trace
+    forward_trace = thunder.core.trace.from_trace(joint_trace)
+    forward_trace.tags.add(thunder.core.trace.TraceTag.AUGMENTED_FORWARD)
+    forward_trace.names = forward_trace.names.copy()  ## ehem
+    forward_trace.bound_symbols += forward_part_bsyms
+
+    # now we create the return value and return bound symbol for  the forward
```
beverlylytle, line 422:
```suggestion
    # now we create the return value and return bound symbol for the forward
```

beverlylytle (approved):
Looks good!  Only some nits.

thunder/transforms/autodiff.py
```diff
@@ -0,0 +1,477 @@
...
+                        ((gg.name in grad_proxy_map) or (gg.name in output_proxy_names))
+                        for gg in (bs.args[0] for bs in bw_bsyms if bs.sym == prims.get_grad)
+                    ):
+                        for nbsym in bw_bsyms:
+                            if nbsym.sym == prims.put_grad:
+                                # put_grad adds something to the gradients
+                                # - if it is an output of the function, we also need to get the
+                                #   "grad_output" (but only once)
+                                # - if we already had a gradient, we need to add the new component
+                                # - structurally (for correctness), there should be no put_grad
```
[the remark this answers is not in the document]
t-vi: yeah, let's do it.

thunder/transforms/autodiff.py
```diff
@@ -0,0 +1,477 @@
...
+                                #   after a get_grad
+                                p = self.env.get(nbsym.args[0].name, nbsym.args[0])
+                                current_grad = grad_proxy_map.get(p.name)
+                                new_grad = nbsym.args[1]
+                                if current_grad is None and p.name in output_proxy_names:
+                                    self.new_trace.push_scope([])
```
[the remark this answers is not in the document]
t-vi: Interesting. So 
```python
with self.new_trace.capture_in_scope() as new_bsyms:
   current_grad = prims.get_grad(p)
self.add_processed_bsyms(new_bsyms)
```
instead of
```python
self.new_trace.push_scope([])
current_grad = prims.get_grad(p)
self.add_processed_bsyms(self.new_trace.pop_scope())
```
?
It emphasizes the block nature.

PR #1957, opened 2025-04-11
ThunderFX: passes all kwargs except torch.compile's kwargs directly to thunder.jit

### LINKED ISSUE #1937: [thunderfx] thunderfx doesn't work with compile options arguments
```python
import thunder
import torch

def fn(x):
    return x + 1

print(thunder.jit(fn, fusion_type="consecutive")(torch.ones(2,)))

# RuntimeError: There are kwargs that are not supported by either thunder.jit or torch.compile: {'fusion_type': 'consecutive'}
thunderfx(fn, fusion_type="consecutive")(torch.ones(2,))

```

This fails as the valid kwargs are deduced from the thunder.jit signature which doesn't consider the kwargs which are passed as part of compile_options.

```python
>>> import inspect
>>> from thunder import jit
>>> thunder_jit_kwarg_names = inspect.getfullargspec(jit).kwonlyargs
>>> print(thunder_jit_kwarg_names)
['langctx', 'executors', 'sharp_edges', 'cache', 'disable_torch_autograd', 'transforms', 'debug_options']
```

https://github.com/Lightning-AI/lightning-thunder/blob/31594068e7df8ae9cbd1dbe27b17cbf76aea0400/thunder/dynamo/utils.py#L820-L824

Potential Solution:

https://github.com/Lightning-AI/lightning-thunder/blob/31594068e7df8ae9cbd1dbe27b17cbf76aea0400/thunder/dynamo/compiler.py#L220-L225

`rest_kwargs` are passed to `thunder.jit` assuming they are from compile options rather than erroring.

kiya00:
<details>

</details>

## What does this PR do?

Fixes #1937.

For a user-friendly experience, thunderfx accepts `kwargs` that may include the kwargs from both `torch.compile` and `thunder.jit`. Since `thunder.jit` uses `**compile_options`, the current check is too tight. This PR updates to separate the kwargs from `torch.compile` options and passes all remaining kwargs directly to `thunder.jit`

### DIFF
--- a/thunder/dynamo/compiler.py
+++ b/thunder/dynamo/compiler.py
@@ -212,23 +212,10 @@ def thunderfx(fn: Callable, /, **kwargs) -> Callable:
     """
     import thunder
 
-    from thunder.dynamo.utils import get_thunder_jit_kwargs, get_torch_compile_kwargs
+    from thunder.dynamo.utils import get_torch_compile_kwargs
 
-    thunder_jit_kwargs = get_thunder_jit_kwargs(**kwargs)
     torch_compile_kwargs = get_torch_compile_kwargs(**kwargs)
-
-    rest_kwargs = {k: v for k, v in kwargs.items() if k not in thunder_jit_kwargs and k not in torch_compile_kwargs}
-    check(
-        not rest_kwargs,
-        lambda: f"There are kwargs that are not supported by either thunder.jit or torch.compile: {rest_kwargs}",
-    )
-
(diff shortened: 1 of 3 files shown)


kshitij12345 (commented):
The fix looks good to me, thanks. It would be nice to add a test.

Probably something like this
```python
def fn(x, w)
    return x @ w

jfn = thunderfx(fn, nv_enable_matmul=True)
jfn(x, w)

# In forward trace verify that we see a nvFusion region with matmul.
```

mruberry (approved):

PR #1958, opened 2025-04-11
Add instance norm

beverlylytle:
Add support for `torch.nn.functional.instance_norm'.

Aside from `torch.nn.functional.gumbel_softmax` which may soon be removed from pytorch, I believe merging this will mean all nonlinear activation functions of `torch.nn.functional` are supported.

Implementation based on https://github.com/pytorch/pytorch/blob/604467de208646f0c3b2663e45f2ff6a655a6716/aten/src/ATen/native/Normalization.cpp#L714


thunder/tests/opinfos.py
```diff
@@ -8325,13 +8325,14 @@ def batch_norm_sample_generator(op, device, dtype, requires_grad, **kwargs):
...
     # TODO: implement running_mean and running_var
     cases = (
-        ((3, 4), {"momentum": 0.2, "eps": 0.5}),
         ((3, 3, 3), {"momentum": 0.2}),
         ((3, 3, 3), {"momentum": -1.2}),
         ((3, 3, 5, 6), {"momentum": 0.0}),
         ((3, 2, 3, 4), {"momentum": -1.0, "eps": 0.5}),
         ((3, 2, 3, 4, 12), {"momentum": -1.0, "eps": 0.5}),
     )
+    if op.name != "instance_norm":
```
mruberry, line 8334: OK... I guess ideally we'd have a common base and then the batch norm sample generator would yield all the common samples and add this sample. This is fine for now

thunder/torch/__init__.py
```diff
@@ -1058,7 +1058,7 @@ def permute(a: TensorLike, /, *dims: int) -> TensorLike:
 def repeat(a: TensorLike, /, *repeats: int) -> TensorLike:
     repeats = utils.extract_shape_from_varargs(repeats)
     utils.check_valid_shape(repeats)
-    utils.check(a.ndim <= len(repeats), f"Expected {a.ndim=} <= {len(repeats)=}")
+    utils.check(a.ndim <= len(repeats), lambda: f"Expected {a.ndim=} <= {len(repeats)=}")
```
mruberry, line 1061: Nice catch!

thunder/torch/__init__.py
```diff
@@ -4123,6 +4123,37 @@ def batch_norm(
     return result
 
 
+@torchsymbol(torch.nn.functional.instance_norm)
+def instance_norm(
+    a: TensorLike,
```
mruberry, line 4128: Nit: `a: TensorLike, /,`

mruberry:
fyi @t-vi , @Borda, looks like the failing CI job is due to a code coverage issue. I'm not sure we use code coverage for anything -- maybe we should remove it so it's no longer a point of failure?

Also interesting to note that we have 15 tests xpassing now. We may want to follow-up on those sometime

mruberry (approved):

thunder/tests/opinfos.py
beverlylytle: Noted.

PR #1959, opened 2025-04-11
Apply update_aliases to viewed input

beverlylytle:
The existent `update_aliases` approach to in-place ops did not consider whether input to a jitted function contained tensors which are views of one another.  This PR attempts to address that reusing mechanisms for detecting and grouping such input that were developed for in-place functionalization.


thunder/core/update_aliases.py
```diff
@@ -42,11 +51,15 @@ def _involves_viewed_args(bsym, viewed):
...
 
-def insert_alias_updates(computation_trace: Trace) -> Trace:
+def insert_alias_updates(computation_trace: Trace, alias_tensor_indices: list[list[int]]) -> Trace:
     swap_map = dict()
     bsyms = []
 
-    # First pass: identify views, their originals, and operands involved in inplace ops
+    # First pass: identify inputs which are views of each other and swap them out with a default,
+    # reshaping if necessary.
+    computation_trace, _ = replace_args_with_alias_map(computation_trace, alias_tensor_indices)
```
beverlylytle, line 60: Now the first step of update_aliases is the same as with in-place functionalization.

thunder/__init__.py
```diff
@@ -474,7 +474,12 @@ def apply_transforms_and_build_cache_entry(cd, cs, cache_info, prologue_trc, com
             computation_traces.append(computation_trc)
 
             if not compile_options.get("skip_inplace_alias_updates", True):
-                computation_traces.append(insert_alias_updates(computation_trc))
+                alias_tensor_indices = []
+                if alias_tensor_indices_str := cache_info["alias_tensor_indices"]:
+                    alias_tensor_indices: list[list[int]] = [
+                        [int(i) for i in s.split(",")] for s in alias_tensor_indices_str.split("-")
+                    ]
```
crcrpar, line 481: Can we prepare `alias_tensor_indices_str` when inplace_functionalization or inplace_alias_updates are enabled?

This is a bit orthogonal to this pull request, but I suppose we should exclude `skip_inplace_functionalization=False` and `skip_inplace_alias_updates=False`.
So in summary I think what I find nice is something like

```python
            utils.check(
                skip_inplace_functionalization or skip_inplace_alias_updates,
                lambda: "We cannot use both `alias_updates` and `functionalization` for in-place ops",
            )
            alias_tensor_indices: list[list[int]] = ...
```

before the two paths of in-place handling

thunder/core/update_aliases.py
```diff
@@ -1,16 +1,25 @@
 from functools import reduce, partial
 
+import thunder
```
crcrpar, line 3: nit-picking as I couldn't see the use of this import.
```suggestion
```

thunder/tests/test_update_aliases.py
```diff
@@ -3,8 +3,10 @@
 import pytest
 import torch.testing
 
+import thunder
```
crcrpar, line 6: ditto, it doesn't seem that we use this import
```suggestion
```

thunder/core/update_aliases.py
```diff
@@ -42,11 +51,15 @@ def _involves_viewed_args(bsym, viewed):
     return any(isinstance(p, TensorProxy) and variableify(p) in viewed for p in bsym.flat_proxy_args)
 
 
-def insert_alias_updates(computation_trace: Trace) -> Trace:
+def insert_alias_updates(computation_trace: Trace, alias_tensor_indices: list[list[int]]) -> Trace:
```
crcrpar, line 54: Could you give a docstring to this function and then add this function to docs by adding `update_aliases.rst` to https://github.com/Lightning-AI/lightning-thunder/tree/4acf8198cfed1761528f9d10d2bc1a5356cf9ce9/docs/source/reference/core, mayb e as a follow-up? 
It'd be really nice if that docstring mentions the difference between this and functionalization.

thunder/__init__.py
beverlylytle: My understanding is that the two strategies are to be blended together eventually.  And since the alias update strategy is still experimental, I think I don't want to put in such a check just yet.

thunder/tests/test_update_aliases.py
beverlylytle: This import is used in `test_inplace_to_alias_func_args`.

thunder/core/update_aliases.py
beverlylytle: This is a good idea.  One that deserves its own PR.

crcrpar (approved):

IvanYashchuk (approved):
@mruberry, could you please help merge this PR?

mruberry (approved):

PR #1969, opened 2025-04-16
Fix TE not computing weight gradient with intermediate inputs (revert #1838)

### LINKED ISSUE #1966: TransformerEngine : May fail to compute weight gradients if weight parameter has some operation before Linear
Repro:
```python
import torch
import thunder
from thunder.executors.transformer_engineex import transformer_engine_ex

def fn(x, w):
        return torch.nn.functional.linear(x, w)

with torch.device("cuda"):
    x = torch.randn(32, 32, requires_grad=True)
    w = torch.randn(32, 32, requires_grad=True)

    # tfn = thunder.jit(fn)
    tfn = thunder.jit(fn, executors=(transformer_engine_ex,))

    o = tfn(x, w)
    o.sum().backward()

    print(w.grad)  # None with TE executor (otherwise there is some value).
    thunder.last_traces(tfn)[0].save_trace("init_trc.py")
    thunder.last_traces(tfn)[-1].save_trace("exec_trc.py")

```

Initial Trace
```python
import thunder
import torch

@torch.no_grad()
@no_autocast
def computation(x, w):
  # x: "cuda:0 f32[32, 32]"
  # w: "cuda:0 f32[32, 32]"
  ltorch.autocast_enter('cuda', torch.bfloat16, True, True)

  # /home/kkalambarkar/lightning-thunder/scratchpad/test.py:317: 	        return torch.nn.functional.linear(x, w)
  t4 = prims.convert_element_type(x, dtypes.bfloat16)  # t4: "cuda:0 bf16[32, 32]"
  t5 = prims.convert_element_type(w, dtypes.bfloat16)  # t5: "cuda:0 bf16[32, 32]"
  t6 = prims.linear(t4, t5, None)  # t6: "cuda:0 bf16[32, 32]"
  ltorch.autocast_exit()
  return (t6,)
```

Problem:
With https://github.com/Lightning-AI/lightning-thunder/pull/1838, we check the `requires_grad` on weight TensorProxy to determine if `TransformerEngine` should compute gradients for weight. But this fails when weight is not directly the input parameter but output of some operation (in above eg, output of conversion to bfloat16). As thunder does not properly propagate `requires_grad` (https://github.com/Lightning-AI/lightning-thunder/issues/1768), the TensorProxy after the conversion has `requires_grad=False`.

NOTE: Directly using the `requires_grad` from the Runtime Tensor also won't work as the conversion happens under `no_grad` context, hence output will not have `requires_grad=True`.

```python
import torch

def fn(x, w):
    with torch.no_grad():
        x_bf16 = x.to(torch.bfloat16)
        print(x_bf16.requires_grad)  # False
        return x_bf16 @ x_bf16

x = torch.randn(16, 16, requires_grad=False, device="cuda")   
w = torch.randn(16, 16, requires_grad=True, device="cuda")

fn(x, w)
```

Thoughts on Solution:
If instead of a single function which computes gradients for `input`, `weight` and `bias` (with FP8 GEMM), we could have a separate function (and symbol) to compute input, weight and bias gradients (need to check feasibility), then it should most likely work fine thanks to thunder's autograd routing the gradients correctly (and discarding computation when not required).

cc: @IvanYashchuk 

Ref: NVBUG 5219259

kshitij12345:
Fixes #1966 

Changes 
* Reverts #1838 
* Adds a new test for verifying fix based on #1966 

Tested with existing and new TE test on RTX 6000, H100 and B200.


thunder/distributed/prims.py
```diff
@@ -184,14 +184,7 @@ def synchronize_meta(
...
             # See [FSDP Sharding] in distributed/__init__.py
             unsharded_shape = a.shape[0] * group.size(), *a.shape[1:]
-            # NOTE: We explicitly pass the `a.requires_grad` as we want to propogate the requires grad correctly on `nn.Parameter`'s
-            # as it maybe queried later.
-            return TensorProxy(
-                shape=unsharded_shape,
-                like=a,
-                distparallel_type=DistParallelType.REPLICATED,
-                requires_grad=a.requires_grad,
-            )
```
kshitij12345, line 194: Should this change be preserved?

IvanYashchuk: `requires_grad` should be propagated only in the meta function of the grad transform and tested. It's incorrect to rely on this attribute for intermediates because there's no guarantee it's used consistently. 

I agree with you here that there's no need to revert this change. We only need to revert the changes that caused performance regression.

t-vi (approved):
@kshitij12345 @IvanYashchuk

PR #1970, opened 2025-04-16
Treat scalar tensors as weak dtypes

riccardofelluga:
When using `torch.maximum` and `torch.minimum` to compare tensors to scalar tensors(zero dims) the output needs to be downcasted to the tensor with ndims != 0 to match torch behaviour. Note thunder still does computation in fp32.

To test this the following snippet:

```python
import torch
import thunder

a = torch.randn(1, 10, dtype=torch.bfloat16, device="cpu")
b = torch.tensor(0.544235342652, dtype=torch.float32, device="cpu")

o = torch.maximum(b, a)
print(o.dtype)

def foo(x, y):
    return torch.maximum(x, y)

jfoo = thunder.jit(foo)
o = jfoo(a, b)
print(o.dtype)

```  

gives the output:
```shell

torch.bfloat16
torch.bfloat16

import torch

@torch.no_grad()
@no_autocast
def computation(x, y):
  # x: "cpu bf16[1, 10]"
  # y: "cpu f32[]"

  # /lightning-thunder/test_maximum.py:13: 	    return torch.maximum(x, y)
  t8 = torch.maximum(x, y)  # t8: "cpu bf16[1, 10]"
    # t8 = ltorch.maximum(x, y)  # t8: "cpu bf16[1, 10]"
      # t6 = prims.convert_element_type(x, dtypes.float32)  # t6: "cpu f32[1, 10]"
      # t7 = prims.maximum(t6, y)  # t7: "cpu f32[1, 10]"
      # t8 = prims.convert_element_type(t7, dtypes.bfloat16)  # t8: "cpu bf16[1, 10]"
  return (t8,)
```

instead of `t8 = torch.maximum(x, y)  # t8: "cpu f32[1, 10]"`


thunder/torch/__init__.py
```diff
@@ -2292,12 +2292,26 @@ def lt_(a, b, /):
 
 @torchsymbol(torch.maximum, is_method=True)
 def maximum(a: TensorProxy, b: TensorProxy) -> TensorProxy:
-    return clang.maximum(a, b)
+    result = clang.maximum(a, b)
+
+    if a.ndim == 0 ^ b.ndim == 0:
+        return result
+
+    result_dtype = a.dtype if b.ndim == 0 else b.dtype
```
mruberry, line 2300: This isn't correct; let me elaborate in a bit

riccardofelluga:
@mruberry hmm that is a good observation! Indeed the same behavior happens for `add`.

Hmm I see the idea of the weak vs true dtypes, but how/were is the conversion carried out? 

I am thinking that this change involves `elementwise_type_promotion` too, because so far it is no checking for weak/true dtype
https://github.com/Lightning-AI/lightning-thunder/blob/2cc90ec34c34681d46c86af65842e33b1b356fd8/thunder/core/utils.py#L421

Am I correct assuming that true dtypes take precedence in ops over weak dtypes? Still we want to keep the computation in fp32 in this case right?

mruberry:
> @mruberry hmm that is a good observation! Indeed the same behavior happens for `add`.
> 
> Hmm I see the idea of the weak vs true dtypes, but how/were is the conversion carried out?
> 
> I am thinking that this change involves `elementwise_type_promotion` too, because so far it is no checking for weak/true dtype
> 
> https://github.com/Lightning-AI/lightning-thunder/blob/2cc90ec34c34681d46c86af65842e33b1b356fd8/thunder/core/utils.py#L421
> 
> Am I correct assuming that true dtypes take precedence in ops over weak dtypes? Still we want to keep the computation in fp32 in this case right?

Let's just start like I suggested. If you take this line:

[2cc90ec/thunder/core/proxies.py#L1286C5-L1286C25](https://github.com/Lightning-AI/lightning-thunder/blob/2cc90ec34c34681d46c86af65842e33b1b356fd8/thunder/core/proxies.py#L1286C5-L1286C25)

```
_true_dtype = _dtype
```

Just try changing it to

```
_true_dtype = _dtype if len(shape) > 0 else dtypes.to_weak_dtype(_dtype)
```

Does that fix the issue?

riccardofelluga:
Hmm also let me add a test for this

thunder/tests/opinfos.py
```diff
@@ -2162,6 +2164,20 @@ def elementwise_binary_generator(op, device, dtype, requires_grad, *, no_rhs_num
...
+        weak_dtype_table = {
+            datatypes.signedinteger: torch.int64,
+            datatypes.unsignedinteger: torch.uint8,
+            datatypes.floating: torch.float32,
+            datatypes.complexfloating: torch.complex64,
+            datatypes.bool_: torch.bool,
+        }
+
+        e = make((4, 4), **kwargs)
+        f = make((), **kwargs, dtype=weak_dtype_table[type(tdtype)])
```
mruberry, line 2178: I think this can use this function:

```
https://github.com/Lightning-AI/lightning-thunder/blob/150d7e0bebda9c2f27bddcf7e21206d0c7d37ee9/thunder/core/dtypes.py#L598
```

mruberry:
> Does this sound right?

Yep! That's correct. 

It looks like this change has affected the numerical consistency of several CPU tests. What's happening with those? If we are numerically inconsistent but more accurate, we should consider modifying the test to detect better numerical accuracy (by using something like `assert_closer`:

https://github.com/Lightning-AI/lightning-thunder/blob/150d7e0bebda9c2f27bddcf7e21206d0c7d37ee9/thunder/tests/framework.py#L106

)

riccardofelluga:
I don't know if this is the best way to approach this, in my mind tho we are lacking a test to check consistency of dtypes between torch and core, I'll open an issue to create that set of tests in another PR

thunder/tests/test_ops.py
```diff
@@ -35,14 +35,32 @@ def test_errors(op, device, dtype, executor, comp):
...
     if isinstance(thunder_result, Exception):
         raise thunder_result
 
-    comp(thunder_result, torch_result)
+    # Try checking strictly, if that does not work, check against reference.
+    try:
+        comp(thunder_result, torch_result)
+    except AssertionError:
+
+        def upcast_tensors(x):
```
mruberry, line 52: Nice use of `assert_closer`.

fyi @beverlylytle

mruberry (approved):

PR #1979, opened 2025-04-22
fix copy_with_setitem_grad to handle broadcasting correctly

### LINKED ISSUE #1962: Tensor shape mismatch when using set item and slices
## 🐛 Bug

The following repro leads to `RuntimeError: Shapes were expected to be the same, but got shapes (1, 2) and (2, 2)!`

```python
import torch
import thunder

key_states = torch.randn(2, 4, device="cuda", dtype=torch.bfloat16, requires_grad=False)
k_embed = torch.randn(1, 2, device="cuda", dtype=torch.bfloat16, requires_grad=True)

def foo(key_states, k_embed):
    key_states[:, 2:] = k_embed
    return key_states

jfoo = thunder.jit(foo)

out = jfoo(key_states, k_embed)
```

kshitij12345:
Fixes #1962 

Grad rule didn't handle broadcasting from unit-dims correctly, this PR fixes it.

### DIFF
--- a/thunder/core/transforms.py
+++ b/thunder/core/transforms.py
@@ -1425,9 +1425,9 @@ def _copy_with_setitem_grad(a: TensorProxy, index, value: Number | TensorProxy):
 
     if isinstance(value, TensorProxy):
         value_grad = g[index]
-        expanded_dims = value_grad.ndim - value.ndim
-        if expanded_dims > 0:
-            value_grad = prims.sum(value_grad, tuple(range(expanded_dims)))
+        # NOTE: `value` could be broadcasted.
+        if not utils.same_shape(value_grad.shape, value.shape):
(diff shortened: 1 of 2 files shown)


thunder/core/transforms.py
```diff
@@ -1425,9 +1425,9 @@ def _copy_with_setitem_grad(a: TensorProxy, index, value: Number | TensorProxy):
 
     if isinstance(value, TensorProxy):
         value_grad = g[index]
-        expanded_dims = value_grad.ndim - value.ndim
-        if expanded_dims > 0:
-            value_grad = prims.sum(value_grad, tuple(range(expanded_dims)))
+        # NOTE: `value` could be broadcasted.
```
mruberry, line 1428: nit: "broadcast"

mruberry (approved):

Issue #1985, opened 2025-04-25
Better error message for not implemented inplace


t-vi:
Famously, inplace operations on subtensors are not supported:

```python
import torch
import thunder

def fn(x):
    x[0].zero_()

jfn = thunder.jit(fn)
x = torch.ones(2, 2)
jfn(x)
```

The current error message is
```
NotImplementedError: in-place op of `torch.zero_` to `<TensorProxy(name="t1", dtype=thunder.dtypes.float32, shape=(2,))>`, a view tensor of `<TensorProxy(name="x", dtype=thunder.dtypes.float32, shape=(2, 2))>` is not supported because 2 != 4
```

while `NotImplementedError` gives a clue that it's a limitation of Thunder and the error message is technically correct and interesting to the thunder developer, not only does it nothing to help the user unblock themselves but can be confusing.

cc @borda @mruberry

PR #1988, opened 2025-04-25
[in-place] Recommend alias-updates when functionalization failed

crcrpar:
## What does this PR do?

Fixes nothing, but this could give one easy-to-try next step.
Related: #1985


t-vi (changes requested):
No, we wantf sane defaults.

PR #1989, opened 2025-04-25
`CompiledObject` uses the argument of `be` instead of the visible `backend`

crcrpar:
## What does this PR do?

As per title, changes `CompiledObject.__init__` to use the argument of `be`.


t-vi (approved):
Thank you @crcrpar

PR #2012, opened 2025-04-30
Better error message for not implemented inplace

### LINKED ISSUE #1985: Better error message for not implemented inplace

lianakoleva:
Fixes #1985. New error message:
```
NotImplementedError: Thunder does not yet support in-place operations of Tensors that are views of a larger Tensor (such as slices). Please modify usage of `torch.zero_` (/Users/liana/thunder/test.py:6).
```

### DIFF
--- a/thunder/core/functionalization.py
+++ b/thunder/core/functionalization.py
@@ -100,8 +100,9 @@ def check_inplace_to_views(computation_trace: Trace) -> dict[VariableInterface, 
         check(
             orig_tensor.numel == in_tensor.numel,
             lambda: (
-                f"in-place op of `{bsym.sym.id}` to `{in_tensor}`, a view tensor of "
-                f"`{orig_tensor}` is not supported because {in_tensor.numel} != {orig_tensor.numel}"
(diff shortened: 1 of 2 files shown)


k223kim (approved):

t-vi (approved):
Thank you @lianakoleva @k223kim

PR #2013, opened 2025-04-30
Forward to original fn when `thunder.examine.examine` called on jitted function

### LINKED ISSUE #1986: thunder.examine.examine should refuse (or properly divert) to work on jit-compiled functions
`thunder.examine.examine` currently needs to be given an unmodified function. If you pass in a non-working jitted function (which you are likely to have when using examine), it claims that the "unmodified function" is not working, which is misleading.

Example:

```python
import torch
import thunder
import thunder.examine

def fn(x):
    x[0].zero_()

jfn = thunder.jit(fn)
x = torch.ones(2, 2)
thunder.examine.examine(jfn, x)
```

We could either grab the original function from the compile data or error out.

cc @borda @mruberry

lianakoleva:
Fixes #1986, adds basic tests for `thunder.examine.examine`.


t-vi (approved):
Thank you @lianakoleva

PR #2021, opened 2025-05-01
nicer printing of Enum objects

t-vi:
While debugging composable FSDP + DDP, I wanted to see the DistParallelType enum value.

Trace refencing an Enum A's VALUE before this PR:
```
def fn():
  return _A_0
```

Trace after this PR:
```
def fn():
  return _A_VALUE_0
```

### DIFF
--- a/thunder/core/trace.py
+++ b/thunder/core/trace.py
@@ -4,6 +4,7 @@ from contextvars import ContextVar
 from contextlib import contextmanager
 import pathlib
 from typing import Optional, Any, Tuple, Type, Dict, List, Union
+from enum import Enum
 from collections.abc import Callable
 from collections.abc import Sequence, Hashable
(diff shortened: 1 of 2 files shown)


thunder/core/trace.py
```diff
@@ -188,6 +189,9 @@ def _make_name(self, *, prefix: str | None = None, is_object_name: bool = False,
                 self.obj_name_ctr += 1
                 if obj is None:
                     name = f"_object_{name}"
+                elif isinstance(obj, Enum):
+                    # even though it should be unique, we need to do the counter suffix for technical problems
```
mruberry, line 193: What technical problems? Should this link an issue?

t-vi: Yeah, once I get a traceback.

PR #1826, opened 2025-03-03
[TE] get the number of saved tensor for meta function with an actual execution

kshitij12345:
TODO

Add description

Verify on TE v2.0, v2.1 and main.

PR #2030, opened 2025-05-02
Fix TransformerEngine executor for all recipes

ksivaman:
<details>

</details>

## What does this PR do?

#1826 introduces a method to dynamically figure out the number of tensors internally saved for the backward pass by running a dummy module during runtime. However, this number is dependent on the FP8 recipe. This PR:

- Adds the relevant recipes to test.
- Executes the dummy module with the exact recipe being run.

### DIFF
--- a/thunder/executors/transformer_engineex.py
+++ b/thunder/executors/transformer_engineex.py
@@ -1,4 +1,4 @@
-from functools import partial, lru_cache
+from functools import partial
 from itertools import chain
 from typing import Any
 from collections.abc import Sequence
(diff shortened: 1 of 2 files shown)


thunder/executors/transformer_engineex.py
```diff
@@ -333,7 +332,10 @@ def _te_functional_linear_meta(
 
             # It's not critical to model the exact shape and dtype of
             # saved_tensors since they are not used in Thunder's meta functions.
-            saved_tensors = tuple(TensorProxy(like=a, shape=a.shape) for _ in range(_get_num_saved_tensors()))
+            saved_tensors = tuple(
+                TensorProxy(like=a, shape=a.shape)
+                for _ in range(_get_num_saved_tensors(get_recipe_from_options_or_default_recipe()))
```
lantiga, line 337: this is more of a design discussion (so, not blocking this PR), I'm wondering if we should make executor-specific options more discoverable for developers by adding them as arguments to the executor itself

In general, transforms do expose arguments, and executors can (and should imo) do it as well.

We can keep the compile options mechanism just like it is today, and have compile options override transform and executor arguments if specified.

lantiga: Certainly, sounds good @IvanYashchuk

lantiga (approved):
Looks great, thank you @ksivaman

PR #2043, opened 2025-05-06
Add custom decompositions for cross entropy loss for the nvfuser executor

protonu:
## What does this PR do?
This PR adds custom decompositions for Cross-Entropy Loss for the nvFuser executor. 
Adding these custom decompositions improves performance and allows further optimization in nvFuser.

For cross-entropy loss forward:

1. We move the take_along_axis computation before the log softmax is computed. This allows us to reduce memory traffic for the inputs.

For cross-entropy loss backward:

1. We replace a scatter-op with a iota and where op as we don't have support for scatter exposed in nvFuser.
2.  We can get rid of a reduction that shows up when backward is computed as nll_loss backward followed by log softmax backward. 

Fixes # (issue).

## PR review

## Did you have fun?


thunder/executors/nvfuserex_impl.py
```diff
@@ -2732,6 +2738,272 @@ def embedding(
...
+    label_smoothing: float = 0.0,
+) -> tuple[TensorLike, TensorLike, TensorLike, TensorLike]:
+    losses: TensorLike
+    if reduction == "mean":
+        losses = TensorLike(like=a, shape=())
+    else:
+        check(
+            False,
+            lambda: f"cross entropy expected reduction to be  'mean' but was given {reduction}",
+        )
```
jjsjann123, line 2798: nitpick
```suggestion
    check(
            reduction == "mean",
            lambda: f"cross entropy expected reduction to be  'mean' but was given {reduction}",
        )
    losses = TensorLike(like=a, shape=())
```

thunder/executors/nvfuserex_impl.py
```diff
@@ -2732,6 +2738,272 @@ def embedding(
...
+
+    max_log_sum_exp: TensorLike = TensorLike(like=target, shape=target.shape, dtype=dtypes.float32)
```
jjsjann123, line 2800:
```suggestion
    max_log_sum_exp: TensorLike = TensorLike(like=target, dtype=dtypes.float32)
```

thunder/executors/nvfuserex_impl.py
```diff
@@ -2732,6 +2738,272 @@ def embedding(
...
+    num_valid_indices: TensorLike = TensorLike(like=losses, shape=(), dtype=dtypes.float32)
+    a_max: TensorLike = TensorLike(like=target, shape=target.shape, dtype=dtypes.float32)
```
jjsjann123, line 2802:
```suggestion
    a_max: TensorLike = TensorLike(like=target, dtype=dtypes.float32)
```

thunder/executors/nvfuserex_impl.py
```diff
@@ -2732,6 +2738,272 @@ def embedding(
...
+    # compute the mean
+    sum_3 = fd.ops.sum(where_1)
+    div = fd.ops.div(sum_3, sum_2_cvt)
+    return div, max, log, sum_2_cvt
+
+
+nv_cross_entropy_fwd = ex.register_operator(
+    "nv_cross_entropy_fwd",
+    meta=cross_entropy_fwd_meta,
+    fn=cross_entropy_fwd,
```
jjsjann123, line 2863: QQ: are we supposed to register the same function twice?

`fn` is supposed to actually run the operator. cross_entropy_fwd doesn't feel right to me here.
tagging @IvanYashchuk, @mruberry for sanity check.

I see we have the same for nv_sdpa, that also felt wrong.

thunder/executors/nvfuserex_impl.py
```diff
@@ -2732,6 +2738,272 @@ def embedding(
...
+    tags=[prims.OpTags.RANDOM_OP],
```
jjsjann123, line 2864: We don't have random op inside cross entropy, so the tag isn't right here.

thunder/executors/nvfuserex_impl.py
```diff
@@ -2732,6 +2738,272 @@ def embedding(
...
+    label_smoothing: float = 0.0,
+) -> Any:
+    return TensorProxy(like=a)
+
+
+def cross_entropy_bwd(
+    g: TensorLike,
+    a: TensorLike,
+    /,
+    *,
```
jjsjann123, line 2888: nitpick: is it necessary to make the follow up position only?
I roughly understand we would want to do that for `fd` and `lc_to_nv_map`

thunder/executors/nvfuserex_impl.py
```diff
@@ -2732,6 +2738,272 @@ def embedding(
...
+    tags=[prims.OpTags.RANDOM_OP],
+)
+register_supported(nv_cross_entropy_fwd.id, cross_entropy_fwd, None)
+
+
+def cross_entropy_bwd_meta(
+    g: TensorLike,
+    a: TensorLike,
+    /,
+    *,
```
jjsjann123, line 2873: ditto

thunder/executors/nvfuserex_impl.py
```diff
@@ -2732,6 +2738,272 @@ def embedding(
...
+    softmax_mul_grad_sum = fd.ops.mul(recomputed_softmax, new_target_bcast)
+    difference = fd.ops.sub(scattered_vals, softmax_mul_grad_sum)
+
+    return difference
+
+
+nv_cross_entropy_bwd = ex.register_operator(
+    "nv_cross_entropy_bwd",
+    meta=cross_entropy_bwd_meta,
+    fn=cross_entropy_bwd,
```
jjsjann123, line 2942: same concern with using `cross_entropy_bwd` for `fn`

thunder/executors/nvfuserex_impl.py
```diff
@@ -2732,6 +2738,272 @@ def embedding(
...
+    /,
+    target: TensorLike,
+    weight: None | TensorLike = None,
+    size_average: None | Any = None,
+    ignore_index: int = -100,
+    reduce: None | Any = None,
+    reduction: str = "mean",
+    label_smoothing: float = 0.0,
+) -> Any:
+    result, _, _, _ = nv_cross_entropy_fwd(
```
jjsjann123, line 2960: Just for my own curiosity (trying my luck with @IvanYashchuk @mruberry) 

When is `execution_transform` used? I assume `grad_transform` would kick-in before that.
But if grad transform already decomposed it to `nv_cross_entropy_fwd`, we don't need an `execution_transform`?

I guess inference would be a case where we wouldn't run grad transform, so it's necessary to have this as well. But I wouldn't expect inference workflow running a loss function.

thunder/executors/nvfuserex_impl.py
```diff
@@ -2732,6 +2738,272 @@ def embedding(
...
+    nv_a = getnv(a, fd, lc_to_nv_map)
+    nv_target = getnv(target, fd, lc_to_nv_map)
+    nv_ignore_index = getnv(ignore_index, fd, lc_to_nv_map)
+
+    zero_scalar = fd.define_scalar(0, dtype=torch_dtype_to_nvfuser_dtype(dtypes.to_torch_dtype(a.dtype)))
+
+    # modify the labels to account for ignore index and then do a
+    # gather/ take_along_axis on the input tensor
+    ne = fd.ops.ne(nv_target, nv_ignore_index)
+    where_0 = fd.ops.where(ne, nv_target, zero_scalar)
```
jjsjann123, line 2829: This feels a bit strange to me. For entries hitting ignore_index, we are clamping them to 0 on the index?

Shouldn't that require padding on the entry of `nv_a` in the first place?

jjsjann123 (commented):
want to double check the forward implementation. kinda looks a bit strange to me.

thunder/executors/nvfuserex_impl.py
jjsjann123: Just to clarify my question,

I think the proper behavior is to clamp the value after take_along_axis to 0, instead of clamping the target to zero. This felt like wrong math to me.

jjsjann123: For context: https://github.com/Lightning-AI/lightning-thunder/blob/4f35fcd2f674071d89a7145312dc5d672e616838/thunder/torch/__init__.py#L5562-L5588

thunder/executors/nvfuserex_impl.py
mruberry: Registering something with the same symbol/id (in this case the symbol/id is `'nv_cross_entropy_fwd'`) will, I believe, clobber any previous registration. See:

https://github.com/Lightning-AI/lightning-thunder/blob/4f35fcd2f674071d89a7145312dc5d672e616838/thunder/extend/__init__.py#L226

I don't see that being registered twice but I could have missed something

thunder/executors/nvfuserex_impl.py
mruberry: I think this could change the name `g` to something like `grad` and not require it be positional-only. Positional-only arguments are for arguments like `a` that don't have reasonable names

thunder/executors/nvfuserex_impl.py
mruberry: The grad transform does happen first and it does decompose to custom operators, like `nv_cross_entropy_fwd`, but those operators aren't assigned to any executors. Then during the execution transform nvFuser will be the only executor that can execute `nv_cross_entropy_fwd` and will be assigned it. 

In general a grad transform could create operations grabbed by a mix of executors, and even an executor's custom grad could create operations that should be executed by a mix of executors.

thunder/executors/nvfuserex_impl.py
```diff
@@ -2732,6 +2738,272 @@ def embedding(
...
+    sub = fd.ops.sub(nv_a, max2)
+    exp = fd.ops.exp(sub)
+    sum_1 = fd.ops.sum(exp, 1)
+    log = fd.ops.log(sum_1)
+
+    # compute S = POST_GATHER(A) - max(A, dim=1) - log(sum(exp(A - max(A, dim=1), dim=1))
+    gather_sub = fd.ops.sub(gather, max)
+    log_softmax_post_gather = fd.ops.sub(gather_sub, log)
+
+    # set the values for ignore index to 0
```
protonu, line 2846: @jjsjann123 this is where we zero out the values that were read for the ignore_index entries.

thunder/executors/nvfuserex_impl.py
jjsjann123: sorry, I meant that we have registered `cross_entropy_fwd` function twice.

Once here with the `ex.register_operator(..., fn=cross_entropy_fwd, ...)`
And then later, we override the translation map with
`register_supported(nv_cross_entropy_fwd.id, cross_entropy_fwd, None)`

The function `cross_entropy_fwd` is a translator function, I don't think it should appear here with `fn`.

thunder/executors/nvfuserex_impl.py
jjsjann123: ah, I was totally blind. :see_no_evil: 

Thanks a lot for kindly pointing it out. :saluting_face:

thunder/executors/nvfuserex_impl.py
jjsjann123: scratch this. I just missed the set-to-zero part later.

thunder/executors/nvfuserex_impl.py
jjsjann123: Yep.

thunder/tests/test_nvfuser.py
```diff
@@ -1071,6 +1071,52 @@ def sdpa_fn(q, k, v, dropout_p, is_causal, scale):
...
+    fwd_trace = thunder.last_traces(compiled_func)[-1]
+    bwd_trace = thunder.last_backward_traces(compiled_func)[-1]
+    fwd_fusion = examine.get_fusions(fwd_trace)
+    bwd_fusion = examine.get_fusions(bwd_trace)
+
+    assert len(fwd_fusion) == 1
+    assert len(bwd_fusion) == 1
+
+    assert "nv_cross_entropy_fwd" in fwd_fusion[-1][-1].name
+    assert "nv_cross_entropy_bwd" in bwd_fusion[-1][-1].name
```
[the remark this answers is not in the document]
protonu: It's something like:
nvFusion0: (nv_cross_entropy_fwd) - not sure if I can check for equality without pattern matching.

thunder/tests/test_nvfuser.py
```diff
@@ -1071,6 +1071,52 @@ def sdpa_fn(q, k, v, dropout_p, is_causal, scale):
...
+
+    ref_inputs = [inp.clone().detach() for inp in inputs]
+    # logits needs to be requires_grad=True for backward
+    ref_inputs[0].requires_grad = True
+
+    ref_loss_out = cross_entropy_fn(*ref_inputs, ignore_index=ignore_index)
+    ref_loss_out.backward()
+
+    torch.testing.assert_close(loss_out, ref_loss_out)
+    torch.testing.assert_close(logits.grad, ref_inputs[0].grad)
```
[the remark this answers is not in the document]
protonu: thanks for pointing that out - done

thunder/tests/test_nvfuser.py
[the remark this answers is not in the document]
protonu: Please let me know if you think of a better method.

thunder/executors/nvfuserex_impl.py
```diff
@@ -2732,6 +2738,272 @@ def embedding(
...
+        return False
+
+    if ignore_index >= 0:
+        return False
+
+    if (weight, size_average, reduce) != (None, None, None):
+        return False
+
+    if label_smoothing != 0.0:
+        return False
```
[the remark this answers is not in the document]
protonu: I wasn't really considering it, but I can definitely look at it if there's a use case.

thunder/executors/nvfuserex_impl.py
protonu: removed

thunder/executors/nvfuserex_impl.py
protonu: @jjsjann123 are you suggesting I change `g` to something like `grad` and not have it position only?
I am not sure about giving `a` a particular name.

thunder/executors/nvfuserex_impl.py
protonu: removed

thunder/executors/nvfuserex_impl.py
IvanYashchuk: I agree it feels wrong. `ex.register_operator` should be enough, and `register_supported` shouldn't be necessary. If it's really needed now, we need to fix it.

thunder/executors/nvfuserex_impl.py
IvanYashchuk: In the no_grad context, the grad transform would not be used, and the execution transform would be used.

thunder/tests/test_nvfuser.py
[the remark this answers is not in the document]
jjsjann123: If there's remaining thunder issue not related to your custom backend, we don't have to patch that in this PR.

I'm wondering if we can have a filter here to skip cases that the registered decomposition doesn't have a support for.

thunder/tests/test_nvfuser.py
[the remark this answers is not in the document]
crcrpar: I see. then I'd not change the current assert's

crcrpar (approved):
just wondering if you have some numbers of cross entropy loss to compare this with the existing ones e.g. https://github.com/Lightning-AI/lightning-thunder/blob/c6928015914fdbdd708fd8e87fbd9d9c1b4a40ef/thunder/executors/triton_crossentropy.py?

protonu:
>just wondering if you have some numbers of cross entropy loss to compare this with the existing ones e.g. https://github.com/Lightning-AI/lightning-thunder/blob/c6928015914fdbdd708fd8e87fbd9d9c1b4a40ef/thunder/executors/triton_crossentropy.py?

@crcrpar I did compare performance against torchcompile (which uses Triton - but is that the same as the link you sent?)
For performance I used these changes in this PR and a pre-segmentation pass in nvfuser (https://github.com/NVIDIA/Fuser/pull/4399). 

This is a benchmark I was going to add:
https://github.com/NVIDIA/Fuser/pull/4472

mruberry:
@IvanYashchuk , @jjsjann123 would you like to review again?

mruberry:
@beverlylytle Would you like to take a look, too?

thunder/executors/nvfuserex_impl.py
```diff
@@ -2732,6 +2737,268 @@ def embedding(
...
+    neg = fd.ops.neg(log_softmax_post_gather)
+    where_1 = fd.ops.where(ne, neg, zero_scalar)
+
+    # sum_2_cvt computes the number of valid indices
+    sum_2 = fd.ops.sum(ne)
+    sum_2_cvt = fd.ops.cast(sum_2, dtype=DataType.Float)
+
+    # compute the mean
+    sum_3 = fd.ops.sum(where_1)
+    div = fd.ops.div(sum_3, sum_2_cvt)
```
beverlylytle, line 2857: Couldn't sum_2_cvt be zero if target is filled with -1s and ignore_index is -1? In pytorch this case returns NaN.

thunder/executors/nvfuserex_impl.py
```diff
@@ -2732,6 +2737,268 @@ def embedding(
...
+    nv_max_log_sum_exp = getnv(max_log_sum_exp, fd, lc_to_nv_map)
+
+    zero = fd.define_scalar(0, dtype=DataType.Int)
+    one = fd.define_scalar(1, dtype=DataType.Int)
+
+    # scatter the gradients (negative) - this is backward of nll los
+    iotas = fd.ops.iota(nv_a.shape()[-1], zero, one, dtype=DataType.Int)
+    iotas_bcast = fd.ops.broadcast_in_dim(iotas, shape=nv_a.shape(), broadcast_dims=[nv_a.ndim - 1])
+    neg_gradients = fd.ops.neg(nv_g)
+    neg_gradients_mod = fd.ops.div(neg_gradients, nv_valid_indices)
```
beverlylytle, line 2910: Again, could nv_valid_indices be zero?

beverlylytle (approved):
👍

thunder/executors/nvfuserex_impl.py
protonu: That's a great point - let me see if I can add this

protonu: I think for `fwd` it should still work.
`div = sum_3 / sum_2_cvt` 

If all entries in the target/labels are ignore index valued then both `sum_3` and `sum_2_cvt` will be zero and we'll return Nan. Please let me know if this is wrong.

thunder/executors/nvfuserex_impl.py
protonu: When there all entries in labels/targets is ignore index, this will still match PyTorch (tested out both fwd and bwd).
This is because `scattered_vals` become 0 and so does `softmax_mul_grad_sum`.

PR #2045, opened 2025-05-07
drop falcon 40b / 7b

t-vi:
seems to be broken with new PyTorch, but likely not worth fixing in itself.


t-vi:
Merging to fix CI.

Issue #2009, opened 2025-04-30
Fix inconsistency in prologue for test_jit_general.py::test_litgpt_variants_kvcache and enable check_trace in test


t-vi:
Currently, the test does not pass the consistency checks, with an error from the prologue, apparently very early on:
```
FAILED thunder/tests/test_jit_general.py::test_litgpt_variants_kvcache[cpu-llama1-like] - AssertionError: proxy name collision p310 in t_mask_cache: "cpu b8[1, 1, 3, 3]" = p310.mask_cache
```
(by replacing `thunder.jit` with `thunder_jit` in the test).
This indicates that there are two distinct proxies called p310.

PR #2048, opened 2025-05-07
handle double proxification of module tensor attribute

ali-alshaar7:
Fixes: #2009

In the test, we have the issue that the `mask_cache` attribute  is wrapped twice by `wrap_attribute` and `wrap_binary_subscr` since its a module attribute directly and not a buffer as well as a tensor which needs to be proxified, it then gets two prologue proxies associated with it, and thus two constraints and this created a double unpack which threw the new `check_trace` requirement off.

```
 ProvenanceRecord(
  i1 = INPUT_FN()
  i2 = LOAD_ATTR(i1, '__dict__')
  i3 = BINARY_SUBSCR(i2, 'mask_cache')
)
```
```
ProvenanceRecord(
  i1 = INPUT_FN()
  i2 = LOAD_ATTR(i1, 'mask_cache')
) 
```

### DIFF
--- a/thunder/core/jit_ext.py
+++ b/thunder/core/jit_ext.py
@@ -210,6 +210,21 @@ class JitCtx:
         return self._computation_trace
 
     def add_constraint(self, constraint):
+        # Avoid adding duplicate constraints
+        for i, c in enumerate(self._constraints):
+            if len(c) != len(constraint):
+                continue
+            if c[0] != constraint[0]:
+                continue
+            for i in range(1, len(c)):
+                if isinstance(c[i], ProxyInterface) and isinstance(constraint[i], ProxyInterface):
+                    if variableify(c[i]) != variableify(constraint[i]):
+                        break
+                elif c[i] != constraint[i]:
+                    break
+            else:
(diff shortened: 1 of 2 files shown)


ali-alshaar7:
I think I would rather fix the interpreter issue causing `mask_cache` unpacking to be interpreted twice than do what's up rn.

t-vi (approved):
Thank you @ali-alshaar7

thunder/core/jit_ext.py
```diff
@@ -210,6 +210,21 @@ def computation_trace(self) -> TraceCtx:
...
+            if c[0] != constraint[0]:
+                continue
+            for i in range(1, len(c)):
+                if isinstance(c[i], ProxyInterface) and isinstance(constraint[i], ProxyInterface):
+                    if variableify(c[i]) != variableify(constraint[i]):
+                        break
+                elif c[i] != constraint[i]:
+                    break
+            else:
+                print(constraint)
```
ali-alshaar7, line 226:
```suggestion
```

PR #2055, opened 2025-05-09
avoid calling transform when checker explicitly rejects symbol

jjsjann123:
The existing logic isn't right.

When calling `can_fuse`, we could have a scenario where all subsymbol is executable.
For which case, if the registered op has an explicit checker to reject it, we should avoid kicking in the transformation, which might rely on assumption checked in the checker.


mruberry (approved):
Did this some up with an actual example?

jjsjann123:
Yeah, it pops up with @protonu 's #2043 
Where the decomposition only supports `reduction==mean` for now, but the transformation also picks up cases other cases, because all subsymbols are fusible.

PR #2057, opened 2025-05-09
removing nv_enable_embedding

jjsjann123:
nvfuser embedding fwd support has been improved to match / outperform other backends. Hence switch to enable it by default.


jjsjann123:
oops, those failures looks like real.

t-vi (approved):
Thanks @jjsjann123 ! 
I'd probably drop the option fully.

PR #2061, opened 2025-05-13
try reenabling skipped test

t-vi:


IvanYashchuk:
Thank you for creating this pull request! If the apex executor is causing a numerical mismatch with PyTorch, it should be removed from the default executors list. Skipping it here just for this test would be hiding the problem, and we would need to spend time debugging it again in the future.

t-vi:
OK, I'm trying to drop apex now.

lantiga (approved):

Issue #2025, opened 2025-05-02
[profile-guided optimization] The input tensor values are lost for the FX subgraphs


kiya00:
## 🐛 Bug

In profile-guided optimization, we collect all FX graphs and benchmark each one to choose the fastest backend. This process requires actually running each graph. Currently, input tensors are randomly generated based on metadata extracted from placeholder `Node` in the `GraphModule`, which contains min/max if possible.

The input to each GraphModule is typically available via either `node.meta["example_value"]` or `node.meta["grapharg"].example`:

`example_value` holds a FakeTensor used for shape/type tracing.

`grapharg.example` is a weak reference to the original input tensor, which contains value range info (e.g. min/max).

In the specific case below, we only have access to the example_value (the FakeTensor), so the generated random tensor lacks value range constraints — which can lead to CUDA errors during execution.

### To Reproduce

```py
from transformers import AutoConfig
import torch
from torch.testing import make_tensor
from functools import partial

config = AutoConfig.from_pretrained("microsoft/Phi-3.5-mini-instruct")
batch_size = 1
seq_length = 256
shape = (batch_size, seq_length)

from transformers import AutoModelForCausalLM

with torch.device("cuda"):
    model = AutoModelForCausalLM.from_config(config, torch_dtype=torch.bfloat16)
    model.requires_grad_(True)
    make = partial(
        make_tensor,
        low=0,
        high=config.vocab_size,
        device="cuda",
        dtype=torch.int64,
        requires_grad=False,
    )
    a = make(shape)
    inputs = {"input_ids": a, "labels": a}

from thunder.dynamo.compiler import thunder_profile, thunder_optimize
pmodel = thunder_profile(model)
pmodel(**inputs)
pmodel(**inputs)
optmodel = thunder_optimize(pmodel)
optmodel(**inputs)
```
it failed at `submod_2`
```py
> /wayan/lightning-thunder/thunder/dynamo/utils.py(888)get_timing()
-> print(f"Error running benchmark {compile_fn.name}: {e}")
(Pdb) l
883         def get_timing(compile_fn, timer_fn):
884             try:
885                 measurement = report.run_benchmark(compile_fn, timer_fn, reset_torch_dynamo=False)
886             except Exception as e:
887                 import pdb; pdb.set_trace()
888  ->             print(f"Error running benchmark {compile_fn.name}: {e}")
889                 return float("inf")
890             return sum(m.median for m in measurement if m is not None)
```
```
GraphModule(
[...]

PR #2062, opened 2025-05-13
Makes the default optimizer on dynamo-segmented graph and handles OOM

### LINKED ISSUE #2026: [profile-guided optimization] The input tensors for the subgraph are newly allocated each time, which leads to OOM, whereas ThunderFX works.

## 🐛 Bug

In profile-guided optimization, we collect subgraphs and run each one on different backends to select the fastest. However, the inputs to each subgraph are newly allocated on every run, which leads to additional memory usage. As a result, profile-guided optimization may run out of memory (OOM), while thunderfx handles the same workload without issue. Currently, the OOM occurs during benchmarking and is caught via exception handling, after which the default Thunder backend is selected.

A potential improvement is to distinguish whether a subgraph input corresponds to the model input or an intermediate tensor. We can then restrict new allocations to only those inputs that are intermediate tensors, reducing unnecessary memory overhead.

### To Reproduce

```py
import torch


from litgpt import Config, GPT
from functools import partial

cfg = Config.from_name("Llama-3-8B")

cfg.n_layer = 10
cfg.block_size = 1024
batch_dim = 4

torch.set_default_dtype(torch.bfloat16)
make = partial(make_tensor, low=0, high=255, device="cuda", dtype=torch.int64)

with torch.device("cuda"):
    model = GPT(cfg)
    shape = (batch_dim, cfg.block_size)
    x = make(shape)

# works fine
# thunderfx_model = thunderfx(model)
# thunderfx_model(x)

from thunder.dynamo.compiler import thunder_profile, thunder_optimize

pmodel = thunder_profile(model)

pmodel(x)
pmodel(x)
# OOM when running the benchmark
optmodel = thunder_optimize(pmodel)
optmodel(x)

```
RTX 6000 50G

kiya00:
<details>

</details>

## What does this PR do?

Fixes #2026  #2025.

## PR review

this PR:
1. Rewrites the default optimizer function, which currently benchmarks each Thunder-split graph and selects the fastest backend for each, to instead benchmark and optimize each Dynamo-segmented graph.

2. Handles OOM issues:

- During benchmarking, uses weakref in `grapharg` to determine whether a graph's input tensor is from the original model input or an intermediate tensor, and only allocates new memory for intermediate tensors.

- Instead of using `retain_graph=True` to benchmark the backward pass separately, introduces a flag to enable benchmarking the forward and backward passes together.

### DIFF
--- a/thunder/dynamo/benchmark_utils.py
+++ b/thunder/dynamo/benchmark_utils.py
@@ -73,6 +73,19 @@ class ThunderCompileSpecification(CompileSpecificationInterface):
         return ["import thunder"]
 
 
+class ThunderCompilerOnGraphModuleSpecification(CompileSpecificationInterface):
+    def __init__(self, specification_name="thunderfx", **kwargs):
+        self.name = specification_name
+        self.thunder_options: dict = kwargs
+
+    def compile(self, gm, **kwargs):
+        from thunder.dynamo import ThunderCompiler
+
+        thunder_compiler = ThunderCompiler(**kwargs)
+        split_gm = thunder_compiler(gm, sample_args=None)
+        return split_gm
+
+
 class TorchCompileSpecification(CompileSpecificationInterface):
     """
     A compile specification for :func:`torch.compile`.
--- a/thunder/dynamo/benchmark_utils.py
+++ b/thunder/dynamo/benchmark_utils.py
@@ -131,22 +131,24 @@ class TorchInductorSpecification(CompileSpecificationInterface):
     https://github.com/Lightning-AI/lightning-thunder/issues/1521
     """
 
-    def __init__(self, specification_name="inductor_backend"):
+    def __init__(self, specification_name="inductor_backend", *, skip_symbolic_trace=False):
         self.name: str = specification_name
+        self.skip_symbolic_trace = skip_symbolic_trace
 
     @staticmethod
-    def torch_inductor(fn, inputs):
+    def torch_inductor(fn, inputs, *, skip_symbolic_trace=False):
         from torch._inductor import compile as inductor_compile
         from torch.fx import symbolic_trace
 
-        fx_graph = symbolic_trace(fn)
-        return inductor_compile(fx_graph, inputs)
+        if not skip_symbolic_trace:
+            fn = symbolic_trace(fn)
+        return inductor_compile(fn, inputs)
 
     def compile(self, fn, *, inputs, **kwargs):
-        return self.torch_inductor(fn, inputs)
+        return self.torch_inductor(fn, inputs, skip_symbolic_trace=self.skip_symbolic_trace)
 
     def to_source(self, fn_name):
-        return f"TorchInductorSpecification.torch_inductor({fn_name}, inputs)"
+        return f"TorchInductorSpecification.torch_inductor({fn_name}, inputs, skip_symbolic_trace={self.skip_symbolic_trace})"
 
     def import_str(self):
         return ["import torch", "from thunder.dynamo.benchmark_utils import TorchInductorSpecification"]
@@ -508,8 +510,8 @@ def check_metrics(
             )
             return None
 
-    measure1 = try_and_log_benchmark(compile_fn1, filename1)
-    measure2 = try_and_log_benchmark(compile_fn2, filename2)
+    _, *measure1 = try_and_log_benchmark(compile_fn1, filename1)
+    _, *measure2 = try_and_log_benchmark(compile_fn2, filename2)
 
     if measure1 is None or measure2 is None:
         return measure1, measure2
--- a/thunder/dynamo/benchmark_utils.py
+++ b/thunder/dynamo/benchmark_utils.py
@@ -142,7 +142,6 @@ class TorchInductorSpecification(CompileSpecificationInterface):
 
         if not skip_symbolic_trace:
             fn = symbolic_trace(fn)
-        assert isinstance(fn, torch.fx.GraphModule)
         return inductor_compile(fn, inputs)
 
     def compile(self, fn, *, inputs, **kwargs):
--- a/thunder/dynamo/benchmark_utils.py
+++ b/thunder/dynamo/benchmark_utils.py
@@ -131,8 +131,9 @@ class TorchInductorSpecification(CompileSpecificationInterface):
     https://github.com/Lightning-AI/lightning-thunder/issues/1521
     """
 
-    def __init__(self, specification_name="inductor_backend", *, skip_symbolic_trace=False):
+    def __init__(self, specification_name="inductor_backend", *, skip_symbolic_trace=True):
         self.name: str = specification_name
+        # self.skip_symbolic_trace decides whether to skip symbolic trace for self.compile
         self.skip_symbolic_trace = skip_symbolic_trace
 
     @staticmethod
@@ -147,8 +148,9 @@ class TorchInductorSpecification(CompileSpecificationInterface):
     def compile(self, fn, *, inputs, **kwargs):
         return self.torch_inductor(fn, inputs, skip_symbolic_trace=self.skip_symbolic_trace)
 
+    # to_source will always use symbolic trace
     def to_source(self, fn_name):
-        return f"TorchInductorSpecification.torch_inductor({fn_name}, inputs, skip_symbolic_trace={self.skip_symbolic_trace})"
+        return f"TorchInductorSpecification.torch_inductor({fn_name}, inputs)"
 
     def import_str(self):
         return ["import torch", "from thunder.dynamo.benchmark_utils import TorchInductorSpecification"]
--- a/thunder/dynamo/benchmark_utils.py
+++ b/thunder/dynamo/benchmark_utils.py
@@ -142,6 +142,7 @@ class TorchInductorSpecification(CompileSpecificationInterface):
 
         if not skip_symbolic_trace:
             fn = symbolic_trace(fn)
+        assert isinstance(fn, torch.fx.GraphModule)
         return inductor_compile(fn, inputs)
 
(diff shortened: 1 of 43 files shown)


thunder/dynamo/utils.py
```diff
@@ -989,16 +948,32 @@ def default_filter(fn: Callable, cutoff: int = 2) -> set[int]:
...
 
 
+def get_or_create_example_inputs_from_placeholders(placeholders: list[torch.fx.Node]) -> list[torch.Tensor]:
+    """
+    Gets the weakref of the inputs if possible, otherwise create inputs for benchmarking
+    """
+    outs = []
+    for p in placeholders:
+        try:
+            input: TensorWeakRef | torch.SymInt = p.meta["grapharg"].example
```
mruberry, line 958: This relies on the assumption that all arguments to the graph are tensors or symbolic integers, but that doesn't have to be true, does it?

thunder/dynamo/utils.py
```diff
@@ -989,16 +948,32 @@ def default_filter(fn: Callable, cutoff: int = 2) -> set[int]:
...
+            if isinstance(input, torch.SymInt):
+                input = input.node.hint
+        except Exception as e:
```
mruberry, line 961: Can this more narrowly catch the exception from attempting to access the example and it not being present, as opposed to catching EVERY exception?

thunder/dynamo/utils.py
```diff
@@ -989,16 +948,32 @@ def default_filter(fn: Callable, cutoff: int = 2) -> set[int]:
...
+            # needs to create a new example input
+            outs.append(_get_example_inputs_from_placeholder(p, only_metadata=False))
+        else:
```
mruberry, line 964: This is a nice use of the `try/except/else` pattern

thunder/dynamo/utils.py
```diff
@@ -1007,32 +982,52 @@ def default_optimizer(gm: torch.fx.GraphModule, stats: ProfileStats) -> Callable
...
+                reset_torch_dynamo=False,
+                example_inputs=example_inputs,
+                measure_fwd_bwd_together=True,
+            )
+        except Exception as e:
+            return float("inf")
+        return sum(m.median for m in measurement if m is not None)
+
+    compiled_gm_to_measurement = []
+    compiled_gm_to_measurement.append(("thunderfx", get_timing(report, thunderfx_compiler, WallTime)))
```
mruberry, line 1020: In the future we should think about making this extensible so that we can easily specify the different backends

thunder/dynamo/utils.py
```diff
@@ -1007,32 +982,52 @@ def default_optimizer(gm: torch.fx.GraphModule, stats: ProfileStats) -> Callable
...
+    compiled_gm_to_measurement.append(("inductor", get_timing(report, torchinductor, WallTime)))
+    compiled_gm_to_measurement.append(("eager", get_timing(report, torcheager, WallTime)))
+
+    sorted_compiled_gm_to_measurement = sorted(compiled_gm_to_measurement, key=itemgetter(1))
+    best_option = sorted_compiled_gm_to_measurement[0][0]
```
mruberry, line 1025: TODO: sort so that the thing to return is sorted, too

Doesn't the benchmarking already produce the compiled versions of the gm for inductor and thunder? Is it hard to acquire the already compiled version of the gm?

mruberry (commented):
Exciting stuff, @kiya00! I left a few comments for your review

kiya00:
some benchmarking results on HF models:
H100 (viking-prod-232), pjnl-20250427

  | torch.compile | thunderfx | prof-guided |  benchmark result in optimizer(sorted by time)
-- | -- | -- | -- | --
google/gemma-7b, | 0.4546394095 | 0.4527971821 | 0.4531846704 | [('inductor', 0.41045997792389244), ('thunderfx', 0.4108308005379513), ('eager', 0.4867668520892039)]
google/gemma-2-9b-it, | 1.404332365 | 1.441357784 | 1.402739552 | [('inductor', 1.3467837269417942), ('thunderfx', 1.3859112485079095), ('eager', inf)]
meta-llama/Llama-3.2-3B, | 0.2008398943 | 0.1910305724 | 0.18195783 | [('thunderfx', 0.16837437450885773), ('inductor', 0.18936991097871214), ('eager', 0.2376473949989304)]
meta-llama/Meta-Llama-3-8B-Instruct, | 0.4182928944 | 0.380578856 | 0.3713191442 | [('thunderfx', 0.34505993756465614), ('inductor', 0.37410450051538646), ('eager', 0.4418665610719472)]
microsoft/Phi-3.5-mini-instruct,   | 0.2388933344 | 0.2263890362 | 0.2149114404 | 6 graphs: [('eager', 0.00045880650158505883), ('inductor', 0.0005389171419665217), ('thunderfx', 0.0006505849305540323)] , <br>[('inductor', 1.1396194354165344e-05), ('eager', 1.7973064514808357e-05), ('thunderfx', 0.00010644900612533092)], <br>[('eager', 2.9640067217405885e-06), ('inductor', 3.5162785544525833e-06), ('thunderfx', 5.415198393166065e-05)], <br>[('inductor', 6.172084249556065e-05), ('eager', 7.013158139307051e-05), ('thunderfx', 0.00017521344125270844)], <br>[('thunderfx', 0.19113125139847398), ('inductor', 0.20647667243611068), ('eager', 0.28697569598443806)], <br>[('inductor', 0.0034934544237330556), ('thunderfx', 0.004427466541528702), ('eager', 0.0056188432965427635)]

mruberry:
The profile-guided results look very promising!

thunder/dynamo/utils.py
kiya00: https://github.com/pytorch/pytorch/blob/8f3d7972ad3e41ce4dcb1e9ff7bd1a3b0a671977/torch/_dynamo/variables/builder.py#L311
The type annotation here indicates is `Union[TensorWeakRef, torch.SymInt]`

kshitij12345 (approved):
The PR looks good to me, thanks @kiya00 

The numbers posted in https://github.com/Lightning-AI/lightning-thunder/pull/2062#issuecomment-2877480003 look good.

However, I am curious about the following

  | torch.compile | thunderfx | prof-guided |  benchmark result in optimizer(sorted by time)
-- | -- | -- | -- | --
meta-llama/Llama-3.2-3B, | 0.2008398943 | 0.1910305724 | 0.18195783 | [('thunderfx', 0.16837437450885773), ('inductor', 0.18936991097871214), ('eager', 0.2376473949989304)]

* is it correct that the numbers are in seconds?
* For this case, dynamo generates only one FXGraph. For this graph, profile-guided optimization will choose thunderfx. So I was wondering why the number for profile-guided is around 10 ms (assuming numbers are in seconds) better than thunderfx. I assumed it would be much closer.

Thanks!

thunder/dynamo/utils.py
mruberry: Would you add a comment to that effect?

thunder/dynamo/benchmark_utils.py
```diff
@@ -118,22 +131,24 @@ class TorchInductorSpecification(CompileSpecificationInterface):
...
-    def torch_inductor(fn, inputs):
+    def torch_inductor(fn, inputs, *, skip_symbolic_trace=False):
         from torch._inductor import compile as inductor_compile
         from torch.fx import symbolic_trace
 
-        fx_graph = symbolic_trace(fn)
-        return inductor_compile(fx_graph, inputs)
+        if not skip_symbolic_trace:
+            fn = symbolic_trace(fn)
+        return inductor_compile(fn, inputs)
```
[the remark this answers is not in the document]
kiya00: Then maybe we don't need to assert it, the `_inductor.compile` will give proper error if the input fn is not the expected type

kiya00:
I run the benchmark on the new commit [db7ccaa](https://github.com/Lightning-AI/lightning-thunder/pull/2062/commits/db7ccaa005de1690036c5b810046de04238bb27b), the result is updated in https://github.com/Lightning-AI/lightning-thunder/pull/2062#issuecomment-2877480003

thunder/dynamo/report.py
```diff
@@ -451,13 +457,16 @@ def run_repro(
     ):
         torch._dynamo.reset()
         example_inputs = self.make_example_inputs()
+        # ref: https://github.com/pytorch/pytorch/blob/0ef5ba43a6e7fe806ea9f27929bf4328ffd1ebf4/torch/_inductor/compile_fx.py#L1921-L1922
```
mruberry, line 460: Great comment and detail!

mruberry (approved):
@IvanYashchuk, would you like to take a look, too?

thunder/dynamo/utils.py
```diff
@@ -1007,32 +971,53 @@ def default_optimizer(gm: torch.fx.GraphModule, stats: ProfileStats) -> Callable
...
-        else:  # For inductor
-            optimized_module = _torch_inductor_compile(graph_module, example_inputs)
-            # Update the node name from "submod_*" to "inductor_*" for more user-friendly names
-            update_node_and_submodule(split_gm, node, node.name.replace("submod", "inductor"), optimized_module)
-
-    recompile_graph(split_gm)
-    return split_gm
+    report = FXGraphReport(gm, "gm", example_inputs_meta)
+    torcheager = TorchEagerSpecification()
+    torchinductor = TorchInductorSpecification(skip_symbolic_trace=True)
```
IvanYashchuk, line 991: Is `skip_symbolic_trace=False` useful to keep?

IvanYashchuk (approved):

thunder/dynamo/utils.py
kiya00: I modified to always skip the symbolic trace for graph module, and use it when generate the script

Issue #2066, opened 2025-05-14
fix list executors recipes


t-vi:
The recipes should require their executors explicitly rather than relying on a default set.
One goal is to fail loudly eg if nvfuser isn't importable rather than having bad perf.

PR #2074, opened 2025-05-15
Change default recipes to have non optional executors argument

KaelanDt:

## What does this PR do?

Fixes https://github.com/Lightning-AI/lightning-thunder/issues/2066 by having BaseRecipe explicitly take an `executors` argument as a list of strings.
The way it works with this PR is that `thunder.compile` and `thunder.jit` may be used without a recipe, but whenever a recipe is passed is has to be of type BaseRecipe or any subclass of it (cannot be string as before), with explicit executors passed to the recipe when initialized.
In addition, if `nvfuser` is passed in `executors`, it fails if nvfuser is not available.

### DIFF
--- a/thunder/recipes/hf_transformers.py
+++ b/thunder/recipes/hf_transformers.py
@@ -40,12 +40,12 @@ class InplaceIndexCopyTransform(thunder.Transform):
 class HFTransformers(BaseRecipe):
     def __init__(
         self,
+        executors,
         show_progress=False,
-        fuser="nvfuser",
         interpreter="thunder.jit",
         plugins=None,
     ):
-        super().__init__(show_progress=show_progress, fuser=fuser, interpreter=interpreter, plugins=plugins)
+        super().__init__(executors=executors, show_progress=show_progress, interpreter=interpreter, plugins=plugins)
         # for kv-cache inplace ops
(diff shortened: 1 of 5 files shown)


k223kim (approved):

KaelanDt:
Will have to update the hf benchmarks too to reflect this change

thunder/recipes/hf_transformers.py
```diff
@@ -40,12 +40,12 @@ def transform_traces_pre_prologue(self, pro, comp, epi, **kwargs):
 class HFTransformers(BaseRecipe):
     def __init__(
         self,
+        executors,
```
ali-alshaar7, line 43: I think the intention with the recipes entrypoint was to avoid users needing to deal with any thunder specific abstractions (such as executors), and instead get some preconfigured class that can instantly be applied to their models.

t-vi: I did not write the issue well, the idea is that the recipes contain a list of executors by name [...,'nvfuser', ....] rather than relying on Thunder's default. As @ali-alshaar7 points out, it's the recipes, not the users that should (need to) know about executors.

KaelanDt:
replaced by https://github.com/Lightning-AI/lightning-thunder/pull/2096

Issue #2023, opened 2025-05-01
Automated transformer coverage testing


t-vi:
We want to systematically assess where we are w.r.t. covering HF transformers cases.

- we want automated e2e tests, maybe you can re-use some of #1984 for how to launch them automatically.
- we want to use the latest version of transformers,
- initially we only run foward, later we need to test both inference (incl static kvcache) and training (fw + bw),
- This is for testing + coverage reporting, not for fixing nor performance!
- Let's start with 10-ish models. (there is a list internally, but I don't know if it is public, yet)

Afterwards, we want to file issues for errors we hit and also look at whether the error messages are OK or need improvement. 

Please reach out to Teja Pulagam (OSS contributor) to collaborate.

You could use the initial phases of the "decomposed" jit:

https://github.com/Lightning-AI/lightning-thunder/blob/3aa706a92092738e24a3f9dc08c96743646105e5/thunder/tests/test_core.py#L3216

IvanYashchuk:
We have a test for Hugging Face CausalLM models (`test_hf_for_nemo` it should be renamed to just `test_hf` maybe), but unfortunately, it's skipped for two months (https://github.com/Lightning-AI/lightning-thunder/commit/a6698fafc0fe652801312860d8d86bf3322f4f6b by @k223kim).

The test could be updated to add the parametrization to use `thunder.jit` in addition to `thunderfx`.

Currently, only two models are there. Here's the list of models we could enable (all of them do not require Hugging Face token): https://gist.github.com/IvanYashchuk/a16fe212ef4c9e42a648049790543f42.

lantiga:
Yes that's definitely the goal @IvanYashchuk, thanks for sharing the list and pointing to `test_hf_for_nemo`.
Now that we have the Studio-backed CI job up and running (daily), we have a way to add coverage as well on the proper machines without disrupting CI.

t-vi:
> it's skipped for two months 

This is tracked as #1920 I will comment there.

KaelanDt:
I believe we can close due to https://github.com/Lightning-AI/lightning-thunder/pull/2281 @t-vi @lantiga

PR #2011, opened 2025-04-30
ci: run quickstart as Lightning jobs

Borda:
<details>
  <summary><b>Before submitting</b></summary>

- [x] Was this discussed/approved via a Github issue? (no need for typos and docs improvements)
- [x] Did you read the [contributor guideline](https://github.com/Lightning-AI/pytorch-lightning/blob/main/.github/CONTRIBUTING.md), Pull Request section?
- [ ] Did you make sure to update the docs?
- [ ] Did you write any new necessary tests?

</details>

## What does this PR do?

resolves #1984

## PR review

Anyone in the community is free to review the PR once the tests have passed.
If we didn't discuss your PR in Github issues there's a high chance it will not be merged.

## Did you have fun?

Make sure you had fun coding 🙃

PR #2086, opened 2025-05-19
E2E Coverage Test for Thunder

tejapulagam:
## What does this PR do?

Initial approach for Issue #2023. Adds script that sets up Hugging Face models with dummy inputs and checks whether an initial trace can be acquired.

### DIFF
--- a/thunder/tests/test_coverage_trace.py
+++ b/thunder/tests/test_coverage_trace.py
@@ -1,0 +1,96 @@
+import torch
+import traceback
+import thunder 
+
+from transformers import (
+    AutoConfig,
+    AutoTokenizer,
+    AutoModel,
+    AutoModelForCausalLM, 
+    AutoModelForSeq2SeqLM,
+    AutoModelForImageClassification
+)
+
+
+MODEL_LIST = [
+    "gpt2",
+    "bert-base-uncased",
+    "google/reformer-enwik8",
+    "facebook/bart-large",
+    "t5-base",
+    "xlnet-base-cased",
+    "facebook/dinov2-small",
+    "albert-base-v2",
+    "google/electra-base-discriminator",
+    "facebook/opt-1.3b", 
+    "google/vit-base-patch16-224"
+ ]
+
+MODEL_TASKS = {
+    "openai/clip-vit-large-patch14": "clip",
+    "facebook/dinov2-small": "vision"
+}
+
+def get_model_class(model_name, config):
+    task = MODEL_TASKS.get(model_name, "text")
+    if task == "vision":
+        return AutoModelForImageClassification
+    elif config.architectures and "CausalLM" in config.architectures[0]:
+        return AutoModelForCausalLM
+    elif config.architectures and "Seq2SeqLM" in config.architectures[0]:
+        return AutoModelForSeq2SeqLM
+    else:
+        return AutoModel
+# custom input depending on model task
+def get_dummy_input(model_name, config):
+    task = MODEL_TASKS.get(model_name, "text")
+
+    if task == "vision":
+        return {
+            "pixel_values": torch.randn(1, 3, 224, 224, device="cpu", dtype=torch.float32)
+        }
+    else:  
+        return {
+            "input_ids": torch.randint(0, 1000, (1, 16), device="cpu")
+        }
+
+@thunder._with_cache_info_ctx
+def run_prologue(jfn, *args, **kwargs):
+    cd = thunder.compile_data(jfn)
+    cs = thunder.compile_stats(jfn)
+    ci = thunder._get_cache_info()
+    cd.populate_cache_info(ci, *args, **kwargs)
+    prologue_trc, computation_trc, epilogue_trc = cd.acquire_initial_trace(
+        cd.fn, args, kwargs, cd, cs, cd.executors_list[0]
+    )
+    cache_entry = cd.apply_transforms_and_build_cache_entry(cd, cs, ci, prologue_trc, computation_trc, epilogue_trc)
+    with thunder.compile_data_and_stats(cd, cs):
+        pro_to_comp, pro_to_epi = cache_entry.prologue_fn(*args, **kwargs)
+    return cache_entry, pro_to_comp, pro_to_epi
+
+def try_model(model_name):
+    print(f"\n=== Testing {model_name} ===")
+
+    try:
+        config = AutoConfig.from_pretrained(model_name)
+        model_class = get_model_class(model_name, config)
+        model = model_class.from_config(config).to("meta")
+        input_sample = get_dummy_input(model_name, config)
+
+    except Exception as setup_err:
+        print(f"[SKIPPED] {model_name} - model setup failed")
+        traceback.print_exc()
+        return  
+    
+    try:
+        jmodel = thunder.jit(model)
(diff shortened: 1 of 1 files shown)


lianakoleva (changes requested):
thank you @tejapulagam!
(1) i added a small request to delete the unused file
(2) when i run this (`python thunder/tests/coverage_trace.py`), it fails on several models: `google/reformer-enwik8`, `facebook/bart-large`, `t5-base`, `xlnet-base-cased`, `facebook/dinov2-small`, `facebook/opt-1.3b`, `google/vit-base-patch16-224`. please share the expected behavior somewhere either in the docs or at least in the PR description. it can be a to-do list of models and/or operators we could look at supporting in the future.
(3) i think it would be good to include this in Github CI to be manually run as per @t-vi's [suggestion](https://github.com/Lightning-AI/lightning-thunder/pull/2086#issuecomment-2890227597) ([reference](https://github.com/Lightning-AI/lightning-thunder/pull/2011)).

thunder/tests/coverage_trace.py
```diff
@@ -0,0 +1,81 @@
...
+def get_dummy_input(model_name, config):
+    task = MODEL_TASKS.get(model_name, "text")
+
+    if task == "vision":
+        return {"pixel_values": torch.randn(1, 3, 224, 224, device="cpu", dtype=torch.float32)}
+    else:
+        return {"input_ids": torch.randint(0, 1000, (1, 16), device="cpu")}
+
+
+def try_model(model_name):
```
Borda: why not run it as a standard test when all these try/except are handled?

thunder/tests/test_coverage_trace.py
```diff
@@ -0,0 +1,95 @@
+import torch
```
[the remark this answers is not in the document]
tejapulagam: done!

thunder/tests/coverage_trace.py
tejapulagam: Yeah I think structuring it as a test case makes sense. Is pytest the preferred framework?

tejapulagam: Updated to use Pytest

thunder/tests/test_core.py
```diff
@@ -39,6 +39,7 @@
 from thunder.core.trace import TraceCtx, set_tracectx, reset_tracectx, tracectx
 from thunder.core.symbol import BoundSymbol
 
+
```
lianakoleva, line 38: Please delete this whitespace

thunder/tests/test_core.py
```diff
@@ -49,6 +50,19 @@
...
+    cs = thunder.compile_stats(jfn)
+    ci = thunder._get_cache_info()
+    cd.populate_cache_info(ci, *args, **kwargs)
+    traces = cd.acquire_initial_trace(cd.fn, args, kwargs, cd, cs, cd.executors_list[0])
+    cache_entry = cd.apply_transforms_and_build_cache_entry(cd, cs, ci, *traces)
+    with thunder.compile_data_and_stats(cd, cs):
+        pro_to_comp, pro_to_epi = cache_entry.prologue_fn(*args, **kwargs)
+    return cache_entry, pro_to_comp, pro_to_epi
+
+
```
lianakoleva, line 61: Why was this moved and modified?

thunder/tests/test_core.py
tejapulagam: done!
https://github.com/Lightning-AI/lightning-thunder/pull/2086/commits/387438b1999925543fea67c1856b1febe905393a

thunder/tests/test_core.py
tejapulagam: Luca mentioned it may be better to make a helper for the acquisition of the trace and I thought test_core.py would be a good location as it contained other helpers for general tests we ran. I can move it back too if it makes more sense for this use case.

thunder/tests/test_core.py
lianakoleva: Looks like it was actually added by pre-commit, my apologies

thunder/tests/test_core.py
lianakoleva: It seems that this existing function was doing the job of the helper function Luca suggested you add, so you hoisted it to be able to import it from another file. Good work finding it and avoiding code duplication!

thunder/tests/test_coverage_trace.py
```diff
@@ -0,0 +1,83 @@
...
+
+    try:
+        config = AutoConfig.from_pretrained(model_name)
+        model_class = get_model_class(model_name, config)
+        model = model_class.from_config(config).to("meta")
+        input_sample = get_dummy_input(model_name, config)
+    except Exception:
+        print(f"[SKIPPED] {model_name} - model setup failed")
+        traceback.print_exc()
+        pytest.skip(f"Model setup failed for {model_name}")
```
Borda: This does not make sense, since you use testing patterns, just let it fail and trace is printed or wrap it with `pytest.mark.xfail` to let them fail

lantiga:
Thank you @tejapulagam for the work and @lianakoleva for the reviews!

We're going to land this as is, and we have worked on some updates to both the workflow and the CI so we'll submit a follow up PR shortly.

lantiga:
CI is breaking with the same failure as `main`, so the issue is unrelated to this PR.

thunder/tests/test_coverage_trace.py
```diff
@@ -0,0 +1,74 @@
...
+        return {"input_ids": torch.randint(0, 1000, (1, 16), device="cpu")}
+
+
+@pytest.mark.parametrize("model_name", MODEL_LIST)
+def test_model_trace(model_name):
+    print(f"\n=== Testing {model_name} ===")
+
+    config = AutoConfig.from_pretrained(model_name)
+    model_class = get_model_class(model_name, config)
+    model = model_class.from_config(config).to("meta")
```
IvanYashchuk, line 68: Real devices will be tested in a follow-up, right?

thunder/tests/test_coverage_trace.py
```diff
@@ -0,0 +1,74 @@
...
+
+    if task == "vision":
+        return {"pixel_values": torch.randn(1, 3, 224, 224, device="cpu", dtype=torch.float32)}
+    else:
+        return {"input_ids": torch.randint(0, 1000, (1, 16), device="cpu")}
+
+
+@pytest.mark.parametrize("model_name", MODEL_LIST)
+def test_model_trace(model_name):
+    print(f"\n=== Testing {model_name} ===")
```
IvanYashchuk, line 64: Verbose mode of pytest (`-v`) should provide the same functionality and it's better to remove the print statement from here.

thunder/tests/test_coverage_trace.py
```diff
@@ -0,0 +1,74 @@
...
+
+    config = AutoConfig.from_pretrained(model_name)
+    model_class = get_model_class(model_name, config)
+    model = model_class.from_config(config).to("meta")
+    input_sample = get_dummy_input(model_name, config)
+
+    jmodel = thunder.jit(model)
+    ce, pro_to_comp, pro_to_epi = run_prologue(jmodel, **input_sample)
```
IvanYashchuk, line 72: In a follow-up, it would be ideal to add a parametrization over inference and training modes.

PR #2088, opened 2025-05-19
Versioning for check_traces

t-vi:
this is needed for adding features to check_traces and then incrementally tighten tests

### DIFF
--- a/thunder/dev_utils/check_trace.py
+++ b/thunder/dev_utils/check_trace.py
@@ -1,3 +1,8 @@
+import thunder
+
+CHECK_VERSION = 3
+
+
 def check_subsymbols(parent_bsym):
     if parent_bsym.sym.is_prim:
         # assert that there are no subsymbols?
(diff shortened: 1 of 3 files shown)


thunder/dev_utils/check_trace.py
```diff
@@ -1,3 +1,8 @@
+import thunder
+
+CHECK_VERSION = 3
```
ali-alshaar7, line 3: Can the version be a labeled enum or so? otherwise its hard to tell what each version does and why pick one over another.

t-vi: Fair enough, I should add a docstring.

The thinking here is
- as a user / developing something new, use `True` to get the latest and greatest.
- for the CI, we want to increment.

The idea is to be able to introduce a new check without fixing all things it detects in the same PR. 
Having feature flags seems over the top, given that we do want to fix the inconsistencies, just not all at once.

lantiga (approved):

PR #2090, opened 2025-05-19
disambiguate renames

t-vi:
...hoping it doesn't happen is a bit of a hack.

### DIFF
--- a/thunder/core/proxies.py
+++ b/thunder/core/proxies.py
@@ -149,6 +149,7 @@ class Proxy(VariableInterface, ProxyInterface):
             cnt = 0
             while trc.has_name(name):
                 name = f"{name_prefix}_{cnt}"
+                cnt += 1
 
         return self.replace(name=name)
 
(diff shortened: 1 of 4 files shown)


thunder/core/proxies.py
```diff
@@ -137,8 +137,19 @@ def replace(self, **changes):
...
+        if disambiguate:
+            trc = get_tracectx()
+        else:
+            trc = None
+
+        if trc is not None:
+            name_prefix = name
+            cnt = 0
+            while trc.has_name(name):
+                name = f"{name_prefix}_{cnt}"
```
crcrpar, line 151: wouldn't we need to increment `cnt`?

t-vi: Great catch. :) Apparently we only needed one level up to now...

lantiga (approved):

crcrpar:
Could you make the title a bit more verbose so that the commit message would be clearer?
e.g. maybe something like Disambiguate proxy names of backward traces

lantiga (approved):

PR #2096, opened 2025-05-19
Make recipes define their executors explicitly

KaelanDt:
<details>

- [] Did you make sure to update the docs?

</details>

## What does this PR do?

Fixes https://github.com/Lightning-AI/lightning-thunder/issues/2066. This means that thunder.compile will require nvfuser to be available to not fail. 

## PR review

## Did you have fun?

### DIFF
--- a/thunder/__init__.py
+++ b/thunder/__init__.py
@@ -294,10 +294,7 @@ def compile(
             plugins_.append(plugin)
     plugins = plugins_
 
-    if recipe is None and not plugins:
-        return thunder.jit(fn)
-
-    if recipe is None and plugins:
+    if recipe is None:
         recipe = thunder.recipes.BaseRecipe()
 
     if recipe == "auto":
--- a/thunder/core/recipe.py
+++ b/thunder/core/recipe.py
@@ -6,7 +6,7 @@ import warnings
 import torch
 
 from thunder.core.transform_common import Transform
-from thunder.extend import Executor, get_default_executors
+from thunder.extend import Executor, get_all_executors
 
 
 @contextmanager
@@ -80,7 +80,7 @@ class Recipe:
         return None
 
     def setup_executors(self) -> list[Executor]:
-        return list(get_default_executors())
+        return list(get_all_executors())
 
     def setup_config(self) -> dict[str, Any]:
         return {}
--- a/thunder/core/recipe.py
+++ b/thunder/core/recipe.py
@@ -6,7 +6,7 @@ import warnings
 import torch
 
 from thunder.core.transform_common import Transform
-from thunder.extend import Executor, get_all_executors
+from thunder.extend import Executor
 
 
 @contextmanager
@@ -80,7 +80,7 @@ class Recipe:
         return None
 
     def setup_executors(self) -> list[Executor]:
-        return list(get_all_executors())
+        return None
 
     def setup_config(self) -> dict[str, Any]:
         return {}
--- a/thunder/core/recipe.py
+++ b/thunder/core/recipe.py
@@ -80,7 +80,7 @@ class Recipe:
         return None
 
     def setup_executors(self) -> list[Executor]:
-        return None
+        return []
 
     def setup_config(self) -> dict[str, Any]:
         return {}
--- a/thunder/recipes/base.py
+++ b/thunder/recipes/base.py
@@ -39,7 +39,7 @@ class BaseRecipe(Recipe):
         plugins=None,
     ):
         super().__init__(interpreter=interpreter, plugins=plugins)
-        self.executors = ["torch", "python", "cudnn", "sdpa"]
+        self.executors = ["cudnn", "sdpa", "torchcompile_xentropy"]
         self.fuser = fuser
         self.setup_fuser()
         self.show_progress = show_progress
--- a/thunder/recipes/base.py
+++ b/thunder/recipes/base.py
@@ -95,7 +95,7 @@ class BaseRecipe(Recipe):
         executors = []
 
         for name in self.executors:
-            executor = thunder.get_executor(name)
+            executor = get_executor(name)
             if executor is None:
 
                 if name == "nvfuser":
--- a/thunder/recipes/base.py
+++ b/thunder/recipes/base.py
@@ -11,11 +11,13 @@ class BaseRecipe(Recipe):
     def __init__(
         self,
         show_progress=False,
+        fuser="nvfuser",
         interpreter="thunder.jit",
         plugins=None,
     ):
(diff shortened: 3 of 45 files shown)


thunder/__init__.py
```diff
@@ -294,10 +294,7 @@ def compile(
             plugins_.append(plugin)
     plugins = plugins_
 
-    if recipe is None and not plugins:
```
ali-alshaar7, line 297: Should the default behaviour here be to use base recipe or fallback to thunder.jit? Deferring to @lantiga

thunder/recipes/base.py
```diff
@@ -10,12 +11,11 @@ class BaseRecipe(Recipe):
     def __init__(
         self,
         show_progress=False,
-        fuser="nvfuser",
```
ali-alshaar7, line 13: So do we never use`torchcompileex ` through this entrypoint? Again, deferring to @lantiga

thunder/tests/test_examine.py
```diff
@@ -11,6 +12,7 @@ def foo(x):
     thunder.examine.examine(foo, x)
 
 
+@pytest.mark.skipif(not nvfuser_available(), reason="NVFuser is not available")
```
ali-alshaar7, line 15: So this will always be skipped now, since it runs on cpu and thus no nvfuser.

thunder/tests/test_recipes.py
```diff
@@ -101,6 +123,8 @@ def test_plugins_basics():
         assert ex.name in [el.name for el in cd.executors_list]
 
 
+# test skipped if nvfuser isn't available because providing plugins calls BaseRecipe
+@pytest.mark.skipif(not nvfuser_available(), reason="NVFuser is not available")
```
ali-alshaar7, line 127: Same as the above comment for all these

ali-alshaar7 (commented):
Few comments mainly about the new default behaviour this introduces

thunder/recipes/base.py
lantiga: Good call out @ali-alshaar7, let's keep exposing the fuser for the time being.

thunder/tests/test_recipes.py
```diff
@@ -7,9 +7,12 @@
 
 from thunder.extend import deregister_executor
 from torch.testing import assert_close, make_tensor
+from thunder.recipes import HFTransformers
+from thunder.executors import nvfuser_available
 from thunder.tests.framework import version_between, IS_WINDOWS
 
 
+@pytest.mark.skipif(not nvfuser_available(), reason="NVFuser is not available")
```
lantiga, line 15:
```suggestion
@pytest.mark.skipif(not nvfuser_available(), reason="nvFuser is not available")
```
similarly for the other cases

lantiga (commented):
Looks good, thank you @KaelanDt! added a few comments

thunder/core/recipe.py
```diff
@@ -80,7 +80,7 @@ def setup_transforms(self) -> list[Transform] | None:
         return None
 
     def setup_executors(self) -> list[Executor]:
-        return list(get_default_executors())
+        return list(get_all_executors())
```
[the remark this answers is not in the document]
lantiga: That's right but then maybe we should just return None or raise NotImplemented

thunder/recipes/base.py
```diff
@@ -1,11 +1,35 @@
...
+from thunder.executors import nvfuser_available
 from thunder.executors.torch_compile import torch_compile_ex
 from thunder.transforms.prune_prologue_checks import PrunePrologueChecks
-
+from thunder.extend import get_all_executors
 from typing import Any
 
 
+def get_nvfuser_package_for_torch(torch_version) -> str:
+    torch_major_minor = torch_version[:4].replace(".", "")
```
t-vi, line 12: Three things:
- I wonder if we should use torch.__version__ and torch.version.cuda here?
- Also I wonder if we should just give typical the package name and "see <link to docs> for details".

The caveats seem to be many for to include in the error message:
- PyTorch has no ABI compatibility between anything, so you do need the latest release pytorch (e.g. 2.7.x) for this to work.
- We don't know which SKUs are supported.
- Let's not error in the error message construction.

thunder/recipes/base.py
```diff
@@ -28,14 +54,48 @@ def setup_transforms(self) -> list[Transform]:
...
+
+    def setup_executors(self) -> list[Executor]:
+        if not isinstance(self.executors, list):
+            raise TypeError(f"self.executors must be a list of executor names, got {type(self.executors).__name__}")
+
+        available_ex = list(get_all_executors())
+        available_ex_map = {ex.name: ex for ex in available_ex}
+        selected_ex = []
+        for name in self.executors:
+            if name not in available_ex_map:
```
t-vi, line 79: why not use `thunder.get_executor(name)` instead of re-doing that?

thunder/recipes/hf_transformers.py
```diff
@@ -41,11 +41,11 @@ class HFTransformers(BaseRecipe):
     def __init__(
         self,
         show_progress=False,
-        fuser="nvfuser",
         interpreter="thunder.jit",
         plugins=None,
     ):
-        super().__init__(show_progress=show_progress, fuser=fuser, interpreter=interpreter, plugins=plugins)
+        super().__init__(show_progress=show_progress, interpreter=interpreter, plugins=plugins)
+        self.executors.append("inplace_index_copy_ex")
```
t-vi, line 48: can we keep the pattern of getting this executor from the transform, please?

t-vi (commented):
Looks great, but there are a couple of details. Thank you @KaelanDt

thunder/recipes/base.py
KaelanDt: addressed

thunder/recipes/hf_transformers.py
```diff
@@ -41,13 +41,14 @@ class HFTransformers(BaseRecipe):
...
-        fuser="nvfuser",
         interpreter="thunder.jit",
         plugins=None,
     ):
-        super().__init__(show_progress=show_progress, fuser=fuser, interpreter=interpreter, plugins=plugins)
+        super().__init__(show_progress=show_progress, interpreter=interpreter, plugins=plugins)
+
         # for kv-cache inplace ops
         self.inplace_index_copy_transform = InplaceIndexCopyTransform()
+        self.executors.append(InplaceIndexCopyTransform().executor.name)
```
t-vi, line 51: would it work if we used the executor here instead of the name? It seems the global lookup is causing trouble somewhere?

t-vi (approved):
Thank you @KaelanDt  Seems there is something up with (multiple) registrations.

PR #2102, opened 2025-05-20
[WIP2]

beverlylytle:
Running CI

### DIFF
--- a/thunder/__init__.py
+++ b/thunder/__init__.py
@@ -559,29 +559,43 @@ def jit(
                 # transform_for_execution and various sorting of symbols,
                 # applying transform_for_execution after this would be
                 # breaking the order of operations
-                computation_trc, backward_trc = split_forward_backward(computation_trc, cd, cs, *computation_trc.args)
+                # computation_trc, backward_trc = split_forward_backward(computation_trc, cd, cs, *computation_trc.args)
                 # Note computation_trc and backward_trc have been appended to cs.last_(backward_)traces
                 # by split_forward_backward
+                from thunder.transforms.autodiff import grad_transform_on_trace
 
-            if not requires_grad:
-                from thunder.executors.passes import transform_for_execution as transform_for_execution_pass
-                from thunder.executors.passes import _transform_for_operator_executor_execution
-                from thunder.distributed.utils import maybe_sort_waits
+                computation_trc = grad_transform_on_trace(computation_trc)
 
-                tmp_comp_trc = _transform_for_operator_executor_execution(computation_trc, cd.executors_list)
-                is_transformed, tmp_comp_trc = maybe_sort_waits(tmp_comp_trc)
-                if is_transformed:
-                    computation_trc = tmp_comp_trc
-                    computation_traces.append(computation_trc)
+            from thunder.executors.passes import transform_for_execution as transform_for_execution_pass
+            from thunder.executors.passes import _transform_for_operator_executor_execution
+            from thunder.distributed.utils import maybe_sort_waits
 
-                extraces = transform_for_execution(
-                    computation_trc,
-                    executors_list=cd.executors_list,
-                    use_del_last_used=False,
-                )
-                computation_traces.extend(extraces)
-                computation_trc = computation_traces[-1]
-                computation_trc = thunder.executors.passes.del_last_used(computation_trc)
+            tmp_comp_trc = _transform_for_operator_executor_execution(computation_trc, cd.executors_list)
+            is_transformed, tmp_comp_trc = maybe_sort_waits(tmp_comp_trc)
+            if is_transformed:
+                computation_trc = tmp_comp_trc
+                computation_traces.append(computation_trc)
+
+            extraces = transform_for_execution(
+                computation_trc,
+                executors_list=cd.executors_list,
+                use_del_last_used=False,
+            )
+            # computation_traces.extend(extraces)
+            computation_trc = extraces[-1]
+
+            if requires_grad:
+                from thunder.core.rematerialization import rematerialize
+                from thunder.transforms.autodiff import split_into_forward_and_backward
+
+                computation_trc = rematerialize(computation_trc)
+                computation_trc, backward_trc = split_into_forward_and_backward(computation_trc)
+
+            computation_trc = thunder.executors.passes.del_last_used(computation_trc)
+            computation_traces.append(computation_trc)
+            if backward_trc is not None:
+                backward_trc = thunder.executors.passes.del_last_used(backward_trc, clear_mutable_collections=True)
+                backward_traces.append(backward_trc)
 
             if not compile_options.get("disable_inplace_copy_check", False):
                 thunder.core.transform_common._inplace_copy_sanity_check(computation_trc)
--- a/thunder/executors/torchex.py
+++ b/thunder/executors/torchex.py
@@ -249,6 +249,24 @@ def _arange_transform(
     return arange(start=start, step=step, end=end, device=torch_device, dtype=torch_dtype)
 
 
+def _get_grad_transform(a: TensorProxy) -> TensorProxy:
+    return prims.get_grad(a)
+
+
+def _put_grad_transform(a: TensorProxy, val: TensorProxy) -> None:
+    # what a hack. _transform_for_operator_executor_execution needs that prims are executable
+    # by some executor, and if they are not, it will descend to the subsymbols.  So, we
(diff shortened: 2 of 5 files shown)


thunder/executors/torchex.py
```diff
@@ -249,6 +249,24 @@ def _arange_transform(
...
+def _get_grad_transform(a: TensorProxy) -> TensorProxy:
+    return prims.get_grad(a)
+
+
+def _put_grad_transform(a: TensorProxy, val: TensorProxy) -> None:
+    # what a hack. _transform_for_operator_executor_execution needs that prims are executable
+    # by some executor, and if they are not, it will descend to the subsymbols.  So, we
+    # force that and recurse so that the symbol is not removed/no error is raised.
+    # This is breaking the intent of transform_for_operator_executor_execution, in that
+    # it is not executable.  But there will be postprocessing that will make it executable.
```
t-vi, line 261: Nothing bad about a good hack, but if you just want get_grad/put_grad to survive you could just special case put grad/get grad in the transform for execution (but there should not be any put_grad left) before you loop over the executors:

https://github.com/Lightning-AI/lightning-thunder/blob/5bdf5bc9f6805d19f5cbb152e24ed300cc109a28/thunder/executors/passes.py#L44-L55

thunder/tests/test_torch_compile_executor.py
```diff
@@ -85,7 +85,7 @@ def test_torch_compile_cat_rope_single_fusion():
 
     backward_execution_trace = thunder.last_backward_traces(jfn)[-1]
     assert len(get_fusions(backward_execution_trace)) == 1
-    assert len(backward_execution_trace.bound_symbols) == 14
+    assert len(backward_execution_trace.bound_symbols) == 17
```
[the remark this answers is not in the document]
IvanYashchuk: Added in https://github.com/Lightning-AI/lightning-thunder/pull/2102/commits/a27f390ded016079c1ce388b0509bd8d9a891d34.

thunder/__init__.py
```diff
@@ -576,9 +586,18 @@ def apply_transforms_and_build_cache_entry(cd, cs, cache_info, prologue_trc, com
                     executors_list=cd.executors_list,
                     use_del_last_used=False,
                 )
-                computation_traces.extend(extraces)
-                computation_trc = computation_traces[-1]
-                computation_trc = thunder.executors.passes.del_last_used(computation_trc)
+                computation_trc = extraces[-1]
```
beverlylytle, line 589: Note that the joint traces are not being appended here, and that is also delayed until after the splitting.  I'm unsure whether this is desired behavior or not.

lantiga:
Thanks for the clarification @IvanYashchuk. I agree we can forego bucketing for now and eventually circle back to it at a later stage.

thunder/core/transform_common.py
```diff
@@ -286,6 +286,9 @@ def cse_single_bsym(
         skip_output=True,
     )
 
+    if bsym.sym.id == prims.PrimIDs.GET_GRAD:
+        return new_bsym
+
```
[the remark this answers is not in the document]
riccardofelluga: Ohh so it looks like this before dce:
```
  nvFusion1(a, t13)
    # t8 = prims.cos(a)  # t8: "cuda:0 f64[2, 3]"
    # t9 = prims.mul(t13, t8)  # t9: "cuda:0 f64[2, 3]"
```
So there has been some dce going on there earlier. The fact that it has no output to me suggests that it suffers from #2132 or similar issue hmmm interesting

beverlylytle:
I've benchmarked this change with the following baseline and result:
```
@main
Model name: Llama-3-8B
Seq Length: 8192
Micro BS: 1
Global BS: 8
Number of Layers: 32
Number of parameters: 1.00B
Distributed Mode: fsdp
Sharding Mode: zero2
Bucketing: block
Compiler: dynamo_thunder
Low Precision Mode: none
Average iter time: 782.73 ms
Memory used: 72.61 GB
Tokens/s: 83707.48
Tokens/s/GPU: 10463.43
TFLOP/s: 4847.74

@reautograd2
Model name: Llama-3-8B
Seq Length: 8192
Micro BS: 1
Global BS: 8
Number of Layers: 32
Number of parameters: 1.00B
Distributed Mode: fsdp
Sharding Mode: zero2
Bucketing: block
Compiler: dynamo_thunder
Low Precision Mode: none
Average iter time: 781.49 ms
Memory used: 72.61 GB
Tokens/s: 83855.93
Tokens/s/GPU: 10481.99
TFLOP/s: 4856.34
```

t-vi:
Does it work with the thunder.jit? Could we also benchmark with "thunder" as compiler?

beverlylytle:
```
@main
Model name: Llama-3-8B
Seq Length: 8192
Micro BS: 1
Global BS: 8
Number of Layers: 32
Number of parameters: 1.00B
Distributed Mode: fsdp
Sharding Mode: zero2
Bucketing: none
Compiler: thunder
Low Precision Mode: none
Average iter time: 799.59 ms
Memory used: 75.75 GB
Saved for backward size: 58448.60 MiB
Saved for backward number of tensors: 775
Tokens/s: 81988.63
Tokens/s/GPU: 10248.58
TFLOP/s: 4748.20

@retrograd2
Model name: Llama-3-8B
Seq Length: 8192
Micro BS: 1
Global BS: 8
Number of Layers: 32
Number of parameters: 1.00B
Distributed Mode: fsdp
Sharding Mode: zero2
Bucketing: none
Compiler: thunder
Low Precision Mode: none
Average iter time: 808.52 ms
Memory used: 71.53 GB
Saved for backward size: 62988.09 MiB
Saved for backward number of tensors: 712
Tokens/s: 81032.54
Tokens/s/GPU: 10129.07
TFLOP/s: 4692.83
```

beverlylytle:
@riccardofelluga has made an excellent point.  The bound symbols which create tensors that are to be recomputed in the backward pass are only inserted once the joint trace is split.  This means that those operations do not get fused optimally.

t-vi:
@beverlylytle Thank you for running the benchmark.

But isn't it then the fusion or part of it that gets duplicated for recomputation?
 At least that should be our target.

The other question is whether everything between the autodiff and the split propagates the tag properly. We would probably need to take a good look.
For Transform for Operator Execution, the trace processor would (want to) do it.
For the fusion passes, this may be more tricky currently.

What we need to do is likely look for recompute tags in the fusion subsymbols and then duplicate the fusion and then dce the bits away that we don't need. WDYT?

thunder/__init__.py
t-vi: No, we would prefer to append things early if only to look at them when transforms error, but we fix this separately, it's not that great in the current torch_autograd code either.

beverlylytle:
Yes, it is the fusion or part of it that gets duplicated, but a fusion region may compute more than what is intended to be recomputed in backward, right?  Moreover, the fusion regions for the basic backward together with the fusion regions for the stuff to be recomputed in the backward may not maximally fused.

t-vi:
- a fusion region may compute more than what is intended to be recomputed in backward, right?

Yes, but this could be DCEd and rerun the codegen?

-  Moreover, the fusion regions for the basic backward together with the fusion regions for the stuff to be recomputed in the backward may not maximally fused.

Indeed, and it is tricky. But our fusion algorithms all seem wild currently. The one used by JIT by default fuses little (only adjacent regions) and the one used by thunderfx reoders operations that can wreck memory consumption (we saw this for checkpointing in particular).
I think if we can rerun the codegen, we could at least re-fuse adjacent regions.

But so I seem to remember that the interaction of fusions with recomputation was one of the key things you wanted to get out of the joint trace logic, so it might be more tricky yet.

The other thing on my list of larger project that might impact our choices is to have something like
```python
def fn(inp, target):
  out = model(inp)
  loss = lossfn(out, target)
  loss.backward()
   
jfn = jit(fn, models=(model,))
```
as one trace. This would let us not split the trace and likely is an important next step (followed by adding `optimizer.step()`) for optimizing training.

beverlylytle:
Not only is it important that the duplicates of the bsyms that should be recomputed exist during the fusion pass, it is also important that they exist during rematerialization. I am going to try an approach where the duplication logic is moved from the splitting up to the creation of the joint trace. I will modify CSE so that no commonality will be found across the forward-backward divide.

t-vi:
To my mind, we could merge this PR as is and do what is needed in a follow up. WDYT? (also @IvanYashchuk )

IvanYashchuk (approved):

Issue #2116, opened 2025-05-22
Activation checkpointing is broken


IvanYashchuk:
## 🐛 Bug

With ba561bf7b11550028f7657db30a075c06d3de10b activation checkpointing got broken resulting in out-of-memory errors.

Command to run
```
torchrun --nproc_per_node=8 thunder/benchmarks/benchmark_litgpt.py --model Llama-3-8B --compile dynamo_thunder --distributed_mode fsdp --checkpoint_activations=True
```
```py
RuntimeError: CUDA out of memory. Tried to allocate 448.00 MiB. GPU 3 has a total capacity of 79.10 GiB of which 83.88 MiB is free.
```

With the commit right before ba561bf7b11550028f7657db30a075c06d3de10b (`git checkout ba561b^`), everything works as expected. Example output:
```
Average iter time: 943.72 ms
Memory used: 32.85 GB
Tokens/s: 69427.74
Tokens/s/GPU: 8678.47
TFLOP/s: 4020.76
```

IvanYashchuk:
Related issue with a smaller example to reproduce the problem with activation checkpointing: https://github.com/Lightning-AI/lightning-thunder/issues/2112.

Borda:
@IvanYashchuk what is the required mechine/GPU for this benchmark?

IvanYashchuk:
Any multi-GPU machine would show the regression. Here's a command for 2x GPUs:
```
torchrun --nproc_per_node=2 thunder/benchmarks/benchmark_litgpt.py --model Llama-3-8B --compile dynamo_thunder --distributed_mode fsdp --checkpoint_activations=True --n_layer=3
```

t-vi:
@IvanYashchuk I tentatively think that #2118 fixes the benchmark command you posted here. How should we go about seeing how long we need #2117 ?

IvanYashchuk:
We don't need https://github.com/Lightning-AI/lightning-thunder/pull/2117 with https://github.com/Lightning-AI/lightning-thunder/pull/2118 merged.

PR #2118, opened 2025-05-22
add tag based checkpointing

t-vi:
Testing for #2116

### DIFF
--- a/thunder/transforms/autodiff.py
+++ b/thunder/transforms/autodiff.py
@@ -10,7 +10,7 @@ from thunder.core.transforms import (
     recompute_saved_for_backward,
 )
 from thunder.core.proxies import ProxyTag
-from thunder.core.symbol import BoundSymbol
+from thunder.core.symbol import BoundSymbol, BoundSymbolTag
 from thunder.core.vjp_utils import make_aug_forward_and_backward
 from thunder.core.pytree import tree_map
 import thunder
@@ -18,7 +18,9 @@ import time
 
 
 def _should_recompute_bsym_in_backward(bsym):
-    return any((ProxyTag.RECOMPUTE_IN_BACKWARD in o.tags) for o in bsym.flat_proxy_outs)
+    return BoundSymbolTag.RECOMPUTE_IN_BACKWARD in bsym.tags or any(
+        (ProxyTag.RECOMPUTE_IN_BACKWARD in o.tags) for o in bsym.flat_proxy_outs
+    )
 
 
 # Transforms a trace by determining which grad transforms to call given the list of executors in priority order
@@ -168,6 +170,7 @@ def grad_transform_on_trace(trace, /, *args, **kwargs):
                 function_arg_names = {a.name for a in bsym.flat_proxy_args}
 
                 for subsym in bsym.subsymbols:
+                    subsym.tags.add(BoundSymbolTag.RECOMPUTE_IN_BACKWARD)
                     for o in subsym.flat_proxy_outs:
                         if o.name not in function_arg_names:
                             o.tags.add(ProxyTag.RECOMPUTE_IN_BACKWARD)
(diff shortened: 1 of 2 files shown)


t-vi:
still OOMs in the example from #2116

thunder/transforms/autodiff.py
```diff
@@ -10,15 +10,17 @@
...
 from thunder.core.pytree import tree_map
 import thunder
 import time
 
 
 def _should_recompute_bsym_in_backward(bsym):
-    return any((ProxyTag.RECOMPUTE_IN_BACKWARD in o.tags) for o in bsym.flat_proxy_outs)
+    return BoundSymbolTag.RECOMPUTE_IN_BACKWARD in bsym.tags or any(
+        (ProxyTag.RECOMPUTE_IN_BACKWARD in o.tags) for o in bsym.flat_proxy_outs
+    )
```
riccardofelluga, line 23: Interesting, there are two types of tags one for tensors and one for bound symbols. I was thinking would we be able to get away with just having bound symbols tags alone? it feels like they already imply that the output, whatever it is, it's going to be recomputed

thunder/transforms/autodiff.py
```diff
@@ -292,6 +299,11 @@ def shallow_copy_if_input(p):
 
             # TODO: check if this is needed: the old impl checked whether len(bsym.subsymbols) > 0 except for the special case "torch.nn.functional.dropout" with p=0...
             # add the decomposition (= the subsymbols) to the front of the symbols to be processed
+
+            # inherit tags, in particular RECOMPUTE_IN_BACKWARD
```
riccardofelluga, line 303:
```suggestion
            # Let the decomposition inherit tags, in particular RECOMPUTE_IN_BACKWARD
```

thunder/transforms/autodiff.py
```diff
@@ -260,6 +263,10 @@ def shallow_copy_if_input(p):
                 self.set_result(result)
                 new_bsyms = self.new_trace.pop_scope()
 
+                # inherit tags, in particular RECOMPUTE_IN_BACKWARD
```
riccardofelluga, line 266:
```suggestion
                # Let the new bound symbols inherit tags, in particular RECOMPUTE_IN_BACKWARD
```

thunder/transforms/autodiff.py
t-vi: Yeah, the idea is to rip out the proxy recompute tag later.

t-vi:
This leaves a number of todos:
- ensure propagation of bound symbol tags through transforms,
- standardise checkpointing on bound symbol tags (i.e. get rid of the proxy tag).

While I'm not seeing it in the benchmark script, I'm seeing differences in the backward. (Will post below.)

t-vi:
This tests has ~38 for eager  ~50 for the old autograd and ~55 for the new.
The forward seems to be identical, but the backward has slightly different nvfusion regions.

```python
    from thunder.tests import litgpt_model
    from torch.distributed.algorithms._checkpoint.checkpoint_wrapper import (
        apply_activation_checkpointing,
        checkpoint_wrapper,
        CheckpointWrapper,
    )

    def forward_backward_peak(m, inp):
        torch.cuda.reset_peak_memory_stats(device=None)
        mem_before = torch.cuda.max_memory_allocated()
        res = m(inp)
        res.sum().backward()
        mem_after = torch.cuda.max_memory_allocated()
        return (mem_after - mem_before) / 2**20

    with torch.device("cuda"):
        m = litgpt_model.GPT.from_name("llama2-like").bfloat16()
        inp = torch.ones((1, 2048), dtype=torch.int64)

    check_fn = lambda submodule: isinstance(submodule, litgpt_model.Block)
    apply_activation_checkpointing(m, checkpoint_wrapper_fn=checkpoint_wrapper, check_fn=check_fn)

    # warmup, allocate grads etc.
    forward_backward_peak(m, inp)
    forward_backward_peak(m, inp)
    jm = thunderfx(m)
    forward_backward_peak(jm, inp)
    forward_backward_peak(jm, inp)

    mem_thunder = forward_backward_peak(jm, inp)
    mem_eager = forward_backward_peak(m, inp)

    print(jm.last_traces)
    print(jm.last_backward_traces)

```

t-vi:
The tag based checkpointing needs a shallow copy at the end in order to remember not to recompute the output of the block when you have consecutive checkpointing blocks.

lantiga (approved):

PR #2069, opened 2025-05-14
Add parity check of shape/dtype/device of runtime and trace

crcrpar:
## What does this PR do?

Adding a check of the consistency aforementioned in the title.

related:
- https://github.com/Lightning-AI/lightning-thunder/issues/2068
- https://github.com/Lightning-AI/lightning-thunder/issues/1983
- https://github.com/Lightning-AI/lightning-thunder/issues/1952

PR #2135, opened 2025-05-25
`prims.where` ignores shape/device of `pred` if it's a CPU scalar tensor

crcrpar:
In order to fix the shape/device mismatch between runtime and trace, this PR adds the excluding `pred` if it's a CPU scalar tensor when deciding shape/device of outputs.

When the `condition` of `torch.where` is a CPU scalar tensor, the broadcast doesn't occur as we can see in the following trace:
```python
def computation(cond, a, b):
  # cond: "cpu b8[]"
  # a: "cpu f32[4, 4]"
  # b: "cpu f32[]"

  # repro.py:11:             return torch.where(cond, a, b)
  t0 = ltorch.where(cond, a, b)  # t0: "cpu f32[]"
    # t0 = prims.where(cond, a, b)  # t0: "cpu f32[]"
  return (t0,)
```
This causes a mismatch between runtime results and trace outputs.

This trace is obtained by running the script below:
```python
import torch
import thunder

def f(cond, a, b):
    return torch.where(cond, a, b)

if __name__ == "__main__":
    jitted = thunder.jit(f)
    with torch.device("cpu"):
        cond = torch.tensor(True, dtype=torch.bool)
        a = torch.tensor([
            [ 6.5077,  7.6115,  8.6770,  6.2419],
            [ 7.1474, -4.7458, -8.1032,  8.7696],
            [ 7.9486,  5.3679, -7.8606, -5.3753],
            [-5.8697,  7.7565, -1.9403,  4.1512],
        ])
        b = torch.tensor(-2.6034)
    out = jitted(cond, a, b)

    print(thunder.last_traces(jitted)[0])
```

If tensors are on CUDA, then the trace is
```python
def computation(cond, a, b):
  # cond: "cuda:0 b8[]"
  # a: "cuda:0 f32[4, 4]"
  # b: "cuda:0 f32[]"

  # repro.py:11:             return torch.where(cond, a, b)
  t2 = ltorch.where(cond, a, b)  # t2: "cuda:0 f32[4, 4]"
    # t0 = prims.broadcast_in_dim(cond, (4, 4), ())  # t0: "cuda:0 b8[4, 4]"
    # t1 = prims.broadcast_in_dim(b, (4, 4), ())  # t1: "cuda:0 f32[4, 4]"
    # t2 = prims.where(t0, a, t1)  # t2: "cuda:0 f32[4, 4]"
  return (t2,)
```
The difference between traces of CPU and CUDA would be attributed to https://github.com/Lightning-AI/lightning-thunder/blob/d0f647479f73787389c4fa6ced3b0b9ada4083c0/thunder/clang/__init__.py#L1410-L1417.

Rel: 
- #2069
- #2187

### DIFF
--- a/thunder/core/prims.py
+++ b/thunder/core/prims.py
@@ -2831,9 +2831,19 @@ def _where_meta(pred: Number | TensorProxy, a: Number | TensorProxy, b: Number |
 
     # Determines output shape
     # NOTE Assumes at least one of pred, a, and b is a TensorProxy because of prior check for Number x Number x Number
-    # `pred` can be a CPU bool scalar tensor, thus to filter it out by checking the numel.
-    shapes = tuple(x.shape for x in (pred, a, b) if isinstance(x, TensorProxy) and x.numel > 1)
-    resultshape = shapes[0]
+    shapes = tuple(x.shape for x in (pred, a, b) if isinstance(x, TensorProxy))
+    # It's possible that `pred` is a CPU bool scalar tensor and either or both of `a` and `b` are a CUDA tensor.
+    # In that case, `shapes[0]`, i.e., `pred.shape` should not be the result shape.
+    if (
+        len(shapes) > 1
+        and isinstance(pred, TensorProxy)
+        and pred.numel == 1
+        and pred.device.devicetype is devices.DeviceType.CPU
+        and (isinstance(a, TensorProxy) or isinstance(b, TensorProxy))
+    ):
+        resultshape = shapes[1]
+    else:
+        resultshape = shapes[0]
 
     return TensorProxy(shape=resultshape, device=resultdevice, dtype=dtype)
 
--- a/thunder/core/prims.py
+++ b/thunder/core/prims.py
@@ -2831,7 +2831,8 @@ def _where_meta(pred: Number | TensorProxy, a: Number | TensorProxy, b: Number |
 
     # Determines output shape
     # NOTE Assumes at least one of pred, a, and b is a TensorProxy because of prior check for Number x Number x Number
-    shapes = tuple(x.shape for x in (pred, a, b) if isinstance(x, TensorProxy))
+    # `pred` can be a CPU bool scalar tensor, thus to filter it out by checking the numel.
+    shapes = tuple(x.shape for x in (pred, a, b) if isinstance(x, TensorProxy) and x.numel > 1)
     resultshape = shapes[0]
 
     return TensorProxy(shape=resultshape, device=resultdevice, dtype=dtype)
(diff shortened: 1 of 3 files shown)


t-vi:
> In order to fix the shape mismatch between runtime and trace, this PR clones compute_broadcast_shape of thunder.clang in thunder.core.prims and call prims.compute_broadcast_shape in the existing compute_broadcast_shape.

I don't think broadcasting in prims is the right thing to do. Why would we not revise the clang op to insert the missing broadcast ops?

thunder/core/prims.py
```diff
@@ -2832,7 +2832,18 @@ def _where_meta(pred: Number | TensorProxy, a: Number | TensorProxy, b: Number |
     # Determines output shape
     # NOTE Assumes at least one of pred, a, and b is a TensorProxy because of prior check for Number x Number x Number
     shapes = tuple(x.shape for x in (pred, a, b) if isinstance(x, TensorProxy))
-    resultshape = shapes[0]
+    # It's possible that `pred` is a CPU bool scalar tensor and either or both of `a` and `b` are a CUDA tensor.
+    # In that case, `shapes[0]`, i.e., `pred.shape` should not be the result shape.
+    if (
```
mruberry, line 2837: This is still not correct. For example:

```
cond =  torch.tensor(True)
a = torch.tensor(1)
b = torch.randn((2, 2))
torch.where(cond, a, b)  # tensor([[1., 1.], [1., 1.]])
```

the shape would be the third shape, not the second.

The same thing happens on CUDA devices. The tensors broadcast. Although on CUDA devices the tensors will already have been broadcast previously before the primitive.

What you can do is take this line from the `check_same_shape` function:

```
non_scalar_shapes = tuple(x.shape for x in args if isinstance(x, TensorProxy) and not is_cpu_scalar_tensor(x))
```

Then:

```
if len(non_scalar_shapes) > 0:
  resultshape = non_scalar_shapes[0]
else:
  resultshape = shapes[0]
```

And I believe that will work. We should add a test for all the inputs being "CPU number tensors," and then each of cond, a, or b being tensors that aren't CPU numbers, and verifying the shape is the same.

crcrpar: I think my initial approach was nice.

IvanYashchuk (approved):

thunder/tests/opinfos.py
```diff
@@ -3736,6 +3721,10 @@ def diagonal_sample_generator(op, device, dtype, requires_grad, **kwargs):
         # input case ((1, 2, 0, 3), -1, 0, -1)
         DecorateInfo(pytest.mark.xfail(strict=True), "test_vjp_correctness"),
         # See: [Fix runtime-trace shape/dtype/device mismatch]
+        # In https://github.com/Lightning-AI/lightning-thunder/pull/2069,
```
mruberry, line 3724: Nice comment

mruberry (approved):

PR #2077, opened 2025-05-15
Reduces the test time

kiya00:
<details>
  <summary><b>Before submitting</b></summary>

- [ ] Was this discussed/approved via a Github issue? (no need for typos and docs improvements)
- [ ] Did you read the [contributor guideline](https://github.com/Lightning-AI/pytorch-lightning/blob/main/.github/CONTRIBUTING.md), Pull Request section?
- [ ] Did you make sure to update the docs?
- [ ] Did you write any new necessary tests?

</details>

## What does this PR do?

Fixes #2051 .

PR #2142, opened 2025-05-27
Enable ruff format in pre-commit

crcrpar:
## What does this PR do?

Enabling `ruff` in pre-commit.
Currently only black is used but https://github.com/Lightning-AI/lightning-thunder/blob/d0f647479f73787389c4fa6ced3b0b9ada4083c0/thunder/dynamo/utils.py#L850 uses `ruff` if it's available.

cc @borda

### DIFF
--- a/thunder/benchmarks/__init__.py
+++ b/thunder/benchmarks/__init__.py
@@ -617,9 +617,9 @@ def run_multiprocess_benchmark(
     print(f"Running distributed benchmark {benchmark.name} with {world_size=}")
     _print_benchmark_arguments(benchmark)
 
-    assert (
-        torch.distributed.is_available()
-    ), "Trying to run a distributed benchmark, but torch.distributed is not available"
+    assert torch.distributed.is_available(), (
+        "Trying to run a distributed benchmark, but torch.distributed is not available"
+    )
 
     # Ensures the benchmark is running on a single CUDA device (which is overridden later)
     assert (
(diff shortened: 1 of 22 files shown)


t-vi:
No, let's stay with black please.

t-vi:
> Why do you prefer to use black? Ruff is a drop-in replacement but faster.

I'm all for material performance gains. But black vs. ruff saves a fraction of the 35 seconds all tests in precommit take, usually run in parallel to other stuff. It's just not worth changing. Things like #2077 and #1874 and if we can disable tests #2144 will be much more impactful. I'm sure there is no shortage of other opportunities.

t-vi (approved):
Let's do it for the better good.

.pre-commit-config.yaml
```diff
@@ -38,12 +38,13 @@ repos:
...
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.11.11
     hooks:
-      - id: black
-        name: Black code
-        exclude: "examples"
+      - id: ruff-check
+        args: ["--fix"]
+      - id: ruff-format
+        types_or: [python, pyi]
```
Borda: 'Let''s split the PR to adding Ruff as flake8 and formatting, so the change is easier to trace/review

crcrpar: This one is for `ruff check` -- https://github.com/Lightning-AI/lightning-thunder/pull/2192

Borda: cool, so making this as a draft until #2192 lands :)

crcrpar:
rebased onto 2192

Borda:
> rebased onto #2192

maybe need some more resolving 🐰

Borda (approved):

PR #2143, opened 2025-05-27
Fix autodiff joint trace dataflow and in-place ops in higher order functions

### LINKED ISSUE #1961: KeyError when running HF `microsoft/GRIN-MoE` model
## 🐛 Bug

I've managed to narrow down the error to this repro by bisecting the model:

```python
import torch

class mp(torch.autograd.Function):
    @staticmethod
    def forward(
        ctx,
        scores,
        multiplier: torch.Tensor,
        selected_experts: torch.Tensor,
        masked_gates: torch.Tensor,
        mask_for_one: torch.Tensor,
    ):
        ctx.save_for_backward(multiplier, selected_experts, masked_gates)
        return multiplier * mask_for_one

    @staticmethod
    def backward(
        ctx,
        grad_at_output: torch.Tensor,
    ):
        _, selected_experts, masked_gates = ctx.saved_tensors

        masked_gates.scatter_add_(
            dim=-1,
            index=selected_experts,
            src=grad_at_output,
        )

        return (
            masked_gates,
            None,
            None,
            None,
            None,
        )

def sparsemixer(scores, multiplier_o, selected_experts, masked_gates, mask_for_one):
    return mp.apply(
        scores,
        multiplier_o,
        selected_experts,
        masked_gates,
        mask_for_one,
    )

scores = torch.randn(128, 16, device="cuda", dtype=torch.bfloat16, requires_grad=True)
multiplier_o = torch.ones(128, 1, device="cuda", dtype=torch.bfloat16, requires_grad=True)
selected_experts = torch.ones(128, 1, device="cuda", dtype=torch.int64)
masked_gates = torch.ones(128, 16, device="cuda", dtype=torch.bfloat16, requires_grad=True)
mask_for_one = torch.ones(128, 1, device="cuda", dtype=torch.bfloat16, requires_grad=False)

jfoo = thunderfx(sparsemixer)
nexp_out = jfoo(scores, multiplier_o, selected_experts, masked_gates, mask_for_one)
```

However, I've hit the same error while working on #1908. It seems to be related to one of the passes that are done when the inputs requires_gradient is true, so it does not happen only when specifying the higher order function like the repro.

riccardofelluga:
## What does this PR do?

Fixes #1961 

In this PR there are two main changes that work together, the first one is the addition of `insert_alias_updates` in the higher order function lookasides that would otherwise have skipped it. 

And the second change is about putting the output of the grad transforms to as output of the join trace. I found that to help with the dataflow because now there is a path from the output up to the put_grad and get_grad. This helps in the subsequent call dce that otherwise can remove symbols that appears after get_grad.

Notes for reviewers:
There has been an error and #2147 has been merged into main instead that on this branch, please feel free to leave the reviews on #2147 instead that here.

- @crcrpar I've put the insert alias updated in the lookasides, however it might also be moved inside the `_convert_pytorchfunc_to_thundertrace` but I am not sure of it, let me know what you think if maybe it should stay out like it is now.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,7 +24,7 @@ repos:
       - id: detect-private-key
 
   - repo: https://github.com/asottile/pyupgrade
-    rev: v3.19.1
+    rev: v3.20.0
     hooks:
       - id: pyupgrade
         args: ["--py310-plus"]
--- a/thunder/core/jit_ext.py
+++ b/thunder/core/jit_ext.py
@@ -948,12 +948,17 @@ def _general_jit_torch_ops_higher_order_autograd_function_apply(fwd, bwd, *fwd_a
     with tracectx(trace_of_forward):
         prims.python_return(*(sequencify(output)))
 
+    from thunder.core.update_aliases import insert_alias_updates
+
+    alias_tensor_indices = [[i] for i in range(len(trace_of_forward.args))]
+    aliased_trace_of_forward = insert_alias_updates(trace_of_forward, alias_tensor_indices)
+
     # See NOTE: `autograd_function_apply` and `no_grad` interaction for details about
     # `thunder.torch.call_higher_order_function_and_consider_outer_autograd_setting`
     @wraps(aug_fwd_trace.python_callable())
     @thunder.torch.call_higher_order_function_and_consider_outer_autograd_setting
     def forward(*args, **kwargs):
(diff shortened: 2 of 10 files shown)


beverlylytle (approved):
FYI: Since #2147 is already merged, the `import thunder` statement is now removed from autodiff, meaning there will be an import error if `not torch_autograd`.  So we better get this merged.

Issue #2060, opened 2025-05-13
TE Integration fails with TE main


kshitij12345:
Running the tests with latest TE main, `pytest thunder/tests/test_transformer_engine_executor.py` results in
```python
      t26 = Tensor.to(x, torch.bfloat16, copy=True)  # t26: "cuda:0 bf16[32, 32]"
      t28 = Tensor.to(w, torch.bfloat16, copy=True)  # t28: "cuda:0 bf16[32, 32]"
>     (t6, (t12, t13, t14, t15, t16, t17, t18, t19), ctx_te_2311) = te_linear_22(t26, t28, None)
E     ValueError: not enough values to unpack (expected 8, got 6)

thunder.computation_37:17: ValueError
```

NOTE: This can't be reproduced on `pjnl-latest` as it seems to be pinned to TE v2.3

IvanYashchuk:
The problem is either in `_get_num_saved_tensors` or `get_recipe_from_options_or_default_recipe`:
https://github.com/Lightning-AI/lightning-thunder/blob/d0f647479f73787389c4fa6ced3b0b9ada4083c0/thunder/executors/transformer_engineex.py#L337

kshitij12345:
Fixed in #2146

PR #2146, opened 2025-05-27
Set input_quantizer.internal to True

beverlylytle:
Addresses #2060.

I determined that https://github.com/NVIDIA/TransformerEngine/pull/1638 added the commit to transformer engine that breaks these tests.   While this change makes the tests pass on my machine, I don't fully understand why.  The result is that more tensors are saved, but, if I read [this docstring](https://github.com/NVIDIA/TransformerEngine/blob/cd37379d24f574f98e74e81d242c1419f1159e6e/transformer_engine/pytorch/tensor/quantized_tensor.py#L74) correctly, the extra saved tensors are internal objects for Transformer Engine.

### DIFF
--- a/thunder/core/jit_ext.py
+++ b/thunder/core/jit_ext.py
@@ -824,6 +824,11 @@ def _general_jit_torch_autograd_function_apply_lookaside(obj: Any, *args, **kwar
     trace_of_augmented_fwd._siginfo = SigInfo.from_name_and_args(custom_fwd_sym.name, unwrapped_custom_forward_args)
     trace_of_augmented_fwd.args = unwrapped_custom_forward_args
 
+    from thunder.core.update_aliases import insert_alias_updates
+
+    alias_tensor_indices = [[i] for i in range(len(trace_of_augmented_fwd.args))]
+    aliased_trace_of_augmented_fwd = insert_alias_updates(trace_of_augmented_fwd, alias_tensor_indices)
+
     # Backward definition
     custom_backward = custom_autograd_function_cls.backward
     grads = tree_map(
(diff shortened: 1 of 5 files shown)


thunder/executors/transformer_engineex.py
```diff
@@ -219,6 +219,8 @@ def forward(self, inp, weight, bias, is_grad_enabled: bool = False):
                 grad_input_quantizer,
             ) = self._get_quantizers(is_grad_enabled)
 
+            input_quantizer.internal = True
```
kshitij12345, line 222: It would be better to update this line

https://github.com/Lightning-AI/lightning-thunder/blob/1ef2b946f25400ed32df83674c41f9ad0b3250ad/thunder/executors/transformer_engineex.py#L293

beverlylytle: oh, yes of course.

t-vi (approved):
Stamping Thank you @beverlylytle @kshitij12345

thunder/executors/transformer_engineex.py
```diff
@@ -290,7 +290,7 @@ def _get_quantizers(self, is_grad_enabled):
         grad_output_quantizer = None
         output_quantizer = None
         input_quantizer = self.quantizers["scaling_fwd"][tex.FP8FwdTensors.GEMM1_INPUT]
-        input_quantizer.internal = False
+        input_quantizer.internal = True
```
beverlylytle, line 293: Leaving an unresolved comment so that this isn't merged before I check this on the various devices.

timmoon10 (approved):

PR #2163, opened 2025-06-02
Use `target` consistently in `thunder.dynamo.utils.is_node_supported_by_thunder` rather than `node.target`

crcrpar:
## What does this PR do?

We seem to have defined the variable of `target` for `node.target` so it'd be reasonable to consistently use `target` instead of `node.target`


crcrpar:
no, it's updated in https://github.com/Lightning-AI/lightning-thunder/blob/97b6a9d9281038f2b3d4fade3f334cff6d8ddd1b/thunder/dynamo/utils.py#L394-L395

PR #2172, opened 2025-06-02
Return tuple of 3 `None`s when `try_and_log_benchmark` fails

crcrpar:
## What does this PR do?

For the sake of parity, `try_and_log_benchmark` returns a tuple of three `None`s and fix the grammar in the following `check`.


crcrpar:
I'm not sure if and where to put the test for this

t-vi (approved):
Thank you @crcrpar @kiya00

Issue #2149, opened 2025-05-29
don't register cudnn executor if cudnn is not available


t-vi:
...as other executors (e.g. nvfuser) do.

This would make it easier to fail if the executor is not available.

@KaelanDt

PR #2174, opened 2025-06-02
Only register cudnn executor if it is available

KaelanDt:
<details>

</details>

## What does this PR do?

Fixes https://github.com/Lightning-AI/lightning-thunder/issues/2149. Also removes cudnn from the BaseRecipe, as it is not always there and is not a core component of thunder (less so than nvfuser). 

## PR review

## Did you have fun?


thunder/recipes/base.py
```diff
@@ -59,7 +59,7 @@ def __init__(
         plugins=None,
     ):
         super().__init__(interpreter=interpreter, plugins=plugins)
-        self.executor_names = ["cudnn", "sdpa", "torchcompile_xentropy"]
+        self.executor_names = ["sdpa", "torchcompile_xentropy"]
         self.fuser = fuser
```
t-vi: I think we might want to keep cudnn required by default (and see that it is installed if we don't have it currently).

thunder/executors/cudnnex.py
```diff
@@ -810,10 +792,27 @@ def _cudnn_sdpa_bwd_wrapper(
...
+        meta=_cudnn_sdpa_bwd_meta,
+        fn=_cudnn_sdpa_bwd_impl,
+    )
+    # Registers the implementation for torch.nn.functional.scaled_dot_product_attention
+    cudnn_ex.register_implementation(
+        ltorch.scaled_dot_product_attention,
+        checker=_cudnn_sdpa_checker,
+        execution_transform=_cudnn_sdpa_fwd_wrapper,
+        grad_transform=_cudnn_sdpa_bwd_wrapper,
+    )
```
riccardofelluga: Being consistent with what's done with nvfuser wouldn't be better to have the same system and return None if cudnn is not available?  
 -> https://github.com/Lightning-AI/lightning-thunder/blob/main/thunder/executors/nvfuserex.py

t-vi:
@riccardofelluga do the updates make you happier about the PR?

t-vi (approved):
Looks good to me, let's give @riccardofelluga @vedaanta a chance to look at it before merging it Wed or Thu.

KaelanDt:
> Since we are modifying two files at this point wouldn't it be better to follow the same style as the other executors that are dependent on another library? Where we dedicate one file to be the gate for the symbols depending on the availability of the library(as in nvfuserex.py)
> 
> So something like cudnnex.py that takes care of the imports and checks and then cudnn_layernorm as is, and maybe creating cudnn_sdpa.py
> 
> @KaelanDt What do you think?

yes, i like that

KaelanDt:
@riccardofelluga I tried a few things by adding a `cudnn_sdpa` file, but it just ended up being messy and relying on a bunch of `cudnn_available` checks and was more confusing in my opinion.

Since the `cudnn_sdpa` operators are registered along with the main cudnn executor, i think it is fine to leave it as is, and to have `cudnn_layernorm` as a separate file (because it creates a new executor)

vedaanta:
the change looks good to me.

@KaelanDt @t-vi I think we should not register cudnn layernorm at all.
It was being bought up a year ago, but there were cpu overheads which made it perform worse than triton/nvfuser.
(Since then cudnn has worked a lot on norms, but their e2e contribution for models I know of has been low enough to not warrant enabling in thunder again.)

vedaanta (approved):

t-vi:
Thank you @vedaanta !

riccardofelluga:
> @riccardofelluga I tried a few things by adding a `cudnn_sdpa` file, but it just ended up being messy and relying on a bunch of `cudnn_available` checks and was more confusing in my opinion.
> 
> Since the `cudnn_sdpa` operators are registered along with the main cudnn executor, i think it is fine to leave it as is, and to have `cudnn_layernorm` as a separate file (because it creates a new executor)

@KaelanDt  Let me come in help here, instead of modifying the cudnn executor and doing a conditional registration of the executor, wouldn't it be better to create a *conditional import* like in nvfuserex.py? 

Would something like this in `cudnnex.py` work?
```python

__all__ = ["cudnn_ex"]

cudnn_ex = None

if cudnn_available():
    import import thunder.executors.cudnn_spda as sdpa_impl
    cudnn_ex = sdpa_impl.cudnn_ex

else:
    warnings.warn("cudnn not available")

``` 

We can also make room for more complex import logic if needed (like say checking versions or disabling cudnn_linearnorm), all without modifying the cudnn executor registration.

thunder/tests/test_recipes.py
```diff
@@ -11,9 +11,12 @@
 from torch.testing import assert_close
 from thunder.recipes import HFTransformers
 from thunder.executors import nvfuser_available
+from thunder.executors.cudnnex import cudnn_available
+from thunder.tests.framework import version_between, IS_WINDOWS
 from thunder.tests.framework import IS_WINDOWS
 
 
+@pytest.mark.skipif(not cudnn_available(), reason="cuDNN is not available")
```
IvanYashchuk, line 19: Is there a plan to implement a CI job without cuDNN support to test these skips?

t-vi: No. These skips are for the convenience of people not having (a compatible version of) cudnn installed  on the development machine.

thunder/executors/cudnnex.py
```diff
@@ -42,783 +37,51 @@ def required_cudnn_version() -> LooseVersion:
...
-        dropout_p,
-        is_causal,
-        query.stride(),
-        key.stride(),
-        value.stride(),
-        grad_query.stride(),
-        grad_key.stride(),
-        grad_value.stride(),
+    cudnn_ex = OperatorExecutor("cudnn", version=cudnn.backend_version())
+    register_executor(cudnn_ex)
```
riccardofelluga: Almost there! We don't actually need to register the executor here, we can keep the registration in the sdpa file and then just export the symbols from there.

```suggestion
    import thunder.executors.cudnn_sdpa as sdpa_impl

    cudnn_ex = sdpa_impl.cudnn_ex
```

thunder/executors/cudnnex.py
```diff
@@ -42,783 +37,51 @@ def required_cudnn_version() -> LooseVersion:
...
-        seed,
-        offset,
-        scale=scale,
-        cat_grad_qkv=cat_grad_qkv,
+    cudnn_ex.register_implementation(
+        sdpa_impl.ltorch.scaled_dot_product_attention,
+        checker=sdpa_impl._cudnn_sdpa_checker,
+        execution_transform=sdpa_impl._cudnn_sdpa_fwd_wrapper,
+        grad_transform=sdpa_impl._cudnn_sdpa_bwd_wrapper,
     )
```
riccardofelluga: Do we need to register this symbols here? I'd rader change the import in the test suite
```suggestion

```

thunder/executors/cudnnex.py
```diff
@@ -42,783 +37,51 @@ def required_cudnn_version() -> LooseVersion:
...
-        is_causal,
-        query.stride(),
-        key.stride(),
-        value.stride(),
-        grad_query.stride(),
-        grad_key.stride(),
-        grad_value.stride(),
+    cudnn_ex = OperatorExecutor("cudnn", version=cudnn.backend_version())
+    register_executor(cudnn_ex)
+    cudnn_sdpa_fwd = cudnn_ex.register_operator(
```
riccardofelluga: What we might need here is to export `torch_to_cudnn_dtype` and do something like this:
```python
__all__ = [..., "torch_to_cudnn_dtype"]

[... the rest of the code ...]

torch_to_cudnn_dtype = sdpa_impl.torch_to_cudnn_dtype
```

thunder/executors/cudnnex.py
```diff
@@ -42,783 +37,51 @@ def required_cudnn_version() -> LooseVersion:
...
-            dK,
-            dV,
-            dBias,
-            graph,
-        )
-    return _cudnnex_cache[cache_key]
-
+cudnn_ex: None | OperatorExecutor = None
+cudnn_sdpa_fwd: Any = None
+cudnn_sdpa_bwd: Any = None
```
riccardofelluga:
```suggestion
from typying import Any, Callable

torch_to_cudnn_dtype: None | Callable
```

riccardofelluga (changes requested):
Almost there!! Thanks for taking the time, I've added some suggestions. Please feel free to ping me if you need any help

thunder/executors/cudnnex.py
```diff
@@ -42,782 +37,24 @@ def required_cudnn_version() -> LooseVersion:
...
-    *,
-    scale: FloatProxy | float | None = None,
-) -> bool:
-    # TODO(#58): make the checker more conservative.
-    if cudnn is None:
-        return False
-
-    if query.device.type != "cuda" or key.device != query.device or value.device != query.device:
+    if v is None:
         return False
```
riccardofelluga, line 41: Nit. a nice warning that cudnn is not available would fit great here

thunder/executors/cudnnex.py
```diff
@@ -42,782 +37,24 @@ def required_cudnn_version() -> LooseVersion:
...
-        (grad_qkv,) = grads
-        grad_query, grad_key, grad_value = grad_qkv.split([query.size(1), key.size(1), value.size(1)], dim=1)
-    else:
-        grad_query, grad_key, grad_value = grads
-    put_grads((query, key, value), (grad_query, grad_key, grad_value))
-
-    return primal
+cudnn_ex: None | OperatorExecutor = None
+torch_to_cudnn_dtype: None | Callable = None
+cudnn = None
```
riccardofelluga, line 54: Not needed anymore ;)
```suggestion

```

thunder/executors/cudnn_sdpa.py
```diff
@@ -0,0 +1,759 @@
...
+    if cat_grad_qkv:
+        grads = (grad_qkv,)
+    else:
+        grads = (grad_query, grad_key, grad_value)
+
+    if attn_mask is not None:
+        grads = grads + (grad_attn_mask,)
+    return grads
+
+
```
riccardofelluga, line 687: Let's keep the registration here because the symbol is needed later on


```suggestion
cudnn_sdpa_bwd = cudnn_ex.register_operator(
    "cudnn_sdpa_bwd",
    meta=_cudnn_sdpa_bwd_meta,
    fn=_cudnn_sdpa_bwd_impl,
)


```

thunder/executors/cudnn_sdpa.py
```diff
@@ -0,0 +1,759 @@
...
+    except cudnn.cudnnGraphNotSupportedError:
+        return False
+    # Otherwise just raise the error.
+    # These errors can be due to internal cudnn bugs, or user error.
+    except Exception:
+        raise
+
+    return True
+
+
```
riccardofelluga, line 389: Also here the registration for the fwd should be here

```suggestion
cudnn_sdpa_fwd = cudnn_ex.register_operator(
    "cudnn_sdpa_fwd",
    meta=_cudnn_sdpa_forward_meta,
    fn=_cudnn_sdpa_fwd_impl,
    tags=(OpTags.DONT_AUTO_RECOMPUTE_IN_BACKWARD,),
)


```

thunder/executors/cudnn_sdpa.py
```diff
@@ -0,0 +1,759 @@
...
+)
+
+cudnn_sdpa_bwd = cudnn_ex.register_operator(
+    "cudnn_sdpa_bwd",
+    meta=_cudnn_sdpa_bwd_meta,
+    fn=_cudnn_sdpa_bwd_impl,
+)
+
+
+cudnn_ex.register_implementation(
```
riccardofelluga: Let's restore comment from old cudnnex.py


```suggestion
# Registers the implementation for torch.nn.functional.scaled_dot_product_attention
cudnn_ex.register_implementation(
```

thunder/executors/cudnn_sdpa.py
```diff
@@ -0,0 +1,759 @@
...
+from thunder.core.compile_data import get_compile_option
+from thunder.core.prims import OpTags
+from thunder.core.transforms import get_grad, put_grad, put_grads
+import thunder.core.dtypes as dtypes
+import thunder.torch as ltorch
+from thunder.core.proxies import pyval
+
+from thunder.extend import OperatorExecutor, register_executor
+
+
```
riccardofelluga, line 23: And here before doing anything let's register the executor and import cudnn for the whole file

```suggestion

import cudnn

cudnn_backend_version = cudnn.backend_version()

cudnn_ex: OperatorExecutor = OperatorExecutor("cudnn", version=cudnn_backend_version)
register_executor(cudnn_ex)


# Cache already built cudnn graphs to save on runtime compilation overhead
```

thunder/executors/cudnn_sdpa.py
```diff
@@ -0,0 +1,759 @@
...
+    import cudnn
+
+    handle = _device_to_cudnn_handle.get(query_device, None)
+    if handle is None:
+        with torch.cuda.device(query_device):
+            handle = cudnn.create_handle()
+            _device_to_cudnn_handle[query_device] = handle
+    cudnn.set_stream(handle=handle, stream=torch.cuda.current_stream(device=query_device).cuda_stream)
+    return handle
+
```
riccardofelluga: Let's keep the comments original from the old file:

```suggestion
# Mapping from device to cudnn handles
_device_to_cudnn_handle = {}


# This function creates a new handle for the device that cudnn should
# run its kernels on. As the suggested approach by cudnn is to make a few handles
# as possible, this function caches these per-device handles.
def _get_cudnn_handle(query_device):
    handle = _device_to_cudnn_handle.get(query_device, None)
    if handle is None:
        with torch.cuda.device(query_device):
            handle = cudnn.create_handle()
            _device_to_cudnn_handle[query_device] = handle

    # Make sure the user stream is set on the handle
    # Fetch the current user stream and pass the data pointer to set_stream API
    cudnn.set_stream(handle=handle, stream=torch.cuda.current_stream(device=query_device).cuda_stream)

    return handle

```

thunder/executors/cudnn_sdpa.py
```diff
@@ -0,0 +1,759 @@
...
+
+def _make_cudnn_sdpa_forward_graph(
+    query, key, value, attn_mask, dropout_p, is_causal, query_stride, key_stride, value_stride
+):
+    import cudnn
```
riccardofelluga: No need for these imports if we import it for the whole file :)

riccardofelluga (approved):
Thanks for the changes, there are some nits but it looks good already!

Well done!

thunder/executors/cudnn_sdpa.py
KaelanDt: yes for this one, the one for torch_to_cudnn_dtype is still needed for cudnn_layernom

thunder/executors/cudnnex.py
KaelanDt: This is not what we do for nvfuser (the executor is just not registered)

PR #1907, opened 2025-03-26
[WIP] Representing DTensor in thunder traces

kshitij12345:
TODO:

1. Add description of the changes
2. Design Doc
3. Sample Generated Traces

PR #2177, opened 2025-06-03
[DO NOT MERGE] WIP DTensor NVFuser Integration

kshitij12345:
Builds on #1907 

TODO:
* Add test


thunder/executors/nvfuserex_impl.py
```diff
@@ -237,6 +244,50 @@ def get_translator(bsym: BoundSymbol) -> Callable:
...
+                return t
+        return None
+
+    def multidevice_schedule(self) -> None:
+        for in_tensor_index, in_dtensor in zip(self.inputs(), in_dtensors):
+            in_tensor = self._find_tensor_by_index(in_tensor_index)
+
+            # Set the device mesh.
+            assert in_dtensor.spec._o.device_mesh.ndim == 1, "nvFuser's Python API only supports 1D meshes."
+            mesh = nvfuser.DeviceMesh(in_dtensor.spec._o.device_mesh.mesh.tolist())
```
kshitij12345: - Add property `device_mesh` to DTensorProxy

thunder/executors/nvfuserex_impl.py
```diff
@@ -237,6 +244,50 @@ def get_translator(bsym: BoundSymbol) -> Callable:
...
+
+            self.sched._set_device_mesh(in_tensor, mesh)
+
+            # Split and parallelize.
+            assert len(in_dtensor.spec._o.placements) == 1, "Expect a 1D mesh"
```
kshitij12345: - Add property `placements` to DTensorProxy

thunder/executors/nvfuserex_impl.py
```diff
@@ -239,6 +246,49 @@ def get_translator(bsym: BoundSymbol) -> Callable:
...
+            if t.index == index:
+                return t
+        return None
+
+    def multidevice_schedule(self) -> None:
+        for in_tensor_index, in_dtensor in zip(self.inputs(), in_dtensors):
+            in_tensor = self._find_tensor_by_index(in_tensor_index)
+
+            # Set the device mesh.
+            assert in_dtensor.device_mesh.ndim == 1, "nvFuser's Python API only supports 1D meshes."
```
IvanYashchuk: What happens if this kind of mesh is passed to nvFuser? It's better to just pass through and let nvFuser raise any errors.
Asserts should be used for something unexpected, not unsupported, because it's possible to skip all asserts in the `-o` execution of Python.

kshitij12345: Makes sense thanks! Without the assert, we will see an error like this
```python
[rank0]: Traceback (most recent call last):
[rank0]:   File "/opt/pytorch/lightning-thunder/test_nvfuser_eg.py", line 155, in <module>
[rank0]:     result = fd([in_dtensor, weight])
[rank0]:              ^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/opt/pytorch/lightning-thunder/test_nvfuser_eg.py", line 124, in __call__
[rank0]:     out_tensors, out_shardings = fusion_def.execute(in_tensors)
[rank0]:                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/opt/pytorch/nvfuser/python/nvfuser/__init__.py", line 311, in execute
[rank0]:     self.multidevice_schedule()
[rank0]:   File "/opt/pytorch/lightning-thunder/test_nvfuser_eg.py", line 83, in multidevice_schedule
[rank0]:     mesh = nvfuser.DeviceMesh(in_dtensor.device_mesh.mesh.tolist())
[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]: TypeError: __init__(): incompatible constructor arguments. The following argument types are supported:
[rank0]:     1. nvfuser._C.DeviceMesh(arg0: list[int])

[rank0]: Invoked with: [[0], [1]]
```

@wujingyue can this error message be improved to clarify that this is not supported. Also curious if nvFuser have supporting multi-dim mesh on roadmap?

wujingyue: > Also curious if nvFuser have supporting multi-dim mesh on roadmap?

nvFuser's C++ API accepts multi-dimensional device meshes: https://github.com/NVIDIA/Fuser/blob/9fbfa14141603ab908791a8af0f074e228d95cbe/csrc/multidevice/device_mesh.h#L36-L38. However, I haven't exposed it to https://github.com/NVIDIA/Fuser/blob/9fbfa14141603ab908791a8af0f074e228d95cbe/python/python_frontend/multidevice_bindings.cpp#L55. Also, that multi-dim mesh support is merely within the DeviceMesh representation. For 2D sharding, I'm sure many other things need to be extended, e.g., sharding propagation and schedulers. 

I don't have any plan for 2D sharding in Q3. The closest thing that requires 2D sharding (and thus 2D mesh) is combining DP and TP. At some point, I was hoping Thunder takes care of DP and further shards a DP'ed trace using nvFuser/TP so I can postpone 2D sharding in nvFuser. However, recent architectures like DeepSeek and Llama4 use DP+TP for attention and replace DP with EP for MoE. I can't think of a way to decompose such sharding between Thunder and nvFuser, making a stronger case for nvFuser to pick up both DP and TP.

wujingyue: > can this error message be improved to clarify that this is not supported

Can you suggest a way? The constructor at https://github.com/NVIDIA/Fuser/blob/9fbfa14141603ab908791a8af0f074e228d95cbe/python/python_frontend/multidevice_bindings.cpp#L55 takes a list of ints. However, your example passes in a list of lists.

thunder/executors/nvfuserex_impl.py
```diff
@@ -325,7 +374,33 @@ def translate_bound_symbol(bsym: BoundSymbol) -> Any:
             nvout = lc_to_nv_map[out]
             fd.add_output(nvout)
 
-    return fd
+    bind(fd, definition, "definition")
```
wujingyue: Instead of binding, can we do the following? 

```
def define_math(fd):
  ...

if any input is DTensor:
  class Model(FusionDefinition):
    def __init__(self):
      super().__init__(max_length=...)

    def definition(self):
      define_math(self)

    def multidevice_schedule(self):
      ...

  return Model()
else:
  fd = FusionDefinition(max_length=...)
  ...
  define_math(fd)
  return fd
```

This is essentially what I wrote at https://github.com/NVIDIA/Fuser/blob/9fbfa14141603ab908791a8af0f074e228d95cbe/tests/python/multidevice/fusion_definition_wrapper.py#L38

thunder/executors/nvfuserex_impl.py
```diff
@@ -325,7 +374,33 @@ def translate_bound_symbol(bsym: BoundSymbol) -> Any:
...
+            "nvfuser: Expected runtime and tracing metadata to be the same for DTensor."
+        )
+
+        _find_tensor_by_index, multidevice_schedule = get_methods_for_dtensor_fd(sorted_unique_inputs)
+
+        bind(fd, multidevice_schedule, "multidevice_schedule")
+        bind(fd, _find_tensor_by_index, "_find_tensor_by_index")
+        is_multigpu_fd = True
+
+    return fd, is_multigpu_fd
```
wujingyue:
```suggestion
    return fd
```
is_multigpu_fd is essentially hasattr(fd, "multidevice_schedule"). It doesn't need to be returned

thunder/executors/nvfuserex_impl.py
wujingyue: There's nothing fundamentally wrong with `bind`. It just feels monkey-patching and is hard to maintain. In general, I prefer creating the right FusionDefinition in one pass instead of 

```
fd = FusionDefinition(...)
bind an attribute
bind more attributes
return fd
```

thunder/executors/nvfuserex_impl.py
kshitij12345: Regarding the error, the original class can be subclassed to handle checking the mesh dimensions.

Something like this -
```python
class NVFuserDeviceMesh(nvfuser.DeviceMesh):  # Subclass the C++ bound class
    def __init__(self, mesh: DeviceMesh):
        if mesh.ndim > 1:
            raise ValueError("nvFuser's Python API only supports 1D meshes.")
        super().__init__(mesh.tolist())
```

thunder/executors/nvfuserex_impl.py
```diff
@@ -575,7 +680,7 @@ def create_fusion_definition_wrapper(
     # TODO (mruberry) We should think how to express "static fusion" that don't need to use
     #   a cache to improve dispatch performance
     @lru_cache(maxsize=2048)
-    def get_fd(input_descriptors) -> FusionDefinition:
+    def get_fd(input_descriptors) -> tuple[FusionDefinition, bool]:
```
IvanYashchuk:
```suggestion
    def get_fd(input_descriptors) -> FusionDefinition:
```

thunder/executors/nvfuserex_impl.py
```diff
@@ -412,6 +482,14 @@ def compute_contiguity(
     return tuple(tuple(x) for x in nv_compute_td(shape, stride))
 
 
+def make_key_from_dtensor(dtensor: DTensor) -> tuple[str]:
```
IvanYashchuk: According to the current type hint, the else branch is never taken.
```suggestion
def make_key_from_dtensor(dtensor: DTensor | Any) -> tuple[str]:
```

thunder/executors/nvfuserex_impl.py
```diff
@@ -412,6 +482,14 @@ def compute_contiguity(
...
+    if isinstance(dtensor, DTensor):
+        key = (repr(dtensor.device_mesh), repr(dtensor.placements))
```
IvanYashchuk: Both DeviceMesh and tuple[Placement, ...] are hashable and comparable. There's no need to use their string representation for comparison.
```suggestion
        key = (dtensor.device_mesh, dtensor.placements)
```

IvanYashchuk (approved):

thunder/executors/nvfuserex_impl.py
```diff
@@ -325,6 +363,41 @@ def translate_bound_symbol(bsym: BoundSymbol) -> Any:
             nvout = lc_to_nv_map[out]
             fd.add_output(nvout)
 
+    MAX_LENGTH = 9999
+
+    if any(map(lambda t: isinstance(t, DTensorProxy), sorted_unique_inputs)):
```
wujingyue:
```suggestion
    if any(isinstance(t, DTensorProxy) for t in sorted_unique_inputs):
```
There's no need for `map` and `lambda`.

thunder/executors/nvfuserex_impl.py
```diff
@@ -325,6 +363,41 @@ def translate_bound_symbol(bsym: BoundSymbol) -> Any:
...
+        # multi-GPU path
+        assert all(map(lambda t: isinstance(t, DTensorProxy), sorted_unique_inputs)), (
```
wujingyue: ditto

kshitij12345:
Ping @t-vi for stamp

thunder/executors/nvfuserex_impl.py
```diff
@@ -239,6 +249,43 @@ def get_translator(bsym: BoundSymbol) -> Callable:
     return _translation_map[bsym.sym.id]
 
 
+class MultiDeviceFusionDefinition(FusionDefinition):
+    def __init__(self, in_dtensors: list[DTensorProxy], define_fn: Callable, max_length: int):
```
wujingyue:
```suggestion
    def __init__(self, define_fn: Callable[[FusionDefinition], None], in_dtensors: list[DTensorProxy], max_length: int):
```
to annotate the input/output type of Callable. 

It's slightly more natural to put `define_fn` in front of `in_dtensors`, because we say "run the function with these inputs" in English.

thunder/executors/nvfuserex_impl.py
```diff
@@ -325,6 +363,41 @@ def translate_bound_symbol(bsym: BoundSymbol) -> Any:
             nvout = lc_to_nv_map[out]
             fd.add_output(nvout)
 
+    MAX_LENGTH = 9999
+
+    if any(isinstance(t, DTensorProxy) for t in sorted_unique_inputs):
+        # multi-GPU path
+        assert all(isinstance(t, DTensorProxy) for t in sorted_unique_inputs), (
+            "nvfuser: Currently we only support Fusion region with all DTensor inputs or all Tensor inputs but not a mix"
+        )
```
crcrpar: why not exceptions?

thunder/executors/nvfuserex_impl.py
```diff
@@ -457,14 +541,37 @@ def __call__(self, *args):
...
+
+                assert len(out_tensors) == len(out_shardings)
+                out_dtensors: list[DTensor] = []
+                for out_tensor, out_sharding in zip(out_tensors, out_shardings):
+                    mesh = dist.device_mesh.init_device_mesh("cuda", (out_sharding.mesh.size,))
+                    placements: list[Placement] = []
+                    for parallel_type in [nvfuser.ParallelType.mesh_x]:
+                        axis: int = out_sharding.axis_sharded_on(parallel_type)
+                        placements.append(Replicate() if axis == -1 else Shard(axis))
+                    out_dtensors.append(DTensor.from_local(out_tensor, mesh, placements))
```
crcrpar, line 565: is `out_tensors` a list of `torch.Tensor`, not `DTensor`?

wujingyue: That's right. I chose that for two reasons: 
1. Using DTensor to nvFuser doesn't have practical benefits. nvFuser has a different representation of distributed tensors. For example, it takes the math as if single-GPU, takes multi-GPU schedules on the side, and takes and produces local tensors. Even if we squeeze DTensor in, it'll be only used at the surface. This makes it less appealing to add this extra dependency and more appealing to convert between DTensor and Tensor in Thunder. 
2. At some point, I created a custom DistributedTensor class (not dist.tensor.DTensor) in nvFuser which wraps the mesh, the placements and local tensor. https://github.com/NVIDIA/Fuser/pull/4035 removed it because Christian found it surprisingly expensive to unpack. 

However, I'm happy to hear your thoughts. I'm sure I haven't spent too much time optimizing nvFuser/multi-GPU for UX.

crcrpar: I've never well thought about the cost of unpacking (and packing) `torch.distributed.tensor.DTensor`, but from my naive perspective, it's possible that we have multiple nvfuser regions of DTensors in a thunder trace. In such case I'm speculating that thunder has multiple packing (and unpacking in nvfuser C++ end?).

So if we do want to eliminate overheads, I think one potential solution of thunder execution trace could be that we first unpack DTensors, delay packing as much as possible, and ask nvfuser multidevice to take tuples of local tensor and its device mesh instead of DTensors.
This is just my random thought and I'm not opposing the current design of nvfuser multidevice at all

wujingyue: > I'm speculating that thunder has multiple packing (and unpacking in nvfuser C++ end?).

Thunder currently needs to pay the overhead of packing/unpacking. nvFuser doesn't take or produce DTensors so packing/unpacking must be done by Thunder. If/When the overhead becomes outstanding for Thunder, we can certainly do something as you recommended.

crcrpar: > nvFuser doesn't take or produce DTensors so packing/unpacking must be done by Thunder.

This feels like a nit-picking but wouldn't nvfuser takes `DTensor`s as inputs to fusion definition and unpack them (or just access local tensor and device mesh) inside (= maybe C++ land)? I admit I'm not familiar with the implementation at all so bear with my poor understanding

kshitij12345:
Ping @t-vi for stamp

thunder/executors/nvfuserex_impl.py
wujingyue: DTensor doesn't exist in **C++**. nvFuser could unpack DTensors in **Python** but I didn't do that for the first reason in https://github.com/Lightning-AI/lightning-thunder/pull/2177#discussion_r2158105626

PR #2181, opened 2025-06-03
Use `to_dtype` and `to_torch_dtype` rather than querying `_thunder_to_torch_dtype_map` and `_torch_to_thunder_dtype_map`

crcrpar:
## What does this PR do?

It seems that for `torch.device` and `thunder.core.devices.Device` we use `to_device` and `to_torch_device` defined in `thunder.core.devices`.
For `torch.dtype` and `thunder.core.dtypes.dtype`, we seem to have used "private" dicts of `_thunder_to_torch_dtype_map` and `_torch_to_thunder_dtype_map` rather than public `to_torch_dtype` and `to_dtype`.
I think it's good to use public ones when available.


t-vi (approved):
Great cleanup thank you @crcrpar

PR #2183, opened 2025-06-03
opt-in for scatter operation in nvfuser

jjsjann123:


t-vi:
Cool to see NVFuser expand the ops!

What would be the circumstances in which one would opt in and not opt in? I can see why linear is a thing that is more tricky in terms of perf, but this seems to be a good option to always activate.

I still think all options are more or less evil, but after #2057 , I'm more optimistic that this one could be temporary.

jjsjann123:
> I still think all options are more or less evil, but after #2057 , I'm more optimistic that this one could be temporary.

Agree that these options are not good user experience. These are going to be a temporary thing.

The reason I'm adding a flag is that we have concurrent work going on.
* nvfuser is adding codegen support for these operations, so eventually we should be able to take them as-is by nvfuser.
* Meanwhile, our multi-gpu folks are working on models that needed these operations inside nvfuser (for parallelization strategy), so we want to have them, even though they are running as a fallback.

We'll see a couple similar ops added in the coming weeks (topk groupedgemm e.t.c.). We are not expecting any user to rely on these flags, since fallbacks don't necessarily provide any kernel perf gain.

jjsjann123:
Looks like scatter related tests run fine. I'm not sure what the failure on stable torch 2.7 is about.

I'll clean up the PR

jjsjann123:
hmm, failure is in dynamo but it doesn't show up in other PRs. so maybe it's real. I'll take a look at that

jjsjann123:
OOM is finally gone. I'm hitting real codegen asserts now, which might be coming from later scatter support plumbed through the code base. I'll follow up with check condition. :rescue_worker_helmet:

thunder/executors/nvfuserex_impl.py
```diff
@@ -2755,56 +2755,56 @@ def embedding(
...
+
+    # index_put is translated to scatter in nvfuser
+    return fd.ops.scatter(nva, nvi_b, nvs, 0)
+
+
+register_supported(PrimIDs.INDEX_PUT, index_put, _index_put_check)
 
 
-def _cross_entropy_check_(
+def _cross_entropy_check(
```
crcrpar: it'd be more desirable to put this change in a different PR
```suggestion
def _cross_entropy_check_(
```

thunder/executors/nvfuserex_impl.py
```diff
@@ -2755,56 +2755,56 @@ def embedding(
...
+    # temporary flag to allow scatter-like operations to be consumed by nvfuserex
+    enable_scatter: None | bool = get_compile_option("nv_enable_scatter", "Enable nvFuser scatter-like operations.")
+    if not enable_scatter:
+        return False
+
+    # TODO: limited support inside nvfuser. remove this when codegen support is generalized.
+    if len(indices) != 1 or indices[0].ndim != 1:
+        return False
+
+    if accumulate == True:
```
crcrpar:
```suggestion
    if accumulate:
```

thunder/executors/nvfuserex_impl.py
```diff
@@ -2756,12 +2756,60 @@ def embedding(
...
 
 
-def _scatter_check(a: TensorProxy, /, index: TensorProxy, src: TensorProxy | Number, dim: int) -> bool:
+def _index_put_check(a: TensorProxy, /, indices: Sequence[TensorProxy], values: TensorProxy, accumulate: bool) -> bool:
     # temporary flag to allow scatter-like operations to be consumed by nvfuserex
     enable_scatter: None | bool = get_compile_option("nv_enable_scatter", "Enable nvFuser scatter-like operations.")
     if not enable_scatter:
         return False
 
+    # TODO: limited support inside nvfuser. remove this when codegen support is generalized.
```
crcrpar, line 2765: Would nvidia/fuser repo have an issue to track the progress of it? With a brief search, I found only this comment https://github.com/NVIDIA/Fuser/blob/7149f8e2aef8d578029c7731e727e267fc34b553/csrc/ops/indexing.cpp#L102-L107. Either way, I suppose we should have some link to refer.

thunder/executors/nvfuserex_impl.py
```diff
@@ -2786,7 +2834,7 @@ def scatter(
 register_supported(PrimIDs.SCATTER, scatter, _scatter_check)
 
 
-def _cross_entropy_check(
+def _cross_entropy_check_(
```
crcrpar: I'm not sure if I suggested this change or not (if so I'm really sorry), but we should avoid having this change as a part of this PR

thunder/tests/test_nvfuser.py
```diff
@@ -969,6 +969,50 @@ def foo(b):
...
+        topk_weight: torch.Tensor  # [seq , top_k]]
+        bmm_out, idxs, topk_weight = inputs
+        out = bmm_out.index_put([idxs], bmm_out)  # [seq*top_k, hidden]
+        # TODO: enable following operation when nvfuser codegen can handle generic scatter
+        # out = out.reshape(*topk_weight.shape, -1)  # [seq, top_k, hidden]
+        # out = out * topk_weight.unsqueeze(-1)  # [seq, top_k, hidden]
+        # out = out.sum(dim=1)  # [seq, hidden]
+        return out
+
+    # use smaller problem size to avoid OOM on dynamo test
```
crcrpar: [question] What would this mean? At glance I couldn't get what this means as this test doesn't use `thunder.dynamo.compiler.thunderfx`

thunder/executors/nvfuserex_impl.py
jjsjann123: No worries, I misunderstood your earlier comment :stuck_out_tongue: I'll revert this.

thunder/tests/test_nvfuser.py
jjsjann123: Apologies, that's out-dated comment. I'll remove it.

FYI, I was hitting OOM issues from dynamo test in CI earlier, but turns out it's not caused by things running in this test specifically.

thunder/executors/nvfuserex_impl.py
jjsjann123: Good call! We should have one since we are moving all jira stuff to github.

jjsjann123: We currently only have a umbrella issue tracking that https://github.com/NVIDIA/Fuser/issues/4857. I added it in the code comment.

crcrpar (approved):
looks good to me

jjsjann123:
@t-vi This PR is ready to go now.

thunder/tests/test_nvfuser.py
```diff
@@ -969,6 +969,49 @@ def foo(b):
...
+    actual = jfoo(inputs)
+    expected = foo(inputs)
+    torch.testing.assert_close(actual, expected)
+
+    fw_trace = thunder.last_traces(jfoo)[-1]
+    fusion_bsyms = tuple(filter(lambda a: a.sym.is_fusion, fw_trace.bound_symbols))
+
+    # assert that everything is merged as a single nvfuser operation
+    assert len(fusion_bsyms) == 1
+    outside_fusion_syms = ["unpack_trivial", "python_return"]
```
crcrpar: nit-picking
```suggestion
    outside_fusion_sym_set = {"unpack_trivial", "python_return"}
    assert {el.sym.name for el in fw_trace.bound_symbols if not el.sym.is_fusion} == outside_fusion_sym_set
```

PR #2131, opened 2025-05-23
Print repro command when `test_core_vs_torch_consistency` fails

crcrpar:
## What does this PR do?

inspired by [PyTorch's `PYTORCH_OPINFO_SAMPLE_INPUT_INDEX`](https://github.com/pytorch/pytorch/blob/2225231a144f182e2e1bd26a619cc1bafad49e6d/torch/testing/_internal/common_utils.py#L270-L279), this gives `test_core_vs_torch_consistency` a mechanism to add repro command to error message.

An example is as follows:

```console
            try:
                result = run_snippet(
                    snippet_torch_consistency,
                    op,
                    device,
                    dtype,
                    tfn,
                    op.torch_reference,
                    sample,
                    lambda a, b, **kwargs: sample_comp(a, b, equal_nan=True, **kwargs),
                )
            except Exception as e:
                if repro_cmd not in str(e):
>                   raise type(e)(f"{str(e)}\n{repro_cmd}") from e
E                   RuntimeError: Raise to get a repro command
E                   command to reproduce the error: THUNDER_OPTEST_SAMPLE_INDEX=0 pytest thunder/tests/test_ops.py -k "test_core_vs_torch_consistency and cpu and thunder.dtypes.float32 and copysign"
```

PR #2190, opened 2025-06-04
[test_ops] Make repro command more specific

crcrpar:
## What does this PR do?

Currently it misses executor name. If a test case is failing on cuda, then the displayed repro command would run both nvfuserex and torchex, which is not expected from a repro command.

With this change, an example of displayed repro commands is 

```
            except Exception as e:
                if repro_cmd not in str(e):
>                   raise type(e)(f"{str(e)}\n{repro_cmd}") from e
E                   RuntimeError: Runtime output has shape of `(4, 4)`, device of `cuda:0`, and dtype of `torch.float64` but compiletime output has shape of `(4, 4)`, device of `cpu`, and dtype of `torch.float64`
E                   command to reproduce the error: THUNDER_OPTEST_SAMPLE_INDEX=1 pytest thunder/tests/test_ops.py::test_core_vs_torch_consistency_floor_divide_torch_cuda_thunder.dtypes.float64

thunder/tests/test_ops.py:148: RuntimeError
```

On main branch it's `THUNDER_OPTEST_SAMPLE_INDEX=1 pytest thunder/tests/test_ops.py -k "test_core_vs_torch_consistency and cuda and thunder.dtypes.float64 and floor_divide"` that does not specify the executor.

Follw-up of #2131

### DIFF
--- a/thunder/tests/test_ops.py
+++ b/thunder/tests/test_ops.py
@@ -129,7 +129,7 @@ def test_core_vs_torch_consistency(op, device: str, dtype: dtypes.dtype, executo
         path = os.path.relpath(__file__)
         repro_cmd = (
             f"command to reproduce the error: THUNDER_OPTEST_SAMPLE_INDEX={i} "
-            f'pytest {path} -k "test_core_vs_torch_consistency and {device} and {dtype} and {op.name}"'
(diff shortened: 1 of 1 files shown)


t-vi (approved):
Thank you @crcrpar

Issue #2196, opened 2025-06-06
[RFC] Propagate `torch.compile` options to fallback path


crcrpar:
`thunder.dynamo.thunderfx` distinguishes torch.compile options from thunder.jit options in https://github.com/Lightning-AI/lightning-thunder/blob/e171edf7bf142d354efb9e2233cd225a13e7fd57/thunder/dynamo/compiler.py#L211-L228.

It however doesn't pass torch.compile options to `torch.compile` used by `ThunderCompilr` https://github.com/Lightning-AI/lightning-thunder/blob/e171edf7bf142d354efb9e2233cd225a13e7fd57/thunder/dynamo/compiler.py#L60-L108 so it could use `torch.compile` with the default options.
I have no idea if and how this potential discrepancy could matter so I just wanted to share

IvanYashchuk:
Yes, definitely, we need to be able to pass options to torch.compile used in `thunderfx`!

crcrpar:
it seems that we had it https://github.com/Lightning-AI/lightning-thunder/commit/177643b75bbf305d52cd0a05acc6755defcde67d and lost it in https://github.com/Lightning-AI/lightning-thunder/commit/8e0faef8350cf871d0132d734f9bcea08cda68dc

kiya00:
The discussion is [here](https://github.com/Lightning-AI/lightning-thunder/pull/1535/files/6307021dbc7fa73f589ac72e2f3deea4c5cabb25#r1878975655), at that time we decided to use the default `torch.compile` without any arguments for the fallback region, it maybe different from the outer `torch.compile` options, since we actually uses the `torch._inductor.compile_fx` part

crcrpar:
One quick question, have we ever tried different "mode"s and/or "option"s?
I especially am interested in the behavior with mode="reduce-overhead", options={"triton.cudagraphs": True}, or others enabling cuda graphs

ref: https://docs.pytorch.org/docs/stable/generated/torch.compile.html

kiya00:
No I've never tried, maybe @kshitij12345 has?

kshitij12345:
No, I haven't investigated them a lot.

PR #2198, opened 2025-06-06
Propagate `torch.compile` options to fallback

crcrpar:
## What does this PR do?

Sketch for https://github.com/Lightning-AI/lightning-thunder/issues/2196

### DIFF
--- a/thunder/dynamo/compiler.py
+++ b/thunder/dynamo/compiler.py
@@ -2,6 +2,8 @@ from __future__ import annotations
 from functools import partial
 from looseversion import LooseVersion
 from typing import TYPE_CHECKING
+from typing import Any
+from typing import overload
 import warnings
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


thunder/dynamo/compiler.py
```diff
@@ -96,6 +119,8 @@ def __init__(self, **thunder_options):
         # Ref to the documentation of `SubgraphInfo` to know more about the information it contains.
         self.subgraph_infos: list[SubgraphInfo] = []
 
+        if thunder_options is None:
+            thunder_options = kwargs
```
kshitij12345, line 123: Will this ignore `kwargs` if `thunder_options` is not None?

I think we should just remove `kwargs` from the signature and update `thunderfx` to explicitly pass `thunder_options` and `torch_compile_options` (if applicable).

PR #2199, opened 2025-06-06
default to hf recipe in thunder.compile for hf models

KaelanDt:
<details>

</details>

## What does this PR do?

This simply defaults to the `HFTransformers` recipe when `thunder.compile` is called if the model comes from HF transformers.

## PR review

## Did you have fun?

### DIFF
--- a/thunder/__init__.py
+++ b/thunder/__init__.py
@@ -295,7 +295,12 @@ def compile(
     plugins = plugins_
 
     if recipe is None:
-        recipe = thunder.recipes.BaseRecipe()
+        model_class = getattr(fn, "__class__", None)
+        if model_class is not None and "transformers" in model_class.__module__:
+            recipe = thunder.recipes.HFTransformers()
+        else:
+            recipe = thunder.recipes.BaseRecipe()
+        print(recipe)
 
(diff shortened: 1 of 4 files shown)


t-vi:
Hey @KaelanDt ! This is great and we want to have something like this, but I'd want to take a different angle, that makes this more generic.
How about something like this:
- we look up up `f"{fn.__class__.__module__}.{fn.__class__.__name__}"` and then iteratively take off suffixes `.foo` and look it up in a dict,
- we make a classmethod `Recipe.is_applicable(cls, fn)` that checks whether the recipe is applicable.
- The '' recipe (when nothing is left) is the default recipe.

WDYT?

@lantiga who also has ideas around it.

(so here the transformers recipe would register  under `transformers` and all would be good, but we can also make specific recipes for classes). The only trick is that typically the module is some submodule, but I guess we live with it.

lantiga:
Hey yes that looks great, @KaelanDt actually the `validate` classmethod in recipes was introduced exactly for this purpose, so let's replace `is_applicable` for `validate`

t-vi:
Oh, right, validate. Sorry about that!

PR #2201, opened 2025-06-06
Dont delay split in benchmarks

beverlylytle:
The recent merge of #2102, while a step in the right direction, doesn't handle activation checkpointing correctly.  The change is to prevent potential big fluctuations in benchmarking until activation checkpointing is better handled.


t-vi:
@IvanYashchuk @beverlylytle do we still need this or is it obsoleted by #2203 ?

beverlylytle:
It's now obsolete.  Closing.

PR #2203, opened 2025-06-06
fix activation checkpointing in the joint trace

beverlylytle:
As discussed in #2102, activation checkpointing is not well handled. This PR moves the duplication logic for the bsyms which need to be recomputed in the backward trace from the splitting up to the creation of the joint trace.  This PR also modifies CSE so that subexpressions common to both the forward section and the backward section of the joint trace do not annihilate each other.  Finally, this PR modifies the `Graph` and `Node` structures used in fusion passes to distinguish (and prevent the merging of) `Node`s with forward operations from `Node`s with backward operations.


t-vi:
@beverlylytle Thank you, seems good in general,  two questions:
- do we need anything to keep CSE from removing it?
- could we make it a separate pass (living in autodiff is cool, but a separate transform)?

Thank you!

thunder/core/transform_common.py
```diff
@@ -287,6 +287,8 @@ def cse_single_bsym(
     )
 
     if bsym.sym.id == prims.PrimIDs.GET_GRAD:
+        # do not cse across the forward-backward boundary of the joint trace
+        rhs_to_bsym_map.clear()
```
beverlylytle: This is the change in CSE to prevent removing the duplicated symbols.

thunder/core/transform_common.py
```diff
@@ -274,7 +274,7 @@ def cse_single_bsym(
         exception_type=AssertionError,
     )
 
-    # `NON_FUNCTIONAL_OPS` are a op that's not deterministic, for example, `torch.nn.functiona.dropout`
+    # `NON_FUNCTIONAL_OPS` are a op that's not deterministic, for example, `torch.nn.functional.dropout`
```
riccardofelluga, line 277: Nice

thunder/core/transform_common.py
riccardofelluga: I was thinking that maybe not in the scope of this issue, but maybe more broadly if we stick with bound symbols tags, then it can be a good idea to add the tags in the RHS hash computation. Just an idea tho, floating around

thunder/transforms/autodiff.py
```diff
@@ -300,8 +301,12 @@ def shallow_copy_if_input(p):
                 new_bsyms = self.new_trace.pop_scope()
 
                 # Let the new bound symbols inherit tags, in particular RECOMPUTE_IN_BACKWARD
+                should_recompute = _should_recompute_bsym_in_backward(bsym)
                 for nbsym in new_bsyms:
                     nbsym.tags |= bsym.tags
+                    if should_recompute:
+                        # potentially taking a recompute proxy tag and making it a bound symbol tag
+                        nbsym.tags.add(BoundSymbolTag.RECOMPUTE_IN_BACKWARD)
```
riccardofelluga: Also maybe not in the scope of this PR but still food for thought, do we really need bound symbol tags? what if we stick with proxy tags only? 

Maybe I am missing some info tho 

cc.ing @t-vi for the idea

thunder/transforms/autodiff.py
```diff
@@ -342,15 +347,53 @@ def shallow_copy_if_input(p):
...
     # run the trace through the processor
     joint_trace, _ = AugmentedForwardProcessor(trace)()
+    joint_trace, _ = InsertRecomputationsProcessor(joint_trace)()
 
     # run through DCE in case some of the gradients of intermediates are not needed.
     joint_trace = dce(joint_trace)
-    # group get_grad symbols together for torch compile fusions
-    # !!! is it preferable to do this here or in the torch compile fusion pass?
+    # group get_grad symbols together for torch compile fusions and to make clear boundary for cse
     _group_get_grad_bsyms(joint_trace)
```
riccardofelluga, line 414: Nice comment!

thunder/core/transform_common.py
t-vi: Sounds like a good idea to mark things as not to participate in cse or so.

thunder/transforms/autodiff.py
t-vi: The bsym tags seemed clearer because they are really the operations we want to work with. Back when we started proxy tags, @IvanYashchuk  suggested using symbol tags instead. But other than consolidating to one eventually (on my list) for recompute, I'm not 100% set.

t-vi:
> While I understand the reason behind the decision of creating InsertRecomputationsProcessor I wonder if it would have been possible to make use of [find_producer_symbol](https://github.com/Lightning-AI/lightning-thunder/blob/62d9535bfeab621b571c3cc8aed974b64381aad0/thunder/core/utils.py#L1088) to avoid creating bound symbol tags and keeping things simple. But that is a non blocking issue for me.

I would prefer to use more processors because they have the chance of standardizing how to propagate various things (names - not perfect yet, tags, source locations etc.) and subsymbol handling etc.

riccardofelluga:
> > While I understand the reason behind the decision of creating InsertRecomputationsProcessor I wonder if it would have been possible to make use of [find_producer_symbol](https://github.com/Lightning-AI/lightning-thunder/blob/62d9535bfeab621b571c3cc8aed974b64381aad0/thunder/core/utils.py#L1088) to avoid creating bound symbol tags and keeping things simple. But that is a non blocking issue for me.
> 
> I would prefer to use more processors because they have the chance of standardizing how to propagate various things (names - not perfect yet, tags, source locations etc.) and subsymbol handling etc.

Hmm but maybe we can use it in combination with the processor, in the end the trace object is stored inside the processor object. It would return the list of bound symbols that we need to add to the `backward_part_bsysms_recomputed` by just looking at the proxies. Probably also would help us move away from the boundsymbol tags to avoid extra complexity there, something like:

```python
if has_recompute_tag(proxy):
    bsysms = find_producer_symbols(self.trace, proxy, inputs)
    self.backward_part_bsysms_recomputed.update(proxy_var, bsysms)
```

t-vi:
Yeah, but why would we re-find the symbols rather than hanging on to the information?
(I could be dense and there is a reason, but I'm not sure.)

Personally, I think tagging stuff and then working with it is something we want to have the option to do in general, so I don't think I'd remove bsym tags even if we don't use them here.

thunder/transforms/autodiff.py
```diff
@@ -342,15 +347,53 @@ def shallow_copy_if_input(p):
...
+                bw_idx = 0
+                while bw_idx < len(processed_bsyms):
+                    backward_part_bsym = processed_bsyms[bw_idx]
+                    modified = False
+                    for n in reversed(backward_part_bsym.flat_proxy_args):
+                        recomp_bsym_rec = self.backward_part_bsyms_recomputed.get(n.name)
+                        if recomp_bsym_rec is not None:
+                            modified = True
+                            recomp_bsym, recomp_output = recomp_bsym_rec
+                            processed_bsyms.insert(bw_idx, recomp_bsym)
```
riccardofelluga: I think we need to update the names of the outputs to avoid name collisions

riccardofelluga:
> Yeah, but why would we re-find the symbols rather than hanging on to the information? (I could be dense and there is a reason, but I'm not sure.)

Hmm my thought is more coming from a correctness standpoint tho, my idea is that by using that find_producers method we  can guarantee that all the needed bsysms are added to the dict.

It is true that if two proxies share the same producers, those producers will be found twice, however it is guaranteed that they will be found(since it works with the dataflow). This idea of reusing what was already found, is actually already built-in in the `find_producers` as it accepts lists of multiple proxies of which it needs to find the producer for, and therefore saves on looking it up twice. 

If we were to use this feature then the code would be different, what I would do in that case would be to collect a list of proxies that have been found with the tag so far and I would call find_producers only once at the start of self.in_backward_region, something like:
 
 ```python
if has_recompute_tag(proxy):
    self.backward_recomputed_proxies += proxy
[...]
if bsym.sym == prims.get_grad:
    self.in_backward_region = True
    self.producer_bsysm = find_producers(self.trace, self.backward_recomputed_proxies, inputs)
```

> Personally, I think tagging stuff and then working with it is something we want to have the option to do in general, so I don't think I'd remove bsym tags even if we don't use them here.

It's also good to keep them, but I think in that case we need to define also the edge cases and make it clear in its design. One I can think rn of would be: if a bsym has the tag, should the inputs to that symbol be tagged too? another: what happens if a transform replaces the symbol, should the new symbol inherit that too? and some other questions of this sort

File: thunder/transforms/autodiff.py (hunk 5 above)
Comment (beverlylytle) [reply thunder/transforms/autodiff.py]: Yes, this is a good point you make.  I've updated the PR to support that.

File: thunder/transforms/autodiff.py [hunk 6]
```diff
@@ -342,15 +358,57 @@ def shallow_copy_if_input(p):
...
+                    for n in reversed(backward_part_bsym.flat_proxy_args):
+                        recomp_bsym_rec = self.backward_part_bsyms_recomputed.get(n.name)
+                        if recomp_bsym_rec is not None:
+                            modified = True
+                            recomp_bsym, recomp_output = recomp_bsym_rec
+                            # To avoid name clashes, we create new output proxies
+                            with tracectx(self.new_trace):
+                                for output in recomp_bsym.flat_proxy_outs:
+                                    new = TensorProxy(like=output)
+                                    self.add_to_swap_map(output, new)
```
riccardofelluga: What happens if the output is not a tensor? Or does this imply that we want to restrict recomputed bsysm to the ones that have tensors as outputs?

This might help in case https://github.com/Lightning-AI/lightning-thunder/blob/58d1585f8d718d7d6fc6fb077a56cca65c4310d3/thunder/core/proxies.py#L140

beverlylytle: Good eye.  No, I don't think that there's a reason to restrict to tensors.

thunder/executors/data_dependent_partition.py
```diff
@@ -20,14 +20,23 @@
...
+        is_backward: bool,
+    ):
         self.ID = ID
         self.start = start
         self.stop = stop
         self.group_bsyms = group_bsyms
         self.group_indices = group_indices
         self.parents: utils.OrderedSet[Node] = utils.OrderedSet()
         self.children: utils.OrderedSet[Node] = utils.OrderedSet()
+        self.is_backward = is_backward
```
riccardofelluga, line 39: To the untrained eye, why is this needed?

riccardofelluga (commented):
Great work, really! 

Alt

Thanks for the effort!

thunder/executors/data_dependent_partition.py
beverlylytle: I had been working on the assumption that the non-fusibility of `get_grad` bsyms was enough to prevent the fusion of operations from the forward part of the joint trace with operations from the backward part of the trace.  When symbols from the forward part of the trace need to be recomputed in the backward part, the `get_grad` boundary is ignored by the topological sorting that happens in `dataflow_merge`.  Adding this extra data to the nodes and preventing forward and backward nodes from merging prevents fusions containing both forward part operations along with operations that need to be recomputed for the backward.

beverlylytle: For context at commit [1246818](https://github.com/Lightning-AI/lightning-thunder/pull/2203/commits/12468189c18c9f8ca55e87544848ad83e5f97d05) the test test_checkpointing_thunderfx was failing with 
```
    @torch.no_grad()
    @no_autocast
      # cotangents: "Collection"
      C0, _, = saved_for_backward
      # C0: "Collection"
      # None
      del saved_for_backward
      t773, = cotangents
      # t773: "cuda:0 f32[1, 2048, 320]"
      del cotangents
      t1045, \
      l_self_modules_transformer_modules_h_modules_1_modules_checkpoint_wrapped_module_modules_mlp_modules_fc_1_parameters_weight_, \
      l_self_modules_transformer_modules_h_modules_1_modules_checkpoint_wrapped_module_modules_mlp_modules_fc_2_parameters_weight_, \
      l_self_modules_lm_head_parameters_weight_, t779, \
      l_self_modules_transformer_modules_ln_f_parameters_weight_, t1006, t752, \
      l_self_modules_transformer_modules_h_modules_1_modules_checkpoint_wrapped_module_modules_mlp_modules_proj_parameters_weight_, \
      l_self_modules_transformer_modules_h_modules_1_modules_checkpoint_wrapped_module_modules_norm_2_parameters_weight_, \
      l_self_modules_transformer_modules_h_modules_1_modules_checkpoint_wrapped_module_modules_attn_modules_proj_parameters_weight_, \
      l_self_modules_transformer_modules_h_modules_1_modules_checkpoint_wrapped_module_modules_attn_modules_attn_parameters_weight_, \
      l_self_modules_transformer_modules_h_modules_1_modules_checkpoint_wrapped_module_modules_norm_1_parameters_weight_, \
      t1000, \
      l_self_modules_transformer_modules_h_modules_0_modules_checkpoint_wrapped_module_modules_mlp_modules_proj_parameters_weight_, \
      t1027, t1028, \
      l_self_modules_transformer_modules_h_modules_0_modules_checkpoint_wrapped_module_modules_mlp_modules_fc_2_parameters_weight_, \
      l_self_modules_transformer_modules_h_modules_0_modules_checkpoint_wrapped_module_modules_mlp_modules_fc_1_parameters_weight_, \
      l_self_modules_transformer_modules_h_modules_0_modules_checkpoint_wrapped_module_modules_norm_2_parameters_weight_, \
      l_self_modules_transformer_modules_h_modules_0_modules_checkpoint_wrapped_module_modules_attn_modules_proj_parameters_weight_, \
      l_self_buffers_sin_, l_self_buffers_cos_, \
      l_self_modules_transformer_modules_h_modules_0_modules_checkpoint_wrapped_module_modules_attn_modules_attn_parameters_weight_, \
      l_self_modules_transformer_modules_h_modules_0_modules_checkpoint_wrapped_module_modules_norm_1_parameters_weight_, \
      l_idx_, = C0
      # t1045: "cuda:0 f32[1, 2048, 64]"
      # l_self_modules_transformer_modules_h_modules_1_modules_checkpoint_wrapped_module_modules_mlp_modules_fc_1_parameters_weight_: "cuda:0 f32[1376, 64]"
      # l_self_modules_transformer_modules_h_modules_1_modules_checkpoint_wrapped_module_modules_mlp_modules_fc_2_parameters_weight_: "cuda:0 f32[1376, 64]"
      # l_self_modules_lm_head_parameters_weight_: "cuda:0 f32[320, 64]"
      # t779: "cuda:0 f32[2048, 64]"
      # l_self_modules_transformer_modules_ln_f_parameters_weight_: "cuda:0 f32[64]"
      # t1006: "cuda:0 f32[1, 2048, 64]"
      # t752: "cuda:0 f32[1, 2048, 1]"
      # l_self_modules_transformer_modules_h_modules_1_modules_checkpoint_wrapped_module_modules_mlp_modules_proj_parameters_weight_: "cuda:0 f32[64, 1376]"
      # l_self_modules_transformer_modules_h_modules_1_modules_checkpoint_wrapped_module_modules_norm_2_parameters_weight_: "cuda:0 f32[64]"
      # l_self_modules_transformer_modules_h_modules_1_modules_checkpoint_wrapped_module_modules_attn_modules_proj_parameters_weight_: "cuda:0 f32[64, 64]"
      # l_self_modules_transformer_modules_h_modules_1_modules_checkpoint_wrapped_module_modules_attn_modules_attn_parameters_weight_: "cuda:0 f32[192, 64]"
      # l_self_modules_transformer_modules_h_modules_1_modules_checkpoint_wrapped_module_modules_norm_1_parameters_weight_: "cuda:0 f32[64]"
      # t1000: "cuda:0 f32[1, 2048, 64]"
      # l_self_modules_transformer_modules_h_modules_0_modules_checkpoint_wrapped_module_modules_mlp_modules_proj_parameters_weight_: "cuda:0 f32[64, 1376]"
      # t1027: "cuda:0 f32[1, 2048, 1376]"
      # t1028: "cuda:0 f32[1, 2048, 1376]"
      # l_self_modules_transformer_modules_h_modules_0_modules_checkpoint_wrapped_module_modules_mlp_modules_fc_2_parameters_weight_: "cuda:0 f32[1376, 64]"
      # l_self_modules_transformer_modules_h_modules_0_modules_checkpoint_wrapped_module_modules_mlp_modules_fc_1_parameters_weight_: "cuda:0 f32[1376, 64]"
      # l_self_modules_transformer_modules_h_modules_0_modules_checkpoint_wrapped_module_modules_norm_2_parameters_weight_: "cuda:0 f32[64]"
      # l_self_modules_transformer_modules_h_modules_0_modules_checkpoint_wrapped_module_modules_attn_modules_proj_parameters_weight_: "cuda:0 f32[64, 64]"
      # l_self_buffers_sin_: "cuda:0 f32[4096, 16]"
      # l_self_buffers_cos_: "cuda:0 f32[4096, 16]"
      # l_self_modules_transformer_modules_h_modules_0_modules_checkpoint_wrapped_module_modules_attn_modules_attn_parameters_weight_: "cuda:0 f32[192, 64]"
      # l_self_modules_transformer_modules_h_modules_0_modules_checkpoint_wrapped_module_modules_norm_1_parameters_weight_: "cuda:0 f32[64]"
      # l_idx_: "cuda:0 i64[1, 2048]"
      clear_mutable_collection(C0)
      del C0
>     t1041 = torch.reshape(t1040, (-1, 64))  # t1041: "cuda:0 f32[2048, 64]"
E     UnboundLocalError: local variable 't1040' referenced before assignment

beverlylytle: The creation of `t1040` which should be one of the duplicated bound symbols appearing in the backward trace has been fused into a forward trace symbol.

PR #2204, opened 2025-06-09
Define `CompiledObject` once at `thunder.dynamo` load time

crcrpar:
## What does this PR do?

- change when (and how many times) `CompiledObject` gets defined
    - Inside `thunderfx`, the class is defined once per `thunderfx` call
- renames it to `ThunderfxCompiledObject`
- add `thunderfx` and `ThunderfxCompiledObject` to docs

cc @borda @lantiga


IvanYashchuk:
@kshitij12345, could you please review this PR?

thunder/dynamo/compiler.py
```diff
@@ -208,16 +208,80 @@ def save_reproducer_to_folder(
...
+
+    This object wraps the function compiled with :func:`torch.compile` using Thunder as a backend
+    (see :class:`~thunder.dynamo.ThunderCompiler`).
+    It provides access to Thunder traces generated during execution, while maintaining the
+    callable interface of the original compiled function.
+
+    Note: This is the return type of ``thunderfx``, not :func:`thunder.jit`.
+    """
+
+    def __init__(self, be, func: Callable):
```
kshitij12345: It would be nice to rename `be` as `backend`. It looks a bit confusing in the generated doc otherwise.

image

thunder/dynamo/compiler.py
```diff
@@ -208,16 +208,80 @@ def save_reproducer_to_folder(
                 )
 
 
-def thunderfx(fn: Callable, /, **kwargs) -> Callable:
-    """Compiles a callable (function or model) by using Thunder as the backend of :func:`torch.compile`
+# We return this object instead of just the raw `compiled` Callable so that
+# we have a place to hang the `last_*traces` properties.
+class ThunderfxCompiledObject:
```
kshitij12345: Optional: I think `ThunderFXCompiledObject` looks better to me. Also, in the fx doc : https://docs.pytorch.org/docs/stable/fx.html, FX is usually capitalized.

PR #2205, opened 2025-06-09
Remove input tensors' `check_tensor_shape_and_metadata` from prologue when inputs are free from `torch.SymInt`

crcrpar:
## What does this PR do?

When `example_inputs` to `thunder.dynamo.ThunderCompiler.__call__` are concrete (meaning no `torch.SymInt`s), it'd be reasonable to rely solely on TorchDynamo caching, without `check_*`s in prologue.


thunder/dynamo/compiler.py
```diff
@@ -41,18 +43,40 @@
...
+def is_in_torch_compile() -> bool:
+    """Returns :obj:`True` if :func:`torch.compile` is active."""
+    return TorchCompileContext.current_compile_id() is not None
 
-    Args:
-        options: The dictionary of options to modify
+
+def is_dynamic_inputs(example_inputs):
+    """Check if inputs dynamic or not by checking the presence of :class:`torch.SymInt`."""
+    from torch.utils import _pytree as pytree
```
kshitij12345: Why not import this at module level?

thunder/dynamo/compiler.py
```diff
@@ -41,18 +43,40 @@
...
+
+    flat_example_inputs, _ = pytree.tree_flatten(example_inputs)
+    return any(
+        isinstance(a, torch.SymInt) or any(isinstance(s, torch.SymInt) for s in a.shape) for a in flat_example_inputs
```
kshitij12345: I think it might be possible that `a` is not a torch.SymInt or a torch.Tensor in which case this would error at `a.shape`.

`a` could be a tuple of `Tensor` or such as well.

thunder/transforms/extraction_only_prologue_transform.py
```diff
@@ -1,9 +1,22 @@
...
+    This transform is mainly used by :class:`~thunder.dynamo.ThunderCompiler` to remove the check of input tensors
+    when either they are :class:`torch.nn.Parameter` or all of them don't have any symbolic shape.
+
+    Args:
+        skip_check_on_input_tensors: If :obj:`True`, remove all the check from the prologue trace as TorchDynamo caching would do enough.
+            Otherwise, remove the checks of tensor proxies with ``ProxyTag.STATIC_MEMORY_LOCATION``.
+    """
 
+    def __init__(self, skip_check_on_input_tensors: bool):
+        self.skip_check_on_input_tensors = skip_check_on_input_tensors
```
kshitij12345, line 24: This isn't used anywhere yet.

kshitij12345 (changes requested):

thunder/dynamo/compiler.py
crcrpar: Just to avoid a potential confusion with `thunder.core.pytree`. I changed this import to `as torch_pytree` and moved it to the top

thunder/transforms/extraction_only_prologue_transform.py
crcrpar: great catch, thank you

thunder/dynamo/compiler.py
crcrpar: I think it could be any type other than `torch.SymInt` or `torch.Tensor` but I think thanks to `tree_flatten`, containers are not there

t-vi (approved):
Thank you @crcrpar @kshitij12345 @IvanYashchuk

PR #2213, opened 2025-06-10
Support `torch.Tensor.view(dtype)`

### LINKED ISSUE #2212: `torch.Tensor.view(dtype)` is not supported
## 🐛 Bug

`torch.Tensor.view` with dtype is not supported while it's seen in e.g. torchao's `to_mx` function here -- https://github.com/pytorch/ao/blob/v0.10.0/torchao/prototype/mx_formats/mx_tensor.py#L146-L329.

#### Code sample

```python
import torch
import thunder


def f(t: torch.Tensor) -> torch.Tensor:
    return t.view(torch.float8_e4m3fn)


if __name__ == "__main__":
    x = torch.testing.make_tensor(
        (4, 4),
        dtype=torch.int8,
        device=torch.device("cuda"),
        requires_grad=False,
    )
    jitted = thunder.jit(f)
    out = jitted(x)
```

This fails as follows:
```
  File "/path/to/c.py", line 18, in <module>
    out = jitted(x)
          ^^^^^^^^^
  File "/path/to/thunder/__init__.py", line 830, in wrapped
    return fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^
  File "/path/to/thunder/__init__.py", line 870, in fn_
                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/thunder/__init__.py", line 809, in wrapped
    cache_entry, inps, pro_to_epi = get_computation_and_inputs_fn(*args, **kwargs)
                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/thunder/core/langctxs.py", line 136, in _fn
    result = fn(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^
  File "/path/to/thunder/__init__.py", line 236, in cache_info_wrapper
    res = fn(*args, **kwargs)
          ^^^^^^^^^^^^^^^^^^^
  File "/path/to/thunder/__init__.py", line 774, in get_computation_and_inputs
    prologue_trc, computation_trc, epilogue_trc = acquire_initial_trace(fn, args, kwargs, cd, cs, ad_hoc_executor)
                                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/thunder/__init__.py", line 434, in acquire_initial_trace
    jit_results: TraceResults = thunder_general_jit(
                                ^^^^^^^^^^^^^^^^^^^^
  File "/path/to/thunder/core/jit_ext.py", line 2133, in thunder_general_jit
    result = jfn(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^
  File "/path/to/thunder/core/interpreter.py", line 7567, in fn_
    raise e
  File "/path/to/thunder/core/interpreter.py", line 7526, in fn_2
    return fn(*args, **kwargs)

  File "/path/to/c.py", line 7, in f
    return t.view(torch.float8_e8m0fnu)
^^^^^^^^^^^^^^^^^^
  File "/path/to/thunder/core/interpreter.py", line 6952, in partial_call_impl
    return partial_function.func(*(partial_function.args + args), **(partial_function.keywords | kwargs))
^^^^^^^^^^
  File "/path/to/thunder/core/interpreter.py", line 1302, in wrapping_wrapper
    res = ufn(*uargs, **ukwargs)
          ^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/thunder/core/jit_ext.py", line 388, in wrapper
    return fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^
  File "/path/to/thunder/core/symbol.py", line 320, in __call__
             ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/thunder/core/langctxs.py", line 136, in _fn
    result = fn(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^
  File "/path/to/thunder/torch/__init__.py", line 1398, in view
    return reshape(a, shape)
           ^^^^^^^^^^^^^^^^^
  File "/path/to/thunder/core/symbol.py", line 320, in __call__
             ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/thunder/core/langctxs.py", line 136, in _fn
    result = fn(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^
  File "/path/to/thunder/torch/__init__.py", line 1138, in reshape
    return clang.reshape(a, shape)
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/thunder/core/langctxs.py", line 136, in _fn
    result = fn(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^
  File "/path/to/thunder/clang/__init__.py", line 1069, in reshape
    if l >= 0:
       ^^^^^^
TypeError: '>=' not supported between instances of 'torch.dtype' and 'int'
```

With thunderfx, thanks to its splitter, it works but it gets fallen back. The module after the split is as follows:
```
    def forward(self, l_t_: "i8[4, 4]"):
        inductor_0 = self.inductor_0(l_t_);  l_t_ = None
        return (inductor_0,)

    class inductor_0(torch.nn.Module):
        def forward(self, l_t_: "i8[4, 4]"):
            view: "f8e4m3fn[4, 4]" = l_t_.view(torch.float8_e4m3fn);  l_t_ = None
            return view

        class _orig_mod(torch.nn.Module):
            def forward(self, l_t_: "i8[4, 4]"):
                view: "f8e4m3fn[4, 4]" = l_t_.view(torch.float8_e4m3fn);  l_t_ = None
                return view
```

### Expected behavior

just works.

crcrpar:
## What does this PR do?

Adding `bitcast` primitive as nvfuser's bitcast op would be used once it's exposed to Python
Fixes #2212

### DIFF
--- a/thunder/tests/opinfos.py
+++ b/thunder/tests/opinfos.py
@@ -3513,6 +3513,20 @@ to_opinfo = OpInfo(
 data_movement_ops.append(to_opinfo)
 
 
+def view_with_dtype_sample_generator(op, device, dtype, requires_grad, **kwargs):
+    make = partial(make_tensor, device=device, dtype=dtype, requires_grad=requires_grad)
+
+    for dst_dtype in {torch.float32, torch.bfloat16, torch.float64} - {dtype}:
+        yield SampleInput(make((8, 8)), dtype)
+
+
+view_with_dtype_opinfo = OpInfo(
+    ltorch.view,
+    sample_input_generator=view_with_dtype_sample_generator,
+    torch_reference=torch.Tensor.view,
+)
+
+
 def cuda_sample_generator(op, device, dtype, requires_grad, **kwargs):
     make = partial(make_tensor, device=device, dtype=dtype, requires_grad=requires_grad)
 
--- a/thunder/tests/opinfos.py
+++ b/thunder/tests/opinfos.py
@@ -3525,6 +3525,7 @@ view_with_dtype_opinfo = OpInfo(
     sample_input_generator=view_with_dtype_sample_generator,
     torch_reference=torch.Tensor.view,
 )
+data_movement_ops.append(view_with_dtype_opinfo)
 
 
 def cuda_sample_generator(op, device, dtype, requires_grad, **kwargs):
--- a/thunder/tests/test_grad.py
+++ b/thunder/tests/test_grad.py
@@ -407,7 +407,11 @@ def test_vjp_correctness(op, device, dtype, executor, comp):
         # sample.thunder() line below attempts to approximate those conversions
         # for non-differentiable arguments like dtypes so that the test will
         # execute properly.
-        sample = sample.thunder()  # converts torch.dtype to thunder.dtype
+        # NOTE: While `convert_element_type` is skipeed as of https://github.com/Lightning-AI/lightning-thunder/pull/2213
+        # as in https://github.com/Lightning-AI/lightning-thunder/blob/dbf6bad3/thunder/tests/opinfos.py#L3324-L3346,
+        # `torch.Tensor.view(dtype)` seems to require `torch.dtype` to be kept as is, opposite to `convert_element_type`.
+        if op.name != "view":
+            sample = sample.thunder()  # converts torch.dtype to thunder.dtype
         sample = sample.remove_singularities(op, eps)
 
         flat_op, flat_args, spec = flatten_func(op.op, sample.args, sample.kwargs)
--- a/thunder/torch/__init__.py
+++ b/thunder/torch/__init__.py
@@ -1498,6 +1498,8 @@ def unsqueeze(a: TensorLike, /, dim: int) -> TensorLike:
 @torchsymbol(torch.Tensor.view, is_method=True)
 def view(a: TensorLike, /, *shape) -> TensorLike:
     shape = utils.extract_shape_from_varargs(shape)
+    if len(shape) == 1 and isinstance(shape[0], torch.dtype):
+        return to(a, dtype=shape[0])
     return reshape(a, shape)
 
 
--- a/thunder/torch/__init__.py
+++ b/thunder/torch/__init__.py
@@ -1500,6 +1500,13 @@ def unsqueeze(a: TensorLike, /, dim: int) -> TensorLike:
 def view(a: TensorLike, /, *shape) -> TensorLike:
     shape = utils.extract_shape_from_varargs(shape)
     if len(shape) == 1 and isinstance(shape[0], torch.dtype):
+        dst_dtype = to_dtype(shape[0])
+        src_dtype = to_dtype(a.dtype)
+        utils.check(
+            dst_dtype.bytes == src_dtype.bytes,
+            lambda: f"`a.view({shape[0]})` is not supported as a's itemsize is {src_dtype.bytes} but target dtype itemsize is {dst_dtype.bytes}",
+            exception_type=NotImplementedError,
+        )
         return to(a, dtype=shape[0])
     return reshape(a, shape)
 
--- a/thunder/torch/__init__.py
+++ b/thunder/torch/__init__.py
@@ -1507,7 +1507,7 @@ def view(a: TensorLike, /, *shape) -> TensorLike:
             lambda: f"`a.view({shape[0]})` is not supported as a's itemsize is {src_dtype.bytes} but target dtype itemsize is {dst_dtype.bytes}",
             exception_type=NotImplementedError,
         )
-        return to(a, dtype=shape[0])
+        return prims.bitcast(a, dtype=to_dtype(shape[0]))
     return reshape(a, shape)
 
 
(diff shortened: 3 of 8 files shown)


thunder/torch/__init__.py
```diff
@@ -1395,6 +1395,8 @@ def unsqueeze(a: TensorLike, /, dim: int) -> TensorLike:
 @torchsymbol(torch.Tensor.view, is_method=True)
 def view(a: TensorLike, /, *shape) -> TensorLike:
     shape = utils.extract_shape_from_varargs(shape)
+    if len(shape) == 1 and isinstance(shape[0], torch.dtype):
+        return to(a, dtype=shape[0])
```
mruberry: Won't this be incorrect with inplace operations, because this will create a copy, but if the result of the view is modified inplace then the original tensor should be changed?

crcrpar: This won't be correct in that this returns a new tensor proxy but I think the support of in-place ops to the output of this would be on par with the existing path given that `view` is included in https://github.com/Lightning-AI/lightning-thunder/blob/62d9535bfeab621b571c3cc8aed974b64381aad0/thunder/torch/__init__.py#L6634-L6657

mruberry: OK; would you just add a `TODO` comment describing the issue?

crcrpar: Could you elaborate what kind of TODO comment you'd expect to see here?
Is it one generally true to the other `_syms_returning_vies` ops such as
```python
# TODO(crcrpar): Ideally we should not create a copy or we should update
# `TensorProxy` so that we can tell with which tensor proxy it shares its underlying storage.
# This is because in pytorch this op would return a tensor of the same storage as the input
```

Or is it one mentioning the necessity of improving in-place support?
Or is it one mentioning view and the other ops in the set would be special-cased in thunder's in-place handling?

mruberry: I think you first TODO is great; I realize that figuring out exactly what to write is tricky. We could just say:

```
TODO: This operation should return a view of the input tensor, but this to semantically returns a copy; we should review this with future inplace support
```

crcrpar:
@t-vi @lantiga @mruberry could you review this?

t-vi (changes requested):
This needs
- inplaciness,
- correct semantics

```python
torch.ones(8, dtype=torch.bool).view(torch.int64)  # tensor([72340172838076673])
```

Maybe we should also add tests, just incase.

crcrpar:
hmm, I now think this deserves a separate independent prim or clang op

crcrpar:
@t-vi I introduced a new primitive so that we call `Tensor.view(dtype)` in torch executor (and nvfuser's bit cast in the near future).
Could you review this again?

thunder/tests/opinfos.py
```diff
@@ -3471,6 +3471,20 @@ def to_sample_generator(op, device, dtype, requires_grad, **kwargs):
...
+        yield SampleInput(make((8, 8)), dtype)
+
+
+view_with_dtype_opinfo = OpInfo(
+    ltorch.view,
+    sample_input_generator=view_with_dtype_sample_generator,
+    torch_reference=torch.Tensor.view,
+)
+
+
```
t-vi, line 3530: I think we need something like
```suggestion

data_movement_ops.append(view_with_dtype_opinfo)
```

other than that, seems good to me.

t-vi:
I think the test which branch to take needs improving .

crcrpar:
I'm working on fixing the failures

thunder/tests/test_grad.py
```diff
@@ -406,7 +406,11 @@ def test_vjp_correctness(op, device, dtype, executor, comp):
         # sample.thunder() line below attempts to approximate those conversions
         # for non-differentiable arguments like dtypes so that the test will
         # execute properly.
-        sample = sample.thunder()  # converts torch.dtype to thunder.dtype
+        # NOTE: While `convert_element_type` is skipeed as of https://github.com/Lightning-AI/lightning-thunder/pull/2213
+        # as in https://github.com/Lightning-AI/lightning-thunder/blob/dbf6bad3/thunder/tests/opinfos.py#L3324-L3346,
+        # `torch.Tensor.view(dtype)` seems to require `torch.dtype` to be kept as is, opposite to `convert_element_type`.
+        if op.name != "view":
+            sample = sample.thunder()  # converts torch.dtype to thunder.dtype
```
crcrpar, line 414: it'd been mysterious to me that `torch.Tensor.view(dtype)` seems to work even when backward is generated but not in this test. it though seems that the test works if the conversion of `torch.dtype` -> `thunder.core.dtypes.dtype` is skipped.
Currently, `convert_element_type` skips this test as in https://github.com/Lightning-AI/lightning-thunder/blob/dbf6bad30cf38606f2feaec629f19357bba0b74d/thunder/tests/opinfos.py#L3324-L3346 so it might be acceptable to remove this line but I chose to apply the change to what really needs the skip.

PR #2215, opened 2025-06-10
Add `torch.float8_e8m0fnu`

crcrpar:
## What does this PR do?

As per title, this adds `torch.float8_e8m0fnu`

### DIFF
--- a/thunder/core/baseutils.py
+++ b/thunder/core/baseutils.py
@@ -380,6 +380,7 @@ _torch_dtype_to_str_map = {
     torch.float8_e4m3fnuz: "torch.float8_e4m3fnuz",
     torch.float8_e5m2: "torch.float8_e5m2",
     torch.float8_e5m2fnuz: "torch.float8_e5m2fnuz",
+    torch.float8_e8m0fnu: "torch.float8_e8m0fnu",
     torch.float16: "torch.float16",
     torch.float32: "torch.float32",
     torch.float64: "torch.float64",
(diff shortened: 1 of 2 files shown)


t-vi:
What's the minimal PyTorch version that has these? We've historically not shied away from bumping minimal PyTorch version, but it'd be good to know if this needs it.

crcrpar:
> What's the minimal PyTorch version that has these? We've historically not shied away from bumping minimal PyTorch version, but it'd be good to know if this needs it.

2.7.0 has it.

IvanYashchuk:
Waiting on code owner review from @lantiga, @mruberry, and/or @t-vi.

IvanYashchuk:
What do you propose for tracking older PyTorch support? Would you add a CI job for PyTorch 2.6? Please create an issue about this.

The only place that has a dependency on `torch.float8_e8m0fnu` object being available is the `_thunder_to_torch_dtype_map` construction.

Changes in https://github.com/Lightning-AI/lightning-thunder/pull/2315 remove this dependency. That PR could be merged first before this one, then I would resolve the merge conflicts here. Alternatively, we merge this PR as is, and Thunder wouldn't be importable with PyTorch 2.6 (which is not tested in CI) until https://github.com/Lightning-AI/lightning-thunder/pull/2315 is merged.

beverlylytle:
[CI is failing](https://github.com/Lightning-AI/lightning-thunder/actions/runs/17157781430/job/48679264069?pr=2396) because the torch version is 2.6.0, but this requires 2.7.0.  Could we roll this back?

@t-vi

t-vi:
@beverlylytle let's bump the requirement. Thank you for pinging.

PR #2222, opened 2025-06-12
TE: fix related to delayed forward-backward split

kshitij12345:
With #2102, we have a different path for creating forward and backward traces.

On the old path, TE integration made a few changes to tag the forward trace to be decorated with fp8_autocast and adding correct amax sync to the backward trace (for DelayedScaling recipe).

This adds similar changes to the new path. Also, some refactor to remove `_include_te_fp8_autocast` from TraceCtx as we can use relevant tag to determine if it is a forward trace.

NOTE: Tested on B200, RTX6000, H100


thunder/executors/transformer_engineex.py
```diff
@@ -544,7 +544,7 @@ def te_sync_fp8_meta_bwd_impl():
 )
 
 
-def _transformer_engine_bwd_fp8_meta_sync(fw_extrace, bw_extrace):
+def _transformer_engine_bwd_fp8_meta_sync(unused_fw_extrace, bw_extrace):
```
IvanYashchuk: Usually, "_" is used for unused variables, and IDEs are able to recognize it correctly with proper syntax highlighting.

thunder/transforms/autodiff.py
```diff
@@ -542,6 +542,12 @@ def backward_fn(saved_for_backward, cotangents):
 
     backward_trace = dce(backward_trace)
 
+    # Importing here to avoid cyclical dependencies in future.
+    # NOTE: This is required only for v1 executor.
+    from thunder.executors.transformer_engineex import transformer_engine_v1_bwd_fp8_meta_sync
```
IvanYashchuk, line 547: Is there an import error now?

kshitij12345: No import error with this patch.

kshitij12345:
Ping @t-vi for stamp.

t-vi (approved):

PR #2210, opened 2025-06-10
Add `bitwise_left_shift` and `bitwise_right_shift`

crcrpar:
## What does this PR do?

Add `bitwise_left_shift` and `bitwise_right_shift` so that `<<` and `>>` would work.

PR #2227, opened 2025-06-12
Avoid negative number rhs values to bitwise shift tests

crcrpar:
## What does this PR do?

A fix following #2210.
As per https://en.cppreference.com/w/cpp/language/operator_arithmetic.html,

> If the value of rhs is negative or is not less than the number of bits in lhs, the behavior is undefined.

See "Built-in bitwise shift operators"

Thank you @jacobhinkle and @naoyam for the help.


mruberry (approved):
Cool!

fyi @beverlylytle

PR #2240, opened 2025-06-14
Add optional profiling to examine() for execution time and speedup reporting

### LINKED ISSUE #2239: Support profiling in examine() to measure execution time and speedup
### Issue Description:

Currently, thunder.examine() helps verify whether Thunder can successfully compile and run a given function. However, it lacks visibility into runtime performance improvements, which can be important for users evaluating the benefits of Thunder JIT.

**Proposed Feature:**

Introduce a profile=True flag to examine() that:

    Measures and prints the execution time of the original (eager) function.

    Measures and prints the execution time of the compiled (Thunder) function.

    Calculates and displays the speedup factor.

**Example Usage:**

```

model = MyModel(...)
examine(model, *args, profile=True)
```

**Expected Output:**

```
The function appears to be working as expected
Execution time - original: 0.1234s
Execution time - compiled: 0.0678s
Speedup: 1.82x
```

**Also Includes:**

    Update to docs/source/fundamentals/examine.rst

    Unit test test_examine_profile using capsys to assert profiling output

**Motivation:**

This feature adds valuable insight into Thunder’s runtime optimization benefits and allows users to quantify performance gains directly using the examine utility.

KAVYANSHTYAGI:
### Description

This PR introduces an optional profile=True argument to the thunder.examine() utility. When enabled, it measures and displays:

-     Execution time of the original (eager) function
-     Execution time of the compiled (Thunder JIT) function
-     Speedup factor comparing the two

This enhancement allows users to quantitatively assess Thunder’s performance improvements with minimal overhead.

**Key Changes**

    thunder/examine/__init__.py:

     Added profile: bool = False argument to examine()

     Wrapped both eager and JIT execution in time.perf_counter() calls

     Printed results conditionally if profile=True

     docs/source/fundamentals/examine.rst:

     Updated with example usage and output

     thunder/tests/test_examine.py:

      Added test_examine_profile() to validate the new functionality via capsys

**Example Output**

```
The function appears to be working as expected
Execution time - original: 0.1234s
Execution time - compiled: 0.0678s
Speedup: 1.82x
```

Motivation

This feature empowers users to:

    Measure Thunder’s runtime performance gains on their models

    Validate whether compilation leads to meaningful speedups

    Use examine() not just for correctness but for lightweight benchmarking

Test Plan

    Verified new test case test_examine_profile passes

    Confirmed expected output in terminal using toy models and compiled functions

    Backward compatibility retained: no changes in output if profile=False

Related Issue

Closes #2239 – Support profiling in examine() to measure execution time and speedup


t-vi:
Hi @KAVYANSHTYAGI . Thank you for looking at improving thunder.
The crux here is that profiling is a bit more involved - you typically want to have some warmup iterations (including compilation etc.) and then run the benchmark. Also, the results are likely not meaningful without detailed information what executors / transforms have been applied etc.
The quickstarts have a profiling function included, so maybe it would be better to use that rather than bolt it onto the examine function which really does not have that much in common with what is needed for profiling.
WDYT?

t-vi:
Let's discuss profiling needs in an issue first and then revisit.

PR #2241, opened 2025-06-14
Support mode 'bilinear' for torch.nn.interpolate

### LINKED ISSUE #1941: implement torch.nn.interpolate mode=bilinear etc.
Currently, we only support "nearest" interpolation, but we should want to support other modes as well.
If doing it via a decomposition is too tedious, we likely should implement a prim mapping back to PyTorch.

kilpelainenj:
<details>

</details>

## What does this PR do?

Fixes #1941 
I created support for torch.nn.functional.interpolate mode = 'bilinear'. I extended the previous opinfos.py tests also for this bilinear mode with the tweak that the mode only allows 2 spatial dimensions. I also created a couple of new tests to ensure that the model only gets 2 spatial dimensions. 

## PR review

As this is my first contribution, I welcome all sorts of feedback and suggestions to make this better. 

## Did you have fun?

Yes. This was a great first issue.


KaelanDt (commented):
Very nice, looks great @kilpelainenj !

It seems to me that we could also add support for linear and trilinear modes using the same underlying, similarly to what is done in pytorch: https://github.com/pytorch/pytorch/blob/main/torch/_decomp/decompositions.py#L3731

t-vi (approved):
Thank you @kilpelainenj @KaelanDt @k223kim 
Let's merge this and then see about trilinear.

PR #209, opened 2024-04-17
docker: compile TE

Borda:
<details>
  <summary><b>Before submitting</b></summary>

- [x] Was this discussed/approved via a Github issue? (no need for typos and docs improvements)
- [x] Did you read the [contributor guideline](https://github.com/Lightning-AI/pytorch-lightning/blob/main/.github/CONTRIBUTING.md), Pull Request section?
- [ ] Did you make sure to update the docs?
- [ ] Did you write any new necessary tests?

</details>

## What does this PR do?

Fixes #196

## PR review

Anyone in the community is free to review the PR once the tests have passed.
If we didn't discuss your PR in Github issues there's a high chance it will not be merged.

## Did you have fun?

Make sure you had fun coding 🙃

cc @borda

PR #2253, opened 2025-06-18
[WIP] TE: Add CI

### LINKED ISSUE #196: [ci] : We should add a CI flow with TransformerEngine installed so that we can run the relevant tests.

kshitij12345:
Fixes #196 

TODO
* [ ] Correctly install TE


Borda:
@kshitij12345, following your own [comment](https://github.com/Lightning-AI/lightning-thunder/pull/209#issuecomment-2077526858), so we have such a GPU?

PR #1978, opened 2025-04-22
Add PEFT benchmarking script in `thunder/benchmarks`

wprazuch:
<details>
  <summary><b>Before submitting</b></summary>

- [ ] Was this discussed/approved via a Github issue? (no need for typos and docs improvements) -> Discussed with @IvanYashchuk 
- [x] Did you read the [contributor guideline](https://github.com/Lightning-AI/pytorch-lightning/blob/main/.github/CONTRIBUTING.md), Pull Request section?
- [ ] Did you make sure to update the docs?
- [ ] Did you write any new necessary tests? -> To verify what kind of tests to add

</details>

## What does this PR do?

It introduces the benchmarking script for PEFT finetuning scenario, which supports:
- compiler setup (inductor/thunder/eager)
- single/multi-gpu setup (fsdp2)

To execute:
```python
python thunder/benchmarks/benchmark_peft.py     --model deepseek-ai/DeepSeek-R1-Distill-Qwen-7B     \
    --devices 1  --trust-remote-code      --attn-implementation sdpa  \
 --max-steps 10        --mbs 1  \
   --seq-length 4096        --jit-backend thunder
```
for multi-gpu:
```python
torchrun --nproc_per_node=8 --master_port=12345 thunder/benchmarks/benchmark_peft.py  \
   --model meta-llama/CodeLlama-34b-Instruct-hf     --strategy fsdp2   \
     --devices 8     --mbs 1     --seq-length 1024     --max-steps 10  \
        --jit-backend eager     --attn-implementation sdpa  \
           --trust-remote-code
```

## PR review

Anyone in the community is free to review the PR once the tests have passed.
If we didn't discuss your PR in Github issues there's a high chance it will not be merged.

## Did you have fun?

Make sure you had fun coding 🙃

PR #2254, opened 2025-06-18
Add PEFT benchmarking script in `thunder/benchmarks`

riccardofelluga:
## What does this PR do?

This PR picks up and replaces #1978 with some tweaks. It introduces PEFT benchmarks for single gpu and multi-gpu with fsdp2 as deistributed strategy.

## PR review

Can be tested by running:

```shell
python thunder/benchmarks/benchmark_peft.py --model deepseek-ai/DeepSeek-R1-Distill-Qwen-7B --trust-remote-code --attn-implementation sdpa --max-steps 10 --mbs 1 --seq-length 4096 --compile thunder --fixed-num-hidden-layers 2
```

Multi-GPU:

```shell
torchrun --nproc_per_node=2 thunder/benchmarks/benchmark_peft.py --model deepseek-ai/DeepSeek-R1-Distill-Qwen-7B --trust-remote-code --attn-implementation sdpa --max-steps 10 --mbs 1 --seq-length 4096 --compile thunder --fixed-num-hidden-layers 2
```

## Side notes

I've extended the logging from the original PR and removed the prints. Now this uses the log levels to print more or less information depending on `--verbose` flag.

@crcrpar in the fsdp2 part I carried it over from the `benchmark_litgpt.py` file, but I've commented out the litgpt specific layers. I've left it there for now but I am planning to remove them after the review, let me know if you think I should find the equivalent layers in transformers and adapt that code.

cc. @wprazuch


thunder/benchmarks/benchmark_peft.py
```diff
@@ -0,0 +1,706 @@
...
+    fixed_num_hidden_layers: int | None = None,
+    trust_remote_code: bool = False,
+):
+    """Load and configure a HuggingFace model."""
+    logger.debug("Starting model configuration")
+
+    config = AutoConfig.from_pretrained(model_name, trust_remote_code=trust_remote_code)
+    logger.debug(f"Config loaded: {config}")
+
+    if fixed_num_hidden_layers is not None:
```
KaelanDt, line 103: Since tweaking the model is allowed here, would it not be better to load the model locally and not rely on the hf hub as is done here https://github.com/Lightning-AI/lightning-thunder/blob/main/thunder/tests/test_networks.py ?

thunder/benchmarks/benchmark_peft.py
```diff
@@ -0,0 +1,706 @@
...
+        logger.info("Resetting cache size for torch.dynamo")
+        torch._dynamo.config.cache_size_limit = 64
+        # With this setting, Dynamo Graphs inline all the modules
+        torch._dynamo.config.inline_inbuilt_nn_modules = True
+
+    if backend == "inductor":
+        logger.info("Compiling model with torch.compile")
+        model = torch.compile(model)
+
+    elif backend == "thunder":
```
KaelanDt: Would perhaps be worth also benchmarking thunder.jit by itself, outside of thunderfx

thunder/benchmarks/benchmark_peft.py
```diff
@@ -0,0 +1,706 @@
...
+            logger.warning(f"Truncating sequence length from {seq_len} to {max_seq_len}")
+            batch = {k: v[:, :max_seq_len] for k, v in batch.items()}
+
+        # Forward pass with SDPA context
+        with sdpa_context():
+            outputs = model(**batch)
+            loss = outputs.loss
+
+        # Backward pass
+        loss.backward()
```
KaelanDt, line 561: I'm not sure the backward is also jitted here

thunder/benchmarks/benchmark_peft.py
riccardofelluga: Sure!

thunder/benchmarks/benchmark_peft.py
riccardofelluga: Could you point me to the specific test you have in mind?

thunder/benchmarks/benchmark_peft.py
riccardofelluga: What do you mean by not jitted?

thunder/benchmarks/benchmark_peft.py
KaelanDt: here https://github.com/Lightning-AI/lightning-thunder/blob/main/thunder/tests/test_networks.py#L430 we initialize the config locally

thunder/benchmarks/benchmark_peft.py
KaelanDt: I meant the backward function is not compiled, but nevermind i think it is

thunder/benchmarks/benchmark_peft.py
riccardofelluga: Oh I see, however in that context we know already the models that we want to test and here that is not always true. This benchmark script is meant to be used as a dev tool to investigate performance, so the fact that one can pick it up and experiment right away is in the main requirements

t-vi (approved):
Thank you @riccardofelluga @KaelanDt 

At some point we might look at splitting out the various requirements here, but not now.

PR #2255, opened 2025-06-18
feat: match no_sync with pytorch

k223kim:
<details>

</details>

## What does this PR do?

Fixes # (issue).

## PR review

## Did you have fun?

### DIFF
--- a/thunder/core/module.py
+++ b/thunder/core/module.py
@@ -379,7 +379,7 @@ class ThunderModule(pytorch.nn.Module):
             yield
         finally:
             reset_skip_data_parallel_grad_sync(token)
-            _sync_grads(self)
+            # _sync_grads(self)
 
     def __getattr__(self, name: str) -> Any:
         if name == "_model":
--- a/thunder/tests/distributed/helper.py
+++ b/thunder/tests/distributed/helper.py
@@ -324,13 +324,14 @@ if torch.distributed.is_available():
 
     def run_test_no_sync_grad_accumulation(
         test_case: DistributedParallelTestCase,
-        get_model_and_optimizer: Callable[[torch.device], tuple[torch.nn.Module, torch.optim.Optimizer]],
+        get_model_and_optimizer: Callable[[torch.device], tuple[torch.nn.Module, torch.nn.Module, torch.optim.Optimizer]],
         is_comm: Callable[[str], bool],
         dataset_size,
     ):
         from collections import defaultdict
         from contextlib import nullcontext
         from thunder.distributed import get_skip_data_parallel_grad_sync
+        from torch.nn.parallel import DistributedDataParallel
 
         device = torch.device("cuda", test_case.rank)
         batch_size = 128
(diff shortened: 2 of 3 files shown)


thunder/core/module.py
```diff
@@ -379,8 +379,8 @@ def no_sync(self):
             yield
         finally:
             reset_skip_data_parallel_grad_sync(token)
-            _sync_grads(self)
 
+    #
```
Borda:
```suggestion
```

thunder/tests/distributed/helper.py
```diff
@@ -431,3 +431,109 @@ def get_ground_truth_loss_grads(device, dataloader):
...
+                if num_grad_accum_steps is not None:
+                    loss /= num_grad_accum_steps
+                loss.backward()
+
+            return loss
+
+        device = torch.device("cuda", test_case.rank)
+
+        gradients = defaultdict(list)
+        for use_no_sync in (True, False):
```
Borda: can we have it rather as an argument ot we really need this loop?

k223kim: Hmm, we do not need this loop. Would it make a difference if we use arguments?

thunder/tests/distributed/helper.py
```diff
@@ -431,3 +431,109 @@ def get_ground_truth_loss_grads(device, dataloader):
...
+                            )
+                        if use_no_sync and i == 0 and iter_count == 0:
+                            import thunder
+
+                            # make sure the backward trace under `no_sync` has actual math computations.
+                            no_sync_bwd_trc = thunder.last_backward_traces(jitted_model)[-1]
+                            test_case.assertGreater(len(no_sync_bwd_trc.bound_symbols), 1)
+                assert torch.allclose(torch_loss, loss, atol=1e-4, rtol=1e-4)
+
+                torch.testing.assert_close(torch_grad, thunder_grad, atol=1e-3, rtol=1e-3)
```
k223kim: @t-vi I think this should fail if thunder and torch are different

thunder/tests/distributed/helper.py
crcrpar: from the perspective of what we test, there's no differences but making `use_no_sync` argument would make test logs way better. It clearly tells us which setting fails or not. I'd have `use_no_sync` as an argument as @Borda suggests.

t-vi: I think we should split the test.
That said, I would expect the no-sync test to roughly do the following:

```
for j in range(2):
    for i in range(2):
        with no_sync():
            fw
            loss
            bw
        grab grad.clones for comparison  # these would not be synced, even outside the context manager.
    fw
    loss
    bw  # sync happens only(!) here and below
    grab grad.clones for comparison  # these are synced
    zero_grad
    fw
    loss
    bw  # sync happens only(!) here and above
    grab grad.clones for comparison  # these are synced
    zero_grad
```

and then the comparison is between the gradients in order between plain PT DDP and Thunder DDP.

t-vi: Note that the grabbing of the grads happens outside the no_sync.

PR #2256, opened 2025-06-18
add float exception to assertion in jit_ext

KaelanDt:
Compilation was crashing when running on modules such as `https://github.com/vllm-project/vllm/blob/8b6e1d639c66d5828d03a7df2c3a500030a5c5cd/vllm/attention/layer.py#L27`, in particular due to the `scale: float` argument.


k223kim (approved):

t-vi (approved):
Thank you @KaelanDt @k223kim @lianakoleva

PR #2262, opened 2025-06-20
add support for full with tensor input

### LINKED ISSUE #2261: support torch.full with (scalar/1 element) tensor as value
## 🚀 Model / language coverage

We currently only accept numbers as value in full.

### Pitch

Apparently `mask = torch.full((tgt_len, tgt_len), torch.tensor(torch.finfo(dtype).min, device=device), device=device)` and is a pattern sometimes found in actual code.

We could support this by mapping full with a single element tensor argument to `.expand(shape).clone().to(...)`.

### Minimal Repro

```python
import torch, thunder
def fn(a):
   return torch.full((2,), a)

a = torch.tensor(1.0, dtype=torch.float64)
print(fn(a))   # tensor([1., 1.])
jfn = thunder.jit(fn)
jfn(a)  # exception (not the nicest error message, either): ValueError: Trying to extract a dtype from a non-numbertype object thunder.dtypes.float64!
```

Note that the dtype is not the original dtype, but roundtripped through dtype->python-type->dtype (unless dtype is specified). Similarly, the result is on the default device (unless otherwise specified).

t-vi:
<details>

</details>

## What does this PR do?

Fixes #2261 .

## PR review

## Did you have fun?

### DIFF
--- a/thunder/tests/test_ops.py
+++ b/thunder/tests/test_ops.py
@@ -10,7 +10,7 @@ import thunder
 import thunder.core.devices as devices
 import thunder.core.dtypes as dtypes
 from thunder.core.pytree import tree_flatten, tree_map
-from thunder.tests.framework import assert_closer, ops, run_snippet, requiresJAX, requiresCUDA
+from thunder.tests.framework import assert_closer, ops, run_snippet, requiresJAX, requiresCUDA, instantiate
 from thunder.tests.opinfos import OpInfo, SampleInput, opinfos
 import thunder.tests.bf16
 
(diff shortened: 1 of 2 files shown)


thunder/tests/test_ops.py
```diff
@@ -557,3 +557,15 @@ def fn(a):
...
     a = torch.randn(5, 5, requires_grad=True)  # trigger grad transform
     assert_close(fn(a), jfn(a))
+
+
+@instantiate()
+def test_full_tensor_value(executor, device, dtype):
+    def fn(a):
+        return torch.full((2,), a)
+
+    jfn = thunder.jit(fn)
```
crcrpar, line 567: I guess it'd be better to use `executor` parameter like
```python
    jfn = executor.make_callable(fn)
```

t-vi: Depends on what the goal is? 
I'm not trying to test various executors here.

crcrpar: Then I think testing one executor, i.e. torch executor would be enough. I think this instantiation could involve multiple executors.

IvanYashchuk: It would be better to specify executors in the `@instantiate` call. Currently, multiple test instances are testing the same thing.

thunder/tests/test_ops.py
```diff
@@ -557,3 +557,15 @@ def fn(a):
     jfn = thunder.jit(fn)
     a = torch.randn(5, 5, requires_grad=True)  # trigger grad transform
     assert_close(fn(a), jfn(a))
+
+
+@instantiate()
+def test_full_tensor_value(executor, device, dtype):
```
IvanYashchuk, line 563: Is there a reason for this to be a separate test and not expanding sample inputs for full OpInfo (`full_sample_generator`)?

Issue #2218, opened 2025-06-11
Reduce `tool.ruff.lint.ignore`


crcrpar:
ae51153db2e623e8b9fb3609de3a6ad976addd0a enabled ruff-check in pre-commit.
I added a bunch of rules to "ignore" list but I think there are too many "ignore"s: 
https://github.com/Lightning-AI/lightning-thunder/blob/ae51153db2e623e8b9fb3609de3a6ad976addd0a/pyproject.toml#L153-L176.

- [ ] "E731",  # Do not assign a lambda expression, use a def
- [ ] "E501",  # todo: Line too long (235 > 120 characters
- [ ] "F841",  # https://docs.astral.sh/ruff/rules/unused-variable/
- [ ] "F821",  # https://docs.astral.sh/ruff/rules/undefined-name/
- [ ] "E402",  # https://docs.astral.sh/ruff/rules/module-import-not-at-top-of-file/
- [ ] "F401",  # https://docs.astral.sh/ruff/rules/unused-import/
- [ ] "F405",  # https://docs.astral.sh/ruff/rules/undefined-local-with-import-star-usage/
- [ ] "E712",  # https://docs.astral.sh/ruff/rules/true-false-comparison/
- [ ] "E711",  # https://docs.astral.sh/ruff/rules/none-comparison/
- [ ] "E721",  # https://docs.astral.sh/ruff/rules/type-comparison/
- [ ] "E722",  # https://docs.astral.sh/ruff/rules/bare-except/
- [ ] "F403",  # https://docs.astral.sh/ruff/rules/undefined-local-with-import-star/
- [ ] "F601",  # https://docs.astral.sh/ruff/rules/multi-value-repeated-key-literal/
- [ ] "F822",  # https://docs.astral.sh/ruff/rules/undefined-export/
- [ ] "F842",  # https://docs.astral.sh/ruff/rules/unused-annotation/
- [x] "F722",  # https://docs.astral.sh/ruff/rules/forward-annotation-syntax-error/
- [x] "E702",  # https://docs.astral.sh/ruff/rules/multiple-statements-on-one-line-semicolon/
- [ ] "E741",  # https://docs.astral.sh/ruff/rules/ambiguous-variable-name/
- [x] "W293",  # https://docs.astral.sh/ruff/rules/blank-line-with-whitespace/
- [ ] "F811",  # https://docs.astral.sh/ruff/rules/redefined-while-unused/
[...]

PR #2265, opened 2025-06-21
[blocked by #2264] remove F811 from ruff ignore rules

crcrpar:
## What does this PR do?

As per title, make changes to make `F811` effective.

Related:
- #2218

### DIFF
--- a/thunder/tests/test_transforms.py
+++ b/thunder/tests/test_transforms.py
@@ -660,7 +660,7 @@ def test_buffer_dtype_casting():
             }
 
             # here we switch the prologue_trace to a copy with new metadata
-            prologue_trace = thunder.transforms.quantization.trace_with_replaced_proxy_metadata(
+            prologue_trace = thunder.transforms.utils.trace_with_replaced_proxy_metadata(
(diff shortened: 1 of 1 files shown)


t-vi (approved):
Seems good, thank you @crcrpar .

In general, we would need to balance creating merge conflicts of PRs in flight with incremental change here. So I would not be unhappy if we could have a think about which things to enable from the list and do so en bloc.

crcrpar:
I'm rechecking which rules can be just removed without actual code changes, to have a PR that changes only pyproject.toml in #2267. Then I think it'd be fair to resolve ignored rules that need actual code changes one by one like this one.
What do you think?

Borda:
@crcrpar or what the "depends on #2264" was meant to be?

crcrpar:
This branch has the commits of 2264 and ones included in this PR. I originally didn't notice that multiple ignored rules could have been removed at once as in 2267. So I thought we should merge 2264 first, then this PR, in the reverse order of the ignore rules in pyproject.toml. 
Then 2267 came up. So now I'm reopening this PR as a new one to remove F811.

PR #2294, opened 2025-07-01
vllm+thunder integration, minimal working example

KaelanDt:
## What does this PR do?

This PR adds an example of using a thunder compiled model with vllm. Right it is about 30% slower than eager when running the script on an L40s. Happy to break this up into another PR that would add a recipe, since that is basically the same as BaseRecipe except for the disable grad argument and the additional executor, and after that identify paths forward to make it cleaner and more robust. A few limitations here:
- The new operator we register has to be flagged to not be dead-code-eliminated, because its outputs are not reused in the rest of the computation. So the traces are not very clean or inspectable. however due to the fast that we are almost as fast as eager and the outputs are correct, i believe the kv cache mechanism works as expected.
- This only works for a batch size = 1.
- This pads the sequences if the sequence length > 1, meaning we waste some computation on short inputs. We should aim to do like what is done for vllm+torch compile with CUDA graphs where a precompilation is done over many input sizes in parallel on startup.
- Speaking of cuda graphs, it would be nice to test how they speed things up/if it works.

## PR review

## Did you have fun?


examples/vllm/thunder_transformers.py
```diff
@@ -0,0 +1,98 @@
...
+
+    torch.ops.vllm.unified_attention_with_output.default(q, k, v, o, layer_name)
+
+    return output, key, value # return tensors that were changed
+
+
+def vllm_unified_attention_meta(query, key, value, output, layer_name):
+    return output, key, value # return tensors that were changed
+
+tmp_exec = TemporaryExecutor()
```
t-vi, line 24: This should be an OperatorExecutor instead.

examples/vllm/thunder_transformers.py
```diff
@@ -0,0 +1,98 @@
...
+
+_vllm_sym = tmp_exec.register_operator(
+    "vllm_unified_attention",
+    meta=vllm_unified_attention_meta,
+    fn=vllm_unified_attention,
+    tags=(prims.OpTags.DONT_DCE,) # needed to avoid DCE of unified_attention executor. TODO: don't rely on this.
+)
+
+tmp_exec._lookasides[torch.ops.vllm.unified_attention_with_output.default] = _vllm_sym
+tmp_exec._lookasides[torch.ops.vllm.unified_attention_with_output._op] = _vllm_sym
```
t-vi, line 37: So for one, there would be the replaces function.

t-vi: Given that it isn't uncommon, should we support passing a tuple for the replaces= arg to register_operator?

PR #2305, opened 2025-07-03
Add `torch._scaled_mm`

crcrpar:
## What does this PR do?

This adds `torch._scaled_mm`.
Since `torch._scaled_mm` has a stride requirements on its first two inputs while `TensorProxy` does not have stride information, `thunder.torch._scaled_mm` cannot have enough check for the op.

follow-up tasks:
- [ ] mx dtype gemm
- [ ] second input matrix's stride check

### DIFF
--- a/thunder/tests/test_ops.py
+++ b/thunder/tests/test_ops.py
@@ -82,7 +82,7 @@ def snippet_torch_consistency(
         if no_fallback_with_double_inputs:
             raise
 
-        def upcast_tensors(x):
+        def upcast(x):
             if isinstance(x, torch.Tensor) and torch.is_floating_point(x):
                 return x.to(torch.double)
 


thunder/tests/opinfos.py
```diff
@@ -7431,6 +7433,49 @@ def tensor_1d_sample_generator(op, device, dtype, requires_grad, **kwargs):
...
+            pytest.mark.xfail,
+            "test_core_vs_torch_consistency",
+            devicetypes=(devices.DeviceType.CPU,),
+        ),
+        DecorateInfo(
+            pytest.mark.xfail,
+            "test_core_vs_torch_consistency",
+            devicetypes=(devices.DeviceType.CUDA,),
+            executors=("nvfuser",),
+        ),
```
riccardofelluga: just for curiosity, is this xfail for nvfuser related to the stride check missing? 👀

crcrpar: no, it's for numerical mismatch

crcrpar:
oh, hi https://github.com/NVIDIA/Fuser/pull/4718

thunder/executors/nvfuserex_impl.py
```diff
@@ -3193,3 +3193,70 @@ def argsort_transform(
...
+    a: TensorLike,
+    b: TensorLike,
+    scale_a: TensorLike,
+    scale_b: TensorLike,
+    bias: TensorLike | None = None,
+    scale_result: TensorLike | None = None,
+    out_dtype: dtypes.dtype | None = None,
+    use_fast_accum: bool = False,
+) -> bool:
+    return nvfuser_version() >= LooseVersion("0.2.27")
```
crcrpar: @jjsjann123 would it be 0.2.28 or 0.2.29?

thunder/core/prims.py
```diff
@@ -4102,6 +4103,59 @@ def matmul_meta(a: TensorProxy, b: TensorProxy, /) -> TensorProxy:
...
+    b: TensorProxy,
+    scale_a: TensorProxy,
+    scale_b: TensorProxy,
+    bias: TensorProxy | None = None,
+    scale_result: TensorProxy | None = None,
+    out_dtype: dtypes.dtype | None = None,
+    use_fast_accum: bool = False,
+) -> TensorProxy:
+    # NOTE: Currently this only supports fp8 a & b and per-tensor scales.
+    # TODO: Devise a way to make sure `a` is row-major and `b` is column-major.
```
IvanYashchuk, line 4147: You can use `prims.stride_order(a, (1, 0))` for conversion to row-major and `prims.stride_order(b, (0, 1)` for conversion to column-major matrices.

thunder/executors/torchex.py
```diff
@@ -1500,12 +1500,58 @@ def _copy_with_setitem_impl(a, key, value):
...
+    result_dtype: torch.dtype = to_torch_dtype(a.dtype if out_dtype is None else out_dtype)
+    utils.check(
+        is_row_major(a),
+        lambda: f"`a` has strides of {a.stride()} but expected to be row-major",
+    )
+    if not is_column_major(b):
+        b = b.t().contiguous().t()
+    utils.check(
+        is_column_major(b),
+        lambda: f"`b` has strides of {b.stride()} but expected to be row-major",
```
IvanYashchuk:
```suggestion
        lambda: f"`b` has strides of {b.stride()} but expected to be column-major",
```

thunder/executors/nvfuserex_impl.py
jjsjann123: That's a good call. I should have bumped up the version..... :face_exhaling:

Right now it's at `0.2.27`, if we want to be safe, we can bump it to `0.2.28` on nvfuser and guard against that.

thunder/executors/nvfuserex_impl.py
```diff
@@ -3193,3 +3193,70 @@ def argsort_transform(
...
+        nv_bias = getnv(bias, fd, lc_to_nv_map)
+    else:
+        nv_bias = None
+    if out_dtype is not None:
+        out_dtype = dtypes.to_dtype(out_dtype)
+    if out_dtype is not None:
+        dtype = lcdtype_to_nvdtype(out_dtype)
+    else:
+        dtype = None
+    # TODO: figure out how to use `scale_result` in `fd.ops.scaled_mm`
```
jjsjann123, line 3317: scale_result is just alph.

jjsjann123: output_block_scale_size and output_block_scale_dtype are meant for outputting quantized tensor.

thunder/core/prims.py
crcrpar: I was rather thinking of having a check to see the first two tensors satisfy the condition rather than having `stride_order` because I think callers of `scaled_mm` are responsible for having the inputs conforming to the condition.

thunder/executors/nvfuserex_impl.py
crcrpar: also, should we want `nv_enable_scaled_mm`-ish option like we have for linear/matmul

thunder/executors/nvfuserex_impl.py
jjsjann123: Note: @crcrpar I'm not sure how *stable* this API would be.

I don't yet know what the cutlass API would look like. I vaguely remember reading some other docs saying that global scaling factor on the output might need to be provided as input to grouped_mm.

nvMelissa:
Hi @crcrpar - this PR looks stale from 2 months ago. Would you please say if this still needs to get merged?

crcrpar:
let me close this one in favor of `torch.nn.functional.scaled_mm` (and `torch.nn.functional.scaled_grouped_mm` - https://github.com/pytorch/pytorch/blob/a3fe1825aa4d95707875ffa8f15fd00c3b94977d/torch/nn/functional.py#L6639 & https://github.com/pytorch/pytorch/blob/a3fe1825aa4d95707875ffa8f15fd00c3b94977d/torch/nn/functional.py#L6723

PR #2310, opened 2025-07-07
Add `torch.scalar_tensor` to `default_torch_ops.py`

crcrpar:
## What does this PR do?

As per title.
It seems that `openbmb/cpm-ant-10b` uses this op, which I noticed while running `benchmark_peft.py`

ref:
- `torch.scalar_tensor` signature: https://github.com/pytorch/pytorch/blob/9e5f4a844c0aebf964a435094005c92713fbe99a/aten/src/ATen/native/native_functions.yaml#L4689-L4693


IvanYashchuk:
@lantiga, @t-vi for merging.

t-vi (approved):
Thank you @crcrpar @IvanYashchuk  Could use a test.

PR #2052, opened 2025-05-09
Make alias updating the default in-place operator approach

beverlylytle:
As a follow-up to the introduction of using `update_aliases` to support in-place operations, this PR aims to make this approach the default instead of the functionalization approach.

PR #2314, opened 2025-07-08
Relax inplace sanity check

beverlylytle:
Changes in NvFuser mean that the sanity check is too strict and results in too many errors thrown.  This PR relaxes the check allowing output of a `copy_` operation to be used again within a fusion definition.  

cc @tfogal


thunder/core/transform_common.py
```diff
@@ -111,6 +111,49 @@ def check(inp, log_str):
             check(copy_to_out, "output")
 
 
+def _replace_copy_output_with_source(extrace: Trace):
```
crcrpar:
```suggestion
def _replace_copy_output_with_source(extrace: Trace) -> None:
```

thunder/core/transform_common.py
```diff
@@ -111,6 +111,49 @@ def check(inp, log_str):
...
+    """The is based on the sharp edge of nvfuser's `add_ouput(output, input)` interface,
```
crcrpar: I'm not quite sure but I guess this is typo of
```suggestion
    """This is based on the sharp edge of nvfuser's `add_ouput(output, input)` interface,
```

crcrpar (commented):
looks good to me, just leaving two non-critical, cosmetic comments

thunder/core/transform_common.py
beverlylytle: Exactly right, thank you.

t-vi:
I must admit that this seems to be something to be fixed in the nvfuser handling rather than in the copy_

beverlylytle:
> I must admit that this seems to be something to be fixed in the nvfuser handling rather than in the copy_

I agree with you, but the sanity check was added to Thunder a while ago, and I'd rather do what I can easily enough in Thunder than wait for a fix in nvfuser.

t-vi:
I'm not against adopting it as a hotfix, but I do think that things like these add up to render the traces a lot less readable than they should be, so I'd be keen to hear about a longer term plan involving fixing NVFuser.

beverlylytle:
@t-vi For readability, would you like that there's a comment included in the bsym header explaining that the source data is being used downstream rather than the output of the `copy_`?  Or do you have another suggestion for a Thunder-side change so that we can avoid disabling the inplace copy check in #2052?

As far as in-place plans for nvfuser are concerned, it would be good to discuss such a feature request with all the relevant parties, either at the weekly meeting or in an Issue.  Would such a discussion make you more comfortable merging this PR?

thunder/core/transform_common.py
```diff
@@ -111,6 +111,49 @@ def check(inp, log_str):
...
+
+        [t2] = nvFusion0(x, y)
+            # result = prims.mul(x, y)
+            # a = prims.copy_(result, x)
+            # t2 = prims.add(result, y)
+    This is only to be used when `skip_inplace_alias_updates` is set to False, and relies on the non-fusibility of prims.UPDATE_ALIASES.
+    """
+    from thunder.core.utils import consumers
+
+    nvfuser_symbols = (bsym for bsym in extrace.bound_symbols if bsym.sym.name.startswith("nvFusion"))
```
IvanYashchuk: @t-vi, what do you think about nvFuser-specific transforms in /core/transform_common.py? Would nvFuser executor file a more appropriate place for hosting these transforms?

thunder/core/transform_common.py
```diff
@@ -111,6 +111,49 @@ def check(inp, log_str):
             check(copy_to_out, "output")
 
 
+def _replace_copy_output_with_source(extrace: Trace) -> None:
+    """This is based on the sharp edge of nvfuser's `add_ouput(output, input)` interface,
```
IvanYashchuk: @jjsjann123, could you please review the changes here? Is this still a sharp edge in nvFuser? Is the target transformation correct?

IvanYashchuk (approved):
Neat fix for the "disable_inplace_copy_check=True" cases!
Let's get this in since it's a strictly positive change for the in-place support in the nvFuser executor in Thunder.

thunder/core/transform_common.py
```diff
@@ -111,6 +111,49 @@ def check(inp, log_str):
...
+    Anti-pattern:
+
+    .. code-block:: python
+
+        [t2] = nvFusion0(x, y)
+            # result = prims.mul(x, y)
+            # a = prims.copy_(result, x)
+            # t2 = prims.add(a, y)
+
+    Do not use `a` after it has been updated, use the `copy_from` variable `result` instead to reflect the dependency:
```
jjsjann123: I don't see the dependency issue there. In currently handling of nvfuser.copy_, the new output is done via a copy. https://github.com/Lightning-AI/lightning-thunder/blob/dbf6bad30cf38606f2feaec629f19357bba0b74d/thunder/executors/nvfuserex_impl.py#L2317-L2319

Can you point me to a repro where that's giving us issues?

crcrpar: `python thunder/benchmarks/benchmark_peft.py --model microsoft/GRIN-MoE --trust-remote-code --attn-implementation eager --max-steps 100 --mbs 1 --seq-length 128 --compile thunder --fixed-num-hidden-layers 2` should reproduce the bug

beverlylytle: It's true that running the peft benchmark will result in `_inplace_copy_sanity_check` raising an error (and that is true too of the smaller examples in `test_update_aliases` that are changed in this pull request).  However, I don't have a clear understanding of why this check checks both the `copy_to` and the output.  The [issue](https://github.com/Lightning-AI/lightning-thunder/issues/265) for adding this check mentions `copy_to` but the [PR](https://github.com/Lightning-AI/lightning-thunder/pull/285) also checks the output without further explanation.  @jjsjann123 is this check on the output unnecessary?

jjsjann123: Let me take a quick look there. Thanks a lot for pointing me to the repro.

jjsjann123: FYI @beverlylytle in the PR you linked, our copy_to implementation was a bit different.
https://github.com/Lightning-AI/lightning-thunder/blob/bcdf5f698edb577ace69002285da0b327e439c61/thunder/executors/nvfuserex_impl.py#L2021C1-L2031C1

It doesn't have an explicit `set` operation like we do now: https://github.com/Lightning-AI/lightning-thunder/blob/dbf6bad30cf38606f2feaec629f19357bba0b74d/thunder/executors/nvfuserex_impl.py#L2317

jjsjann123: > However, I don't have a clear understanding of why this check checks both the copy_to and the output.

nvfuser in-place update support isn't robust in the sense that we don't have dependency checks in the codegen for aliases. But we do have some segmentation logic to force kernel break up to avoid that. I'm not sure if it's bullet-proof though. tagging @Priya2698 

However, the issue we are seeing in thunder feels more like a transformation problem on the thunder/nvfuserex stack, since we are creating an illegal trace on the thunder side.

Though I'm having difficulty getting a minimum repro or following through with the transformation.

Priya2698: The segmentation pass resolves the RW race occuring when the aliased tensorview is in the path for broadcasts. We may have other patterns where a similar RW race can be triggered, and the pass is not a general dependency check for aliases in that sense.

beverlylytle: A small example that fails without this change is:
```
import torch
import thunder

def h(x):
    x.add_(1).sin().mul_(5)
    return x

a = torch.randn(5, device="cuda")

jfn = thunder.jit(h, skip_inplace_alias_updates=False, skip_inplace_functionalization=True)

actual = jfn(a)
```
with error
```
NotImplementedError: t2 = prims.sin(t1)  # t2: "cuda:0 f32[5]" trying to use <TensorProxy(name="t1", dtype=thunder.dtypes.float32, shape=(5,))> (the output of 'prims.copy_') as input, which is not safe. There is a risk of accessing the wrong memory. If you are sure you don't want to use this check, it can be disabled by setting `disable_inplace_copy_check=True` in `thunder.jit`.
```

jjsjann123: Thanks a lot for that example @beverlylytle . Thanks for refreshing the status of nvfuser inplace update support @Priya2698 .

The few question I have is:
1. the error is coming from a thunder check, which is probably out-dated at this point. since we have changed the behavior of `copy_` in nvfuser translation.

2. this PR doesn't change the generated kernel vs `main` with the check removed.
```
diff --git a/thunder/core/transform_common.py b/thunder/core/transform_common.py
index 14443b6c..d3bd6098 100644
--- a/thunder/core/transform_common.py
+++ b/thunder/core/transform_common.py
@@ -107,8 +107,8 @@ def _inplace_copy_sanity_check(extrace: Trace):
                             f" There is a risk of accessing the wrong memory. If you are sure you don't want to use this check, it can be disabled by setting `disable_inplace_copy_check=True` in `thunder.jit`."
                         )
 
-            check(copy_to_arg, "'copy_to' argument")
-            check(copy_to_out, "output")
+            #check(copy_to_arg, "'copy_to' argument")
+            #check(copy_to_out, "output")
```

beverlylytle: Summarizing an outside discussion, changes made in NvFuser since the addition of the sanity check mean that the check can be relaxed and the swapping of output and source is no longer necessary.

@t-vi

thunder/core/transform_common.py
beverlylytle: Resolving as out of date.

thunder/core/transform_common.py
beverlylytle: @jjsjann123 If you would please re-review this change...

PR #2317, opened 2025-07-08
Fix dataflow ordering for recomputed symbols

### LINKED ISSUE #2316: Autodiff does not place recomputation in the correct order
## 🐛 Bug

While debugging TEv2 checkpointing I've noticed that some symbols are placed in the wrong order and proxies are used before they are called

### To Reproduce

<details>

<summary> executor for repro </summary>

```python

from functools import partial

import torch
import torch.utils.checkpoint

from thunder.core.prims import get_grad, put_grad
from thunder.core.proxies import TensorProxy
from thunder.extend import OperatorExecutor, register_executor

repro_ex = OperatorExecutor("repro_ex")
register_executor(repro_ex)


def _pass_token_meta(tokens):
    return (*(TensorProxy(like=token) for token in tokens),)


def _pass_token_impl(tokens):
    return tokens


_pass_token = repro_ex.register_operator("pass_token", meta=_pass_token_meta, fn=_pass_token_impl)


def exec_transform(a, w, bias):
    primal, _ = _pass_token((a, w))
    return primal


def grad_transform(a, w, bias):
    aw, bw = _pass_token((a, w))
    primal, cd = _pass_token((bw, aw))
    grad_out = get_grad(primal)
    grad_a, grad_w, _ = _pass_token((grad_out, cd, aw))
    put_grad(a, grad_a)
    put_grad(w, grad_w)
    return primal


repro_ex.register_implementation(
    ltorch.linear,
    checker=lambda a, w, bias: True,
    execution_transform=exec_transform,
    grad_transform=grad_transform,
)

checkpoint_fn = partial(torch.utils.checkpoint.checkpoint, use_reentrant=False)


def fn(a, w0, w1):
    b = checkpoint_fn(torch.nn.functional.linear, a, w0)
    return torch.nn.functional.linear(b, w1)


x = torch.randn((16, 16), device="cuda", dtype=torch.float32, requires_grad=True)
w0 = torch.randn((16, 16), device="cuda", dtype=torch.float32, requires_grad=True)
w1 = torch.randn((16, 16), device="cuda", dtype=torch.float32, requires_grad=True)


jf = thunderfx(fn, executors=[repro_ex])
out = jf(x, w0, w1)
print(jf.last_backward_traces[-1])

```

</details>

Produces the following trace where `bw_t4` is used before it is created

```python
  # cotangents: "Collection"
  C0, _, = saved_for_backward
  # C0: "Collection"
  # None
  del saved_for_backward
  t18, = cotangents
  # t18: "cuda:0 f32[16, 16]"
  del cotangents
  t17, t14, l_a_, l_w0_, = C0
  # t17: "cuda:0 f32[16, 16]"
  # t14: "cuda:0 f32[16, 16]"
  # l_a_: "cuda:0 f32[16, 16]"
  # l_w0_: "cuda:0 f32[16, 16]"
  clear_mutable_collection(C0)
  del C0
  (t19, t20, _) = dummy_pass_token((t18, t17, t14))
  del t18, t17, t14
  (_, bw_t7) = dummy_pass_token1((bw_t5, bw_t4))
  del bw_t5
  (bw_t4, _) = dummy_pass_token((l_a_, l_w0_))
  del l_a_, l_w0_
  (t9, t10, _) = dummy_pass_token((t19, bw_t7, bw_t4))
  del t19, bw_t7, bw_t4
  return (t9, t10, t20)
```

## Analysis

For what I've understood the main problem lies here 

https://github.com/Lightning-AI/lightning-thunder/blob/addb7036289ba8f5ec34cc077b23dee35cf6bc1d/thunder/transforms/autodiff.py#L383 

where we use the order the args appears in the boundsymbol to "decide" the insertion order. The easies fix would be to just sort `processed_bsyms` before the `self.add_processed_bsyms(processed_bsyms)` call. However, I'd go a step further and avoid using `bw_idx` too. 

Draft PR with fix here #2317 

cc. @beverlylytle

riccardofelluga:
## What does this PR do?

Fixes #2316.

This patch finds producer symbols and sorts them in the dataflow order so the repro in #2316 outputs the following(bw_t4 is created before it's used):

```python
  # cotangents: "Collection"
  C0, _, = saved_for_backward
  # C0: "Collection"
  # None
  del saved_for_backward
  t18, = cotangents
  # t18: "cuda:0 f32[16, 16]"
  del cotangents
  t17, t14, l_a_, l_w0_, = C0
  # t17: "cuda:0 f32[16, 16]"
  # t14: "cuda:0 f32[16, 16]"
  # l_a_: "cuda:0 f32[16, 16]"
  # l_w0_: "cuda:0 f32[16, 16]"
  clear_mutable_collection(C0)
  del C0
  (t19, t20, _) = pass_token((t18, t17, t14))
  del t18, t17, t14
  (bw_t4, bw_t5) = pass_token((l_a_, l_w0_))
  del l_a_, l_w0_
  (_, bw_t7) = pass_token((bw_t5, bw_t4))
  del bw_t5
  (t9, t10, _) = pass_token((t19, bw_t7, bw_t4))
  del t19, bw_t7, bw_t4
  return (t9, t10, t20)
```

### DIFF
--- a/thunder/transforms/autodiff.py
+++ b/thunder/transforms/autodiff.py
@@ -363,9 +363,49 @@ def grad_transform_on_trace(trace, /, *args, **kwargs):
         def __init__(self, trace):
             super().__init__(trace)
             self.backward_part_bsyms_recomputed = {}
+            self.already_processed_recomputations = set()
+
+        def find_recomputation_symbols_for_bsym(self, bsym) -> list[tuple[str, BoundSymbol]]:
+            need_sorting = list()
+            queue: list[Proxy] = list(
+                filter(lambda p: p.name in self.backward_part_bsyms_recomputed, bsym.flat_proxy_args)
+            )
+            visited = set(map(lambda p: p.name, queue))
+            while queue:
+                proxy = queue.pop()
+                producer_record = self.backward_part_bsyms_recomputed.get(proxy.name)
+                if producer_record is not None:
+                    producer, _ = producer_record
+                    need_sorting.append((proxy.name, producer))
+                    for arg in producer.flat_proxy_args:
+                        arg_name = arg.name
+                        if arg_name not in visited:
+                            queue.append(arg)
+                            visited.add(arg_name)
+
+            if need_sorting:
+                remaining = dict(need_sorting)
+                sorted_recomputation = []
+                while remaining:
+                    sorted_recomputation_names = []
+                    for name, producer in remaining.items():
+                        ready = True
+                        for dep in producer.flat_proxy_args:
+                            if dep.name in remaining:
+                                ready = False
+                                break
+                        if ready:
+                            sorted_recomputation_names.append(name)
+                    for name in sorted_recomputation_names:
+                        sorted_recomputation.append((name, remaining[name]))
+                        del remaining[name]
+
+                return sorted_recomputation
+
+            return []
 
         def process_bsym(self, bsym: BoundSymbol) -> None:
-            processed_bsyms = [bsym.from_bsym()]
+            processed_bsyms = []
             if _should_recompute_bsym_in_backward(bsym) and BoundSymbolTag.BACKWARD not in bsym.tags:
                 bsym_out_names = {o.name for o in bsym.flat_proxy_outs}
                 nbsym = bsym.from_bsym()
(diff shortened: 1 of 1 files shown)


thunder/transforms/autodiff.py
```diff
@@ -363,40 +363,78 @@ class InsertRecomputationsProcessor(TraceSubstitutionProcessor):
...
+                for name, rec_bsym in recomputed_bsyms:
+                    if name in self.already_processed_recomputations:
+                        continue
+                    # To avoid name clashes, we create new output proxies.
+                    # This relies on the fact that all backward operations occur after get_grad,
+                    # which is no longer true after fusion passes.
+                    with tracectx(self.new_trace):
+                        for output in rec_bsym.flat_proxy_outs:
+                            self.already_processed_recomputations.add(output.name)
+                            # Avoid double-prefixing if the name already starts with "bw_"
```
beverlylytle: oh... if there are already things prefixed with "bw_" does that mean there might be a clash between an existing "bw_*" variable and one that is being created here?  Well, not clash exactly because `replace_name` will postpend a `_1` giving something like `bw_t406_1` if `bw_t406` already exists.

thunder/transforms/autodiff.py
```diff
@@ -363,40 +363,78 @@ class InsertRecomputationsProcessor(TraceSubstitutionProcessor):
...
+            return []
 
         def process_bsym(self, bsym: BoundSymbol) -> None:
-            processed_bsyms = [bsym.from_bsym()]
+            processed_bsyms = []
             if _should_recompute_bsym_in_backward(bsym) and BoundSymbolTag.BACKWARD not in bsym.tags:
                 bsym_out_names = {o.name for o in bsym.flat_proxy_outs}
                 nbsym = bsym.from_bsym()
                 nbsym.tags.add(BoundSymbolTag.BACKWARD)
                 bsym_rec = (nbsym, bsym_out_names)
```
beverlylytle: Looks like `bsym_out_names` isn't used anymore, so I think `backward_part_bsyms_recomputed` can be simplified to something of the form `{arg.name: bsym for arg in bsym.flat_proxy_outs}` for each bsym that should be recomputed..

thunder/transforms/autodiff.py
```diff
@@ -363,40 +363,78 @@ class InsertRecomputationsProcessor(TraceSubstitutionProcessor):
         def __init__(self, trace):
             super().__init__(trace)
             self.backward_part_bsyms_recomputed = {}
+            self.already_processed_recomputations = set()
+
+        def find_recomputation_symbols_for_bsym(self, bsym) -> list[tuple[str, BoundSymbol]]:
+            need_sorting = list()
```
beverlylytle: could be a `dict` from the start

beverlylytle (approved):
Looks great!

thunder/transforms/autodiff.py
riccardofelluga: Ohh you're right!

thunder/transforms/autodiff.py
riccardofelluga: My bad, it was a leftover from earlier experimentation. Actually that condition will never be met because of the check earlier in the `already_processed_recomputations` set.

thunder/transforms/autodiff.py
```diff
@@ -362,41 +362,75 @@ def shallow_copy_if_input(p):
...
+                    need_sorting[proxy.name] = producer
+                    for arg in producer.flat_proxy_args:
+                        arg_name = arg.name
+                        if arg_name not in visited:
+                            queue.append(arg)
+                            visited.add(arg_name)
+
+            sorted_recomputation = []
+
+            if need_sorting:
```
beverlylytle: Probably don't need this conditional...

riccardofelluga:
Ping @t-vi @lantiga what do you think of this solution?

t-vi (approved):
Thank you @riccardofelluga @beverlylytle 

I'm rather unconvinced that we should need or have three loops here instead of one, but I don't have a better fix.

PR #2318, opened 2025-07-09
Add scalar tensor input to full_sample_generator

IvanYashchuk:
Expands full's sample generator for OpInfo-based testing with a scalar tensor input.

cc @mruberry


t-vi (approved):
Thank you @IvanYashchuk

PR #2319, opened 2025-07-09
update TE test

kshitij12345:

PR #2320, opened 2025-07-09
Run TE tests in CI

### LINKED ISSUE #196: [ci] : We should add a CI flow with TransformerEngine installed so that we can run the relevant tests.

kshitij12345:
Fixes #196 

TODO
* [x] Correctly install TE stable

NOTE: 
* This only adds test for TE stable. Will follow-up for the latest main in a separate PR.
* The failure for the TE job is because of #2319 
* We don't run the distributed tests as the instance seems to have 1 GPU (will tackle this separately later).


kshitij12345:
This has caught another error which occurs on platform where default recipe is DelayedScaling - have PR #2322 fixing that.

t-vi (approved):
Look great, thank you @kshitij12345

riccardofelluga (approved):
Thanks for pinging! LGTM

PR #2326, opened 2025-07-11
Creates a thin way to plum _grouped_mm to the corresponding API exposed in nvFuser.

protonu:
This PR adds a thin layer in Thunder to call _grouped_mm in nvFuser (also the torch executor).


thunder/core/prims.py
```diff
@@ -3763,6 +3764,65 @@ def sort_meta(a: TensorProxy, /, dim: int, descending: Number, sorted: Number) -
 sort = make_prim(PrimIDs.SORT, "sort", meta=sort_meta)
 
 
+def _grouped_mm_meta(
+    a: TensorProxy, b: TensorProxy, offs: None | TensorProxy = None, bias: None | TensorProxy = None, dtype=None
```
crcrpar: nit-picking. type annotation for `dtype`.

I don't remember how this file imports `thunder.core.dtypes` though
```suggestion
    a: TensorProxy, b: TensorProxy, offs: None | TensorProxy = None, bias: None | TensorProxy = None, dtype: dtyps.dtype | None = None
```

thunder/clang/__init__.py
```diff
@@ -2042,3 +2042,15 @@ def argsort(a: TensorProxy, /, dim: None | int = None, descending: bool = False,
...
     return prims.argsort(a, dim, descending, stable)
+
+
+@clangop()
+def _grouped_mm(
+    a: TensorProxy,
+    b: TensorProxy,
+    offsets: TensorProxy = None,  # Unused in meta, but type-checked
+    bias: TensorProxy = None,
+    dtype=None,
```
crcrpar: nit-picking: type annotation. note that I've not checked how `dtypes` is imported
```suggestion
    offsets: TensorProxy | None = None,  # Unused in meta, but type-checked
    bias: TensorProxy | None = None,
    dtype: dtypes.dtype | None = None,
```

thunder/executors/nvfuserex_impl.py
```diff
@@ -3188,6 +3188,35 @@ def argsort_transform(
...
 
+
+def _grouped_mm_check_(
+    a: TensorProxy,
+    b: TensorProxy,
+    offs: TensorProxy,
+    bias: TensorProxy = None,
+    out_dtype=None,
+) -> bool:
+    return True
```
crcrpar: I guess it'd be better to have nvfuser version check and `are_supported_tensors` here 

https://github.com/Lightning-AI/lightning-thunder/blob/dbf6bad30cf38606f2feaec629f19357bba0b74d/thunder/executors/nvfuserex_impl.py#L222

thunder/executors/nvfuserex_impl.py
```diff
@@ -3188,6 +3188,35 @@ def argsort_transform(
...
+    bias: TensorProxy = None,
+    out_dtype=None,
+    *,
+    fd: FusionDefinition,
+    lc_to_nv_map: dict,
+) -> list[TensorLike]:
+    nva = getnv(a, fd, lc_to_nv_map)
+    nvb = getnv(b, fd, lc_to_nv_map)
+    nvoffs = getnv(offs, fd, lc_to_nv_map) if offs is not None else None
+    return fd.ops.grouped_mm(nva, nvb, nvoffs)
```
crcrpar: it seems that `bias` is not used.
Could you make use of it if it's supported or update the checker to return when `bias is not None`?

thunder/executors/torchex.py
```diff
@@ -1819,6 +1821,8 @@ def _pad_prim_impl(
 
 _register_implementation(ltorch.baddbmm, baddbmm, checker=_always_executable)
 _register_implementation(ltorch.bmm, bmm, checker=_always_executable)
+_register_implementation(prims._grouped_mm, _grouped_mm, checker=_always_executable)
+_register_implementation(ltorch._grouped_mm, _grouped_mm, checker=_always_executable)
```
crcrpar: As per https://github.com/pytorch/pytorch/blob/b0556110e58e3bcf2c872b933e4fd4a0d34398ad/aten/src/ATen/native/cuda/Blas.cpp#L1677-L1686, the current implementation disallows BFlo16 which can be reflected here as well (though I feel just letting the implementation fail not that bad)

thunder/core/prims.py
```diff
@@ -3763,6 +3764,65 @@ def sort_meta(a: TensorProxy, /, dim: int, descending: Number, sorted: Number) -
...
+        dtype: Optional dtype for output (unused in meta, but type-checked)
+
+    Returns:
+        TensorProxy with shape (groups, m, n) or (m, n)
+    """
+    # Validate types
+    utils.check_type(a, TensorProxy)
+    utils.check_type(b, TensorProxy)
+    if offs is not None:
+        utils.check_type(offs, TensorProxy)
```
kshitij12345: I think we should also add similar checks for `offset` if present.

https://github.com/pytorch/pytorch/blob/85320336799e38411d15c0e159b41248cda01218/aten/src/ATen/native/cuda/Blas.cpp#L1692-L1697

thunder/core/prims.py
```diff
@@ -3763,6 +3764,65 @@ def sort_meta(a: TensorProxy, /, dim: int, descending: Number, sorted: Number) -
...
+        utils.check(a.shape[1] == b.shape[1], lambda: f"Inner dimension mismatch: {a.shape[2]} vs {b.shape[1]}")
+        out_shape = (a.shape[0], b.shape[2])
+    # 3D case: grouped matmul
+    elif a.ndim == 3 and b.ndim == 3:
+        utils.check(a.shape[0] == b.shape[0], lambda: f"Group count mismatch: {a.shape[0]} vs {b.shape[0]}")
+        utils.check(a.shape[2] == b.shape[1], lambda: f"Inner dimension mismatch: {a.shape[2]} vs {b.shape[1]}")
+        out_shape = (a.shape[0], a.shape[1], b.shape[2])
+    else:
+        raise AssertionError(
+            f"grouped_mm expects both inputs to be 2D or both to be 3D, got shapes {a.shape} and {b.shape}"
```
kshitij12345: Based on the `if` and `elif` above, we would hit this else branch if both `a` and `b` are 2D but error message says `grouped_mm expects both inputs to be 2D` which seems confusing.

thunder/executors/torchex.py
kshitij12345: Isn't it the opposite that it only allows BFloat16 inputs? We should update the checker accordingly.

kshitij12345:
Looking at the CI [error](https://github.com/Lightning-AI/lightning-thunder/actions/runs/16226276918/job/45818910468?pr=2326#step:8:32), we will need to guard the registration of the `torchsymbol` based on PyTorch version as `torch._grouped_mm` seems to be available only in the nightlies.

thunder/executors/torchex.py
crcrpar: you're correct. That was my silly typo. Thank you

wujingyue:
I removed several YAGNIs and fixed several bugs. It's ready to review again.

thunder/core/prims.py
```diff
@@ -3763,6 +3764,62 @@ def sort_meta(a: TensorProxy, /, dim: int, descending: Number, sorted: Number) -
...
+    else:
+        utils.check(False, lambda: f"Unexpected shape combination: {a.shape} and {b.shape}")
+
+    utils.check_same_dtype(a, b)
+    utils.check(a.dtype in dtypes.float_math_dtypes, lambda: f"`a` must be 16-bit float or higher, got {a.dtype}")
+    utils.check(utils.is_integer_dtype(offs.dtype), lambda: f"`offs` must be integers, got {offs.dtype}")
+
+    utils.check_same_device(a, b)
+
+    return TensorProxy(like=a, shape=out_shape, dtype=a.dtype)
```
crcrpar: nit-picking `dtype=a.dtype` can be obviated

thunder/torch/__init__.py
```diff
@@ -5620,6 +5621,13 @@ def linear(a: TensorLike, w: TensorLike, /, bias: None | TensorLike = None) -> T
     return prims.linear(a, w, bias)
 
 
+if LooseVersion(torch.__version__) >= "2.8":
+
+    @torchsymbol(torch._grouped_mm)
+    def _grouped_mm(a: TensorProxy, b: TensorProxy, offs: TensorProxy) -> TensorProxy:
```
crcrpar: It seems that `torch._grouped_mm` can take bias and out_dtype as per https://github.com/pytorch/pytorch/blob/c6b479bc09090336dcd1c97f5e368e76490065ea/tools/autograd/derivatives.yaml#L1209. Could you match the signature?

thunder/executors/torchex.py
crcrpar, line 1: I think it'd be better to have grouped_mm in torchex as well.

thunder/executors/nvfuserex_impl.py
crcrpar: I'm afraid I don't think this comment is resolved. What's the plan about `bias` and `out_dtype`?

thunder/torch/__init__.py
wujingyue: Thanks for the pointer! 

> Could you match the signature?

Yes. However, in this PR, I don't plan to support bias or out_dtype so `utils.check`. This is for YAGNI. Sounds good to you?

thunder/executors/nvfuserex_impl.py
wujingyue: I checked bias and out_dtype in `torch/__init__.py`. Is that sufficient?

thunder/torch/__init__.py
crcrpar: I think my ask is to update torchex to have `grouped_mm` and move the check of `bias` and `dtype` to `_grouped_mm_check_` of nvfuserex_impl so that the coverage would be a bit clearer.
Would it sound reasonable?

thunder/executors/torchex.py
wujingyue: Done. I had to add several version checks though.

thunder/clang/__init__.py
```diff
@@ -2042,3 +2042,12 @@ def argsort(a: TensorProxy, /, dim: None | int = None, descending: bool = False,
     dim = utils.canonicalize_dim(a.ndim, dim)
 
     return prims.argsort(a, dim, descending, stable)
+
+
+@clangop()
+def _grouped_mm(
```
wujingyue: Does this need to match torch's signature as well? What's the difference between `clang` and `torch`?

thunder/torch/__init__.py
wujingyue: Following the YAGNI principle, I recommend we do that in the future. It's unclear to me what the semantics are when offsets=None and/or bias!=None and/or dtype!=None. So it's unclear to me how to write [_grouped_mm_meta](https://github.com/Lightning-AI/lightning-thunder/pull/2326/files#diff-16f33141471a03112ba59caeb9fc36095d5df9b1564e79994e63d8d4ac175606R3767) for those cases. I'd rather we keep the Thunder prims simple for the cases we need. Also note that `torch._grouped_mm` is an experimental API so it's more likely to change than other APIs.

thunder/clang/__init__.py
kshitij12345: `clangop` for `_grouped_mm` is not used elsewhere and I don't think we need it. We should remove it.

thunder/executors/nvfuserex_impl.py
```diff
@@ -3188,6 +3188,36 @@ def argsort_transform(
...
+        return False
+
+    return nvfuser_version() >= LooseVersion("0.2.28")
+
+
+def _grouped_mm_transform(
+    a: TensorProxy,
+    b: TensorProxy,
+    offs: TensorProxy,
+    bias: TensorProxy = None,
```
kshitij12345: Since `prims._grouped_mm` doesn't have `bias` and `out_dtype` argument, I don't think the transform needs them as well.

kshitij12345 (approved):
LGTM, thanks @wujingyue @protonu

thunder/torch/__init__.py
crcrpar: okay, understood.

kshitij12345:
Have filed a follow-up issue to add grad rule for `prims._grouped_mm` - https://github.com/Lightning-AI/lightning-thunder/issues/2362

wujingyue:
@mruberry could you review and merge this?

thunder/tests/opinfos.py
```diff
@@ -347,6 +347,7 @@ def __init__(
         singularity_fn_producer=None,
         test_torch_compile_executor=False,
         instantiate_complex_tests=False,
+        no_fallback_with_double_inputs=False,
```
IvanYashchuk: It would be useful to have some documentation for this flag. What does it do, and why is it necessary for tests?

wujingyue: I rewrote the reference implementation and removed this flag completely.

jjsjann123:
tagging @mruberry for a royal decree.

wujingyue:
@t-vi I believe this is ready to merge. ~The `grads` test error is unrelated, but could you check `testing-guardian`?~

The only CI failure seems to be an unrelated timeout. 
```
FAILED thunder/tests/test_dynamo.py::test_WallTime_KernelTime - Failed: Timeout (>240.0s) from pytest-timeout.
= 1 failed, 2138 passed, 196 skipped, 24 xfailed, 7 xpassed, 81299 warnings in 524.00s (0:08:44) =
```

IvanYashchuk:
@t-vi, could you please approve this for merging? Thank you!

PR #2328, opened 2025-07-14
[ruff linter] Enable isort for consistent import order

crcrpar:
## What does this PR do?

As per title, this enables `isort` via ruff.


crcrpar:
I'm overwhelmed by the failures and don't have enough bandwidth to make it clean.

PR #2331, opened 2025-07-16
[SMOKE_TEST][DO_NOT_REVIEW]

jjsjann123:
Just testing OOM on CI with embedding test.


jjsjann123:
Looks like OOM is gone. I'll move back to the original PR.

PR #2337, opened 2025-07-17
[WIP] [thunderfx] Intermediate mark non differentiable

### LINKED ISSUE #1950: [thunderfx] Problem with buffer updated during the forward - RuntimeError: Trying to backward through the graph a second time (or directly access saved tensors after they have already been freed).
This was found while running `BAAI/Bunny-v1_1-4B` (see `Repro with Model` below)

Minimal Repro
```python
import torch
import thunder

class Model(torch.nn.Module):
    def __init__(self, *args, **kwargs):
        self.fc = torch.nn.Linear(2, 2)
        self.register_buffer("buf", None, False)

        # Ok if buffer is already populated.
        # self.buf = torch.ones(2, 2, device=x.device)

    def forward(self, x):
        if self.buf is None:
            self.buf = torch.randn(1, 2, device=x.device)
        return self.fc(x) + self.buf
    
with torch.device("cuda"):
    org_m = Model()
    # m = org_m
    m = thunderfx(org_m)
    # m = thunder.jit(org_m)

    x = torch.randn(2, 2)
    m(x).sum().backward()

    # With thunderfx, tensor([[1., 1.]], device='cuda:0', grad_fn=<ThunderFunctionBackward>)
    # With eager, tensor([[1., 1.]], device='cuda:0')
    # With thunder, None
    print(org_m.buf)

    m(x).sum().backward()

```

With thunderfx path, the buffer that is saved has `grad_fn=<ThunderFunctionBackward>` and hence in the second iteration during backward, we incorrectly try to backward on this already freed part of the computation graph.

thunderfx's torch autograd integration should mark output from no_grad region or requires_grad=False (need to propagate this correctly see https://github.com/Lightning-AI/lightning-thunder/issues/1733 and https://github.com/Lightning-AI/lightning-thunder/issues/1768) as [non-differentiable](https://pytorch.org/docs/stable/generated/torch.autograd.function.FunctionCtx.mark_non_differentiable.html) so that buffer won't hold onto the previously freed graph.

For thunder.jit, it just ignores the buffer being updated and generates new value for `self.buf` for every invocation (will file a seperate issue)

Repro with Model
```python
import torch
import thunder
import transformers

with torch.device("cuda"):
    model_name = "BAAI/Bunny-v1_1-4B"
    config = transformers.AutoConfig.from_pretrained(model_name, trust_remote_code=True)
    config.num_hidden_layers = 1
    model = transformers.AutoModelForCausalLM.from_config(
                        config,
                        torch_dtype=torch.float16,
                        trust_remote_code=True,
                        attn_implementation="eager",
                    )

    model = thunderfx(model)

    ids = torch.randint(0, 10, (1, 128))
    output = model(ids, labels=ids)
    output.loss.backward()

    ids = torch.randint(0, 10, (1, 128))
    output = model(ids, labels=ids)
    output.loss.backward()

```

cc: @riccardofelluga

kshitij12345:
Fixes #1950

### DIFF
--- a/thunder/executors/torch_autograd.py
+++ b/thunder/executors/torch_autograd.py
@@ -1,5 +1,6 @@
 from dataclasses import replace
-from typing import TYPE_CHECKING, Sequence
+from typing import TYPE_CHECKING
+from collections.abc import Sequence
 
 import torch
 


kshitij12345:
Closing in favour of #2348

PR #2342, opened 2025-07-18
Specify torch.randint's default dtype

### LINKED ISSUE #2341: Wrong dtype for randint
The operator `randint` in PyTorch returns int64 when no dtype is specified.  However when `randint` is Thunder-jitted, it returns float32.  The following example shows where this inconsistency between PyTorch and Thunder can result in an error.

```
import thunder
import torch


def f(x, y):
    x.scatter_add_(index=torch.randint(0, 1, x.shape), src=y, dim=-1)
    return x

x = torch.randn(2, 3)
y = torch.randn(2, 3)

x_ = x.clone().detach()
y_ = y.clone().detach()

z_ = f(x_, y_)  # this is fine.

jf = thunder.jit(f)
z = jf(x, y)  # this throws an error `RuntimeError: index dtype=thunder.dtypes.float32 was not an integer dtype`
```

shino16:
Fixes #2341. Unlike most tensor creation ops, `torch.randint`'s default type is `torch.int64`. This is now specified in `ltorch.randint`.

cc @beverlylytle


thunder/torch/__init__.py
```diff
@@ -850,7 +850,7 @@ def randint(
     utils.check(generator is None, lambda: "generator is not None which is currently unsupported", NotImplementedError)
     utils.check(out is None, lambda: "out is not None which is currently unsupported", NotImplementedError)
     device = to_device(maybe_get_default_device(device))
-    dtype = to_dtype(maybe_get_default_dtype(dtype))
+    dtype = to_dtype(dtype or torch.int64)
```
beverlylytle: Very tidy!

thunder/tests/test_core.py
```diff
@@ -2864,6 +2864,16 @@ def fn():
     assert jfn() == torch.int64
 
 
+def test_randint_default_dtype():
+    # If any of start, end, or stop are floating-point, the dtype is inferred to be the default dtype, see get_default_dtype().
```
beverlylytle: This comment looks like copypasta from `arange`, and it doesn't seem relevant here.  Could you remove it?

thunder/tests/test_core.py
```diff
@@ -2864,6 +2864,16 @@ def fn():
...
+    # Otherwise, the dtype is inferred to be torch.int64.
+    def fn():
+        return torch.randint(0, 5, (2, 3))
+
+    jfn = thunder.jit(fn)
+    assert jfn().dtype == fn().dtype == torch.int64
```
beverlylytle, line 2871: It would be good to have a second test case where the dtype is specified to something other than an int type.

beverlylytle:
There's a lot of red CI here, but none look related to this change.

thunder/tests/test_core.py
shino16: Thanks, it was indeed irrelevant.

thunder/tests/test_core.py
shino16: In my understanding, the diff below represents the test cases where dtype is specified. I added `dtype=datatypes.floating`.

https://github.com/Lightning-AI/lightning-thunder/pull/2342/commits/09441303e1b58f4da178364f48709aa1bf553ba8#diff-039a4ceefc0c6622f561125634d69c81b24012cfc7be6bc92f0ecc14d6ba37bcR6817-R6823

beverlylytle: Does that test compare the dtypes of the outputs?

shino16: It is tested by `thunder.tests.test_ops.test_core_vs_torch_consistency`. This internally uses `torch.testing.assert_close` for comparison by default, which compares dtypes too. Its Opinfo is:
https://github.com/shino16/lightning-thunder/blob/695beefe564126c28a2a6c7bad72104a3fe57d3f/thunder/tests/opinfos.py#L6817-L6824

beverlylytle: Oh boy, there's a lot of indirection in trying to figure out how these results are checked.  I was initially thrown off because `check_dtype=False` [in a couple of places here](https://github.com/Lightning-AI/lightning-thunder/blob/19e6fa415d3021e9d96e8e2ccaa72856affec97a/thunder/tests/framework.py#L105), but it looks like as long as there is not a custom comparator set in the opinfo, we will [fall back](https://github.com/Lightning-AI/lightning-thunder/blob/19e6fa415d3021e9d96e8e2ccaa72856affec97a/thunder/tests/framework.py#L354) to `assert_close`.  So, great, thanks for looking into that!

beverlylytle:
CI failing due to #2343

t-vi (approved):
Thank you @shino16 @beverlylytle

Issue #2330, opened 2025-07-15
[benchmark_peft] Convoluted Type Error from `torch.baddbmm` in `thunderfx`'d `stabilityai/japanese-stablelm-base-alpha-7b`


crcrpar:
### To Reproduce

Run the following code sample, the error will appear.
Interestingly, `thunder.jit` with the same executors doesn't cause the error.

```
1. Initializing Attention module...
   ...module initialized.
2. Compiling with Thunder...
   Thunder used executors: ['cudnn', 'sdpa', 'torchcompile_xentropy', 'nvfuser']
   ...compilation finished.
3. Preparing dummy input...
   ...dummy input prepared with shape: torch.Size([1, 128, 4096])
4. Running forward pass...

--- SCRIPT FAILED ---
Forward pass failed with exception:
Traceback (most recent call last):
  File "/opt/pytorch/lightning-thunder/repro_attention_bug.py", line 210, in <module>
    outputs = model(hidden_states, attention_mask, position_ids)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/pytorch/lightning-thunder/thunder/dynamo/compiler.py", line 236, in __call__
    return self._func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/torch/_dynamo/eval_frame.py", line 372, in __call__
    return super().__call__(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/torch/nn/modules/module.py", line 1767, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/torch/nn/modules/module.py", line 1778, in _call_impl
    return forward_call(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/torch/_dynamo/eval_frame.py", line 699, in compile_wrapper
    return fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^
[...]

PR #2348, opened 2025-07-22
[thunderfx] mark output non-differentiable based on FXGraph output node inspection

### LINKED ISSUE #1950: [thunderfx] Problem with buffer updated during the forward - RuntimeError: Trying to backward through the graph a second time (or directly access saved tensors after they have already been freed).

kshitij12345:
Fixes #1950 #2330 

Problem - Thunder doesn't propagate requires_grad correctly and hence when we connect to PyTorch's autograd, we mark all outputs as differentiable which can lead to errors.

Solution - For thunderfx, we check the `grad_fn` on outputs of the FXGraph (which propagate the requires_grad correctly) to determine if an output should be marked as differentiable or not while connecting with PyTorch's autograd. We pass this information regarding which output is differentiable using compile options to thunder.jit interface.

Other approach (which should work for thunder.jit as well) that I tried was to transform initial trace acquired by thunder to execution trace with no decorators and PyTorch executor and run it with Fake inputs, but this will not work correctly when we have `torch.no_grad`  as the corresponding symbols for `no_grad` act as a marker during trace time (see Sample trace below) and currently don't have an executor claiming them. We can potentially register them with PyTorch executor to do the correct thing which should most likely get it working.

Sample trace
```python
def computation(l_x_):
  # l_x_: "cpu bf16[3]"

  # <eval_with_key>.4:5:            _set_grad_enabled = torch._C._set_grad_enabled(False);  _set_grad_enabled = None
  ltorch._set_grad_enabled_with_warning(False)

  # <eval_with_key>.4:6:            y = l_x_ + 1
  y = ltorch.add(l_x_, 1, alpha=1)  # y: "cpu bf16[3]"
    # t0 = prims.convert_element_type(l_x_, dtypes.float32)  # t0: "cpu f32[3]"
    # _ = prims.convert_element_type(1, float)
    # t1 = prims.add(t0, 1.0)  # t1: "cpu f32[3]"
    # y = prims.convert_element_type(t1, dtypes.bfloat16)  # y: "cpu bf16[3]"

  # <eval_with_key>.4:7:            _set_grad_enabled_1 = torch._C._set_grad_enabled(True);  _set_grad_enabled_1 = None
  ltorch._set_grad_enabled_with_warning(True)

  # <eval_with_key>.4:8:            sin = l_x_.sin()
  sin = ltorch.sin(l_x_)  # sin: "cpu bf16[3]"
    # t3 = prims.convert_element_type(l_x_, dtypes.float32)  # t3: "cpu f32[3]"
    # t4 = prims.sin(t3)  # t4: "cpu f32[3]"
    # sin = prims.convert_element_type(t4, dtypes.bfloat16)  # sin: "cpu bf16[3]"

  # <eval_with_key>.4:9:            add_1 = l_x_ + sin;  l_x_ = sin = None
  add_1 = ltorch.add(l_x_, sin, alpha=1)  # add_1: "cpu bf16[3]"
    # t6 = prims.convert_element_type(l_x_, dtypes.float32)  # t6: "cpu f32[3]"
    # t7 = prims.convert_element_type(sin, dtypes.float32)  # t7: "cpu f32[3]"
    # t8 = prims.add(t6, t7)  # t8: "cpu f32[3]"
    # add_1 = prims.convert_element_type(t8, dtypes.bfloat16)  # add_1: "cpu bf16[3]"
  return (add_1, y)
```

NOTE: The ideal solution will be to propagate requires_grad correctly in thunder (see https://github.com/Lightning-AI/lightning-thunder/issues/1768).

### DIFF
--- a/thunder/core/codeutils.py
+++ b/thunder/core/codeutils.py
@@ -152,12 +152,10 @@ def to_printable(
     if isinstance(x, ProxyInterface):
         return x
 
-    from thunder.torch.experimental.dtensor_codeutils import populate_object_ctx_for_dtensor_spec
+    from thunder.torch.experimental.dtensor_codeutils import is_dtensor_spec
 
-    if populate_object_ctx_for_dtensor_spec(x, object_ctx):
-        return x
-
-    if dataclasses.is_dataclass(x):
+    # NOTE: DTensorSpec is a dataclass but we want it to be handled differently from other dataclasses.
+    if dataclasses.is_dataclass(x) and not is_dtensor_spec(x):
         # Add `class` to the object_ctx so that we can reuse it during the trace execution.
         if isinstance(x, type):  # dataclass type
             cls = x
@@ -240,11 +238,6 @@ def prettyprint(
 
     if isinstance(x, ContextObject):
         return m(x.name)
-
-    from thunder.torch.experimental.dtensor_codeutils import prettyprint_dtensor_spec
-
-    if dtensor_repr := prettyprint_dtensor_spec(x):
-        return m(dtensor_repr)
 
     if dataclasses.is_dataclass(x):
         # For a dataclass instance of class
--- a/thunder/tests/distributed/test_dtensor.py
+++ b/thunder/tests/distributed/test_dtensor.py
@@ -34,7 +34,6 @@ functions_to_test = {
     "DTensor test requires CUDA and NCCL `torch.distributed` backend",
 )
 class DTensorTest(DistributedParallelTestCase):
-    @pytest.mark.skip(reason="https://github.com/Lightning-AI/lightning-thunder/issues/2355")
     @common_utils.parametrize("executor, fn_key", product(tuple(executors_map.keys()), functions_to_test.keys()))
     def test_dtensor_basic_op(self, executor, fn_key):
         num_devices = self.world_size
@@ -106,7 +105,6 @@ class DTensorTest(DistributedParallelTestCase):
         with pytest.raises(AssertionError):
             tmodel(in_dtensor, w)
 
-    @pytest.mark.skip(reason="https://github.com/Lightning-AI/lightning-thunder/issues/2355")
     def test_dtensor_incorrect_cotangent(self):
         num_devices = self.world_size
         mesh = DeviceMesh("cuda", list(range(num_devices)))
--- a/thunder/tests/opinfos.py
+++ b/thunder/tests/opinfos.py
@@ -695,6 +695,13 @@ abs_opinfo = ElementwiseUnaryOpInfo(
             dtypes=(datatypes.complex32,),
             devicetypes=(devices.DeviceType.CPU,),
         ),
+        # Ref - https://github.com/Lightning-AI/lightning-thunder/issues/2363
+        DecorateInfo(
+            pytest.mark.skip,
+            "test_vjp_correctness",
+            dtypes=(datatypes.float64,),
+            executors=("nvfuser",),
+        ),
     ),
 )
 
@@ -1931,7 +1938,15 @@ softsign_opinfo = OpInfo(
     sample_input_generator=elementwise_unary_generator,
     torch_reference=_elementwise_unary_torch(torch.nn.functional.softsign),
     singularity_fn=lambda x: x,
-    test_directives=(),
+    test_directives=(
+        # Ref - https://github.com/Lightning-AI/lightning-thunder/issues/2363
+        DecorateInfo(
+            custom_comparator(partial(assert_close, atol=5e-4, rtol=5e-4)),
+            "test_vjp_correctness",
+            dtypes=(datatypes.float64,),
+            executors=("nvfuser",),
+        ),
+    ),
 )
 elementwise_unary_ops.append(softsign_opinfo)
 
@@ -7125,6 +7140,13 @@ normalize_opinfo = OpInfo(
             "test_phantom_grad_vs_torch_consistency",
             dtypes=(datatypes.float16,),
             devicetypes=(devices.DeviceType.CPU, devices.DeviceType.CUDA),
+        ),
+        # Ref -https://github.com/Lightning-AI/lightning-thunder/issues/2363
+        DecorateInfo(
+            custom_comparator(partial(assert_close, atol=5e-4, rtol=5e-4)),
+            "test_vjp_correctness",
+            dtypes=(datatypes.float64,),
+            executors=("nvfuser",),
         ),
     ),
 )
--- a/thunder/torch/experimental/dtensor_codeutils.py
+++ b/thunder/torch/experimental/dtensor_codeutils.py
@@ -3,39 +3,9 @@ import torch
 from thunder.torch.experimental.dtensor_utils import run_only_if_distributed_is_available
 
 if torch.distributed.is_available():
-    from torch.distributed.tensor._dtensor_spec import DTensorSpec, TensorMeta
-    from torch.distributed.tensor import DeviceMesh, Partial, Placement, Replicate, Shard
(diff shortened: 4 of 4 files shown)


IvanYashchuk:
> Other approach (which should work for thunder.jit as well) that I tried was to transform initial trace acquired by thunder to execution trace with no decorators and PyTorch executor and run it with Fake inputs, but this will not work correctly when we have torch.no_grad as the corresponding symbols for no_grad act as a marker during trace time and currently don't have an executor claiming them.

Alternatively, instead of running the acquired by Thunder Interpreter program with FakeTensor inputs, run the program that Thunder is tracing (the callable input `fn` to `thunder.jit`) with FakeTensor inputs.

thunder/dynamo/splitter.py
```diff
@@ -167,6 +167,16 @@ def is_thunder_supported_partition(node: torch.fx.Node) -> bool:
         node_name = node.name
         if is_thunder_supported_partition(node):
             graph_module = getattr(split_gm, node.name)
+
+            is_differentiable_outputs = []
+            for n in graph_module.graph.nodes:
+                if n.op == "output":
+                    for n in n.all_input_nodes:
+                        if n.meta["example_value"].grad_fn is None:
```
IvanYashchuk, line 175: It's surprising that they store `grad_fn` in example_value and not just "requires_grad=True/False". Great that it works!

thunder/executors/torch_autograd.py
```diff
@@ -123,7 +128,18 @@ def detach_if_tensor(t):
...
             return torch.randn(1, device="meta", requires_grad=True)
         else:
+            if is_differentiable_outputs is None:
+                # Default to original behavior of marking all outputs as differentiable.
+                is_differentiable_outputs = tuple(True for _ in flat_output)
+
             ctx.save_for_backward(*saved_tensors)
+
+            assert len(flat_output) == len(is_differentiable_outputs)
+            filter_non_differentiable = [
```
IvanYashchuk, line 138: Why do you prefer to propagate "differentiable" output information to invert this right before the `ctx.mark_non_differentiable` call? It seems more straightforward, especially for those who know about `mark_non_differentiable` to propagate "non-differentiable" info.

IvanYashchuk (approved):
Very creative solution!

thunder/executors/torch_autograd.py
kshitij12345: While developing and debugging, `is_differentiable` info seemed to make a lot more sense while comparing against the trace. I am open to propagating `non-differentiable` info instead, wdyt?

thunder/__init__.py
```diff
@@ -852,6 +852,8 @@ def maybe_connect_to_autograd(cache_entry, result):
             # ThunderFunction (which is a torch.autograd.Function subclass)
             data_for_autograd, (saved_tensors, saved_other) = result
 
+            is_differentiable_outputs = compile_options.get("is_differentiable_outputs", None)
```
t-vi, line 855: Sigh.

kshitij12345: Would adding something like DebugOptions [PR](https://github.com/Lightning-AI/lightning-thunder/pull/1447) in a follow-up PR for passing this information be ok?

kshitij12345:
Not sure why but GPU tests are consistently failing with OOM with this PR even though they look unrelated.

thunder/executors/torch_autograd.py
IvanYashchuk: Makes more sense is a perfect reason. I don't have a preference. I was only curious about the choice.

thunder/dynamo/report.py
```diff
@@ -1098,6 +1098,12 @@ def foo(x):
     # Dynamo uses lazy generation of the underlying Python code, so we need to
     # force recompilation of the GraphModule before passing it to Thunder.
     recompile_graph(gm)
+
+    thunder_options["fusion_type"] = thunder_options.get("fusion_type", _DEFAULT_THUNDER_FUSION_TYPE)
+    thunder_options["thunderfx_disable_split_autograd"] = thunder_options.get(
+        "thunderfx_disable_split_autograd", _DEFAULT_THUNDERFX_DISABLE_SPLIT_AUTOGRAD
+    )
```
kiya00: maybe we can get the default thunder_options from an unused ThunderCompiler, like `thunder_options.update(ThunderCompiler().thunder_options)`

kshitij12345: That is a nice idea, thanks!

thunder/__init__.py
t-vi: This is like a third layer hack.
"we don't propagate gradients" is the problem.
Then we added the "let's mark everything as requiring grad" hack.
Now we add "let's pass information from a random place around in weird ways".
We could add a `THUNDER_OVERRIDE_DIFFERENTIABLE_OUTPUTS` environment variable somewhere, too.

kshitij12345: I agree that fixing propagating requires_grad correctly is the appropriate solution and the issue is tracked in #1768 (we can/should prioritise it for next quarter).

However, I am not clear on the requested change here.

t-vi:
Not tracking requires grad properly also adds to  #2361

PR #2349, opened 2025-07-22
Make inplace flags defaults consistent

beverlylytle:
The default values for inplace flags are mixed: `False` for skip_inplace_alias_updates [here](https://github.com/Lightning-AI/lightning-thunder/blob/95b76454ffb4365fc8b320d4c502ad29361f8a3b/thunder/__init__.py#L491) but `True` [here](https://github.com/Lightning-AI/lightning-thunder/blob/95b76454ffb4365fc8b320d4c502ad29361f8a3b/thunder/__init__.py#L687).  This PR fixes that inconsistency.


beverlylytle:
I agree with you @t-vi.  I will make a PR to remove the other path soon, but I wanted to address this probably invisible and tiny bug first.

Issue #2300, opened 2025-07-02
fix BINARY_SUBSCR to use __class_getitem__ as needed  (__getitem__() missing 1 required positional argument: 'idx' )


t-vi:
*Note*: If you have a model or program that is not supported yet but should be, please use the program coverage template.

## 🐛 Bug

`InterpreterError: Encountered exception TypeError: SequenceWrapperMethods.__getitem__() missing 1 required positional argument: 'idx' while tracing `

### To Reproduce

```
def fn():
   return list[int]
thunder.jit(fn)()
```

fails.

### Additional context

The error is becauee the _binary_subscr_handler in thunder/core/interpreter.py always uses __getitem__ instead of __class_getitem__ when the tos1 is a type.

It would want a test along the lines of the above in test_interpreter.py .

The relevance is that Transformers 4.53 introduced something of that in the jitted codepath.

@tejapulagam if you want to work on this, please self-assign.

cc @lantiga

tejapulagam:
I can work on this!

t-vi:
closed by @tejapulagam in #2366  Thank you  Teja

PR #2366, opened 2025-07-28
Bugfix/fix binary subscr class getitem

tejapulagam:
#  Fix type subscript handling in binary_subscr_handler

## What does this PR do?

This PR addresses issue [#2300](https://github.com/Lightning-AI/lightning-thunder/issues/2300) by attempting to fix how the interpreter handles type subscripts (e.g., `list[int]`) in the `binary_subscr_handler`.

- Adds logic in the BINARY_SUBSCR handler to check if tos1 is a type. If it is, it uses __class_getitem__ instead of __getitem__.

- Adds a test to test_interpreter.py.

**Current Problem:**
When using `thunder.jit()` on functions that return parameterized types, the interpreter incorrectly processes type subscripts through the standard `getitem` path instead of recognizing them as type operations.

**Example failing case:**
```python
def fn():
    return list[int]

thunder.jit(fn)()  # Currently fails
```
Open questions: 
- Is `class_getitem` the correct fix because currently its usage for this results in a notImplementedError?
- Is there a better way check for the type subscripts because the current check does not work?


thunder/core/interpreter.py
```diff
@@ -3620,11 +3620,20 @@ def impl(container, start, end):
...
-    def impl(tos1, tos):
-        return tos1.__getitem__(tos)
+    def class_getitem_impl(cls, index):
+        return cls.__class_getitem__(index)
 
-    res = _interpret_call(impl, tos1, tos)
+    def getitem_impl(obj, index):
+        return obj.__getitem__(index)
+
+    if isinstance(tos1, type):
```
t-vi:
```suggestion
    if isinstance(unwrap(tos1), type):
```

thunder/core/interpreter.py
```diff
@@ -3620,11 +3620,20 @@ def impl(container, start, end):
 def _binary_subscr_handler(inst: dis.Instruction, /, stack: InterpreterStack, **kwargs) -> None | INTERPRETER_SIGNALS:
     tos = stack.pop_wrapped()
     tos1 = stack.pop_wrapped()
+    print(str(isinstance(tos1, type)) + "is working")
+    print(type(tos1))
+    print(type(tos1).__mro__)
```
t-vi:
```suggestion
```

thunder/tests/test_interpreter.py
```diff
@@ -3572,3 +3572,12 @@ def fn():
         jfn()
     except Exception as e:
         assert 'print("A" * "A")' in "\n".join(traceback.format_tb(e.__traceback__))
+
+
+def test_binary_subscr_on_types():
+    def fn():
+        return list[int]
```
t-vi: I think we should avoid needing to have to deal with the list class getitem (which is more elaborate) for this PR.
```suggestion
    class A:
        @classmethod
        def __class_getitem__(cls, index):
            return index

    def fn():
        return A[int]
```

thunder/tests/test_interpreter.py
```diff
@@ -3572,3 +3572,12 @@ def fn():
...
+
+    jfn = thunder.jit(fn)
+    out = jfn()
+    assert out is list[int]
```
t-vi:
```suggestion
    assert out is int
```

t-vi (approved):
Supergood, thank you @tejapulagam

PR #2368, opened 2025-07-29
Remove functionalization path

beverlylytle:
Now that alias updating is the default path for in-place op support, the functionalization path can be removed.  However, the aliasing strategy is not a silver bullet for in-place op support.  While the functionalization path is being removed from the code base, it will still live in the git history.  Hopefully ideas developed in the removed code can be revived to blend with the aliasing strategy.


crcrpar (approved):

t-vi (commented):
Hurray for not having alternative codepaths!
Thank you @beverlylytle @crcrpar

Issue #2338, opened 2025-07-17
[DTensor] Execute Llama4 DecorderLayer with DTensor via nvFuser


kshitij12345:
This is a tracker issue for adding support for executing a ~DeepSeek~ Llama4 TransformerLayer in distributed setting using nvFuser multi-device support.

Tasks
* [ ] Fix https://github.com/Lightning-AI/lightning-thunder/issues/2234
* [x] Add support for nvfuser direct bindings (to avoid cache reset issues) - see https://github.com/NVIDIA/Fuser/pull/4689
* [ ] Operator coverage
  * [ ] Add operator coverage for MOE Layer. Reference Impl [nvfuser/test_moe.py](https://github.com/NVIDIA/Fuser/blob/main/tests/python/test_moe.py)
    * [x] prims.reshape #2383
    * [ ] ~~prims.empty~~ (updated implementation doesn't require this)
    * [x] prims.linear #2422 
    * [x] prims.convert_element_type #2382
    * [x] prims.neg #2552 
    * [x] prims.exp #2496 
    * [x] prims.add
    * [x] prims.reciprocal #2552 
    * [x] prims._grouped_mm #2503 
    * [x] prims.mul
    * [x] prims.broadcast_in_dim #2382
    * [ ] prims.scatter
    * [ ] prims.lt
    * [ ] prims.where
    * [ ] prims.take #2430 
    * [ ] ~~prims.copy_~~  (updated implementation doesn't require this)
    * [ ] prims.sum
    * [ ] ltorch.cumsum
    * [ ] ltorch.argsort
    * [ ] prims.topk
    * [ ] prims.full

  * [ ] Add operator coverage to execute the complete TransformerDecoderLayer
* [x] Add proper testing infrastructure  #2412 

cc @tfogal

nvMelissa:
@kshitij12345 - just checking if you got the message that we should focus on Llama 4  first.

kshitij12345:
Yes, thanks for the reminder, have updated the issue title and description to reflect this.

nvMelissa:
Hi @kshitij12345 - I forgot if you told me if all of these will also work for nvFP4. Would you please clarify?

kshitij12345:
For nvFP4, we will mostly likely require update around only `linear` and `_grouped_mm`.

nvMelissa:
@kshitij12345 - would you please close this and leave the nvFP4 for tracking?

kshitij12345:
[...]

PR #2370, opened 2025-07-29
[dtensor] use nvfuser_direct for nvfuser dtensor execution

kshitij12345:
Ref #2338

Tested with existing `thunder/tests/distributed/test_dtensor.py`

Current nvFuser Python interface  leads to an error when creating multiple FusionDefinitions with the same math but different device schedules - [issue](https://github.com/NVIDIA/Fuser/issues/4507).

For this, nvFuser has a new Python front-end (nvfuser_direct) which doesn't face this issue. In this PR, we update the multi-device path to use this interface. Also, in future, this will be the default interface and we will update to using this for single-device path as well.

EDIT: ~~Fix for https://github.com/NVIDIA/Fuser/issues/4806 needs to land first~~ This has been fixed

cc @tfogal


thunder/executors/nvfuserex_impl.py
```diff
@@ -383,10 +382,13 @@ def check_dtensor_tracing_and_runtime_metadata(inp):
             lambda: "nvfuser: Expected runtime and tracing metadata to be the same for DTensor.",
         )
 
-        fd = MultiDeviceFusionDefinition(definition, sorted_unique_inputs, max_length=MAX_LENGTH)
+        fd = DirectFusionDefinition()
         # Device may be set in one of the "factory" methods like full, iota, or uniform
         # NOTE: This should be called before defining because a factory method may look-up at `_selected_device` while being defined.
         fd._selected_device = None
+        with fd:
```
rdspring1, line 382: Unrelated: How is `_selected_device` used in `FusionDefinition`?

thunder/executors/nvfuserex_impl.py
```diff
@@ -542,28 +544,10 @@ def __call__(self, *args):
         if self.store_inputs:
             self.last_inputs = args
 
-        if hasattr(fd, "multidevice_schedule"):
+        if IS_TORCH_DISTRIBUTED_AVAILABLE and any(isinstance(t, torch.distributed.tensor.DTensor) for t in args):
```
wujingyue:
```suggestion
        if dist.is_available() and any(isinstance(t, torch.distributed.tensor.DTensor) for t in args):
```
`dist.is_available()` is short and cheap already, and readers immediately know it's simply calling torch rather than some Thunder magic.

wujingyue (approved):
Great work!

thunder/executors/nvfuserex_impl.py
kshitij12345: `_selected_device` is used to set the device for factory ops like `full`, `iota`, etc.

It is optional set in `_select_device`
https://github.com/Lightning-AI/lightning-thunder/blob/29d39bcffa449c72ce9a5dc46be8ce90fbe01cb7/thunder/executors/nvfuserex_impl.py#L1097-L1108

And `_select_device` is called in the lowering for `full`, `iota`, etc.
https://github.com/Lightning-AI/lightning-thunder/blob/29d39bcffa449c72ce9a5dc46be8ce90fbe01cb7/thunder/executors/nvfuserex_impl.py#L1137-L1151

kshitij12345:
Ping @t-vi @IvanYashchuk

t-vi (approved):
Thank you @kshitij12345 @wujingyue @rdspring1

PR #2372, opened 2025-07-29
Remove unnecessary underscores

wujingyue:


.pre-commit-config.yaml
```diff
@@ -1,5 +1,5 @@
 default_language_version:
-  python: python3.10
+  python: python3
```
crcrpar: I'm not quite sure if this change should be included. I think it's worth thinking about bumping the version here, but let's do it in a separate pull request.

```suggestion
  python: python3.10
```

t-vi (approved):
Thank you @wujingyue @crcrpar

PR #2374, opened 2025-07-29
Lower cumsum to nvfuser

wujingyue:
so nvFuser gets a larger portion of the MoE layer. Context: https://github.com/NVIDIA/Fuser/pull/4866#issuecomment-3126652527

### DIFF
--- a/thunder/executors/nvfuserex_impl.py
+++ b/thunder/executors/nvfuserex_impl.py
@@ -2944,7 +2944,7 @@ def cross_entropy_fwd(
     nv_target = getnv(target, fd, lc_to_nv_map)
     nv_ignore_index = getnv(ignore_index, fd, lc_to_nv_map)
 
-    zero_scalar = fd.define_scalar(0, dtype=torch_dtype_to_nvfuser_dtype(dtypes.to_torch_dtype(a.dtype)))
+    zero_scalar = fd.define_scalar(0, dtype=lcdtype_to_nvdtype(a.dtype))
 
     # modify the labels to account for ignore index and then do a
     # gather/ take_along_axis on the input tensor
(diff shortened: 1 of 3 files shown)


wujingyue:
Side note: the sample data is non-deterministic, making debugging hard. In nvFuser, we call torch.manual_seed before each test so things like `torch.randn` and `torch.testing.make_tensor` will generate deterministic values. cc @IvanYashchuk

wujingyue:
> The CI tests for `cumsum` are failing due to accuracy issues.

Thanks for pointing that out! 

nvFuser actually computes more accurate results than torch. This is because torch.cumsum calls cub::DeviceScan which in this case accumulates in float16. 

https://github.com/Lightning-AI/lightning-thunder/blob/main/thunder/tests/test_ops.py#L91 fails to detect this because it doesn't up the dtype. I'm experimenting with https://github.com/Lightning-AI/lightning-thunder/pull/2374/commits/aa1126d06e4280b5c3bcb673493a154a6ff97742...

wujingyue:
> However, I was wondering whether nvFuser can just add `fd.ops.cumsum`.

cc @jacobhinkle I believe this is getting slightly closer with your recent ScanOp? 

Anyhow, I don't expect cumsum of 128 (i.e. number of experts) integers to be a bottleneck. So triu+matmul or triu+where+sum as you suggested should be fine.

wujingyue:
The only CI failures are caused by cumsum_transform sometimes taking torch.dtype and sometimes thunder.dtypes.dtype. I think it's related to `register_supported(ltorch.cumsum, ...)` and that `dtype` is an argument. Any idea how to solve this? cc @IvanYashchuk and @kshitij12345

kshitij12345:
> The only CI failures are caused by cumsum_transform sometimes taking torch.dtype and sometimes thunder.dtypes.dtype. I think it's related to register_supported(ltorch.cumsum, ...) and that dtype is an argument. Any idea how to solve this?

You can use - `lcdtype_to_nvdtype(dtypes.to_dtype(dtype))` as `dtypes.to_dtype` handles converting `torch.dtype` to thunder's dtype and preserves thunder's dtype. This should unblock the PR.

But I am surprised that `dtype` can be torch.dtype or thunder.dtypes.dtype, I will check to see why that is the case and file an issue accordingly.

wujingyue:
> You can use - `lcdtype_to_nvdtype(dtypes.to_dtype(dtype))`

That works -- thanks! 

The PR is ready to review now.

thunder/tests/test_ops.py
```diff
@@ -78,13 +78,23 @@ def snippet_torch_consistency(op: OpInfo, torch_op, sample: SampleInput, comp: C
...
+            return torch.tensor([], dtype=dtype).is_floating_point()
+
+        def upcast(x):
             if isinstance(x, torch.Tensor) and torch.is_floating_point(x):
                 return x.to(torch.double)
+
+            # Some torch APIs (e.g. cumsum) take a `dtype` argument and use it
+            # as the **compute** dtype. Replace that with torch.double so the
+            # reference result is more likely to be computed in double.
+            if isinstance(x, torch.dtype) and is_floating_point_dtype(x):
```
kshitij12345: You can use `is_floating_point` on `torch.dtype`.

```python
>>> import torch
>>> torch.float.is_floating_point
True
>>> torch.float16.is_floating_point
True
```

thunder/executors/nvfuserex_impl.py
```diff
@@ -3188,6 +3181,51 @@ def argsort_transform(
 # Register argsort with NVFuser
 register_supported(prims.argsort, argsort_transform, _argsort_check_)
 
+
+def _cumsum_check(a: TensorProxy, dim: int, /, dtype: dtypes.dtype | None = None) -> bool:
+    if a.ndim != 1:
```
kshitij12345, line 3186: We should document (maybe add a TODO) that this lowering needs to be updated once nvfuser supports cumsum

wujingyue:
@t-vi and @mruberry, this is ready to merge

t-vi (approved):
Thank you @wujingyue @kshitij12345

PR #2375, opened 2025-07-30
Remove early split trace path

beverlylytle:
This PR removes the boolean compile option "delay_trace_split" and code used only in the `False` path.  The `False` path was previously used for creating and processing the backward pass, and has been replaced by the contents of thunder.transforms.autodiff, the `True` path, if you will.


t-vi (approved):
Supercool. Thank you @beverlylytle

PR #2344, opened 2025-07-21
TransformerEngine executor checkpointing 

riccardofelluga:
## What does this PR do?

Replaces #2094 and updates it. This now uses the joint trace to do the transformations requires for TE to work correctly when linears are checkpointed.

Will close #2027

## Design decisons

What all of this does is to look at the joint trace and identify which `te_fp8_amax_and_scale_update` need to be removed. Since we want to update the state only once per run of the joint trace, this identifies which states are updated twice and removes all but the last occurrence of the `te_fp8_amax_and_scale_update` for the checkpointed states.

The steps are:
1. Look at the unique symbol names that create the state and associate them with the output of their first occurrence.
2. Swap all the other proxies with the proxy that was seen first.
3. On the reverse bound symbols list remove all the second occurances of the  `te_fp8_amax_and_scale_update` for the same states.
4. reverse the list again and swap the proxies that were output to the deleted `te_fp8_amax_and_scale_update` with the relative inputs.
5. Profit.

I've tried to make it part of the transform but I was not able to because this transformation needs to look at least three times at the trace(maybe can be lowered but still it needs to go back to the forward trace to remove the amax updates). We cannot simply use the current transform infrastructure because it does not allow to manipulate the joint trace but only the splitted ones. 
Arguably, being able to apply the transform on the joint trace would remove the need for this double transform pass(i.e. `_te_activation_checkpointing_transform` and `TransformerEngineTransformV2` would be joined into one]

## Trace

<details>

<summary>Joint trace before transformation</summary>

```python
import thunder
import thunder.core.prims as prims
import torch
from thunder.executors.torchex import no_autocast

@torch.no_grad()
@no_autocast
def computation(l_x_, l_w_, l_w2_):
  # l_x_: "cuda:0 f32[64, 64]"
  # l_w_: "cuda:0 f32[64, 64]"
  # l_w2_: "cuda:0 f32[64, 64]"

  # <eval_with_key>.7:6: 	    tag_activation_checkpoint = torch.ops.higher_order.tag_activation_checkpoint(wrap_body_0, l_x_, l_w_, use_reentrant = False);  wrap_body_0 = l_x_ = l_w_ = None
  r7 = get_te_fp8_recipe_0()  # r7: "<class 'NoneType'>"
  s8 = get_te_fp8_state_1(r7, 'forward', 2)  # s8: "<class 'NoneType'>"
  (q9, q10) = get_te_fp8_quantizers_2(s8, 2)
[...]

PR #2376, opened 2025-07-30
Update TE v2 executor tests

riccardofelluga:
While working on #2344 I've noticed that when using mse we cannot do more than 3 iterations otherwise the loss goes outside well fp8 range. Using cross entropy helps testing a more realistic scenario and leaves more iterations for numerical errors to accumulate in case there is a mismatch with TE behavior

### DIFF
--- a/thunder/executors/transformer_engine_v2ex_impl.py
+++ b/thunder/executors/transformer_engine_v2ex_impl.py
@@ -140,6 +140,7 @@ def _linear_fwd_impl(a, w, bias, input_quantizer: Quantizer, weight_quantizer: Q
         input=a,
         weight=w,
         bias=bias,
+        dtype=w.dtype,
         with_quantized_compute=True,
         input_quantizer=input_quantizer,
         weight_quantizer=weight_quantizer,
--- a/thunder/tests/test_transformer_engine_v2_executor.py
+++ b/thunder/tests/test_transformer_engine_v2_executor.py
@@ -113,7 +113,7 @@ def test_te_linear_forward_backward_multiple_iteration(fp8_recipe: recipe.Recipe
     # Running more iterations leads to `nan` for both eager and thunder
     # with BlockScaling.
     # Potentially because we are training on dummy data and task
-    iterations = 3
+    iterations = 6
 
     # TE inputs
     input_shape = (768, 4096)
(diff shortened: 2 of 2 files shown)


thunder/tests/test_transformer_engine_v2_executor.py
```diff
@@ -126,16 +126,15 @@ def clone_params(*params):
...
 
     inputs = tuple(torch.rand(*input_shape, device=device, dtype=dtype) for _ in range(iterations))
 
     def train_model(model, optimizer):
         # Run for `iterations`.
         for iter_n in range(iterations):
             x = inputs[iter_n]
-            result = model(x)
-            loss = torch.nn.functional.mse_loss(result.sum(), target_value)
+            loss = torch.nn.functional.cross_entropy(result, target_value)
```
kshitij12345, line 138:
```suggestion
            result = model(x)
            loss = torch.nn.functional.cross_entropy(result, target_value)
```

thunder/executors/transformer_engine_v2ex_impl.py
```diff
@@ -140,6 +140,7 @@ def _linear_fwd_impl(a, w, bias, input_quantizer: Quantizer, weight_quantizer: Q
         input=a,
         weight=w,
         bias=bias,
+        dtype=w.dtype,
```
kshitij12345: Is this necessary or is this a stray change?

riccardofelluga: I wanted to add it in the checkpointing PR but it is needed also here for new versions of TE since they changed the input validation for the _functional_forward call. So it can go up with this or the other PR

kshitij12345: It looks like a bug in TE.

https://github.com/NVIDIA/TransformerEngine/blob/858755c0c3898d5bc971fcec031b26ff26e7e4fb/transformer_engine/pytorch/ops/basic/basic_linear.py#L404-L408

`elif` should be `elif weight is not None and isinstance(weight, torch.Tensor):`.

Will file a bug in TE.

kshitij12345: Filed https://github.com/NVIDIA/TransformerEngine/issues/2011. I think we should add a comment referring to this issue above this line.

riccardofelluga:
> LGTM, thanks @riccardofelluga
> 
> I will send a similar patch for TE v1 executor test.

I can add it with this PR if it's ok @kshitij12345

kshitij12345:
> I can add it with this PR if it's ok @kshitij12345

Yes, that'd be great, thanks! 🙇

riccardofelluga:
Updated and tested on both H100 and B200.

t-vi (approved):
Thank you @riccardofelluga @kshitij12345

PR #2381, opened 2025-07-31
Fix example in README.md

zasdfgbnm:
[`assert_close`](https://docs.pytorch.org/docs/stable/testing.html#torch.testing.assert_close) returns None. Mismatches are reported by raising errors. So it makes no sense to assert on the return value of `assert_close`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,7 +146,7 @@ x = torch.randn(64, 2048)
 
 y = thunder_model(x)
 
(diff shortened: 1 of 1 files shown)


crcrpar (approved):

t-vi (approved):

t-vi:
Thank you @zasdfgbnm @crcrpar

Issue #2234, opened 2025-06-13
DTensor: Get it working on the thunderfx path.


kshitij12345:
This currently doesn't work.

TODO - Add example

IvanYashchuk:
Reproducer in https://github.com/Lightning-AI/lightning-thunder/issues/1898 has necessary environment variables set to be able to instantiate DTensor object without torchrun or multiple GPUs.

PR #2382, opened 2025-07-31
[DTensor] Add prims for dtype conversion and broadcasting

kshitij12345:
Ref: https://github.com/Lightning-AI/lightning-thunder/issues/2338
These are required to support type promotion and broadcasting for other torch operations.

NOTE: grad rule for `broadcast_in_dim` requires implementing `sum` which will be added in a follow-up.

cc @tfogal


thunder/executors/nvfuserex_impl.py
```diff
@@ -1088,6 +1092,7 @@ def convert_element_type(
 
 
 register_supported(PrimIDs.CONVERT_ELEMENT_TYPE, convert_element_type, _convert_element_type_check)
+register_supported(dtensor_convert_element_type_prim, convert_element_type, _convert_element_type_check)
```
IvanYashchuk, line 977: If for the nvFuser executor `PrimIDs.CONVERT_ELEMENT_TYPE` and `dtensor_convert_element_type_prim` are the same, how are they different from each other in Thunder?

thunder/torch/experimental/dtensor_torch_and_prims.py
```diff
@@ -134,5 +135,60 @@ def dtensor_mul(a: TensorLike, b: TensorLike) -> TensorLike:
     return dtensor_mul_prim(a, b)
 
 
+def dtensor_convert_element_type_meta(a, dtype):
+    tdtype = ltorch.to_torch_dtype(dtype)
+    output = run_with_fake_tensor(lambda x, dt: x.to(dt), a, tdtype)
+    local_tensor_proxy = TensorProxy(like=a.local_tensor, shape=output._local_tensor.shape, dtype=dtype)
+    spec = output._spec
+    spec_proxy = AnyProxy(spec, history=a.history)
```
IvanYashchuk, line 181: What does history do? It's an optional kwarg for AnyProxy and if it's not used anywhere in the DTensor support it should be dropped.
```suggestion
    spec_proxy = AnyProxy(spec)
```

thunder/torch/experimental/dtensor_torch_and_prims.py
```diff
@@ -134,5 +135,60 @@ def dtensor_mul(a: TensorLike, b: TensorLike) -> TensorLike:
...
+
+
+def _dtensor_convert_element_type_prim_grad(a: TensorLike, dtype) -> TensorLike:
+    fwd = dtensor_convert_element_type_prim(a, dtype)
+
+    g = get_grad(fwd)
+    g_converted = dtensor_convert_element_type_prim(g, a.dtype)
+    put_grad(a, g_converted)
+
+    return fwd
```
IvanYashchuk, line 205: It's the same function as we have in https://github.com/Lightning-AI/lightning-thunder/blob/84c76b44ef23dc3483ab1caea51dfbf80e93bc2c/thunder/core/transforms.py#L667-L674

What are the options for reducing code duplication?

thunder/torch/experimental/dtensor_torch_and_prims.py
```diff
@@ -134,5 +135,60 @@ def dtensor_mul(a: TensorLike, b: TensorLike) -> TensorLike:
     return dtensor_mul_prim(a, b)
 
 
+def dtensor_convert_element_type_meta(a, dtype):
+    tdtype = ltorch.to_torch_dtype(dtype)
+    output = run_with_fake_tensor(lambda x, dt: x.to(dt), a, tdtype)
+    local_tensor_proxy = TensorProxy(like=a.local_tensor, shape=output._local_tensor.shape, dtype=dtype)
+    spec = output._spec
```
IvanYashchuk, line 180: Is there a SpecProxy in Thunder?

thunder/torch/experimental/dtensor_torch_and_prims.py
```diff
@@ -134,5 +135,60 @@ def dtensor_mul(a: TensorLike, b: TensorLike) -> TensorLike:
     return dtensor_mul_prim(a, b)
 
 
+def dtensor_convert_element_type_meta(a, dtype):
+    tdtype = ltorch.to_torch_dtype(dtype)
+    output = run_with_fake_tensor(lambda x, dt: x.to(dt), a, tdtype)
+    local_tensor_proxy = TensorProxy(like=a.local_tensor, shape=output._local_tensor.shape, dtype=dtype)
```
IvanYashchuk, line 179: Local tensor could be obtained by running prims.convert_element_type.

Why is it better to run with fake tensors? What happens if the shape returned by fake tensor ops is different from prims.convert_element_type?

thunder/executors/nvfuserex_impl.py
kshitij12345: They are different in thunder as `dtensor_convert_element_type_prim ` also understands DTensor meta-data propagation.

It lowers to the same math operation for nvFuser as we specify the maths ops and input layout (/multidevice schedule) separately in nvFuser.

Here is where we specify the multidevice schedule - 

https://github.com/Lightning-AI/lightning-thunder/blob/84c76b44ef23dc3483ab1caea51dfbf80e93bc2c/thunder/executors/nvfuserex_impl.py#L258-L278

https://github.com/Lightning-AI/lightning-thunder/blob/84c76b44ef23dc3483ab1caea51dfbf80e93bc2c/thunder/executors/nvfuserex_impl.py#L382-L384

thunder/torch/experimental/dtensor_torch_and_prims.py
kshitij12345: At the moment, no. We just use `AnyProxy` to represent `spec`.

thunder/torch/experimental/dtensor_torch_and_prims.py
kshitij12345: I think it is used for Provenance tracking. Not directly required for DTensor support.

thunder/torch/experimental/dtensor_torch_and_prims.py
kshitij12345: One option is to update the existing prims one by one to support propagating DTensor layout and providing the executors information about the supported types (Tensor, DTensor) which the executor opt in or out of.

I will add a note about this in the tracking issue for DTensor support #2338 

EDIT: Have updated the description of #2338 to track this.

thunder/torch/experimental/dtensor_torch_and_prims.py
kshitij12345: > Local tensor could be obtained by running prims.convert_element_type.

I think this is true for these primitives but may not always be True (as sometimes the inputs could be redistributed). So, I feel it is safer to rely on FakeTensors.

Eg. when we can't obtain with local tensor.

```python
# torchrun --local-ranks-filter=0 --nnodes 1 --nproc-per-node 2 test_dtensor.py

import torch
from torch.distributed._tensor import DeviceMesh, Shard, distribute_tensor
from torch.distributed.tensor.placement_types import Placement, Shard, Replicate
import os

num_devices = 2
mesh = DeviceMesh("cuda", list(range(num_devices)))
torch.cuda.set_device(LOCAL_RANK)

input_size = 4
hidden_size = 16
output_size = 8

def model(x, w):
    return x @ w
weight = distribute_tensor(torch.randn(hidden_size, output_size, requires_grad=True), mesh, [Shard(1)])
in_dtensor = distribute_tensor(torch.randn(input_size, hidden_size, requires_grad=True), mesh, [Shard(0)])

print(weight._local_tensor.shape, "WEIGHT LOCAL SHAPE")  # torch.Size([16, 4]) WEIGHT LOCAL SHAPE
print(weight.shape, "WEIGHT GLOBAL SHAPE")  # torch.Size([16, 8]) WEIGHT GLOBAL SHAPE

print(in_dtensor._local_tensor.shape, "INP LOCAL SHAPE")  # torch.Size([2, 16]) INP LOCAL SHAPE
print(in_dtensor.shape, "INP GLOBAL SHAPE")  # torch.Size([4, 16]) INP GLOBAL SHAPE

output = torch.compile(model)(in_dtensor, weight)
print(output._local_tensor.shape, "OUTPUT LOCAL SHAPE")  # torch.Size([4, 4]) OUTPUT LOCAL SHAPE
print(output.shape, "OUTPUT GLOBAL SHAPE")  # torch.Size([4, 8]) OUTPUT GLOBAL SHAPE

torch.distributed.destroy_process_group()

```

thunder/tests/distributed/test_dtensor.py
crcrpar, line 1: quick question: would we make the added test OpInfo based at some point?

thunder/tests/distributed/test_dtensor.py
```diff
@@ -161,6 +161,55 @@ def fn(x, w):
         with pytest.raises(RuntimeError, match="has changed for cotangent between tracing and runtime"):
             torch.autograd.grad(actual, (in_dtensor, w_dtensor), g_o)
 
+    @common_utils.parametrize("executor", tuple(executors_map.keys()))
+    def test_dtensor_convert_element_type(self, executor):
+        from thunder.torch.experimental.dtensor_torch_and_prims import dtensor_convert_element_type_prim
+        from thunder.core.dtypes import bfloat16
```
crcrpar: now most environments would be compatible with bfloat16 but how about using https://github.com/Lightning-AI/lightning-thunder/blob/68b4c4cbba9be58c9b3404a7847b2e1fa8256cd4/thunder/tests/bf16.py#L5?

Would this be a bit too paranoid/meticulous?

thunder/tests/distributed/test_dtensor.py
kshitij12345: I have found a problem with OpInfo based test, will fix that and migrate these test to be OpInfo based.

thunder/tests/distributed/test_dtensor.py
kshitij12345: Right, it should be safe. Have used `dtypes.bfloat16` where `dtypes` is imported at module level. Thanks!

thunder/tests/distributed/test_dtensor.py
```diff
@@ -136,6 +136,54 @@ def fn(x, w):
         with pytest.raises(RuntimeError, match="has changed for cotangent between tracing and runtime"):
             torch.autograd.grad(actual, (in_dtensor, w_dtensor), g_o)
 
+    @common_utils.parametrize("executor", tuple(executors_map.keys()))
+    def test_dtensor_convert_element_type(self, executor):
```
crcrpar, line 140: Just curious, can we use executors defined in `thunder.tests.framework` here?

kshitij12345: `executors_map` also uses the same executors defined in `thunder.tests.framework`.

https://github.com/Lightning-AI/lightning-thunder/blob/cbfaf03716f6a11f1060b017ee75e3420a511206/thunder/tests/distributed/helper.py#L43-L47

At the moment, DTensor doesn't with DynamoExecutor when backward is also required (see https://github.com/Lightning-AI/lightning-thunder/issues/2234).

crcrpar: I think I was curious if `@common_utils.parametrize("executor", tuple(executors_map.values()))` works

kshitij12345: That works, but then the test name would be `test_dtensor_convert_element_type_executor0`. So, we can either add `ids` to `parametrize` or pass the executor names like we have now.

IvanYashchuk (approved):

t-vi (approved):
Thank you @kshitij12345 @crcrpar @IvanYashchuk

PR #2393, opened 2025-08-01
Check `var is not None` before `add_edges(var)`

crcrpar:
## What does this PR do?

This would be rather a workaround than a proper fix for https://github.com/Lightning-AI/lightning-thunder/issues/2365


IvanYashchuk:
@t-vi, I need your approval to merge it. Thank you!

PR #2396, opened 2025-08-04
[WIP]

beverlylytle:


beverlylytle:
Some grad tests are failing in a non-trivial way.  In the case of `atleast_2d`, test_grad is being supplied an input which is then directly returned as output (`z, y = atleast_2d(x, y)`).  Before this change the output `y` would have been replaced by `None` and we wouldn't hit the error raised [here](https://github.com/Lightning-AI/lightning-thunder/blob/88182a02de063a35d5cad4839ef7877e6a54d758/thunder/core/trace_interpreter.py#L55).

IvanYashchuk:
Could you please check whether the error goes away if `atleast_2d` and all similar operations use `prims.clone` and not return the input directly?

t-vi:
I would not expect clone (which is not what PyTorch does, it really does identity.) Maybe what is missing is something in `__iter__`. 
Note that the grad checks are still using the old logic, which has much more elaborate handling of complex outputs, so maybe moving those from running vjp to comparing to torch grad after `torch.autograd.grad(thunder_function(x), x)` fixes it. I would recommend looking into whether it is test specific before trying to fix. I vaguely recall a similar problem before  but I don't remember the exact issue.

beverlylytle:
You are right, @t-vi , that `prims.clone` is not the right approach here because it does not mimic PyTorch.  However I don't think this is just limited to the tests. For example, 
```
import torch
import thunder

class MyFunction(torch.autograd.Function):
    @staticmethod
    def forward(ctx, x: torch.Tensor) -> torch.Tensor:
        x = x.sin()
        return torch.atleast_2d(x)

    @staticmethod
    def backward(ctx, grad_output) -> torch.Tensor:
        return grad_output

class Model(torch.nn.Module):
    def __init__(self):
        super().__init__()

    def forward(self, x) -> torch.Tensor:
        return MyFunction.apply(x)

model = Model().to(dtype=torch.float64)

jitted = thunder.jit(model)

y = torch.randn(10, 10, requires_grad=True, device="cuda")
z = jitted(y)
```
still fails with "Encountered exception ValueError: Variable t1 is being overwritten this is not allowed while tracing Model()", though this failure is unrelated to removing this helper function.  Using `prims.shallow_copy` fixes both issues.

beverlylytle:
Btw, which `__iter__` do you mean @t-vi ?

t-vi:
Couldn't we del _ instead when it is assigned?

beverlylytle:
```
import torch
import thunder

def f(x):
    y, z = torch.var_mean(x, dim=1)
    return z

jitted = thunder.jit(f)

x = torch.randn(10, 10)
z = jitted(x)

last_trace = thunder.last_traces(jitted)[-1]
var_mean_bsym = last_trace.bound_symbols[1]
print(var_mean_bsym.flat_outs)         # [None, <TensorProxy(name="t11", dtype=thunder.dtypes.float32, shape=(10,))>]

print(var_mean_bsym)                   # (_, t11) = torch.var_mean(x, 1, keepdim=False, correction=1)
                                       #   # (t10, t11) = ltorch.var_mean(x, 1, keepdim=False, correction=1)
                                       #     # (t10, t11) = prims.var_mean(x, (1,), correction=1)
```
The underscore doesn't appear until the [printing process](https://github.com/Lightning-AI/lightning-thunder/blob/5c18b33917586e388e9526eb28cfd48cf8755915/thunder/core/symbol.py#L85) and is otherwise represented as `None`.  I attempted to modify `del_last_used` to also add a bsym with op prims.DEL and arg `None`, but this meant also changing the `del_printer` from `arg_string = ", ".join(codeutils.prettyprint(x, literals_allowed=False) for x in arg_printables)` to `arg_string = ", ".join(codeutils.prettyprint(x, literals_as_underscores=True) for x in arg_printables)`.  But that gives me the howling fantods.

t-vi:
I think the del_last_used could check if the return is a tuple and if it is it can just iterate over that instead of using flat proxy outs. This would allow you to detect none properly.

beverlylytle:
@t-vi I'm not understanding your last comment, or at least the motivation behind it.  Correct me if I'm wrong, but it seems like you are suggesting that unused TensorProxies should continue to be replaced with None in the dce pass and that del_last_used should then tidy things up.  What is the advantage of replacing an unused TensorProxy with None in a bsym's outs?

thunder/core/transform_common.py
```diff
@@ -162,29 +162,14 @@ def dce(trace: Trace, needed_proxies: None | set[Variable] = None) -> Trace:
...
-            if some_unused:
-
-                def _helper(x):
-                    if isinstance(x, Proxy) and (variableify(x) not in needed_proxies or producer_map[x] != bsym):
-                        return None
-                    return x
-
-                nbsym_output = tree_map(_helper, bsym.output)
-                nbsym = bsym.from_bsym(output=nbsym_output)
-
```
IvanYashchuk, line 187: This definitely does not belong to "dead code elimination" since we already have "del last used" that would delete unused Proxy outputs.

thunder/torch/__init__.py
```diff
@@ -1436,7 +1436,7 @@ def warn_ndim_not_2():
 def reverse_dims_T(a: TensorLike, /) -> TensorLike:
     if a.ndim != 2:
         warn_ndim_not_2()
-    return a if a.ndim < 2 else prims.transpose(a, tuple(reversed(range(a.ndim))))
+    return prims.shallow_copy(a) if a.ndim < 2 else prims.transpose(a, tuple(reversed(range(a.ndim))))
```
IvanYashchuk: An alternative location for this change is to add a check whether `result` in Symbol.\_\_call__ is a pass-through from args or kwargs and use `prims.shallow_copy` if it is.
https://github.com/Lightning-AI/lightning-thunder/blob/105b8caa04b572cb9aa54f18b49d063e92598e77/thunder/core/symbol.py#L316-L316

beverlylytle: Done.

thunder/tests/test_core.py
```diff
@@ -2111,28 +2111,6 @@ def func(qkv):
     assert "scaled_dot_product_attention" in tuple(bsym.sym.id for bsym in traces[-1].bound_symbols)
 
 
-@instantiate(dtypes=NOTHING)
-def test_no_passthrough_symbol(executor, device, _):
```
beverlylytle, line 2115: On second thought about using `shallow_copy` in `Symbol.__call__`: There's no way to make this test sensible, but it seems to have been added for a reason.  I'm not sure where to find the issue(?) "backward trace contains symbols not present in forward that cause # NotImplementedError" to see what this test is really trying to guard against. @IvanYashchuk

notebooks/liger_kernel.ipynb
```diff
@@ -322,11 +322,12 @@
     "        while bound_symbols:\n",
     "            bsym = bound_symbols.pop(0)\n",
     "            if bsym.sym == litgpt_apply_rope:\n",
-    "                for i, bsym2 in enumerate(bound_symbols):\n",
+    "                while bound_symbols:\n",
```
beverlylytle, line 325: The change here is necessary because the bsym which produces `bsym2.args[0]` is sitting in between `bsym` and `bsym2`.

thunder/core/transform_common.py
t-vi: This is to not create proxies for unused outputs.

beverlylytle: Yes, it is clear that this prevents making proxies for unused inputs.  In the final code output, this leaves us with things of the form `y, _ = f(x)` which is never followed by `del _`, an obvious memory leak.   To insert these `del` statements, we would need to add a corresponding logical branch in `del_last_used`, or worse, in the printing logic itself.  If we delete this selected logical branch and allow proxies to be created for unused inputs, then `del_last_used` takes care of this as it is.  To me, having two dependent but non-obviously related logical branches spread over `dce` and `del_last_used` makes the code hard to read, hard to reason about and hard to maintain.  What is the disadvantage of allowing these proxies to be created?

thunder/tests/test_grad.py
```diff
@@ -44,6 +44,7 @@
     "index_put",
     "batch_norm",
     "instance_norm",
+    "torch_type",
```
beverlylytle, line 48: This opinfo is normally skipped, having been filtered out by `check_bsym_for_vjp`.  But with the shallow copy change, it is no longer filtered with this check and so we force its exclusion here.

thunder/tests/test_examine_memory.py
```diff
@@ -113,5 +113,5 @@ def test_nanogpt_block():
     # We are checking the estimated memory against a fixed value for consistency.
     assert max_mem_fw[0] == 381754368
     assert sum(max_mem_fw[1].values()) == 375462912
-    assert max_mem_bw[0] == 641097728
-    assert sum(max_mem_bw[1].values()) == 440474624
+    assert max_mem_bw[0] == 741761024
+    assert sum(max_mem_bw[1].values()) == 541137920
```
beverlylytle, line 117: Here the output of softmax and the shallow_copy of it (coming from a dropout) are both being saved for backward.  The memory allocation computations have no way of knowing that the inputs to the backward are views of each other.  So, memory allocation is not, in practice, increasing.

thunder/tests/opinfos.py
```diff
@@ -695,6 +695,13 @@ def _abs_torch(x: torch.Tensor | Number):
             dtypes=(datatypes.complex32,),
             devicetypes=(devices.DeviceType.CPU,),
         ),
+        # PyTorch doesn't support abs for bool on cpu
+        DecorateInfo(
+            pytest.mark.skip,
```
crcrpar: Can't we specify reason argument?
```suggestion 
        DecorateInfo(
            pytest.mark.skip(reason="PyTorch doesn't support abs for bool on cpu"),
```

ref: https://github.com/Lightning-AI/lightning-thunder/blob/54baafb4892c001625269414d213b4ae3e77b881/thunder/tests/opinfos.py#L6510

beverlylytle: That's nicer, thank you.

PR #2409, opened 2025-08-09
[do not review] cudnn-frontend `rms_norm`

crcrpar:
## What does this PR do?

many `nan` values seem to be produced. so don't merge this.

EDIT: the nans seem to come from pytorch reference outputs


crcrpar:
it doesn't feel intuitive that a PR gets closed if its target branch is deleted, but TIL

PR #2414, opened 2025-08-11
re-enable cuda-python nb

kshitij12345:
Was disabled in https://github.com/Lightning-AI/lightning-thunder/pull/2410

Few modules from cuda-python were updated - https://nvidia.github.io/cuda-python/cuda-bindings/13.0.0/release/13.0.0-notes.html#breaking-changes

cc @borda

### DIFF
--- a/requirements/notebooks.txt
+++ b/requirements/notebooks.txt
@@ -4,5 +4,5 @@
 ipython[all] ~=8.37.0
 numpy
 liger-kernel == 0.4.0
-cuda-python
+cuda-python >=12.6.1, <14.0.0
 litgpt == 0.5.1
--- a/requirements/notebooks.txt
+++ b/requirements/notebooks.txt
@@ -4,5 +4,5 @@
 ipython[all] ~=8.37.0
 numpy
 liger-kernel == 0.4.0
-cuda-python >=12.0, <13.0.0  # todo
+cuda-python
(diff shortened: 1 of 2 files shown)


requirements/notebooks.txt
```diff
@@ -4,5 +4,5 @@
 ipython[all] ~=8.37.0
 numpy
 liger-kernel == 0.4.0
-cuda-python >=12.0, <13.0.0  # todo
+cuda-python
```
Borda:
```suggestion
cuda-python >=13.0, <14.0.0
```

Borda: btw, would this be compatible with `cuda-python ~12.0`?

t-vi: I would expect some deprecation cycle where the old stuff works, maybe?

kshitij12345: Looks like these modules were introduced in 12.6.1 release with deprecation warnings to the older namespace. We just missed them.

Ref: https://nvidia.github.io/cuda-python/cuda-bindings/13.0.0/release/12.6.1-notes.html#cuda-namespace-cleanup-with-a-new-module-layout

kshitij12345:
```suggestion
cuda-python >=12.6.1, <14.0.0
```

PR #2435, opened 2025-08-18
Fix nvfuserex scatter translation to use compile time scatter dim.

jjsjann123:


thunder/tests/test_nvfuser.py
```diff
@@ -974,34 +974,36 @@ def foo(b):
...
+
+        def foo(inputs: list):
+            inp: torch.Tensor
+            idxs: torch.Tensor
+            src: torch.Tensor
+            inp, idxs, src = inputs
+            out = torch.scatter(inp, dim, idxs, src)
+            return out
+
+        bsz = 4
```
crcrpar: nit-picking: how about moving these constants outside the loop?

thunder/tests/test_nvfuser.py
```diff
@@ -974,34 +974,36 @@ def foo(b):
...
-    fw_trace = thunder.last_traces(jfoo)[-1]
-    fusion_bsyms = tuple(filter(lambda a: a.sym.is_fusion, fw_trace.bound_symbols))
-
-    # assert that everything is merged as a single nvfuser operation
-    assert len(fusion_bsyms) == 1
-    outside_fusion_syms = ["unpack_trivial", "python_return"]
-    assert {el.sym.name for el in fw_trace.bound_symbols if not el.sym.is_fusion} == set(outside_fusion_syms)
+    for dim in (0, 1):
+
+        def foo(inputs: list):
```
crcrpar: Question: would things work even if dim is foo's argument?

jjsjann123: since we don't have dynamic shape/scalar yet, even when we feed that as input here, we are still baking it in as constants. So that wouldn't change how the trace looks like.

But if you think it's cleaner to write the test in that manner. I'll change it to that.

thunder/tests/test_nvfuser.py
```diff
@@ -974,34 +974,38 @@ def foo(b):
...
-    fw_trace = thunder.last_traces(jfoo)[-1]
-    fusion_bsyms = tuple(filter(lambda a: a.sym.is_fusion, fw_trace.bound_symbols))
+        # NOTE nv_enable_scatter to allow scatter operation to go through nvfuser
+        jfoo = thunder.jit(foo, nv_enable_scatter=True)
 
-    # assert that everything is merged as a single nvfuser operation
-    assert len(fusion_bsyms) == 1
-    outside_fusion_syms = ["unpack_trivial", "python_return"]
-    assert {el.sym.name for el in fw_trace.bound_symbols if not el.sym.is_fusion} == set(outside_fusion_syms)
+        inputs = [x, ind, src, dim]
```
kshitij12345: I think the inputs are in the order of `fd.ops.scatter`'s parameters but it would be more readable to have them in same order as `torch.scatter`. That way we can also directly do `thunder.jit(torch.scatter, nv_enable_scatter=True)` and get rid of `foo`.

thunder/tests/test_nvfuser.py
crcrpar: understood.

I've literally no tastes here

thunder/tests/test_nvfuser.py
jjsjann123: good call. Let me re-work that.

kshitij12345 (approved):
LGTM, thanks @jjsjann123

jjsjann123:
@t-vi tagging for review by code owner.

jjsjann123:
failure seems to be unrelated. Merging main hoping it would resolve.

crcrpar (approved):

Issue #2439, opened 2025-08-19
[TE] CI failing with version `GLIBCXX_3.4.32' not found


kshitij12345:
https://github.com/Lightning-AI/lightning-thunder/actions/runs/17073354947/job/48408022731?pr=2404#step:5:180

t-vi:
This is an issue with the environment it runs on (needs an update of libstdc++ in the conda environment)

t-vi:
Fixed in #2443

PR #2440, opened 2025-08-19
[TE] Disable TE CI as it is already failing

kshitij12345:
Related #2439 

cc @borda


riccardofelluga (approved):
Thanks for filing the issue and adding the comment! 👍

t-vi:
With #2443 we are back to the normal, I think (though there are numerical things).

kshitij12345:
Closing in favor of #2443, thank you @t-vi!

PR #2441, opened 2025-08-20
Remove `op_name_to_fn: dict[str, Callable]` from test_cudnn_executor

crcrpar:
## What does this PR do?

As per title.
I tried removing it because the NOTE comment felt a bit outdated to me.


t-vi (approved):
Thank you @crcrpar

PR #2449, opened 2025-08-22
Fix memory leak due to CompileData being in a cycle

kshitij12345:
In `thunder.jit`, there is a pattern that function where inner function like `populate_cache_info` captures `cd` (CompileData) and later there is `cd.populate_cache_info = populate_cache_info` which leads to a reference cycle leading to memory leak. (This also happens with other inner functions like `get_computation_and_inputs`, `acquire_initial_trace`, etc.

https://github.com/Lightning-AI/lightning-thunder/blob/9918ec734cfc838df41483b63b3bb90b3c7dd31d/thunder/__init__.py#L607
https://github.com/Lightning-AI/lightning-thunder/blob/9918ec734cfc838df41483b63b3bb90b3c7dd31d/thunder/__init__.py#L783

Example Cycle: 
image

Repro:
```python
import torch
import thunder
import weakref


memory_at_start = torch.cuda.memory_allocated()

def _allocate_model_in_function():
    model = torch.nn.Linear(256, 256, device="cuda")

    tfn = thunder.jit(model)
    cd = tfn._lc_cd
    return weakref.ref(cd), weakref.ref(model)

refs = _allocate_model_in_function()

assert torch.cuda.memory_allocated() == memory_at_start
for ref in refs:
    assert ref() is None
```

### DIFF
--- a/thunder/__init__.py
+++ b/thunder/__init__.py
@@ -779,10 +779,10 @@ def jit(
         return wrapped
 
     get_computation_and_inputs = decorate_computation_function(get_computation_and_inputs, host_execution_timer)
-    cd.get_computation_and_inputs = get_computation_and_inputs
-    cd.populate_cache_info = populate_cache_info
-    cd.acquire_initial_trace = acquire_initial_trace
-    cd.apply_transforms_and_build_cache_entry = apply_transforms_and_build_cache_entry
+    # cd.get_computation_and_inputs = get_computation_and_inputs
+    # cd.populate_cache_info = populate_cache_info
+    # cd.acquire_initial_trace = acquire_initial_trace
+    # cd.apply_transforms_and_build_cache_entry = apply_transforms_and_build_cache_entry
 
     def update_call_statistics(fn):
         def wrapped(*args, **kwargs):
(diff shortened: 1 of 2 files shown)


thunder/__init__.py
```diff
@@ -779,10 +779,10 @@ def wrapped(*args, **kwargs):
...
 
     get_computation_and_inputs = decorate_computation_function(get_computation_and_inputs, host_execution_timer)
-    cd.get_computation_and_inputs = get_computation_and_inputs
-    cd.populate_cache_info = populate_cache_info
-    cd.acquire_initial_trace = acquire_initial_trace
-    cd.apply_transforms_and_build_cache_entry = apply_transforms_and_build_cache_entry
+    # cd.get_computation_and_inputs = get_computation_and_inputs
+    # cd.populate_cache_info = populate_cache_info
+    # cd.acquire_initial_trace = acquire_initial_trace
+    # cd.apply_transforms_and_build_cache_entry = apply_transforms_and_build_cache_entry
```
kshitij12345: Some tests rely on these and they seem publicly available.

t-vi: Ugh. Thorny. We could make cd an object with these methods if it helps. Or we could replace the capture of cd with a weakref.

kshitij12345: Thanks, will try the weakref approach.

PR #2460, opened 2025-08-25
Remove JAX from tests

crcrpar:
## What does this PR do?

As per title. Because it seems that not many of us have frequently run the tests we have with JAX.

cc @borda


t-vi (approved):
Thank you @crcrpar

PR #2474, opened 2025-08-27
Add full support for `set_grad_enabled`

### LINKED ISSUE #2398: [thunderfx] Bug when no_grad region is split between inductor and thunder
This affects `tomg-group-umd/huginn-0125` model which fails with "element 0 of tensors does not require grad and does not have a grad_fn"

Ref: NVBUG 5422785

Repro
```python
import torch
import random

def fn(x):
    with torch.no_grad():
        # This fallsback inductor because it is autoregistered
        sinc = x.sinc()
    return x + 1 + sinc

x = torch.randn(10, requires_grad=True, device="cuda")

print(fn(x))

o = thunderfx(fn)(x)
print(o)

```

Output (NOTE: Thunder output doesn't have `grad_fn`).
```python
tensor([ 1.3002,  1.8726,  2.0150,  0.7574,  2.0463,  2.1055,  1.7366,  0.7835,
         2.1306, -0.6469], device='cuda:0', grad_fn=<AddBackward0>)
tensor([ 1.3002,  1.8726,  2.0150,  0.7574,  2.0463,  2.1055,  1.7366,  0.7835,
         2.1306, -0.6469], device='cuda:0')
```

Split Graph:
It can be observed that the `no_grad` region is split between inductor fallback and thunder.
Inductor actually sets the `torch._C._set_grad_enabled(False)` which disables grad for PyTorch. And given how thunder implements `torch._C._set_grad_enabled`, it doesn't actually set the PyTorch state but only maintains it internal (see below). Hence while populating the prologue cache for execution (see below), when thunder grabs the PyTorch `torch.is_grad_enabled`, it sees False and incorrectly detects rest of the operations being run under `no_grad` mode.

```python
    def forward(self, l_x_: "f32[10]"):
        submod_0 = self.submod_0();  submod_0 = None
        submod_1 = self.submod_1(l_x_)
        submod_2 = self.submod_2(l_x_, submod_1);  l_x_ = submod_1 = None
        return (submod_2,)
        
    # This goes to thunder
    # Thunder just looks at `_set_grad_enabled` and tracks it internally (and PyTorch is not aware of it).
        def forward(self):
            _set_grad_enabled = torch._C._set_grad_enabled(False);  _set_grad_enabled = None
            return ()
            
    # This goes to inductor
    # This actually sets the _set_grad_enabled to False in PyTorch
        def forward(self, l_x_: "f32[10]"):
            _set_grad_enabled = torch._C._set_grad_enabled(False);  _set_grad_enabled = None
            
             # File: /opt/pytorch/lightning-thunder/test_disable_dynamo.py:7 in fn, code: sinc = x.sinc()
            sinc: "f32[10]" = l_x_.sinc();  l_x_ = None
            return sinc
            
    # This goes to thunder
    # Thunder just looks at `_set_grad_enabled` and tracks it internally (and PyTorch is not aware of it).
    class submod_2(torch.nn.Module):
        def forward(self, l_x_: "f32[10]", sinc: "f32[10]"):
            _set_grad_enabled_1 = torch._C._set_grad_enabled(True);  _set_grad_enabled_1 = None
            
             # File: /opt/pytorch/lightning-thunder/test_disable_dynamo.py:8 in fn, code: return x + 1 + sinc
            add: "f32[10]" = l_x_ + 1;  l_x_ = None
            add_1: "f32[10]" = add + sinc;  add = sinc = None
            return add_1
```

Ref:
https://github.com/Lightning-AI/lightning-thunder/blob/43dc767deb6d6a98cf03f4f1ed64511401f84a4b/thunder/torch/__init__.py#L6056-L6067

https://github.com/Lightning-AI/lightning-thunder/blob/43dc767deb6d6a98cf03f4f1ed64511401f84a4b/thunder/__init__.py#L645-L648

shino16:
Fixes #2398. Thunder has support for `torch.set_grad_enabled` calls that come from `torch.no_grad()` context by managing `cd.is_grad_enabled` without actually calling `torch.set_grad_enabled`. As in #2398, when Thunder's graph splitter splits the region, `torch.set_grad_enabled(True/False)` appears in each segment independently without its counterpart. We should actually call it, and it should be done before connecting the computation to Autograd so that it will populate the `grad_fn` attribute appropriately.

### DIFF
--- a/thunder/__init__.py
+++ b/thunder/__init__.py
@@ -865,13 +865,14 @@ def jit(
         cache_entry, inps, pro_to_epi = get_computation_and_inputs(*args, **kwargs)
 
         result = cache_entry.computation_fn(*inps)
-        # We must do this before connecting to autograd so that
-        # grad_fn attribute will be set on the outputs accordingly
-        pytorch.set_grad_enabled(cd.is_grad_enabled)
         result = maybe_connect_to_autograd(cache_entry, result)
         result = call_epilogue(cache_entry, result, pro_to_epi)
 
         cs.last_computation = cache_entry.computation_fn
+
+        # Reflect the state of is_grad_enabled, as its changes have been recorded only inside Thunder
+        pytorch.set_grad_enabled(cd.is_grad_enabled)
+
         return result
 
     if isinstance(fn, pytorch.nn.Module):
--- a/thunder/__init__.py
+++ b/thunder/__init__.py
@@ -865,6 +865,9 @@ def jit(
         cache_entry, inps, pro_to_epi = get_computation_and_inputs(*args, **kwargs)
 
         result = cache_entry.computation_fn(*inps)
+        # We must do this before connecting to autograd so that
+        # grad_fn attribute will be set on the outputs accordingly
+        pytorch.set_grad_enabled(cd.is_grad_enabled)
         result = maybe_connect_to_autograd(cache_entry, result)
         result = call_epilogue(cache_entry, result, pro_to_epi)
 
--- a/thunder/core/symbol.py
+++ b/thunder/core/symbol.py
@@ -327,11 +327,13 @@ class Symbol:
 
         cd = get_compile_data()
         if cd is not None and not cd.is_grad_enabled:
+            flat_args, _ = tree_flatten((args, kwargs))
+            flat_arg_ids = {id(arg) for arg in flat_args}
             # If grad is disabled using `torch.no_grad` or `torch._C._set_grad_enabled(False)`,
             # tag the results with `DETACHED_AUTOGRAD_GRAPH` which makes this Symbol a constant for
             # vjp transform (applied later).
             def tag_tensorproxy_output_as_detached(proxy):
-                if isinstance(proxy, TensorProxy):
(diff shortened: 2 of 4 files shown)


shino16:
Currently, the following does not work:
```py
import torch, thunder

torch.set_grad_enabled(False)
def fn(x):
    with torch.enable_grad():
        return x.sin()

x = torch.randn((), requires_grad=True, device="cuda")
assert fn(x).grad_fn is not None
assert thunder.jit(fn)(x).grad_fn is not None  # AssertionError
assert thunder.dynamo.thunderfx(fn)(x).grad_fn is not None  # AssertionError
```

This is because:
1. the epilogue calls `unpack_trivial(t0)`, where `t0` represents `x.sin()`
2. the output of `unpack_trivial` is also `t0`
3. autograd is disabled when tracing the epilogue
4. so the output of `unpack_trivial` is tagged `DETACHED_AUTOGRAD_GRAPH` meaning it will be treated as constant

This maybe a similar issue to https://github.com/Lightning-AI/lightning-thunder/pull/2396#issuecomment-3150831485.

shino16:
Let me close this PR as the purpose has somewhat changed. I'll reopen as a new PR.

PR #2475, opened 2025-08-28
Skip TE test on SM120+ as Float8BlockScaling is currently unsupported in thunder

kshitij12345:
Reference NVBUG: 5400059

On SM120/121, TE defaults to Float8BlockScaling recipe which is currently unsupported.

### DIFF
--- a/thunder/tests/test_transformer_engine_v2_executor.py
+++ b/thunder/tests/test_transformer_engine_v2_executor.py
@@ -6,6 +6,7 @@ from torch.testing import assert_close
 
 import thunder
 from thunder.tests.framework import requiresCUDA
+from thunder.tests.utils import skip_on_sm120_and_sm121, is_sm120_orsm121
 
 transformer_engine_module = pytest.importorskip(
     "transformer_engine", reason="transformer_engine was not found, skipping the tests."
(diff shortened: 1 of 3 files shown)


thunder/tests/test_transformer_engine_v2_executor.py
```diff
@@ -381,6 +391,7 @@ def fn(x, w):
 
 @requiresCUDA
 @pytest.mark.parametrize("fp8_recipe", recipes, ids=recipe_ids)
+@skip_on_sm120_and_sm121
```
kshitij12345, line 397: Need to disable even if recipe other than default is specified because of https://github.com/Lightning-AI/lightning-thunder/issues/2477

thunder/tests/utils.py
```diff
@@ -36,3 +38,20 @@ def filter_differentiable_outputs(outputs):
...
     return list(filter(is_output_differentiable, outputs))
+
+
+def is_sm120_orsm121():
+    return torch.cuda.get_device_capability() in ((12, 1), (12, 0))
+
+
+def skip_on_sm120_and_sm121(fn):
+    # NOTE: On SM120/121, TE defaults to using Float8BlockScaling
+    #       which is currently unsupported in thunder.
```
riccardofelluga: nit. I'd move this in TE's related usage, since here it is supposed to be a general util

thunder/tests/utils.py
```diff
@@ -36,3 +38,20 @@ def filter_differentiable_outputs(outputs):
         outputs = [outputs]
 
     return list(filter(is_output_differentiable, outputs))
+
+
+def is_sm120_orsm121():
+    return torch.cuda.get_device_capability() in ((12, 1), (12, 0))
```
riccardofelluga, line 44: In https://github.com/NVIDIA/TransformerEngine/pull/1939 they do >= 12, should we do the same or keep the == 120/121 ?

riccardofelluga (approved):
Couple of nits, for the rest thanks for keeping an eye out on TE.

thunder/tests/utils.py
kshitij12345: I think it is better to check only for 120/121 as there is nothing else which is above that at the moment. So, if something comes up, it would be better to get signal than this being skipped automatically.

thunder/tests/utils.py
kshitij12345: Have moved the comment to TE tests. Have kept the function here as it doesn't have anything TE specific.

t-vi (approved):
Thank you @kshitij12345 @riccardofelluga

thunder/tests/utils.py
riccardofelluga: okok good by me

PR #2478, opened 2025-08-28
Add MOE TP example

kshitij12345:


test_moe_distributed.py
```diff
@@ -0,0 +1,339 @@
...
+        return distribute_module(
+            module,
+            device_mesh,
+            self._partition_fn,
+            partial(self._prepare_input_fn, self.input_layouts, self.desired_input_layouts),
+            partial(self._prepare_output_fn, self.output_layout, self.use_local_output),
+        )
+
+
+def parallelize_moe_model(model: Llama4MoE, device_mesh: torch.distributed.DeviceMesh):
```
kshitij12345, line 277: Parallel Plan

test_moe_distributed.py
```diff
@@ -0,0 +1,339 @@
...
+
+    yield
+
+    # Restore
+    torch.set_default_dtype(prev_dtype)
+    torch.set_default_device(prev_device)
+
+
+# Referred from torchtitan: https://github.com/pytorch/torchtitan/blob/827255bb484d0f0a97fe5bec22b70f8b4750f685/torchtitan/experiments/llama4/infra/expert_parallel.py#L25
+class GroupedLinearColwiseParallel(ParallelStyle):
```
wujingyue, line 181:
```suggestion
class GroupedLinearColwiseParallel(ColwiseParallel):
```

IIUC, ColwiseParallel almost works except https://github.com/pytorch/pytorch/blob/4fd761fecca86693c8cd7755ea6d13d61bb94090/torch/distributed/tensor/parallel/style.py#L155. We can probably let GroupedLinearColwiseParallel inherit from ColwiseParallel. This way, all we need to override is `_apply`.

kshitij12345: I think we will also need to override, `_preparse_input_fn` as ColwiseParallel expects only one input but for GroupedLinear, we also have `offsets`.

https://github.com/pytorch/pytorch/blob/4fd761fecca86693c8cd7755ea6d13d61bb94090/torch/distributed/tensor/parallel/style.py#L99-L117

kshitij12345: We will also have to update `_partition_fn` to consider the fact that the weights are pre-transposed and shard accordingly. We also need to update `_apply` to not error on `GroupedLinear` as mentioned above. So, it doesn't seem like inheriting from `ColwiseParallel` will be very useful.

wujingyue: > ColwiseParallel expects only one input

Thanks -- I missed that and I'm convinced it makes sense to write GroupedLinearColwiseParallel from scratch. 

However, I'd try to follow YAGNI and minimize the number of `if`s. For example, IIUC, 
1. We can remove the `__init__` arguments `input_layouts`, `output_layouts` and `use_local_output` because we only use the defaults. 
2. We can assume input_layouts are replicated so input processing is simply a from_local. 
3. We can assume output_layouts are `Shard(-1)` so output processing is simply a to_local. 

Also, for the upcoming nvFuser integration, I'd simply use ColwiseParallel and RowwiseParallel. nvFuser only needs these ParallelStyles as annotations. It doesn't need to call `_apply`, `_prepare_input_fn` or `_prepare_output_fn` at all. Refer to https://github.com/NVIDIA/Fuser/blob/9bbe0f8c090a505c34a467a8c0afc8f4b65ec5b3/tests/python/multidevice/test_deepseek_v3.py#L121

kshitij12345: Right, those can be simplified with stricter constraints on input and output layouts.

kshitij12345:
Merged in #2582

PR #2479, opened 2025-08-28
Ensure autograd is enabled before connecting Thunder-compiled fn to autograd

### LINKED ISSUE #2398: [thunderfx] Bug when no_grad region is split between inductor and thunder

shino16:
Fixes #2398. This PR ensures the interaction with autograd when it is disabled globally but gets enabled inside the traced region.

### What this PR does
* Puts `ThunderFunction.apply` call under `torch.enable_grad()` region
* Disallows tagging a `TensorProxy` as `DETACHED_AUTOGRAD_GRAPH` if the proxy is also an argument of the trace
  This is correct because that tensor is created somewhere else. This fixes [#2474 (comment)](https://github.com/Lightning-AI/lightning-thunder/pull/2474#issuecomment-3232852332)

### What this means
This PR enables calling `torch.set_grad_enabled(True/False)` arbitrarily in a Thunder-traced region, while populating `grad_fn` attribute on appropriate tensors to enable backward. Another PR will reflect `cd.is_grad_enabled` (which is maintained only inside Thunder) to PyTorch after execution, which will make a full support for `torch.set_grad_enabled)`.


thunder/tests/test_core.py
```diff
@@ -2559,6 +2560,41 @@ def foo5(x):
...
+            torch.set_grad_enabled(next_enable)
+            next_enable = not next_enable
+        if ends_with_op:
+            x = x.sin()
+        return x
+
+    for global_grad_enabled in [False, True]:
+        for n_flips, next_enable, starts_with_op, ends_with_op in itertools.product(
+            range(4), [False, True], [False, True], [False, True]
+        ):
```
crcrpar: At glance it seems that this nested loop can be `pytest.mark.parametrize` and each test instance is responsible for one set of parameters.
What do you think?

shino16: I thought making 4\*2\*2\*2 separate pytest tests may be too time-consuming, but it is surely doable.

thunder/core/symbol.py
```diff
@@ -327,11 +327,14 @@ def __call__(self, *args, **kwargs):
 
         cd = get_compile_data()
         if cd is not None and not cd.is_grad_enabled:
+            flat_args, _ = tree_flatten((args, kwargs))
+            flat_arg_ids = {id(arg) for arg in flat_args}
```
IvanYashchuk: In Thunder we think of two different Proxy objects with the same name as identical. Let's use the `.name` attribute here as in many other places in Thunder codebase.

thunder/executors/torch_autograd.py
```diff
@@ -168,17 +168,18 @@ def connect_to_autograd(
...
+            saved_tensors,
+            saved_other,
+            is_differentiable_outputs,
+            flat_output,
+            *flat_args,
+        )
+        if side_channel is not None:
+            # we need to pass the inputs to avoid "leave has moved inside the graph"
+            # if the function returns an argument as is
+            ThunderOutputFunction.apply(dummy_res, side_channel, *flat_args)
```
IvanYashchuk: Indentation changes might make it a little more difficult to trace back what commit introduced the code. Smaller diff would be with `torch.enable_grad` applied as decorator on `connect_to_autograd`.

thunder/tests/test_core.py
IvanYashchuk: It simply aids debugging when failing parametrization is seen in the list of tests and not implicit to the test implementation itself. Runtime implications of this are interesting. Would you check what is the different in reality to verify your guess?

t-vi (approved):
Thank you @shino16 @crcrpar @IvanYashchuk

thunder/tests/test_core.py
shino16: The test time changed from 3.22s to 5.22s, which seems totally acceptable. Thank you for your comments!

thunder/core/symbol.py
shino16: Updated. Thank you!

thunder/executors/torch_autograd.py
shino16: True, this makes so much sense. Updated.

thunder/tests/test_core.py
crcrpar: As for the execution time, it might make a difference by not covering multiple executors, devices, and dtypes

shino16:
I'm having trouble reproducing the test failures in my environment. I'll work on this later.

shino16:
The issue was the new test `test_set_grad_enabed` not restoring the original state of `torch.is_grad_enabled`. I fixed using [`request.addfinalizer`](https://docs.pytest.org/en/6.2.x/reference.html?highlight=finalizer#pytest.FixtureRequest.addfinalizer).

thunder/tests/test_core.py
shino16: Now it's not parametrized by executors and devices.

PR #2480, opened 2025-08-28
Reflect `cd.is_grad_enabled` to PyTorch after everything is done

shino16:
This PR:
* Gives full support for `torch.set_grad_enabled`.
* Ensures the external state (`torch.is_grad_enabled`) is updated correctly after execution.
* Ensures correct behavior when the graph splitter splits `torch.no_grad()` region or `torch.enable_grad()` region into multiple segments.

This PR builds upon
- [ ] #2479


shino16:
Outdated.

PR #2485, opened 2025-08-29
Replace linear checker for TEv2

### LINKED ISSUE #2477: [TE] TE v2 should not use checker from TE v1
Checker from TE v1 tries determines if a Linear is valid based on the recipe passed as a compile option. This is different from TE v2 which gets the recipe from `FP8GlobalStateManager.get_fp8_recipe()` which can lead to mismatch.

https://github.com/Lightning-AI/lightning-thunder/blob/38733f021db4e0108f5c11c52fcc52d005462d5c/thunder/executors/transformer_engineex.py#L471-L480

Sample Program where there might be a mismatch in recipe (to be run B200 to see the mismatch)
```python
import torch
import thunder
from thunder.executors.transformer_engine_v2ex import transformer_engine_v2_ex, TransformerEngineTransformV2
from transformer_engine.common import recipe

x = torch.randn(16, 16, device="cuda")

linear = te.Linear(16, 16)

tfn = thunder.jit(torch.nn.functional.linear, executors=[transformer_engine_v2_ex], transforms=[TransformerEngineTransformV2()])

with te.fp8_autocast(fp8_recipe=recipe.DelayedScaling()):
    linear(x)
    
    # On B200, the default recipe is MXFP8, so the `_linear_checker` will determine based on `MXFP8` recipe but based on the fp8_autocast, it should have been for DelayedScaling.
    tfn(x, linear.weight)

```

cc: @riccardofelluga

riccardofelluga:
This PR fixes #2477 by copying the _linear_checker function from TEv1 to TEv2 implementation. It also updates it by using recipe methods to identify the recipe type.

### DIFF
--- a/thunder/executors/transformer_engine_v2ex_impl.py
+++ b/thunder/executors/transformer_engine_v2ex_impl.py
@@ -7,10 +7,10 @@ from thunder.core.prims import linear as linear_prim
 from thunder.core.prims import get_grad, put_grad
 from thunder.core.proxies import AnyProxy, TensorProxy
 from thunder.extend import StatefulExecutor, register_executor
-from thunder.executors.transformer_engineex import _linear_checker
 import thunder.torch as ltorch
 from thunder import Transform
 from thunder.core import prims
+import thunder.core.devices as devices
 from thunder.core.proxies import Proxy, Variable, unvariableify, variableify
 from thunder.core.trace import from_trace, TraceProvenance, TraceTag, TraceCtx
 from thunder.core.transforms import (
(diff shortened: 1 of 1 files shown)


thunder/executors/transformer_engine_v2ex_impl.py
```diff
@@ -250,6 +252,39 @@ def _te_linear_grad_transform(a, w, bias):
...
+        # DelayedScaling and MXFP8BlockScaling have different shape requirements.
+        if fp8_recipe.delayed():
+            return check_dim_for_fp8_exec(a)
+
+        assert fp8_recipe.mxfp8()
+        shape = a.shape
+        return shape[0] % MXFP8_BLOCK_SCALING_SIZE == 0 and shape[1] % MXFP8_BLOCK_SCALING_SIZE == 0
+
+    # Inputs must be on CUDA and
+    # input sizes must satisfy -> dim0 is divisible by 8 and dim1 is divisible by 16.
```
kshitij12345: IIRC, this constraint only applies to the DelayedScaling. So this should be updated to `input sizes must satisfy size constraints based on the recipe.` (also true for v1 checker)

t-vi (approved):
Thank you @riccardofelluga @kshitij12345

PR #2486, opened 2025-08-30
thunderfx: Avoid failure when example_value does not have attr of grad_fn

crcrpar:
## What does this PR do?

This is reopened 2419.
Some values, such as `torch.SymInt` would not have `grad_fn` attr.

Signed-off-by: Masaki Kozuki <mkozuki@nvidia.com>

### DIFF
--- a/thunder/dynamo/splitter.py
+++ b/thunder/dynamo/splitter.py
@@ -175,7 +175,7 @@ def _splitter(
             for n in graph_module.graph.nodes:
                 if n.op == "output":
                     for n in n.all_input_nodes:
-                        if "example_value" not in n.meta or n.meta["example_value"].grad_fn is None:
+                        if "example_value" not in n.meta or getattr(n.meta["example_value"], "grad_fn") is None:
                             is_differentiable_outputs.append(False)
                         else:
                             is_differentiable_outputs.append(True)
--- a/thunder/dynamo/splitter.py
+++ b/thunder/dynamo/splitter.py
@@ -175,7 +175,7 @@ def _splitter(
             for n in graph_module.graph.nodes:
                 if n.op == "output":
                     for n in n.all_input_nodes:
-                        if "example_value" not in n.meta or getattr(n.meta["example_value"], "grad_fn") is None:
+                        if "example_value" not in n.meta or getattr(n.meta["example_value"], "grad_fn", None) is None:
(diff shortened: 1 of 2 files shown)


crcrpar:
When closing 2419, I misunderstood the change

thunder/dynamo/splitter.py
```diff
@@ -175,7 +175,7 @@ def is_thunder_supported_partition(node: torch.fx.Node) -> bool:
             for n in graph_module.graph.nodes:
                 if n.op == "output":
                     for n in n.all_input_nodes:
-                        if "example_value" not in n.meta or n.meta["example_value"].grad_fn is None:
+                        if "example_value" not in n.meta or getattr(n.meta["example_value"], "grad_fn") is None:
```
shino16: Maybe this is meant to be `getattr(n.meta["example_value"], "grad_fn", None)`?

thunder/tests/test_dynamo.py
```diff
@@ -1697,3 +1697,26 @@ def test_fn(x):
...
+
+            add: "Sym(s27 + 1)" = l_self_cumulative_length_1_ + 1
+            l_self_cumulative_length_1_ = None
+            return (add,)
+
+    module = GraphModule()
+    ref_inputs = (100,)
+    reference = module(*ref_inputs)
+
+    jitted = thunderfx(GraphModule(), dynamic=True)
```
shino16, line 1716: I'm trying to understand the `dynamic=True` option here. [The official document](https://docs.pytorch.org/docs/main/generated/torch.compile.html) has some explanation, but would teach me why it matters here?

thunder/dynamo/splitter.py
crcrpar: good catch

thunder/tests/test_dynamo.py
crcrpar: As in the issue 2420, the error this PR tries to fix is caused by [`torch.SymInt`](https://docs.pytorch.org/docs/stable/torch.html#symbolic-numbers) inside `torch.fx.GraphModule` and `torch.SymInt` would never be there when there's no dynamism.

shino16: Ah! That makes sense, thank you.

PR #2487, opened 2025-08-30
Inference benchmark of "meta-llama/Llama-4-Maverick-17B-128E"

crcrpar:
## What does this PR do?

- [ ] Figure out why thunderfx_benchmark_report fails
- [ ] Figure out why `--mode=inductor` fails
- [ ] Think about how to override `SomeModule.forward`, e.g. [`Llama4TextMoe.forward`](https://github.com/huggingface/transformers/blob/ff8b88a948fc2f6aba421ca64ad165291928dcee/src/transformers/models/llama4/modeling_llama4.py#L147)


thunder/benchmarks/benchmark_inference.py
```diff
@@ -0,0 +1,605 @@
...
+                p.requires_grad_(False)
+
+        # `thunderfx` seems to hide the access to vocab_size somewhere so
+        # store it here before any compiler is applied.
+        self.vocab_size = model.vocab_size
+
+        if not self.config.disable_moe_replacement:
+            _replace_llama4_moe(model)
+        if self.config.enable_nvfp4:
+            _quantize_llama4(model)
```
jjsjann123, line 177: sweet~~~

thunder/benchmarks/benchmark_inference.py
```diff
@@ -0,0 +1,605 @@
...
+            "*.layers.*.feed_forward.shared_expert.up_proj": ColwiseParallel(use_local_output=False),
+            "*.layers.*.feed_forward.shared_expert.down_proj": RowwiseParallel(use_local_output=True),
+        }
+
+        if self.config.dtensor_single_gpu or WORLD_SIZE > 1:
+            model = parallelize_module(model, mesh, tp_plan)
+
+            # Required as that doesn't understand inference mode
+            for p in model.parameters():
+                p.requires_grad_(False)
```
jjsjann123, line 168: I thought we always needed to mark gradient as `False`, just to avoid segmentation from autograd...

thunder/benchmarks/benchmark_inference.py
```diff
@@ -0,0 +1,605 @@
...
+
+        # `thunderfx` seems to hide the access to vocab_size somewhere so
+        # store it here before any compiler is applied.
+        self.vocab_size = model.vocab_size
+
+        if not self.config.disable_moe_replacement:
+            _replace_llama4_moe(model)
+        if self.config.enable_nvfp4:
+            _quantize_llama4(model)
+        self.model = self._compile_model(model)
```
jjsjann123, line 178: note for myself, for debugging purpose, we might want to compile only a section of the entire model...

thunder/benchmarks/layers_for_inference_benchmark.py
```diff
@@ -0,0 +1,584 @@
...
+    bias: torch.Tensor | None,
+) -> torch.Tensor:
+    hp_weight = dequantize_to_dtype(
+        fp4_weight, weight_scaling_factor, weight_global_scale, activation.dtype, fp4_weight.device, 16
+    )
+    return activation @ hp_weight + bias
+
+
+@torch.library.register_fake("nvf_cutlass::f16a_nvfp4weight_scaled_mm")
+def _(
```
jjsjann123, line 218: dumb question, fake function here is just for type/shape inference? Is this needed because we are using torch.compile?

thunder/benchmarks/layers_for_inference_benchmark.py
```diff
@@ -0,0 +1,584 @@
...
+        c_strides: torch.Tensor,
+    ) -> None:
+        self.register_buffer("fp4_weight", fp4_weight)
+        self.register_buffer("weight_scaling_factor", weight_scaling_factor)
+        self.register_buffer("weight_global_scale", weight_global_scale)
+        self.register_buffer("ab_strides", ab_strides)
+        self.register_buffer("c_strides", c_strides)
+
+    def forward(self, hidden_states: torch.Tensor, offsets: torch.Tensor) -> torch.Tensor:
+        tokens_per_group = offsets[1:] - offsets[:-1]
```
jjsjann123, line 425: Note for myself. we might not like this in the decomposition. These offsets might be adding more complexity.

I know I said that functionally, we could compute these inside each grouped_mm. Since these are identical for all grouped_mm in the moe, hoist `blockscale_offsets` and `problem_sizes` might give us some perf gain.

thunder/benchmarks/layers_for_inference_benchmark.py
```diff
@@ -0,0 +1,584 @@
...
+        problem_sizes = torch.stack(
+            [
+                tokens_per_group,
+                torch.full_like(tokens_per_group, hidden_states.size(0)),
+                torch.full_like(tokens_per_group, self.fp4_weight.size(2) * 2),
+            ],
+            dim=1,
+        )
+        blockscale_offsets = offsets
```
jjsjann123: BTW, the computation here isn't right. It's roughly

`blockscale_offsets = cumsum(ceildiv(tokens_per_group, 128) * 128);`

thunder/benchmarks/layers_for_inference_benchmark.py
jjsjann123: I understand that changing the signature might be tricky since we are doing a monkey patching.

Luckily we are also defining `GroupedLinear`, I think we can get away with changing the api for fp16/bf16 grouped linear to keep some stub args as well.

thunder/benchmarks/layers_for_inference_benchmark.py
crcrpar: I seem to have misunderstood what https://github.com/NVIDIA/Fuser/blob/54a9d813d2ea0fbeb0bfa94eec2963c403cc192e/tests/python/test_narrow_precision.py#L161-L162 means

jjsjann123: I don't think you read it wrong, that code comment is wrong and it's definitely my fault. Thanks for pointing it out.

thunder/benchmarks/benchmark_inference.py
```diff
@@ -0,0 +1,605 @@
...
+        config = AutoConfig.from_pretrained(model_id)
+
+        if hasattr(config, "text_config"):
+            config = config.text_config
+        if self.config.num_layers:
+            config.num_hidden_layers = self.config.num_layers
+
+        self.hf_config = config
+
+        # TODO: Apply NVFP4 quantization to weights if requested
```
crcrpar: this should be removed since I've `_quantize_llama4` above
```suggestion
```

thunder/benchmarks/layers_for_inference_benchmark.py
crcrpar: this "fake" is necessary for torch.compile not to do graph breaks around the op.
Yes we use TorchDynamo through thunderfx, but also the manual registration translates fake functions into thunder meta functions

thunder/benchmarks/layers_for_inference_benchmark.py
crcrpar: the code here needs to be fixed following J's comment

crcrpar:
on gb200, pjnl-20250910

### ThunderFX
```
============================================================
BENCHMARK RESULTS - meta-llama/Llama-4-Maverick-17B-128E thunder
============================================================

Throughput Metrics:
  Overall Throughput: 65.08 tokens/sec
  Prefill Throughput: 44501.49 tokens/sec
  Decode Throughput: 65.34 tokens/sec
  Latency: 38.40 ms/token

Latency Breakdown:
  Time to First Token (TTFT): 46.02 ms
  Time Between Output Tokens (TBOT): 38.38 ms
  Prefill Time: 46.02 ms
  Decode Time: 19613.65 ms
  Total Generation Time: 19659.67 ms

Memory Usage:
  Current Memory: 37.44 GB
  Peak Memory: 50.57 GB

Variance Analysis:
  Throughput Std Dev: 118693.20 ms
  TTFT Std Dev: 0.33 ms
```

### Eager
```
============================================================
BENCHMARK RESULTS - meta-llama/Llama-4-Maverick-17B-128E eager
============================================================

Throughput Metrics:
  Overall Throughput: 146.23 tokens/sec
  Prefill Throughput: 36188.80 tokens/sec
  Decode Throughput: 148.34 tokens/sec
  Latency: 6.84 ms/token

Latency Breakdown:
  Time to First Token (TTFT): 56.59 ms
  Time Between Output Tokens (TBOT): 6.74 ms
  Prefill Time: 56.59 ms
  Decode Time: 3445.55 ms
  Total Generation Time: 3502.14 ms

Memory Usage:
  Current Memory: 37.39 GB
  Peak Memory: 57.82 GB

Variance Analysis:
  Throughput Std Dev: 53.66 ms
  TTFT Std Dev: 0.09 ms
```

thunder/benchmarks/layers_for_inference_benchmark.py
```diff
@@ -0,0 +1,591 @@
...
+) -> torch.Tensor:
+    return torch.empty((activation.size(0), fp4_weight.size(0)), device=activation.device, dtype=activation.dtype)
+
+
+# TODO: Update this accordingly to the progress of nvfp4 kernel implementation.
+# An alternative is to use `_register_nvfuser_translator` of https://github.com/Lightning-AI/lightning-thunder/pull/2481
+# instead of updating this function itself.
+@torch.library.custom_op("nvf_cutlass::f16a_nvfp4weight_scaled_grouped_mm", mutates_args=())
+def nvfuser_f16a_nvfp4weight_scaled_grouped_mm(
+    activation: torch.Tensor,
```
jjsjann123, line 233: Note formyself. I think we are missing global_scale for activation. (which uses heuristics, instead of computed from live data).

thunder/benchmarks/layers_for_inference_benchmark.py
```diff
@@ -0,0 +1,591 @@
...
+
+    def __init__(
+        self,
+        in_features: int,
+        out_features: int,
+        *,
+        fp4_weight: torch.Tensor | nn.Parameter,
+        weight_scaling_factor: torch.Tensor | nn.Parameter,
+        weight_global_scale: torch.Tensor | nn.Parameter | None,
+        bias: torch.Tensor | nn.Parameter | None,
```
jjsjann123, line 284: same note here, global scale for activation needs to be passed in during init.

thunder/benchmarks/layers_for_inference_benchmark.py
```diff
@@ -0,0 +1,591 @@
...
+        tokens_per_group = offsets[1:] - offsets[:-1]
+        problem_sizes = torch.stack(
+            [
+                tokens_per_group,
+                torch.full_like(tokens_per_group, hidden_states.size(0)),
+                torch.full_like(tokens_per_group, self.fp4_weight.size(2) * 2),
+            ],
+            dim=1,
+        )
+        blockscale_offsets = torch.cumsum(torch.ceil(tokens_per_group, 128) * 128)
```
jjsjann123, line 434: sweet~~~

t-vi:
As this is code outside CI, I'm merging.

PR #2495, opened 2025-09-01
[DTensor] Use enum for PrimIDs similar to prims.PrimIDs

kshitij12345:
Add `DTensorPrimIDs` enum similar to `PrimIDs` enum to make importing and supporting DTensor primitives simpler (look at the update in `nvfuserex_impl.py`)

Tested with existing tests in `test_dtensor.py` - `pytest thunder/tests/distributed/test_dtensor.py`

### DIFF
--- a/thunder/core/symbol.py
+++ b/thunder/core/symbol.py
@@ -327,11 +327,14 @@ class Symbol:
 
         cd = get_compile_data()
         if cd is not None and not cd.is_grad_enabled:
+            flat_args, _ = tree_flatten((args, kwargs))
+            flat_arg_names = {arg.name for arg in flat_args if isinstance(arg, TensorProxy)}
+
             # If grad is disabled using `torch.no_grad` or `torch._C._set_grad_enabled(False)`,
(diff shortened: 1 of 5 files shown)


thunder/torch/experimental/dtensor_torch_and_prims.py
```diff
@@ -26,6 +27,16 @@
 
 import torch
 
+
+class DTensorPrimIDs(Enum):
+    # DTensor-specific primitives
+    CHECK_DTENSOR_SPEC_REPR = auto()
+    MUL_PRIM = auto()
```
shino16: I thought this could be `MUL` as "prim" is in the class name, but I understand it as being aligned with `dtensor_mul_prim`.

shino16 (commented):
Thanks for requesting me a review!

thunder/torch/experimental/dtensor_torch_and_prims.py
kshitij12345: Agreed that `PRIM` in `MUL_PRIM` is redundant. Updating, thanks!

PR #2502, opened 2025-09-04
Enable direct bindings in Thunder

rdspring1:
This PR enables the direct bindings for nvfuser executor if `DIRECT_BINDINGS_SUPPORTED_VERSION >= 0.2.34`.

* Legacy bindings will remain in the nvfuser pip wheel under `import nvfuser`, but it will be deprecated and no longer receive new operations.
* `dockers/ubuntu-cuda/Dockerfile`, `scripts/validate_build.py`, `thunder/tests/conftest.py`, `thunder/executors/nvfuserex.py` contain `import nvfuser` for version checks and `FusionCache.reset()`. These files will need to be updated when legacy bindings are removed.

## What is direct bindings?
* TL;DR: It is direct pybind11 mapping from nvfuser CPP API to python.
* The python API is the same as legacy bindings, so Thunder is minimally affected.
* Direct bindings does not have a static Fusion Definition cache like legacy bindings.

## Why direct bindings?
* The legacy bindings utilized an intermediate representation from nvFuser's underlying FusionIR where only the definition ops were exposed in this intermediate representation to allow for graph caching.  By moving to direct bindings, we are allowing for the possibility of adding more bindings to expose the scheduling of operations from python to allow users to more directly interact with nvFuser internals from python.  One example where exposing scheduling operations is useful is multi-gpu support where the mesh dimensions of a tensor are annotated via a scheduling operation.
* It is difficult to expose multi-gpu scheduling primitives with legacy bindings.
* The static Fusion Definition cache can be a memory hog and cumbersome to update.

### DIFF
--- a/thunder/executors/nvfuserex_impl.py
+++ b/thunder/executors/nvfuserex_impl.py
@@ -70,7 +70,7 @@ from thunder.executors.nvfuserex import nvfuser_version
 # NOTE This impl file is here because nvFuser may not be available, so it's imported conditionally
 #   by nvfuserex.py when nvFuser is available.
 
-DIRECT_BINDINGS_SUPPORTED_VERSION = LooseVersion("0.2.32")
+DIRECT_BINDINGS_SUPPORTED_VERSION = LooseVersion("0.2.34")
 DTENSOR_SUPPORTED_VERSION = LooseVersion("0.2.28")
 if nvfuser_version() >= DIRECT_BINDINGS_SUPPORTED_VERSION:
     import nvfuser_direct as nvfuser


thunder/executors/nvfuserex_impl.py
```diff
@@ -370,19 +369,15 @@ def check_dtensor_tracing_and_runtime_metadata(inp):
...
         with fd:
             definition(fd)
             multidevice_schedule(fd, sorted_unique_inputs)
     else:
-        # NOTE nvFuser's default max length is 1024 operations at the time of this writing
-        #   This arbitrarily increases it to 9999
-        # TODO Review splititng very large fusions or removing the max length restriction completely
-        #   See "Very large nvFuser fusions hit max_length"
-        fd = FusionDefinition(max_length=MAX_LENGTH)
+        fd = FusionDefinition()
```
rdspring1, line 389: nvfuser `0.2.32` bumped the default max length for legacy bindings to 9999 while the direct bindings does not have a fusion cache.

thunder/executors/nvfuserex_impl.py
```diff
@@ -629,42 +624,6 @@ def __init__(self):
         else:
             self.set_fuel(FUEL_LEVEL.UNLIMITED)
 
-        env_var_save_serde = os.getenv("ENABLE_NVFUSER_SERIALIZATION", None)
```
rdspring1, line 637: Direct bindings does not support serialization. Removing it for simplicity.

kshitij12345: Just curious to know if there is a plan to support this in future.

rdspring1: Serialization would need to be added on top of [[RFC] Create LRU cache for direct bindings. #4893](https://github.com/NVIDIA/Fuser/pull/4893). It isn't on the roadmap but it has been accounted for.

rdspring1:
@t-vi This PR is ready for your review. Also, what are the 10 required tests that haven't run yet?

t-vi:
@rdspring1  unfortunately, the CI is not up currently. I'm running the tests manually now and will merge when they pass.

PR #2507, opened 2025-09-08
torch.cumsum api change

jjsjann123:
switch to native cumsum in nvfuser instead of decomposition via matmul op.

### DIFF
--- a/examples/quickstart/requirements.txt
+++ b/examples/quickstart/requirements.txt
@@ -1,4 +1,4 @@
-transformers
+transformers==4.55.4
 accelerate
 nvfuser-cu128-torch27
 nvidia-cudnn-frontend
--- a/thunder/__about__.py
+++ b/thunder/__about__.py
@@ -1,7 +1,7 @@
-__version__ = "0.2.5.dev0"
+__version__ = "0.2.6.dev0"
 __author__ = "Lightning-AI et al"
 __author_email__ = "community@lightning.ai"
-__copyright__ = f"2024 {__author__}"
+__copyright__ = f"2024, 2025 {__author__}"
 __homepage__ = "https://github.com/Lightning-AI/lightning-thunder"
 __docs__ = "Lightning Thunder project."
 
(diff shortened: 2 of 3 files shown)


thunder/executors/nvfuserex_impl.py
```diff
@@ -3248,25 +3241,13 @@ def cumsum_transform(
...
         out_dtype = lcdtype_to_nvdtype(dtypes.to_dtype(dtype))
 
     nv_a = getnv(a, fd, lc_to_nv_map)
-    nv_a = fd.ops.cast(nv_a, compute_dtype)
-
-    mask = fd.ops.full((a.numel, a.numel), fd.define_scalar(1), compute_dtype)
-    mask = fd.ops.triu(mask)
-
-    out = fd.ops.matmul(nv_a, mask)
+    out = fd.ops.cumsum(nv_a, dim)
```
jjsjann123: we can guard on nvfuser version if public CI has older nvfuser version running without cumsum. :crossed_fingers:

jjsjann123:
errrr. pytorch cumsum does type promotion for integer/bool... so nvfuser hits type mismatch... I need to go fix that. :face_exhaling:

jjsjann123:
https://github.com/NVIDIA/Fuser/pull/5140

fixing it in this one ^

Meanwhile, I'll add a version guard on the existing code change.

jjsjann123:
nvfuser patch is in. I think this PR is good for review.

I verified it locally on my machine, but I could be missing stuff in CI. I'll wait until tomorrow when the new container is out to ask bump nvfuser nightly version before asking to merge this one.

t-vi (approved):
Thank you @jjsjann123

jjsjann123:
Awesome. I wonder if we are comfortable with merge this as-is....

Because the added code requires last night's nvfuser package to be tested. So if we bump nvfuser nightly package, we will be able to verify that (even though we are not publishing stable wheel yet)

jjsjann123:
Having said that, this is also a tiny PR. We can certainly revert is something is broken and found later. :wink:

PR #2508, opened 2025-09-09
drop nll_loss shapes forbidden by PT 2.9

t-vi:
- CI fix needed for PT nightly changing
- Thank you @Borda for pointing this out.


t-vi:
Merging this fast to fix CI.

PR #2510, opened 2025-09-09
TEv2 as default TE executor

### LINKED ISSUE #2170: Use TE v2 as default TE executor
## 🚀 Feature

Make the TE v2 executor default and replace v1.

To minimize any potential issue, the following must be closed first:

- [ ] #2027 
    - [ ] Depends on #2094  
- [ ] #2081 

Using it as default will potentially solve the following:

- [ ] #1424 
- [ ] #1050

Will update this comment as PRs get merged and issue closed.

riccardofelluga:
## What does this PR do?

Closes #2170 by renaming the current TE executor to TEv1 and replace it with TEv2

## PR review

I've tried to check for all spots where the TEv1 executor was used but keep an eye out for missing renaming.

## TODO
- [x] Run benchmarks with option transformerengine and trasnformerengine_v1
- [x] Pass all tests

### DIFF
--- a/thunder/__about__.py
+++ b/thunder/__about__.py
@@ -1,7 +1,7 @@
-__version__ = "0.2.5.dev0"
+__version__ = "0.2.6.dev0"
 __author__ = "Lightning-AI et al"
 __author_email__ = "community@lightning.ai"
-__copyright__ = f"2024 {__author__}"
+__copyright__ = f"2024, 2025 {__author__}"
 __homepage__ = "https://github.com/Lightning-AI/lightning-thunder"
 __docs__ = "Lightning Thunder project."
 


riccardofelluga:
@t-vi @lantiga for the recipes[ I've kept the v1 there,](https://github.com/Lightning-AI/lightning-thunder/pull/2510/files#diff-54e2eb80d709850f5d1c5f6f450918794f8178ded2df4349e4f1c0cf5840103f) but maybe it can be updated in a follow-up PR? (need to add also the transform to the recipe)

crcrpar (commented):
Overall looks good to me. I'be a question: would we have a depreciation plan for transformer engine executor of v1?

riccardofelluga:
> Overall looks good to me. I'be a question: would we have a depreciation plan for transformer engine executor of v1?

We don't have a plan yet, but a follow up PR might introduce some warnings when using TEv1 and we could add some quick documentation for the migration(really one page no more as all it changes is that the user needs to add the transform as an argument). 

Then maybe later another decision would be to completely replace it and remove it from the repo. However to this I don't have strong opinions about it. 

cc. for visibility @IvanYashchuk @kshitij12345

t-vi (approved):
Yay @riccardofelluga 

To my mind, we can drop v1 as soon as we switched the recipe, but your use-cases might be different.

PR #2511, opened 2025-09-10
pin transformers for quickstart

t-vi:
This pins transformers for the quickstart.

Transformers 4.56 seems to use torch.compile more aggressively and also change something in the generation in an incompatible way.


t-vi:
Merging to fix quickstart CI

PR #2522, opened 2025-09-23
Refactor quantization.py to use TSP

tejapulagam:
## What does this PR do?
Refactors thunder/transforms/quantization.py to build the computation trace using TraceSubstitutionProcessor instead of manual bound-symbol rewriting. 

Test test_networks.py::test_quantization is still passing.


thunder/transforms/quantization.py
```diff
@@ -218,48 +219,68 @@ def transform_traces_pre_prologue(self, prologue_trace, computation_trace, epilo
         }
 
         new_computation_trace = trace_with_replaced_proxy_metadata(computation_trace, computation_proxy_map)
```
t-vi, line 242: Do we still need this btw or could we absorb it into the QuantizationProcessor?

thunder/transforms/quantization.py
```diff
@@ -218,48 +219,68 @@ def transform_traces_pre_prologue(self, prologue_trace, computation_trace, epilo
...
+                    self.add_processed_bsyms([bsym.from_bsym()])
+                    self.set_result(bsym.output)
+
+        # Process the trace using the QuantizationProcessor
+        processor = QuantizationProcessor(
+            new_computation_trace, quantized_proxies, additional_proxies, self.quant_states, new_compute_inputs
+        )
+
+        # Add new compute inputs to the trace args before processing
         new_computation_trace.args = (*new_computation_trace.args, *new_compute_inputs)
```
t-vi: I wonder if we can do this before the processing and also add the new bindings.

thunder/transforms/quantization.py
```diff
@@ -218,48 +219,68 @@ def transform_traces_pre_prologue(self, prologue_trace, computation_trace, epilo
...
-                    sym=bnb_matmul_nf4,
-                    subsymbols=[],
-                    args=new_args,
-                )
-
-                new_computation_trace.bound_symbols.append(mm_bsym)
-            else:
-                new_computation_trace.bound_symbols.append(bsym.from_bsym())
+        # Insert the new bindings at the beginning of the trace
+        new_computation_trace.bound_symbols = new_bindings + new_computation_trace.bound_symbols
```
t-vi: I would prefer to keep the order of the bindings aligned with the arg order.

thunder/transforms/quantization.py
tejapulagam: I think it might be better to keep them separate because the TraceSubstitutionProcessor is only concerned with symbol-by-symbol replacement, i tried to experiment with it and ran into more errors.

thunder/transforms/quantization.py
tejapulagam: 6c6286f26261be9de92d8472c3c994bcfaffa3cb

thunder/transforms/quantization.py
tejapulagam: 37b595ba2e8b8ab019f814fbedbfb655729ba1f9

not sure if this is what you had in mind

thunder/transforms/quantization.py
t-vi: Fair.

thunder/transforms/quantization.py
t-vi: I thought it could fit into the TSP but let's keep it this way for now. Thank you.

t-vi:
Manually checked GPU tests, merging.

PR #2523, opened 2025-09-23
Initial TE NVFP4 recipe support

riccardofelluga:
This PR changes the input checks to enable NVFP4 using TE executor after PR https://github.com/NVIDIA/TransformerEngine/pull/2177 will be merged. 

This change makes Thunder ready to accept NVFP4 recipe, however a follow-up PR will be needed to add tests when the recipe API will be finalized on TE side(i.e. when their PR will merge) 

cc. @ksivaman shape requirement for the last dimension changed to be half of MXFP8. 

To test with the latest TransformerEngine repo on branch `nvfp4_recipe`(https://github.com/ksivaman/TransformerEngine-1/tree/nvfp4_recipe) run the snippet here:

```python
import torch
import thunder

from thunder.executors.transformer_engineex import transformer_engine_ex, TransformerEngineTransform
from transformer_engine.common import recipe

def nvfp4_vanilla():
    nvfp4_recipe = recipe.NVFP4BlockScaling()
    nvfp4_recipe.fp4_quant_fwd_inp = recipe.QParams()
    nvfp4_recipe.fp4_quant_fwd_weight = recipe.QParams()
    nvfp4_recipe.fp4_quant_bwd_grad = recipe.QParams()
    return nvfp4_recipe

fp8_recipe = nvfp4_vanilla()

def foo(x, w):
    return thunder.torch.linear(x, w)

device = "cuda"
x = torch.randn(32, 32, device=device, requires_grad=True)
w = torch.randn(32, 32, device=device, requires_grad=True)

cfunc = thunder.jit(
    foo,
    executors=[transformer_engine_ex],
    transforms=[TransformerEngineTransform()],
)

with te.fp8_autocast(fp8_recipe=fp8_recipe):
    out = cfunc(x, w)

print(out)

print(thunder.last_traces(cfunc)[-1])
```

### DIFF
--- a/thunder/executors/transformer_engineex_impl.py
+++ b/thunder/executors/transformer_engineex_impl.py
@@ -282,7 +282,10 @@ def _linear_checker(
             return shape[0] % MXFP8_BLOCK_SCALING_SIZE == 0 and shape[1] % MXFP8_BLOCK_SCALING_SIZE == 0
 
         if hasattr(fp8_recipe, "nvfp4") and fp8_recipe.nvfp4():
-            return shape[0] % MXFP8_BLOCK_SCALING_SIZE == 0 and shape[1] % MXFP8_BLOCK_SCALING_SIZE / 2 == 0
+            import math
+            from transformer_engine.pytorch.constants import NVFP4_BLOCK_SCALING_SIZE
+            # https://github.com/ksivaman/TransformerEngine-1/blob/1af7dd88aae5afb45e82148089038e1d1de9675d/transformer_engine/pytorch/tensor/nvfp4_tensor.py#L176-L184 
+            return len(shape) > 2 and shape[-1] % NVFP4_BLOCK_SCALING_SIZE == 0 and math.prod(shape[:-1]) % NVFP4_BLOCK_SCALING_SIZE == 0
 
         return False
 
--- a/thunder/executors/transformer_engineex_impl.py
+++ b/thunder/executors/transformer_engineex_impl.py
@@ -271,6 +271,18 @@ def _linear_checker(
 
     fp8_recipe = FP8GlobalStateManager.get_fp8_recipe()
 
+    import transformer_engine.common.recipe as recipe
+
+    suported_recipes = [recipe.DelayedScaling, recipe.MXFP8BlockScaling]
+    if hasattr(recipe, "NVFP4BlockScaling"):
+        suported_recipes.append(recipe.NVFP4BlockScaling)
+
+    if not isinstance(fp8_recipe, suported_recipes):
+        import warnings
+
+        warnings.warn(f"{type(fp8_recipe)} is not supported by TE executor, TE wont be used.")
+        return False
+
     def check_valid_fp8_shapes(a):
         # DelayedScaling and MXFP8BlockScaling have different shape requirements.
         if fp8_recipe.delayed():
--- a/thunder/executors/transformer_engineex_impl.py
+++ b/thunder/executors/transformer_engineex_impl.py
@@ -277,7 +277,7 @@ def _linear_checker(
     if hasattr(recipe, "NVFP4BlockScaling"):
         suported_recipes.append(recipe.NVFP4BlockScaling)
 
-    if not isinstance(fp8_recipe, suported_recipes):
+    if not isinstance(fp8_recipe, tuple(suported_recipes)):
         import warnings
 
         warnings.warn(f"{type(fp8_recipe)} is not supported by TE executor, TE wont be used.")
--- a/thunder/executors/transformer_engineex_impl.py
+++ b/thunder/executors/transformer_engineex_impl.py
@@ -1,5 +1,6 @@
 import time
 from typing import TYPE_CHECKING
+import warnings
 
 import torch.distributed as torch_dist
 
@@ -27,6 +28,7 @@ if TYPE_CHECKING:
     from thunder.core.proxies import TensorProxy
 
 import transformer_engine.pytorch as te
+import transformer_engine.common.recipe as te_recipe
 from transformer_engine.pytorch.constants import MXFP8_BLOCK_SCALING_SIZE
 from transformer_engine.pytorch.fp8 import (
     _amax_and_scale_update,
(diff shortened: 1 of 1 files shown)


thunder/executors/transformer_engineex_impl.py
```diff
@@ -276,9 +276,15 @@ def check_valid_fp8_shapes(a):
         if fp8_recipe.delayed():
             return check_dim_for_fp8_exec(a)
 
-        assert fp8_recipe.mxfp8()
```
kshitij12345, line 279: Removing this assert also means that we support - 

https://docs.nvidia.com/deeplearning/transformer-engine/user-guide/api/common.html#transformer_engine.common.recipe.Float8CurrentScaling

https://docs.nvidia.com/deeplearning/transformer-engine/user-guide/api/common.html#transformer_engine.common.recipe.Float8BlockScaling

Looking at the PR title and description, I think we should `assert recipe in (mxfp8, nvfp4)` and support the above recipes later.

thunder/executors/transformer_engineex_impl.py
```diff
@@ -276,9 +276,15 @@ def check_valid_fp8_shapes(a):
...
         shape = a.shape
-        return shape[0] % MXFP8_BLOCK_SCALING_SIZE == 0 and shape[1] % MXFP8_BLOCK_SCALING_SIZE == 0
+
+        if fp8_recipe.mxfp8():
+            return shape[0] % MXFP8_BLOCK_SCALING_SIZE == 0 and shape[1] % MXFP8_BLOCK_SCALING_SIZE == 0
+
+        if hasattr(fp8_recipe, "nvfp4") and fp8_recipe.nvfp4():
+            return shape[0] % MXFP8_BLOCK_SCALING_SIZE == 0 and shape[1] % MXFP8_BLOCK_SCALING_SIZE / 2 == 0
```
kshitij12345: It would be great if you can link the relevant shape check code from TE [PR](https://github.com/NVIDIA/TransformerEngine/pull/2177). I couldn't find the relevant bit after a quick glance.

thunder/executors/transformer_engineex_impl.py
riccardofelluga: I've removed the assert in favor of the if statements and a default return False statement for the function. Maybe it'd be cleaner to do the recipe check earlier before looking at the input shape. 

The `assert recipe in (mxfp8, nvfp4)` would stop the program in case the recipe is not supported, would it be better maybe to just have the checker to return False and not use TE linear instead?

kshitij12345: > Maybe it'd be cleaner to do the recipe check earlier before looking at the input shape.

Yes, that sounds better.

> would it be better maybe to just have the checker to return False and not use TE linear instead?

IMO it would be better if the user knows upfront that the given recipe is unsupported than inspecting the trace and trying to figure out why the linear were not replaced.

riccardofelluga: > IMO it would be better if the user knows upfront that the given recipe is unsupported than inspecting the trace and trying to figure out why the linear were not replaced

Hmm yess that is true, tho the same should be valid for the other executors like cudnn, so maybe a warning might be enough?

kshitij12345: As long as user knows that this recipe won't work (via warning or error), it should be fine.

thunder/executors/transformer_engineex_impl.py
```diff
@@ -271,14 +271,40 @@ def _view_input_as_2d(x):
 
     fp8_recipe = FP8GlobalStateManager.get_fp8_recipe()
 
+    import transformer_engine.common.recipe as recipe
+
+    suported_recipes = [recipe.DelayedScaling, recipe.MXFP8BlockScaling]
```
kshitij12345: `suported_recipes` -> `supported_recipes`

Also, `supported_recipes` can already be a tuple, so that we don't need to construct a tuple below.

thunder/executors/transformer_engineex_impl.py
```diff
@@ -271,14 +271,40 @@ def _view_input_as_2d(x):
 
     fp8_recipe = FP8GlobalStateManager.get_fp8_recipe()
 
+    import transformer_engine.common.recipe as recipe
```
kshitij12345: It would be nicer to import at module level.

thunder/executors/transformer_engineex_impl.py
```diff
@@ -271,14 +271,40 @@ def _view_input_as_2d(x):
...
+
+    suported_recipes = [recipe.DelayedScaling, recipe.MXFP8BlockScaling]
+    if hasattr(recipe, "NVFP4BlockScaling"):
+        suported_recipes.append(recipe.NVFP4BlockScaling)
+
+    if not isinstance(fp8_recipe, tuple(suported_recipes)):
+        import warnings
```
kshitij12345: It would be nicer to import at module level.

thunder/executors/transformer_engineex_impl.py
```diff
@@ -271,14 +271,40 @@ def _view_input_as_2d(x):
...
 
-        assert fp8_recipe.mxfp8()
         shape = a.shape
-        return shape[0] % MXFP8_BLOCK_SCALING_SIZE == 0 and shape[1] % MXFP8_BLOCK_SCALING_SIZE == 0
+
+        if fp8_recipe.mxfp8():
+            return shape[0] % MXFP8_BLOCK_SCALING_SIZE == 0 and shape[1] % MXFP8_BLOCK_SCALING_SIZE == 0
+
+        if hasattr(fp8_recipe, "nvfp4") and fp8_recipe.nvfp4():
+            import math
```
kshitij12345: It would be nicer to import at module level.

thunder/executors/transformer_engineex_impl.py
```diff
@@ -271,14 +271,40 @@ def _view_input_as_2d(x):
...
+            from transformer_engine.pytorch.constants import NVFP4_BLOCK_SCALING_SIZE
+
+            # Check inherited from TE https://github.com/ksivaman/TransformerEngine-1/blob/1af7dd88aae5afb45e82148089038e1d1de9675d/transformer_engine/pytorch/tensor/nvfp4_tensor.py#L176-L184
+            return (
+                len(shape) >= 2
+                and shape[-1] % NVFP4_BLOCK_SCALING_SIZE == 0
+                and math.prod(shape[:-1]) % NVFP4_BLOCK_SCALING_SIZE == 0
```
kshitij12345: Since we use `_view_input_as_2d`, shape should be 2d and `shape[0] % NVFP4_BLOCK_SCALING_SIZE == 0` should be ok as well.

thunder/executors/transformer_engineex_impl.py
```diff
@@ -271,14 +271,40 @@ def _view_input_as_2d(x):
```
kshitij12345, line 273: Have tested this patch on RTX6000 Ada with existing `test_transformer_engine_v2_executor.py`. Can you please confirm if this was also tested with existing tests on other hardware which supports MXFP8 recipe?

thunder/executors/transformer_engineex_impl.py
riccardofelluga: removed in favor of the other way to check shapes

thunder/executors/transformer_engineex_impl.py
riccardofelluga: Tested on B200, all passing :)

riccardofelluga:
Hi @t-vi @mruberry @KaelanDt  can I ask for stamp?

PR #427, opened 2024-05-16
Makes cudnn a default executor

vedaanta:
<details>
  <summary><b>Before submitting</b></summary>

- [x] Was this discussed/approved via a Github issue? (no need for typos and docs improvements)
- [x] Did you read the [contributor guideline](https://github.com/Lightning-AI/pytorch-lightning/blob/main/.github/CONTRIBUTING.md), Pull Request section?
- [x] Did you make sure to update the docs?
- [x] Did you write any new necessary tests?

</details>

## What does this PR do?
Cudnn is now a default executor.
The only operation targeted is sdpa.

The main change is a stricter checker function. Both forward and backward graph support are ensured before claiming sdpa operation.
(The checker was previously made lenient in #57)

Fixes #418.

## PR review

Anyone in the community is free to review the PR once the tests have passed.
If we didn't discuss your PR in Github issues there's a high chance it will not be merged.

## Did you have fun?

Make sure you had fun coding 🙃

PR #2525, opened 2025-09-24
fix: working README.md example, support nvfuser for torch==2.8

lianakoleva:
## What does this PR do?

Provides working "Quick start" example for new users. The default fuser for `thunder.compile` is `nvFuser`, so this change creates a working example that [runs on any machine](https://github.com/Lightning-AI/lightning-thunder/pull/427#discussion_r1604777732). Further, `thunder.compile` has `cudnn` in its default executor list, while cuDNN is listed as an optional executor in the "Advanced install options" section. An alternative would be to change the default fuser for `thunder.compile` to `torch.compile` and move `pip install nvidia-cudnn-frontend` to the main `pip install` list, but I assumed functionality is more important to maintain than documentation. Happy to make any change!

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,7 @@ pip install nvfuser-cu128-torch28 nvidia-cudnn-frontend  # if NVIDIA GPU is pres
   <summary>For older versions of <code>torch</code></summary>
 
 <code>torch==2.7</code> + CUDA 12.8
+
 ```bash
 pip install lightning-thunder
 
@@ -95,6 +96,7 @@ pip install nvfuser-cu128-torch27 nvidia-cudnn-frontend  # if NVIDIA GPU is pres
 ```
 
 <code>torch==2.6</code> + CUDA 12.6
+
 ```bash
 pip install lightning-thunder
 
@@ -103,6 +105,7 @@ pip install nvfuser-cu126-torch26 nvidia-cudnn-frontend  # if NVIDIA GPU is pres
 ```
 
 <code>torch==2.5</code> + CUDA 12.4
+
 ```bash
 pip install lightning-thunder
 
--- a/README.md
+++ b/README.md
@@ -79,10 +79,38 @@ Install Thunder via pip ([more options](https://lightning.ai/docs/thunder/latest
 ```bash
 pip install lightning-thunder
 
-pip install torch==2.7.0 torchvision==0.22  # for torch==2.7
-pip install torch==2.6.0 torchvision==0.21  # for torch==2.6
+pip install -U torch torchvision
+pip install nvfuser-cu128-torch28 nvidia-cudnn-frontend  # if NVIDIA GPU is present
+```
+
+<details>
+  <summary>For older versions of <code>torch</code></summary>
+
+<code>torch==2.7</code> + CUDA 12.8
+```bash
+pip install lightning-thunder
+
+pip install torch==2.7.0 torchvision==0.22
 pip install nvfuser-cu128-torch27 nvidia-cudnn-frontend  # if NVIDIA GPU is present
 ```
+
+<code>torch==2.6</code> + CUDA 12.6
+```bash
+pip install lightning-thunder
+
+pip install torch==2.6.0 torchvision==0.21
+pip install nvfuser-cu126-torch26 nvidia-cudnn-frontend  # if NVIDIA GPU is present
+```
+
+<code>torch==2.5</code> + CUDA 12.4
+```bash
+pip install lightning-thunder
+
+pip install torch==2.5.0 torchvision==0.20
+pip install nvfuser-cu124-torch25 nvidia-cudnn-frontend  # if NVIDIA GPU is present
+```
+
+</details>
 
 <details>
   <summary>Advanced install options</summary>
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -77,7 +77,8 @@ For **performance experts**, Thunder is the most ergonomic framework for underst
 Install Thunder via pip ([more options](https://lightning.ai/docs/thunder/latest/fundamentals/installation.html)):
 
 ```bash
-pip install torch==2.6.0 torchvision==0.21 nvfuser-cu124-torch26
+pip install torch==2.7.0 torchvision==0.22 nvfuser-cu128-torch27  # for torch==2.7
+pip install torch==2.6.0 torchvision==0.21 nvfuser-cu124-torch26  # for torch==2.6
```
KaelanDt: I don't think this will work if you don't have a gpu. Tried on my mac for example, got 

```
nvidia-smi command not found. Ensure NVIDIA drivers are installed.
```

If we want it to run on any gpu we may as well also keep thunder.compile and add cudnn in the quickstart, since we also want first users to see good performance, cudnn executors will help for that. 

Also, i'm not sure the two lines cover all gpu machines: what if you have torch 2.5 installed?

KaelanDt: I also think keeping nvfuser as the default is important, as we don't want to rely as much as fusions coming from torch.compile

lantiga:
Hey @lianakoleva thanks for the fix!

I recommend the following:
- we keep `thunder.compile` as the entrypoint
- we handle the case where cuda is not available by avoiding executors that require it
- we have cudnn frontend as a default dependency (installing nvfuser and cudnn frontend on a CPU-only machine doesn't hurt, so we can either have two different install instructions for CPU and GPU or include both by default and that's it)

With these changes: https://github.com/Lightning-AI/lightning-thunder/compare/main...luca/fix-cpu-experience
you can run the quickstart on a CPU machine without warnings. Feel free to pull them in.

cc @KaelanDt

README.md
lianakoleva: I attempted on Ubuntu CPU and Mac, the `nvfuser` package does not install but the others do, so I think this is fine.
```
$ pip install torch==2.7.0 torchvision==0.22 nvfuser-cu128-torch27
Collecting torch==2.7.0
  Downloading torch-2.7.0-cp313-none-macosx_11_0_arm64.whl.metadata (29 kB)
Collecting torchvision==0.22
  Downloading torchvision-0.22.0-cp313-cp313-macosx_11_0_arm64.whl.metadata (6.1 kB)
ERROR: Could not find a version that satisfies the requirement nvfuser-cu128-torch27 (from versions: none)

[notice] A new release of pip is available: 25.0.1 -> 25.2
[notice] To update, run: pip install --upgrade pip
ERROR: No matching distribution found for nvfuser-cu128-torch27
```

Our [required version](https://github.com/Lightning-AI/lightning-thunder/blob/main/requirements/base.txt#L1) of `torch` is already `>=2.7.1`, so I don't think we need to be concerned with `2.5`.

lianakoleva:
@lantiga I made the corresponding README fix, but switched the install order because `pip install lightning-thunder` currently overwrites the torch install from the previous step (`2.7->2.8`), which invalidates the `nvfuser` install. The only compatible version I found for `torch>=2.8.0` is `nvfuser-cu128` with torch nightly: `pip install --pre "nvfuser-cu128[torch]" --extra-index-url https://pypi.nvidia.com`, but I don't think we want to suggest this.

t-vi:
@lianakoleva I'm seeing https://pypi.org/project/nvfuser-cu128-torch28/ which works with torch28 and cu128 (which is the default version for torch28 from pip per https://pytorch.org/get-started/locally/)

t-vi (approved):
Sounds good. Thank you @lianakoleva  Given that this supports blackwell allright, do we want to drop the advanced installation instruction's _Blackwell Support_ section?

thunder/executors/transformer_engineex.py
```diff
@@ -8,18 +8,21 @@
...
 
 transformer_engine_ex: None | StatefulExecutor = None
 TransformerEngineTransform: None | Transform = None
 _te_activation_checkpointing_transform: None | Callable[[TraceCtx], TraceCtx] = None
 
-if package_available("transformer_engine"):
-    import thunder.executors.transformer_engineex_impl as impl
+if torch.cuda.is_available():
+    if package_available("transformer_engine"):
+        import thunder.executors.transformer_engineex_impl as impl
```
riccardofelluga, line 21: @lianakoleva  What happens if torch.cuda.is_available() returns False? Will this error silently?

thunder/executors/triton_crossentropy.py
```diff
@@ -1,16 +1,20 @@
...
 triton_ex: None | OperatorExecutor = None
-if triton_version is not None:
-    try:
-        from thunder.executors.triton_crossentropy_impl import triton_ex as impl_ex
 
-        triton_ex = impl_ex
-    except Exception:
-        import warnings
+if torch.cuda.is_available():
+    if triton_version is not None:
```
riccardofelluga, line 11: Same consideration as earlier here.

riccardofelluga (commented):
What use case made you hit an import issue?

Since I made recent changes to one of these files, could it be possible to be added to the reviewers next time?

PR #2503, opened 2025-09-05
[WIP] DTensor: Add _grouped_mm torch and prim

kshitij12345:
Related #2338 

TODO:
* [ ] Figure out why we see `Error from segmentation group 1: The singleton Communicator isn't available. This is most likely because the instance wasn't successfully initialized due to lack of a multi-process running (e.g. mpirun or torchrun).` only when running this primitive.

PR #2526, opened 2025-09-24
Run test_dtensor with torchrun

kshitij12345:
Context:
When running the `test_dtensor.py` like we do now (`pytest thunder/tests/distributed/test_dtensor.py`), we see `Error from segmentation group 1: The singleton Communicator isn't available.` for #2503 

Changes:

* Move all the tests under `class DTensorTest(DistributedParallelTestCase)` to free functions with `test` prefix and run the test with `torchrun --nproc-per-node 2 --no-python pytest thunder/tests/distributed/test_dtensor.py` (This is required to correctly run the nvFuser tests)

cc @borda

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -4,14 +4,14 @@
 # @global-owner1, @global-owner2, and @global-owner3 will be requested for review when someone opens a pull request.
 
 # Thank you, our previous code owners for their service:
-#   @carmocca
+#   @carmocca @borda
 
-* @mruberry @lantiga @t-vi
+* @mruberry @lantiga @t-vi @KaelanDt
 
 # CI/CD and configs
-/.azure/        @borda @lantiga @t-vi
-/.github/       @borda @lantiga @t-vi
-/.lightning/    @borda @lantiga @t-vi
-/dockers/       @borda @lantiga @t-vi
-Makefile        @borda @lantiga @t-vi
-*.yml           @borda @lantiga @t-vi
+/.azure/        @KaelanDt @lantiga @t-vi
+/.github/       @KaelanDt @lantiga @t-vi
+/.lightning/    @KaelanDt @lantiga @t-vi
+/dockers/       @KaelanDt @lantiga @t-vi
+Makefile        @KaelanDt @lantiga @t-vi
+*.yml           @KaelanDt @lantiga @t-vi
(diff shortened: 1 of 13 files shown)


wujingyue (approved):
LGTM, but can you clarify the context? What makes it hard for nvFuser DTensor tests to work with DistributedParallelTestCase?

kshitij12345:
> but can you clarify the context?

(Have also added this to PR description)
When running the `test_dtensor.py` like we do now (`pytest thunder/tests/distributed/test_dtensor.py`), we see `Error from segmentation group 1: The singleton Communicator isn't available.` for #2503 

> What makes it hard for nvFuser DTensor tests to work with DistributedParallelTestCase?

I was seeing hang-up in some tests when `DistributedParallelTestCase` is used with `torchrun --nproc-per-node 2 --no-python pytest thunder/tests/distributed/test_dtensor.py`. Haven't found the exact cause for the same.

KaelanDt (approved):
Thank you @kshitij12345

crcrpar:
> I was seeing hang-up in some tests when `DistributedParallelTestCase` is used with `torchrun --nproc-per-node 2 --no-python pytest thunder/tests/distributed/test_dtensor.py`. Haven't found the exact cause for the same.

we can just run `DistributedParallelTestCase`-based tests with `pytest` and I think it's the way to run such tests. PyTorch distributed elastic launch command is not needed. (I know you have a problem with `pytest` and dtensor test)

t-vi:
I would prefer if we could adapt the DistributedParallelTestCase to run the test with PyTest...

kshitij12345:
Have updated https://github.com/Lightning-AI/lightning-thunder/pull/2503 to work with `DistributedParallelTestCase`, so this PR is not required anymore. I did notice that tests run faster with this approach, so in future, if required we can try this.

crcrpar:
We already have two ways of writing distributed tests. One is inheriting DistributedParallelTestCase, and the other is using https://github.com/Lightning-AI/lightning-thunder/blob/602fdbdc2e3257e5dc3bd993c162f6a20f9f5d0f/thunder/tests/distributed/helper.py#L235.

I though was wondering a bit if the other way might work better than the first one

t-vi:
Thank you @kshitij12345

PR #2531, opened 2025-09-25
thunderfx: pass bare `torch.compile` to `_splitter`

crcrpar:
## What does this PR do?

Since we use torch.compile w its default values, it doesn't look necessary to register it as an attribute of `ThunderCompiler`.

### DIFF
--- a/thunder/dynamo/compiler.py
+++ b/thunder/dynamo/compiler.py
@@ -128,7 +128,6 @@ class ThunderCompiler:
             "thunderfx_disable_split_autograd", _DEFAULT_THUNDERFX_DISABLE_SPLIT_AUTOGRAD
         )
         self.thunder_options = thunder_options
-        self._torch_compile = torch.compile
 
     def __call__(self, gm: torch.fx.GraphModule, sample_args: list[torch.SymInt, torch.Tensor]):
         from thunder import jit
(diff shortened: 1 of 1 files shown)


t-vi (approved):
Thank you @crcrpar 
Anyone from the ThunderFX side who want's to take a look before merging?

shino16 (approved):

shino16:
Thank you for pointing this out, I've been wondering why this exists too. This seems to be left off in #1535

shino16:
Why don't we pass this `torch_compile_kwargs` to `torch.compile` for submodules? I think it makes more sense to propagate pytorch options for submodule compilation too.
https://github.com/Lightning-AI/lightning-thunder/blob/5ece7ef52e7e4e04dea6cd61f8dfd3a3150c7fc9/thunder/dynamo/compiler.py#L326

crcrpar:
> Why don't we pass this `torch_compile_kwargs` to `torch.compile` for submodules? I think it makes more sense to propagate pytorch options for submodule compilation too.
> 
> https://github.com/Lightning-AI/lightning-thunder/blob/5ece7ef52e7e4e04dea6cd61f8dfd3a3150c7fc9/thunder/dynamo/compiler.py#L326

I've #2196 and #2198. The PR is a zombie now. Feel free to take it over if you want.

shino16:
Thank you for the reference. If this was an intended design choice, I won't argue against it.

shino16:
This will conflict with my PR, https://github.com/Lightning-AI/lightning-thunder/pull/2600, which directly passes `torch._inductor.compile` (instead of `torch.compile`) directly to `_splitter`.

crcrpar:
close this in favor of 2600

PR #2532, opened 2025-09-25
Add `float4_e2m1fn_x2` to lcdtype_to_nvdtype_map

crcrpar:
## What does this PR do?

As per title.


crcrpar:
@jjsjann123 from which version the packed fp4 dtype is supported? 0.2.27 would be sufficing?

jjsjann123:
> @jjsjann123 from which version the packed fp4 dtype is supported? 0.2.27 would be sufficing?

Nope. 0.2.27 doesn't have that.  (i.e.  `e798e06` when moving files around (with 0.2.27) doesn't have the type)
Looks like 0.2.28 is safe though.  (i.e. `ed57b64` has the type when bumping version to 0.2.28)

t-vi:
GPU tests passed manually. Merging.

PR #2533, opened 2025-09-25
avoid `torch.float4_e2m1fn_x2` in `_get_min_and_val`

crcrpar:
## What does this PR do?

As per title. We might want to upcast packed fp4 tensors to 16/32-bit to get min and max but for now just avoiding errors from thunderfx `_splitter`


t-vi:
CI says we need to make the addition conditional on such a dtype existing in PyTorch (would be my favourite) or bump the PT requirement.
https://github.com/Lightning-AI/lightning-thunder/actions/runs/18001272157/job/51210920505?pr=2533

t-vi:
Manual checks passed, merging.

PR #2534, opened 2025-09-25
Propagate `disable_torch_autograd` to thunderfx's `_splitter`

### LINKED ISSUE #2606: Avoid vjp as much as possible in ThunderFX splitter. Because for now the focus is on custom operators for inference most of them would not have backward def

crcrpar:
## What does this PR do?

Fixes somewhat unexpected graph splits in thunderfx path with programs that use `torch.library.custom_op` even when it's manually registered through `_register_custom_op`.
For example, when there's a `torch.library.custom_op` that's registered to Thunder without backward definition, that custom_op could result in a graph split. This is because splitter only sees input proxies `requires_grad` -- https://github.com/Lightning-AI/lightning-thunder/blob/280c57ec289d09eacad0a29cf97a332e02ecebaa/thunder/dynamo/utils.py#L312-L322.

close #2606

### DIFF
--- a/.lightning/workflows/notebooks.yaml
+++ b/.lightning/workflows/notebooks.yaml
@@ -43,5 +43,5 @@ run: |
   # iterate over all listed notebooks and execute them with jupyter
   while read -r line; do
       echo "Processing $line"
-      jupyter execute $line --timeout=300
+      jupyter execute $line --timeout=450
   done <<< $(cat ci.txt)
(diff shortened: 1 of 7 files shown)


shino16 (approved):

thunder/dynamo/utils.py
```diff
@@ -309,7 +312,12 @@ def _run_with_cache_info():
                 exception=str(e),
             )
 
-        function_to_run = value_and_grad(thunder_symbol) if requires_grad else thunder_symbol
+        function_to_run = thunder_symbol
```
shino16: Is this line redundant?

crcrpar: ah yes, great catch.

Issue #2501, opened 2025-09-04
ThunderFX OOM error on Gemma 2 27B, FSDP


shino16:
ThunderFX gives an cuda OOM error in the following benchmark.

**Reproduction**
```sh
torchrun \
    --nnodes 1 \
    --nproc-per-node 8 \
    thunder/benchmarks/benchmark_litgpt.py \
        --max_iters 20 \
        --warmup_iters 5 \
        --model_name Gemma-2-27b \
        --distributed_mode fsdp \
        --shard_mode zero3 \
        --compile dynamo_thunder \
        --checkpoint_activations True \
        --low_precision_mode none \
        --micro_batch_size 1 \
        --bucketing_mode block
```

This OOM error does not happen with eager nor Inductor.

NVBugs 5043188

Issue #2527, opened 2025-09-24
Activation checkpoint not working inside Inductor-compiled submodules


shino16:
## 🐛 Bug

When ThunderFX delegates activation checkpointing operations to Inductor-compiled submodules, apparently the checkpointing mechanism fails to execute properly, causing excessive memory usage due to retained intermediate activations. This may apply to all models that containing some Thunder-unsupported ops in activation checkpointed regions, causing e.g. #2501.

Note. When Thunder encounters an unsupported op in a checkpointed region, the entire region will go to Inductor. Huge memory usage in status quo could be a good sign of Thunder's poor coverage.

#### Code sample

```py
import torch
from torch.utils.checkpoint import checkpoint
from thunder.dynamo import thunderfx

def fn(x):
    return x.sinc().sinc().sinc()

def checkpoint_fn(x):
    return checkpoint(fn, x, use_reentrant=False)

for compiler in [thunderfx, torch.compile]:
    torch.cuda.reset_peak_memory_stats()
    assert torch.cuda.memory_allocated() == 0

    x = torch.randn((1024, 1024, 1024), device="cuda", requires_grad=True)  # 4 GB
    y = compiler(checkpoint_fn)(x)
    del x, y

    peak_mem_usage = torch.cuda.max_memory_allocated()
    print(f"{compiler.__name__}: {peak_mem_usage / 1024 / 1024 / 1024} GB")
```
```
thunderfx: 16.0 GB
compile: 8.0 GB
```
Input for `thunder.dynamo.ThunderCompiler`:
```py
class GraphModule(torch.nn.Module):
    def forward(self, L_x_: "f32[1024, 1024, 1024]"):
        l_x_ = L_x_
        
         # File: /opt/pytorch/lightning-thunder/tmp/repro.py:9 in checkpoint_fn, code: return checkpoint(fn, x, use_reentrant=False)
        wrap_body_0 = self.wrap_body_0
        tag_activation_checkpoint = torch.ops.higher_order.tag_activation_checkpoint(wrap_body_0, l_x_, use_reentrant = False);  wrap_body_0 = l_x_ = None
        getitem: "f32[1024, 1024, 1024]" = tag_activation_checkpoint[0];  tag_activation_checkpoint = None
        return (getitem,)
        
    class wrap_body_0(torch.nn.Module):
        def forward(self, l_x_: "f32[1024, 1024, 1024]"):
[...]

PR #2538, opened 2025-09-25
Convert activation checkpointing tag with eager checkpointing function

shino16:
This is a workaround for #2527. `torch.ops.higher_order.tag_activation_checkpoint` does not perform activation checkpointing when run in eager mode, so we convert it back to `torch.utils.checkpoint.checkpoint`.

### DIFF
--- a/thunder/benchmarks/layers_for_inference_benchmark.py
+++ b/thunder/benchmarks/layers_for_inference_benchmark.py
@@ -300,7 +300,15 @@ class NVFP4InferenceLinear(nn.Module):
         )
 
     @staticmethod
-    def from_linear(linear: nn.Linear) -> NVFP4InferenceLinear:
+    def from_linear(linear: nn.Linear, fqn: str | None = None) -> NVFP4InferenceLinear:
+        """
+        Creates an NVFP4InferenceLinear layer from a standard nn.Linear layer.
+
+        Args:
+            linear (nn.Linear): The source linear layer.
+            fqn (str | None, optional): Fully qualified name of the layer. Currently unused,
+                but retained for compatibility with interfaces that require it or for future use.
+        """
         weight = linear.weight
         bias = linear.bias
         out_features, in_features = weight.size()
@@ -420,6 +428,7 @@ class NVFP4InferenceGroupedLinear(nn.Module):
         ab_strides: torch.Tensor,
         c_strides: torch.Tensor,
     ) -> None:
+        super().__init__()
         self.register_buffer("fp4_weight", fp4_weight)
         self.register_buffer("weight_scaling_factor", weight_scaling_factor)
         self.register_buffer("weight_global_scale", weight_global_scale)
@@ -451,7 +460,14 @@ class NVFP4InferenceGroupedLinear(nn.Module):
         )
 
     @staticmethod
-    def from_grouped_linear(grouped_linear: GroupedLinear) -> NVFP4InferenceGroupedLinear:
+    def from_grouped_linear(grouped_linear: GroupedLinear, fqn: str | None = None) -> NVFP4InferenceGroupedLinear:
+        """
+        Create an NVFP4InferenceGroupedLinear from a GroupedLinear.
+
+        Args:
+            grouped_linear (GroupedLinear): The source GroupedLinear.
+            fqn (str or None): Fully qualified name. Currently unused; reserved for future use or compatibility.
+        """
         weight = grouped_linear.weight
         (
             fp4_weight,
--- a/thunder/core/transforms.py
+++ b/thunder/core/transforms.py
@@ -1767,6 +1767,19 @@ def digamma_backward(a: Proxy, g):
     return g * polygamma(1, a)
 
 
+def _silu_grad(a: Proxy, inplace: bool = False):
+    from thunder.torch import silu
+
+    fwd = silu(a, inplace)
+    g = get_grad(fwd)
+    sigmoid = 1 / (1 + clang.exp(-a))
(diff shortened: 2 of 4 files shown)


shino16:
This fixes #2501.

on `main` (280c57ec289d09eacad0a29cf97a332e02ecebaa)
```
[rank2]: torch.OutOfMemoryError: CUDA out of memory. Tried to allocate 8.00 GiB. GPU 2 has a total capacity of 139.72 GiB of which 3.71 GiB is free. Including non-PyTorch memory, this process has 0 bytes memory in use. Of the allocated memory 127.71 GiB is allocated by PyTorch, and 6.76 GiB is reserved by PyTorch but unallocated. If reserved but unallocated memory is large try setting PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True to avoid fragmentation.  See documentation for Memory Management  (https://pytorch.org/docs/stable/notes/cuda.html#environment-variables)
```
^ happening on each rank

in this PR (ac5508a99df68cbffb9a0b73d9f5d4069e97eb99)
```
Model name: Gemma-2-27b
Seq Length: 8192
Micro BS: 1
Global BS: 8
Number of Layers: 46
Number of parameters: 3.55B
Distributed Mode: fsdp
Sharding Mode: zero3
Bucketing: block
Compiler: dynamo_thunder
Low Precision Mode: none
Average iter time: 9991.12 ms
Memory used: 81.76 GB
Tokens/s: 6555.90
Tokens/s/GPU: 819.49
TFLOP/s: 1192.38
```

thunder/dynamo/compiler.py
```diff
@@ -135,9 +135,8 @@ def __call__(self, gm: torch.fx.GraphModule, sample_args: list[torch.SymInt, tor
 
         remove_empty_autocast(gm)
 
-        # Dynamo uses lazy generation of the underlying Python code, so we need to
-        # force recompilation of the GraphModule before passing it to Thunder.
-        recompile_graph(gm)
```
shino16, line 140: Recompiling here was added in commit 0338afef1bab7552a7e1a730579f8e74ccfb90ff when we did not have the graph splitting logic. Now we break the graph down in the subsequent code, so no need for recompile.

thunder/dynamo/splitter.py
```diff
@@ -186,8 +185,6 @@ def is_thunder_supported_partition(node: torch.fx.Node) -> bool:
                 partial(_get_example_inputs_from_placeholder, only_metadata=True), placeholders
             )
             example_input_metadatas.append(list(example_input_metadata))
-            # Replace PyTorch operators within the checkpointed function with the corresponding Thunder operators
-            checkpoint_converter(split_gm, graph_module)
```
shino16, line 199: `torch.utils.checkpoint.checkpoint` is Thunder-tracible.

thunder/dynamo/compiler.py
```diff
@@ -135,9 +135,9 @@ def __call__(self, gm: torch.fx.GraphModule, sample_args: list[torch.SymInt, tor
...
+        # Convert tag_activation_checkpoint operators, which is meaningless in eager mode, to actual checkpoint calls
+        # This will not be needed when we have found a way to make tag_activation_checkpoint fall back to PyTorch's backend
+        convert_checkpoint_tags(gm)
```
kshitij12345, line 141: It would be helpful to refer https://github.com/Lightning-AI/lightning-thunder/issues/2539 in the comment.

thunder/tests/test_dynamo.py
```diff
@@ -844,6 +844,37 @@ def find_target_module(model, target_module_name):
...
+
+    def fn(x):
+        return op(op(op(op(x))))
+
+    def checkpoint_fn(x):
+        return checkpoint.checkpoint(fn, x, use_reentrant=False)
+
+    initial_mem = torch.cuda.memory_allocated()
+
+    x = torch.randn((1024 // 4, 1024, 1024), device="cuda", requires_grad=True)
```
kshitij12345: It would be better to use smaller input as these tests run in parallel.

thunder/dynamo/utils.py
```diff
@@ -630,57 +631,15 @@ def _get_example_inputs_from_placeholder(
...
 
-def _checkpoint_function_converter(gm: torch.fx.GraphModule):
+def convert_checkpoint_tags(gm: torch.fx.GraphModule):
     """
-    Replace PyTorch operators in ``gm`` representing a checkpointed function with corresponding Thunder operators. The input ``gm`` is modified inplace.
+    Replaces tag_activation_checkpoint operators with torch.utils.checkpoint.checkpoint calls.
 
-    Args:
-        gm (torch.fx.GraphModule): The GraphModule of the checkpointed function, which is modified inplace.
+    tag_activation_checkpoint only marks nodes for torch.compile stack but does not execute actual checkpointing in eager mode.
```
kshitij12345, line 648: It would be nice to mention that this function mutates the `gm`.

kshitij12345 (approved):

thunder/tests/test_dynamo.py
shino16: Ah that's a good point, thank you!

kiya00 (approved):
Thank you @shino16 for the fix, there are 2 tests `test_checkpoint_converter`, `test_checkpoint_converter_submodule` that test the old `converter`, but I think we can keep them, as they also seem to validate the functionality of `convert_checkpoint_tags`.

shino16:
Let me block this PR for now, since #2600 seems like a better way forward.

shino16:
Resolved by #2600.

PR #2542, opened 2025-09-26
Use signature (*args, **kwargs) when signature is unavailable

shino16:
We use `inspect.signature(fn)` to get the function's signature, particularly the names of its parameters. This raises ValueError when the signature is unavailable, so we guard this and give it a default signature `(*args, **kwargs)`.

## What this fixes
Although Thunder has been able to trace functions containing `getattr(a, "mT")` or `a.mT`, tracing `getattr` itself was impossible.
```py
import torch, thunder

a = torch.randn((2, 3))
print(thunder.trace()(getattr, a, "mT"))
```
`main` 70c0421126b92450c3c3945b7213325f478be2c1
```
  File "/usr/lib/python3.12/inspect.py", line 2224, in _signature_fromstr
    raise ValueError("{!r} builtin has invalid signature".format(obj))
ValueError: <built-in function getattr> builtin has invalid signature
```
This PR (56f4b7cb2901c7714799f6308b32faa6b338ff8c)
```py
import thunder
import torch

@torch.no_grad()
@no_autocast
def getattr(*args):
  # args: "Collection"
  t0, _, = args
  # t0: "cpu f32[2, 3]"
  # 'mT'
  t1 = prims.transpose(t0, (1, 0))  # t1: "cpu f32[3, 2]"
  return t1
```

## Why this matters
`value_and_grad(getattr)(a, "mT")` raises the same ValueError, which makes ThunderFX think that it doesn't support `getattr`, causing graph break every time it sees `a.mT`. This PR removes such graph breaks.
https://github.com/Lightning-AI/lightning-thunder/blob/70c0421126b92450c3c3945b7213325f478be2c1/thunder/dynamo/utils.py#L312-L322

### Note
1. Trace body and runtime behavior are unaffected.
2. Default signature does not affect what can actually be passed to the function. For example,
```py
thunder.trace()(getattr, a, "mT", None, None)  # TypeError: getattr expected at most 3 arguments, got 4
thunder.trace()(getattr, a, attr="mT")  # TypeError: getattr() takes no keyword arguments
```

### DIFF
--- a/thunder/tests/test_core.py
+++ b/thunder/tests/test_core.py
@@ -1620,6 +1620,19 @@ def test_nested_trace_no_name_collision(executor, device, _):
     thunder.trace()(bar, a, b)
 
 
+# Tests if Thunder can trace a function without a valid signature
+@instantiate(dtypes=NOTHING)
+def test_no_signature(executor, device, _):
+    a = make_tensor((2, 3), device=device, dtype=torch.float32)
+
+    fn_trace = thunder.trace()(getattr, a, "mT")
+
+    jfn = executor.make_callable(fn_trace)
+    actual = jfn(a, "mT")
+    expected = getattr(a, "mT")
+    assert_close(actual, expected)
+
+
 @instantiate(dtypes=NOTHING)
 def test_trace_args_no_name_collision(executor, device, _):
     from thunder.core.trace import detached_trace
--- a/thunder/core/codeutils.py
+++ b/thunder/core/codeutils.py
@@ -436,7 +436,10 @@ def get_siginfo(fn: Callable, args, kwargs, *, _make_named_inputs: bool = False)
         fn_ = fn_.meta
 
     # Binds args and kwargs to signature
-    sig = inspect.signature(fn_)
+    try:
+        sig = inspect.signature(fn_)
+    except ValueError:
+        sig = inspect.signature(lambda *args, **kwargs: None)
     kwargs.update(partial_kwargs)
     ba = sig.bind(*args, **kwargs)
 


shino16:
Repro for ThunderFX's graph break
```py
import torch

@thunderfx
def fn(x):
    return x.mT

fn(torch.randn(2, 3, requires_grad=True))
fn._backend.subgraph_infos[-1].split_graph_module.print_readable()
```
`main` (70c0421126b92450c3c3945b7213325f478be2c1)
```py
    def forward(self, l_x_: "f32[2, 3]"):
        inductor_0 = self.inductor_0(l_x_);  l_x_ = inductor_0 = None
        return ()
        
    class inductor_0(torch.nn.Module):
        def forward(self, l_x_: "f32[2, 3]"):
             # File: /opt/pytorch/lightning-thunder/tmp/main.py:6 in fn, code: return x.mT
            getattr_1: "f32[3, 2]" = l_x_.mT;  l_x_ = getattr_1 = None
            return ()
            
        class _orig_mod(torch.nn.Module):
            def forward(self, l_x_: "f32[2, 3]"):
                 # File: /opt/pytorch/lightning-thunder/tmp/main.py:6 in fn, code: return x.mT
                getattr_1: "f32[3, 2]" = l_x_.mT;  l_x_ = getattr_1 = None
                return ()
```
This PR (56f4b7cb2901c7714799f6308b32faa6b338ff8c)
```py
    def forward(self, l_x_: "f32[2, 3]"):
        thunder_0 = self.thunder_0(l_x_);  l_x_ = thunder_0 = None
        return ()
        
    class thunder_0(torch.nn.Module):
        def forward(self, l_x_: "f32[2, 3]"):
             # File: /opt/pytorch/lightning-thunder/tmp/main.py:6 in fn, code: return x.mT
            getattr_1: "f32[3, 2]" = l_x_.mT;  l_x_ = getattr_1 = None
            return ()
            
        class _model(torch.nn.Module):
            def forward(self, l_x_: "f32[2, 3]"):
                 # File: /opt/pytorch/lightning-thunder/tmp/main.py:6 in fn, code: return x.mT
                getattr_1: "f32[3, 2]" = l_x_.mT;  l_x_ = getattr_1 = None
                return ()
```

thunder/tests/test_core.py
```diff
@@ -1620,6 +1620,19 @@ def bar(__a, __b):
     thunder.trace()(bar, a, b)
 
 
+# Tests if Thunder can trace a function without a valid signature
+@instantiate(dtypes=NOTHING)
```
crcrpar, line 1624: Q: with this setting, how many instances are created? or how many instances would you think are needed?

crcrpar (approved):

thunder/tests/test_core.py
shino16: It creates 3 test instances: `torch_cpu`, `torch_cuda`, `nvFuser`. I admit I don't think we need to test this (and many others in `test_core.py`) for multiple executors. So my personal answer is just one, without cuda device.

shino16: To clarify, I think the necessity level to test this three times is the same as its adjacent tests.

t-vi (approved):
Thank you @shino16 @crcrpar

thunder/tests/test_core.py
crcrpar: then maybe we should audit this test file, to make tests faster, in a separate pr/issue

PR #2545, opened 2025-09-26
fix: transformer_engine install in ci

lianakoleva:
[WIP] DO NOT MERGE


t-vi:
@lianakoleva thanks for looking into  it, @kshitij12345 's PR #2536 works, so we'll go with that.

lianakoleva:
Great, thanks!

PR #2552, opened 2025-09-30
[DTensor] Add prim and torch sym for neg and reciprocal

kshitij12345:
Related #2338


t-vi:
We can do it, but I can't help feel like it should be general for pointwise unary and binary ops.

t-vi (approved):
Thank you @kshitij12345

Issue #841, opened 2024-07-23
Prologue error: check_slice_value fails with function wrappers and closures


IvanYashchuk:
## 🐛 Bug

```py
import torch
import thunder

from thunder.tests.test_grad import _make_differentiable_wrapper
from thunder.core.utils import flatten_func

x = torch.randn((5, 5))

s1 = slice(1, 3, 1)
s2 = slice(2, 4, 2)

func = thunder.torch.getitem
args = (x, (s1, s2))

flat_op, flat_args, spec = flatten_func(thunder.torch.getitem, args, {})

filtered_op, filtered_args = _make_differentiable_wrapper(flat_op, flat_args)
jf = thunder.jit(filtered_op, disable_torch_autograd=True)
jf(*filtered_args)
```
```py
RuntimeError                              Traceback (most recent call last)
Cell In[1], line 20
     18 filtered_op, filtered_args = _make_differentiable_wrapper(flat_op, flat_args)
     19 jf = thunder.jit(filtered_op, disable_torch_autograd=True)
---> 20 jf(*filtered_args)

File ~/dev/lightning-thunder/thunder/__init__.py:686, in jit.<locals>.fn_(*args, **kwargs)
    683 cs.last_trace_host_start = time.perf_counter_ns()
    684 cs.calls += 1
--> 686 cache_entry, inps, pro_to_epi = get_computation_and_inputs(*args, **kwargs)
    687 cs.last_trace_host_execution_start = time.perf_counter_ns()
    689 result = cache_entry.computation_fn(*inps)

File ~/dev/lightning-thunder/thunder/__init__.py:225, in _with_cache_info_ctx.<locals>.cache_info_wrapper(*args, **kwargs)
    223 tok = _cache_info_ctx.set({})
    224 try:
--> 225     res = fn(*args, **kwargs)
    226 finally:
    227     _cache_info_ctx.reset(tok)

File ~/dev/lightning-thunder/thunder/__init__.py:563, in jit.<locals>.get_computation_and_inputs(*args, **kwargs)
    561 cs.last_prologue_execution_start = time.perf_counter_ns()
    562 if interpretation is INTERPRETATION_OPTIONS.TRANSLATE_PYTHON:
--> 563     inps, pro_to_epi = pro(*args, **kwargs)
    564 else:
    565     inps = pro(*args, **kwargs)

[...]

PR #2578, opened 2025-10-01
tests: clean up xfails in VJP tests

aobolensk:
Clean up for test_vjp_correctness xfail marks:
- Remove xfail for 'expand', 'expand_as' ops, since they are already implemented
- Remove xfail for 'getitem', since referenced issue (https://github.com/Lightning-AI/lightning-thunder/issues/841) is already addressed and the test passes
- Limit xfail for 'addcdiv' operation on CUDA only, since the issue is present only there

### DIFF
--- a/thunder/tests/opinfos.py
+++ b/thunder/tests/opinfos.py
@@ -2969,6 +2969,7 @@ addcdiv_opinfo = OpInfo(
         DecorateInfo(
             pytest.mark.xfail,
             "test_vjp_correctness",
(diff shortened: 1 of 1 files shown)


t-vi:
Hi @aobolensk , thank you for your PR! Great to see you here. I'll try to prod the CI and then it's good to go.
If you want to reach out via mail or so, do give me a shout, I'd be keen to here more about your interest in Thunder.

PR #2580, opened 2025-10-01
fix output dtype for nvfuserex cumsum

jjsjann123:
### Context:

nvfuser can't do cumsum math in reduced precision. The implicit type promotion happening inside needs to be corrected to avoid promoting output dtype as a side effect.


jjsjann123:
cc'ing @naoyam

PR #2551, opened 2025-09-29
Wrap fallback submodules for Inductor in a new `nn.Module`

shino16:
Fixes #2539, which in turn fixes #2527 and #2501. This makes PR #2538 obsolete.

As in https://github.com/Lightning-AI/lightning-thunder/issues/2527#issuecomment-3334005768, when we pass a GraphModule to `torch.compile` it doesn't get lowered properly. As a workaround, this PR wraps the fallback GraphModule in a new `nn.Module`. This workaround was found in https://github.com/Lightning-AI/lightning-thunder/issues/2527#issuecomment-3345877210 (credit @mattteochen).

Alternatively we could call `torch._inductor.compile_fx.compile_fx` directly to compile the GraphModule, but as it returns a bare `forward` function instead of a `nn.Module` instance, it has troubles with registering as a submodule to the outer GraphModule.

PR #2429, opened 2025-08-14
thunderfx: handle output node with no example_value

kshitij12345:
It is possible to have nodes with no `example_value` which could end up being an output node after splitter. This is happens for "ibm-granite/granite-3.1-3b-a800m-instruct"

NOTE: There is a graph-break at `tolist` if the tensor is floating type.

Example of graph where node may have no example value
```python
import torch

def test_fn(x):
    y = (x + 10)
    z = y.tolist()[0]
    return z + 2

x = torch.tensor([1, 2, 3, 4, 5])

def backend(gm, inps):
    gm.print_readable()
    # def forward(self, L_x_: "i64[5]"):
    #     l_x_ = L_x_
        
    #      # File: /opt/pytorch/lightning-thunder/test.py:4 in test_fn, code: y = (x + 10)
    #     y: "i64[5]" = l_x_ + 10;  l_x_ = None
        
    #      # File: /opt/pytorch/lightning-thunder/test.py:5 in test_fn, code: z = y.tolist()[0]
    #     getitem = y[0];  y = None
    #     item: "Sym(u0)" = getitem.item();  getitem = None
        
    #      # File: /opt/pytorch/lightning-thunder/test.py:6 in test_fn, code: return z + 2
    #     add_1: "Sym(u0 + 2)" = item + 2;  item = None
    #     return (add_1,)

    for node in gm.graph.nodes:
        if "example_value" not in node.meta:
            print(node)  # prints `getitem` and `output`
    return gm

torch.compile(test_fn, backend=backend)(x)

from thunder.dynamo import thunderfx

model = thunderfx(test_fn)

#   File "/opt/pytorch/lightning-thunder/thunder/dynamo/splitter.py", line 175, in _splitter
#     if n.meta["example_value"].grad_fn is None:
#        ~~~~~~^^^^^^^^^^^^^^^^^
# torch._dynamo.exc.BackendCompilerFailed: backend='<thunder.dynamo.compiler.ThunderCompiler object at 0x70b657217800>' raised:
# KeyError: 'example_value'
model(x)
```

Issue #1326, opened 2024-10-17
Re-enable inductor tests on windows


tfogal:
We had to disable a couple tests on windows because they now fail on windows:

```
FAILED thunder/tests/test_torch_compile_executor.py::test_torch_compile_litgpt - torch._dynamo.exc.BackendCompilerFailed: backend='inductor' raised:
RuntimeError: Compiler: cl is not found.

FAILED thunder/tests/test_autocast.py::test_torch_compile_autocast - 
torch._dynamo.exc.BackendCompilerFailed: backend='inductor' raised:
RuntimeError: Compiler: cl is not found.
```

## 🐛 Bug

Re-enable the tests.

### To Reproduce

Just drop the `or platform.system() == Windows` in the existing `skipIf` clause on these two tests.

### Additional context

Example CI run where this failed: https://github.com/Lightning-AI/lightning-thunder/actions/runs/11390861866/job/31693384602?pr=1324

t-vi:
I think the main thing is to get the environment set up "vcvarsall.bat", unfortunately, it seems a bit of a hassle.

PR #2600, opened 2025-10-03
Use `torch._inductor.compile` for ThunderFX fallback entrypoint

### LINKED ISSUE #2539: ThunderFX's fallback is not using Inductor compilation
## 🐛 Bug

ThunderFX fallback submodules are NOT being compiled by Inductor and apparently executed in eager mode instead.

This is causing missed opportunities of fusion (such as @kshitij12345 found in https://github.com/Lightning-AI/lightning-thunder/issues/2527#issuecomment-3334618852) and memory spikes (#2501).

The impact will be huge when combined with activation checkpointing. In the current implementation, if Thunder cannot support all ops in a checkpointed region, the entire region falls back to `torch.compile`. When checkpointing is applied in the granularity of blocks, a huge part of a model will not be compiled by neither Thunder nor Inductor.

cc @kshitij12345 @IvanYashchuk

shino16:
Fixes #2539 and other issues it caused.

## What this fixes

`torch.compile` skips the lowering of `GraphModule` that Thunder's splitter emits. This PR passes the `GraphModule` to `torch._inductor.compile` instead, which then calls `torch._inductor.compile_fx.compile_fx`, to make sure Inductor lowers it.

Although it is found in https://github.com/Lightning-AI/lightning-thunder/issues/2527#issuecomment-3345877210 that wrapping the `GraphModule` in a new `Module` is sufficient, it was found in https://github.com/Lightning-AI/lightning-thunder/pull/2551#issuecomment-3347809603 that Dynamo cannot trace the enter/exit of `torch.autocast()` region.

#### (A) Graph Splitting
Inductor assumes that the output node of the `GraphModule` has the form `return (t0, ..., tN)`. Usually `torch.fx.passes.split_module.split_module` (the underlying algorithm of the graph splitter) makes sure this holds for the submodules it creates, but when the splitter cuts out a node that returns a tuple, the output value will look like `return tuple_value`. This breaks Inductor.

<details><summary>Example</summary>

```py
import torch, thunder

def fn(x): return x.sinc()

@thunder.dynamo.thunderfx
def checkpointed_fn(x):
    return torch.utils.checkpoint.checkpoint(fn, x)

checkpointed_fn(torch.randn(2, 3))
```
```py
    def forward(self, l_x_: "f32[2, 3]"):
        submod_0 = self.submod_0(l_x_);  l_x_ = None
        submod_1 = self.submod_1(submod_0);  submod_0 = None
        return (submod_1,)
        
        def forward(self, l_x_: "f32[2, 3]"):
             # File: /opt/pytorch/lightning-thunder/tmp/main.py:8 in checkpointed_fn, code: return torch.utils.checkpoint.checkpoint(fn, x)
            wrap_body_0 = self.wrap_body_0
            tag_activation_checkpoint = torch.ops.higher_order.tag_activation_checkpoint(wrap_body_0, l_x_);  wrap_body_0 = l_x_ = None
            return tag_activation_checkpoint
            
        class wrap_body_0(torch.nn.Module):
            def forward(self, l_x_: "f32[2, 3]"):
                 # File: /opt/pytorch/lightning-thunder/tmp/main.py:5 in fn, code: def fn(x): return x.sinc()
                sinc: "f32[2, 3]" = l_x_.sinc();  l_x_ = None
                return (sinc,)
                
        def forward(self, tag_activation_checkpoint):
             # File: /opt/pytorch/lightning-thunder/tmp/main.py:8 in checkpointed_fn, code: return torch.utils.checkpoint.checkpoint(fn, x)
            getitem: "f32[2, 3]" = tag_activation_checkpoint[0];  tag_activation_checkpoint = None
            return getitem
```
```
  File "/usr/local/lib/python3.12/dist-packages/torch/_functorch/_aot_autograd/utils.py", line 555, in call_and_expect_output_descs
    assert out_spec == out_desc_spec, (
           ^^^^^^^^^^^^^^^^^^^^^^^^^
torch._dynamo.exc.BackendCompilerFailed: backend='<thunder.dynamo.compiler.ThunderCompiler object at 0x7c1a8fe84410>' raised:
AssertionError: ([<function aot_stage1_graph_capture.<locals>.orig_flat_fn2 at 0x7c1a5434ede0>, <function create_functional_call.<locals>.functional_call at 0x7c1a5423ab60>, GraphModule(
  (wrap_body_0): GraphModule()
)], [(FunctionalTensor(_to_functional_tensor(FakeTensor(..., size=(2, 3)))),)], [PlainAOTOutput(idx=0)], TreeSpec(immutable_list, None, [TreeSpec(tuple, None, [*])]), TreeSpec(immutable_list, None, [*]))
```
This is a complaint for `return tag_activation_checkpoint`, which `compile_fx.make_graph_return_tuple` converts to `return [tag_activation_checkpoint]`.

</details>

We work around this by grouping the subsequent `getitem` node together into the node that was cut out. This hack is applied to nodes whose `example_value` is a tuple, which includes `torch.ops.higher_order.tag_activation_checkpoint` and `torch.ops.higher_order.autograd_function_apply`.

#### (B) Graph break inside Inductor's fallback submodule

Inductor is not responsible for breaking the graph, so it raises an exception when it encounters something it cannot compile. 

<details><summary>Example</summary>

```py
import torch, thunder.dynamo

@thunder.dynamo.thunderfx
def fn(x, idx, val):
    x = x.clone()
    x[idx] = val
    return x

x = torch.randn(3, requires_grad=True)
idx = x > 0.5
val = torch.randn(torch.count_nonzero(idx), requires_grad=True)

fn(x, idx, val)
```
```py
    def forward(self, l_x_: "f32[3]", l_idx_: "b8[3]", l_val_: "f32[2]"):
        submod_0 = self.submod_0(l_x_);  l_x_ = None
        submod_1 = self.submod_1(submod_0, l_idx_, l_val_);  l_idx_ = l_val_ = submod_1 = None
        return (submod_0,)
        
        def forward(self, l_x_: "f32[3]"):
             # File: /opt/pytorch/lightning-thunder/tmp/main.py:7 in fn, code: x = x.clone()
            x: "f32[3]" = l_x_.clone();  l_x_ = None
            return x
            
        def forward(self, x: "f32[3]", l_idx_: "b8[3]", l_val_: "f32[2]"):
             # File: /opt/pytorch/lightning-thunder/tmp/main.py:8 in fn, code: x[idx] = val
            x[l_idx_] = l_val_;  setitem = x;  x = l_idx_ = l_val_ = setitem = None
            return ()         
```
```
torch._subclasses.fake_tensor.DynamicOutputShapeException: aten.nonzero.default
```

</details>

We must catch the exception and let Dynamo handle it by falling further back to `torch.compile`. I except there are a lot more exceptions that `torch._inductor.compiler` may raise. There are no common supertype for such exceptions, so we will have to cover each of them when it pops up.

#### (C) Lack of `example_value`

Inductor requires instances of `FakeTensor`s representing the inputs. This is sometimes unavailable: see #2429. We will simply fall back to `torch.compile` in such cases.

## Concern

This PR does not cover cases where a submodule containing `torch.autocast()` region falls back to `torch.compile` (when (B) or (C) applies). `AssertionError` is raised in such cases (see the new test which is xfailed).

Should we catch such errors and resort to eager mode? I wonder how big the impact will be.

### DIFF
--- a/thunder/dynamo/splitter.py
+++ b/thunder/dynamo/splitter.py
@@ -238,6 +238,9 @@ def _splitter(
                     # This exception is meant to be handled by Dynamo, which is responsible for graph break
                     jit_fn = fallback_torch_compile(f"Dynamic output shape operator encountered: {e}.")
 
+            # This is for ease of debugging. We add graph attribute so GraphModule.print_readable will print it
+            jit_fn.graph = graph_module.graph
+
             # Update the node name from "submod_*" to "inductor_*" for more user-friendly names
             update_node_and_submodule(split_gm, node, node.name.replace("submod", "inductor"), jit_fn)
             submodule_to_compiled_fns[getattr(original_split_gm, node_name)] = CompiledFunction(
--- a/thunder/dynamo/splitter.py
+++ b/thunder/dynamo/splitter.py
@@ -1,4 +1,5 @@
 from __future__ import annotations
+import operator
 from typing import TYPE_CHECKING
 import copy
 from functools import partial
@@ -96,11 +97,12 @@ def _splitter(
     partition_cnt = 0
     supported_partitions: set[int] = set()
     split_reasons: list[SplitReason] = []
+    unsupported_collection_users: set[torch.fx.Node] = set()
 
     nodes_in_unsupported_ctx_regions = get_nodes_in_unsupported_ctx_regions(gm)
 
     def callback(node) -> int:
-        nonlocal prev_value, partition_cnt, split_reasons, supported_partitions
+        nonlocal prev_value, partition_cnt, split_reasons, supported_partitions, unsupported_collection_users
 
         assert node.op not in (
             "placeholder",
@@ -118,10 +120,18 @@ def _splitter(
                 info=f"node with name: {node.name} and target: {node.target} is not supported probably because it is in unsupported context.",
             )
             split_reasons.append(split_reason)
+        elif node in unsupported_collection_users:
+            is_thunder_supported = False
+            split_reason = None
         else:
             is_thunder_supported, split_reason = is_node_supported_by_thunder(node)
             if split_reason is not None:
                 split_reasons.append(split_reason)
+
+        if not is_thunder_supported and baseutils.is_collection(node.meta.get("example_value", None)):
+            for user in node.users:
+                assert user.target is operator.getitem
+                unsupported_collection_users.add(user)
 
         if prev_value == is_thunder_supported:  # We are in the same region.
             return partition_cnt
--- a/thunder/dynamo/splitter.py
+++ b/thunder/dynamo/splitter.py
@@ -3,10 +3,13 @@ import operator
 from typing import TYPE_CHECKING
 import copy
 from functools import partial
+import warnings
 
 import torch
+from torch._subclasses.fake_tensor import DynamicOutputShapeException
 from torch.fx.passes.split_module import split_module
 
+from thunder.core import baseutils
 from thunder.dynamo.utils import (
     SubgraphInfo,
     CompiledFunction,
@@ -18,6 +21,7 @@ from thunder.dynamo.utils import (
     update_node_and_submodule,
     recompile_graph,
     checkpoint_converter,
+    make_fake_arguments,
     _get_example_inputs_from_placeholder,
     _ThunderSplitGraphModule,
 )
@@ -208,7 +212,32 @@ def _splitter(
             )
         elif node.name.startswith("submod"):  # For inductor
             graph_module = getattr(split_gm, node.name)
-            jit_fn = torch_inductor(graph_module)
+
+            class ModuleWrapper(torch.nn.Module):
+                def __init__(self, fn):
+                    super().__init__()
+                    self.fn = fn
+
+                def forward(self, *args, **kwargs):
+                    return self.fn(*args, **kwargs)
+
+            def fallback_torch_compile(reason: str) -> torch.nn.Module:
+                warnings.warn(f"{reason} Falling back to torch.compile.")
+                # torch.compile does not lower GraphModule properly. See https://github.com/Lightning-AI/lightning-thunder/issues/2539
+                # We work around this by wrapping it in a Module
+                return torch.compile(ModuleWrapper(graph_module))
+
+            fake_args = make_fake_arguments(graph_module)
+            if fake_args is None:
+                jit_fn = fallback_torch_compile("Example values for arguments are not available.")
+            else:
+                try:
+                    # torch._inductor.compile returns a function, but update_node_and_submodule expects a Module
+                    jit_fn = ModuleWrapper(torch_inductor(graph_module, fake_args))
+                except DynamicOutputShapeException as e:
+                    # This exception is meant to be handled by Dynamo, which is responsible for graph break
+                    jit_fn = fallback_torch_compile(f"Dynamic output shape operator encountered: {e}.")
+
             # Update the node name from "submod_*" to "inductor_*" for more user-friendly names
             update_node_and_submodule(split_gm, node, node.name.replace("submod", "inductor"), jit_fn)
             submodule_to_compiled_fns[getattr(original_split_gm, node_name)] = CompiledFunction(
--- a/thunder/tests/test_dynamo.py
+++ b/thunder/tests/test_dynamo.py
@@ -1633,7 +1633,8 @@ def test_spliter_bwd():
     val = torch.randn(nz, dtype=torch.bfloat16, requires_grad=True)
 
     cfn = thunderfx(fn)
-    cfn(x, idx, val)
+    with pytest.warns(match="Dynamic output shape operator"):
+        cfn(x, idx, val)
     reason = cfn._backend.subgraph_infos[0].split_reasons
     assert len(reason) == 1
     assert "Failed while running meta for node with name: setitem" in reason[0].info
--- a/thunder/tests/test_dynamo.py
+++ b/thunder/tests/test_dynamo.py
@@ -155,6 +155,21 @@ def test_basic_splitter(executor, device: str, dtype: dtypes.dtype, dynamic: boo
     assert any(target.startswith("thunder_") for target in targets)  # Verify that the submodules have name `thunder_*`
 
 
+@instantiate(dtypes=NOTHING)
+def test_fallback_to_inductor(executor, device, dtype):
+    x = torch.randn(3, 3, device=device, dtype=dtype)
+
+    def func(x):
+        return x.sinc().cos().sinc().sinc()
+
+    cfunc = thunderfx(func)
+    with patch("torch._inductor.compile", side_effect=torch._inductor.compile) as mock_inductor:
+        cfunc(x)
+
+    # Once for sinc() and once for sinc().sinc()
+    assert mock_inductor.call_count == 2
+
+
 @instantiate(
     dtypes=NOTHING,
     executors=[DynamoThunderExecutor],
@@ -844,6 +859,35 @@ def test_checkpoint_converter_submodule():
             assert isinstance(n.target, Symbol) or callable(n.target)
 
 
+@requiresCUDA
+@pytest.mark.parametrize("op", [torch.sin, torch.sinc])
+def test_checkpoint_memory_use(op):
+    import torch.utils.checkpoint as checkpoint
+
+    def fn(x):
+        return op(op(op(op(x))))
+
+    def checkpoint_fn(x):
+        return checkpoint.checkpoint(fn, x, use_reentrant=False)
+
+    initial_mem = torch.cuda.memory_allocated()
+
+    x = torch.randn((128, 128), device="cuda", requires_grad=True)
+    jfn = thunderfx(checkpoint_fn)
+    y = jfn(x)
+
+    peak_mem_usage = torch.cuda.max_memory_allocated() - initial_mem
+
+    y_ref = fn(x)
+    torch.testing.assert_close(y, y_ref)
+
+    assert peak_mem_usage == x.nbytes * 2
+    if op == torch.sinc:
+        # Make sure the checkpointed region falled back to PyTorch
+        sinfo = jfn._backend.subgraph_infos[-1]
+        assert any(n.name.startswith("inductor") for n in sinfo.split_graph_module.graph.nodes)
+
+
 @instantiate(
     dtypes=NOTHING,
     executors=[DynamoThunderExecutor],
@@ -935,6 +979,8 @@ def test_dynamo_reproducer_submodules(use_pytest_benchmark, tmp_path):
 
 
 def test_deepcopy_graph_module():
+    import thunder
+
     class MyModule(torch.nn.Module):
         def __init__(self):
             super().__init__()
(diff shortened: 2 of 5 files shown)


shino16:
Note on (C) Lack of `example_value`: Thanks to https://github.com/pytorch/pytorch/pull/163807, `example_value` no longer becomes unavailable in our reproducer #2429, unless we set `torch._dynamo.config.capture_scalar_outputs` to `True`. I am not aware of other possibilities that makes `example_value` unavailable, but I will keep the mechanisms and the tests for (C) for safety.

shino16:
Tests are failing on Windows with:
```
compiler = 'cl'

    @functools.cache
    def check_compiler_exist_windows(compiler: str) -> None:
        """
        Check if compiler is ready, in case end user not activate MSVC environment.
        """
        try:
            subprocess.check_output([compiler, "/help"], stderr=subprocess.STDOUT)
        except FileNotFoundError as exc:
>           raise RuntimeError(f"Compiler: {compiler} is not found.") from exc
E           torch._inductor.exc.InductorError: RuntimeError: Compiler: cl is not found.
E           
E           Set TORCHDYNAMO_VERBOSE=1 for the internal stack trace (please do this especially if you're reporting a bug to PyTorch). For even more developer context, set TORCH_LOGS="+dynamo"

C:\hostedtoolcache\windows\Python\3.10.11\x64\lib\site-packages\torch\_inductor\cpp_builder.py:140: InductorError
```
I'm trying to prepare a Windows environment for investigation.

kiya00:
Hi @shino16 , IIRC, all the torch.compile related tests are skipped on Windows by the `IS_WINDOWS` flag, https://github.com/Lightning-AI/lightning-thunder/issues/1326 I think the situation remains unchanged.
I'm curious whether this PR could affect performance?

mattteochen:
Thank you, @shino16. It seems that this should solve the torch compile `GraphModule` fallback issues that we were seeing. In theory, this shouldn't introduce perf regression, as before we were running submodules in eager mode, but some benchmarks could be useful.

shino16:
Thank you for your comments! As @mattteochen says, we should gain performance benefit from Inductor for fallback submodules, but latency will increase for the same reason.

kshitij12345:
>  but latency will increase for the same reason.

IIUC, this would happen only for the first run, right?

mattteochen:
> > but latency will increase for the same reason.
> 
> IIUC, this would happen only for the first run, right?

Yes, torch  inductor has its own cache (very random snapshot):
proxy_gm

thunder/dynamo/splitter.py
```diff
@@ -118,11 +124,19 @@ def callback(node) -> int:
...
             split_reasons.append(split_reason)
+        elif node in unsupported_collection_users:
+            is_thunder_supported = False
+            split_reason = None
         else:
             is_thunder_supported, split_reason = is_node_supported_by_thunder(node)
             if split_reason is not None:
                 split_reasons.append(split_reason)
 
+        if not is_thunder_supported and baseutils.is_collection(node.meta.get("example_value", None)):
```
kshitij12345, line 145: It would be useful to add a comment here similar to the explanation in the description.

thunder/dynamo/splitter.py
```diff
@@ -118,11 +124,19 @@ def callback(node) -> int:
                 info=f"node with name: {node.name} and target: {node.target} is not supported probably because it is in unsupported context.",
             )
             split_reasons.append(split_reason)
+        elif node in unsupported_collection_users:
+            is_thunder_supported = False
+            split_reason = None
```
kshitij12345, line 134: Do we skip specifying the split_reason here because it has already been captured?

thunder/dynamo/splitter.py
```diff
@@ -198,7 +212,35 @@ def is_thunder_supported_partition(node: torch.fx.Node) -> bool:
...
+
+            class ModuleWrapper(torch.nn.Module):
+                def __init__(self, fn):
+                    super().__init__()
+                    self.fn = fn
+
+                def forward(self, *args, **kwargs):
+                    return self.fn(*args, **kwargs)
+
+            def fallback_torch_compile(reason: str) -> torch.nn.Module:
```
kshitij12345: > Should we catch such errors and resort to eager mode? I wonder how big the impact will be.

As we are not clear about what the behavior of calling `torch.compile` here will be, I think falling back to eager would be ok. We can understand this behavior better and then fallback to `torch.compile` in a follow-up PR. Wdyt?

shino16: Sounds very reasonable, thanks!

thunder/dynamo/splitter.py
shino16: Yes, my idea was that the`split_reason` should be specified when `node` was added to `unsupported_collection_users`. I'll add a comment to explain this.

shino16:
Here's some benchmarks on `benchmark_peft.py`, run on NVIDIA H100 80GB HBM3. Options are taken from our weekly peft benchmark pipelines.

HEAD~2 refers to 545ac48953892c0d4137b55c943b920b68373c17, where Inductor submodule falls back to `torch.compile` instead of eager mode when Inductor fails.

Average iteration time after 10 warmup (ms)
|  | HEAD | HEAD~2 | main | Has Inductor submodule | main / HEAD |
| --- | --- | --- | --- | --- | --- |
| Qwen/Qwen-7B-Chat | 11.15 | 11.15 | 15.05 | Yes | 1.350 |
| ibm-research/gpt2-medium-multiexit | 10.09 | 10.25 | 9.37 | Yes | 0.929 |
| facebook/incoder-6B | 8.66 | 9.09 | 8.86 | Yes | 1.023 |
| MBZUAI/MobiLlama-05B | 10.17 | 10.18 | 10.21 | Yes | 1.004 |
| smallcloudai/Refact-1_6B-fim | 7.41 | 7.30 | 7.44 | No | 1.004 |

Peak Memory Allocation (GB)

|  | HEAD | main | Has Inductor submodule | main / HEAD |
| --- | --- | --- | --- | --- |
| Qwen/Qwen-7B-Chat | 9.37 | 26.11 | Yes | 0.359 |
| ibm-research/gpt2-medium-multiexit | 0.82 | 0.82 | Yes | 1.000 |
| facebook/incoder-6B | 11.14 | 11.14 | Yes | 1.000 |
| MBZUAI/MobiLlama-05B | 2.15 | 2.15 | Yes | 1.000 |
| smallcloudai/Refact-1_6B-fim | 9.82 | 9.82 | No | 1.000 |

mattteochen:
Do we expect patch issues like https://github.com/Lightning-AI/lightning-thunder/issues/2501 ?

shino16:
Yes, this fixes #2501.

HEAD (927b58a70c20ed087fda03f3983b304d6965d8a2)
```
Loading model with {'name': 'Gemma-2-27b', 'hf_config': {'org': 'google', 'name': 'gemma-2-27b'}, 'scale_embeddings': True, 'attention_scores_scalar': 144, 'block_size': 8192, 'sliding_window_size': 4096, 'sliding_window_layer_placing': 2, 'vocab_size': 256000, 'padding_multiple': 512, 'padded_vocab_size': 256000, 'n_layer': 46, 'n_head': 32, 'head_size': 128, 'n_embd': 4608, 'rotary_percentage': 1.0, 'parallel_residual': False, 'bias': False, 'lm_head_bias': False, 'n_query_groups': 16, 'shared_attention_norm': False, 'norm_class_name': 'RMSNorm', 'post_attention_norm': True, 'post_mlp_norm': True, 'norm_eps': 1e-05, 'mlp_class_name': 'GemmaMLP', 'gelu_approximate': 'tanh', 'intermediate_size': 36864, 'rope_condense_ratio': 1, 'rope_base': 10000, 'rope_adjustments': None, 'n_expert': 0, 'n_expert_per_token': 0, 'attention_logit_softcapping': 50.0, 'final_logit_softcapping': 30.0, 'rope_n_elem': 128}
(...)
Average iter time: 11371.18 ms
Memory used: 62.15 GB
Tokens/s: 5760.30
Tokens/s/GPU: 720.04
TFLOP/s: 1047.68
```

main (ab067fdfbe11fa83b779d7b09ad8c8c1ccf24dc0)
```
[rank6]:   File "/usr/local/lib/python3.12/dist-packages/torch/nn/functional.py", line 2135, in softmax
[rank6]:     ret = input.softmax(dim, dtype=dtype)
[rank6]:           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank6]: torch.OutOfMemoryError: CUDA out of memory. Tried to allocate 8.00 GiB. GPU 6 has a total capacity of 79.14 GiB of which 3.57 GiB is free. Including non-PyTorch memory, this process has 0 bytes memory in use. Of the allocated memory 68.51 GiB is allocated by PyTorch, and 5.95 GiB is reserved by PyTorch but unallocated. If reserved but unallocated memory is large try setting PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True to avoid fragmentation.  See documentation for Memory Management  (https://pytorch.org/docs/stable/notes/cuda.html#environment-variables)
```

thunder/tests/test_dynamo.py
```diff
@@ -935,6 +991,8 @@ def forward(self, x):
 
 
 def test_deepcopy_graph_module():
+    import thunder
```
kshitij12345: Should be safe to import `thunder` at module level.

thunder/tests/test_dynamo.py
```diff
@@ -962,7 +1019,13 @@ def forward(self, x):
 @instantiate(
     dtypes=NOTHING,
     executors=[DynamoThunderExecutor],
-    decorators=(pytest.mark.parametrize("use_pytest_benchmark", (True, False), ids=("benchmark", "repro")),),
+    decorators=(
+        pytest.mark.parametrize("use_pytest_benchmark", (True, False), ids=("benchmark", "repro")),
+        pytest.mark.skipif(
+            condition=IS_WINDOWS,
```
kshitij12345, line 1018: Curious to know if this previously ran successfully on Windows CI?

thunder/dynamo/splitter.py
```diff
@@ -205,7 +223,34 @@ def is_thunder_supported_partition(node: torch.fx.Node) -> bool:
...
+            class ModuleWrapper(torch.nn.Module):
+                def __init__(self, fn):
+                    super().__init__()
+                    self.fn = fn
+
+                def forward(self, *args, **kwargs):
+                    return self.fn(*args, **kwargs)
+
+            def fallback_eager(reason: str) -> torch.nn.Module:
+                warnings.warn(f"{reason} Falling back to torch.compile.")
```
kshitij12345:
```suggestion
                warnings.warn(f"{reason} Falling back to eager.")
```

thunder/tests/test_dynamo.py
shino16: It was skipped in the function body only when `IS_WINDOWS` and `use_pytest_benchmark` are True. Now fallback submodules go to Inductor, so we now need to always skip on Windows, otherwise it fails:
https://github.com/Lightning-AI/lightning-thunder/actions/runs/18272047785/job/52016121038?pr=2600#step:11:5112

And I have just realized `pytest.skip` in the function body is now redundant...

shino16:
Hi @t-vi, this is ready for your review. Thanks!

shino16:
Added the table of peak memory usage to the previous benchmark results. https://github.com/Lightning-AI/lightning-thunder/pull/2600#issuecomment-3386346123

ThunderFX used to consume 493.15% of what eager mode consumes on Qwen/Qwen-7B-Chat, and 333.73% on MBZUAI/MobiLlama-05B. The former will decrease by 64%, while the latter will stay the same.

PR #2607, opened 2025-10-07
fix benchmarks job

t-vi:
Another bump of nvfuser, but let's re-use a requirements.txt we already have.


KaelanDt:
Seems like another error, i will investigate @t-vi

PR #2598, opened 2025-10-03
[DTensor] Add test with parallelize_module

kshitij12345:
Related: #2338

PR #2611, opened 2025-10-07
Enable MoE TP with thunderfx

kshitij12345:
Depends on #2598 
Related: #2338 

Changes
* Allow nvFuser executor to handle AsyncCollectiveTensor as it may show up when using programs with DTensor.
* Update nvFuser executor to create separate fusion region (and FusionDefinition) for multi-device and single-device.
* Update dtensor_mul to correctly handle broadcasting and type promotion.
* Add `dtensor_silu` dtensor symbol composed of existing dtensor prims.
* Enable thunderfx path for the distributed test_moe.py.

Generated trace from thunderfx - https://gist.github.com/kshitij12345/c02ecef47a9974216d8350bb0e0db4ca

### DIFF
--- a/.lightning/workflows/notebooks.yaml
+++ b/.lightning/workflows/notebooks.yaml
@@ -43,5 +43,5 @@ run: |
   # iterate over all listed notebooks and execute them with jupyter
   while read -r line; do
       echo "Processing $line"
-      jupyter execute $line --timeout=300
+      jupyter execute $line --timeout=450
   done <<< $(cat ci.txt)
--- a/notebooks/dev_tutorials/fsdp_tutorial.ipynb
+++ b/notebooks/dev_tutorials/fsdp_tutorial.ipynb
@@ -352,6 +352,7 @@
     "import torch.distributed\n",
     "import thunder\n",
     "import thunder.distributed\n",
+    "from thunder.distributed.transforms.fsdp_v2 import FSDPTransform\n",
     "import os\n",
     "\n",
     "# # # # # # # #\n",
@@ -370,6 +371,7 @@
     "\n",
     "# Model\n",
     "model = create_model()\n",
+    "\n",
     "# Input\n",
     "x = torch.randn(dim, dim, device=device)\n",
     "\n",
@@ -383,6 +385,11 @@
     "device = f\"cuda:{rank}\"\n",
     "\n",
     "# # # # # # # #\n",
+    "# Move model to correct device\n",
+    "# # # # # # # #\n",
+    "model.to(device)\n",
+    "\n",
+    "\n",
     "# Move inputs to correct device\n",
     "# # # # # # # #\n",
     "x = x.to(device)\n",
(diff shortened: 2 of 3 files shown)


thunder/torch/experimental/dtensor_torch_and_prims.py
```diff
@@ -507,6 +473,45 @@ def dtensor_add(a: TensorLike, b: TensorLike) -> TensorLike:
...
+
+    return fwd
+
+
+register_grad(dtensor_mul_prim, _dtensor_mul_prim_grad)
+
+
+@dtensor_torchsymbol(torch.mul, id="dtensor.torch.mul")
+def dtensor_mul(a: TensorLike, b: TensorLike) -> TensorLike:
+    return _elementwise_binary_wrapper(
```
kshitij12345, line 507: This is the main change for `dtensor_mul` instead of directly calling `dtensor_mul_prim`

thunder/tests/distributed/test_moe.py
```diff
@@ -170,3 +171,13 @@ def test_llama4_moe_distributed(self):
         expected = model(inp)
 
         torch.testing.assert_close(actual, expected, atol=1e-5, rtol=1e-5)
+
+        tmodel = thunderfx(model, nv_enable_linear=True, nv_enable_scatter=True)
```
wujingyue:
```suggestion
        tmodel = thunderfx(parallelized_model, nv_enable_linear=True, nv_enable_scatter=True)
```
Do you mean this instead? I assume you want to thunderfx the distributed model.

wujingyue: Hmm... when I `print(tmodel.last_traces)`, I do see parameters are sharded as expected. What magic am I missing? cc @crcrpar

kshitij12345: Nice catch, thank you!

`parallelize_module` updates the passed model inplace. So, `model` also points to the distributed module. I will add `org_model = copy.deepcopy(model)` to get reference result.

```python

import torch.nn as nn
import torch
from torch.distributed._tensor import DeviceMesh, Shard, distribute_tensor
from torch.distributed.tensor.placement_types import Placement, Shard, Replicate
import os
import thunder
import torch.distributed as dist
from torch.distributed.tensor.parallel import parallelize_module, ColwiseParallel

num_devices = 2
mesh = DeviceMesh("cuda", list(range(num_devices)))

torch.cuda.set_device(f"cuda:{LOCAL_RANK}")

linear = torch.nn.Linear(10, 10)
parallel_linear = parallelize_module(linear, mesh, ColwiseParallel(use_local_output=False))

print(type(linear.weight))  # DTensor
print(type(parallel_linear.weight))  # DTensor

torch.distributed.destroy_process_group()
exit()
```

kshitij12345:
@t-vi can you check if the distributed failure is related to the PR? - https://github.com/Lightning-AI/lightning-thunder/pull/2611/checks?check_run_id=52429299655

Thanks!

PR #2615, opened 2025-10-08
Refactor low precision option handling in `benchmark_litgpt.py`

riccardofelluga:
TODO: refactor adding te as compile option at line 382-382

### DIFF
--- a/thunder/benchmarks/benchmark_litgpt.py
+++ b/thunder/benchmarks/benchmark_litgpt.py
@@ -69,6 +69,93 @@ device = torch.device("cuda", local_rank)
 torch.cuda.set_device(device)
 
 
+@dataclass
+class LowPrecisionHandler:
+    mode: str
+    fp8_recipe: recipe.Recipe | None = None
+
+    use_thunder_te: bool = True
+    use_legacy_thunder_te: bool = False
+
+    @property
+    def use_te_autocast(self) -> bool:
+        return self.enabled and not self.use_legacy_thunder_te
+
+    def __post_init__(self) -> None:
+        self.enabled = self.mode != "none" and self.mode in ["fp8-default-te", "fp8-default-te-wo_layernorm"]
+        if not self.enabled:
+            return
+
+        if not transformer_engine_available:
+            raise ImportError(
+                "Selected benchmark config is for TransformerEngine but could not import the TransformerEngine library!"
+            )
+
+        check_fp8_compute_capability()
+
+        self.fp8_recipe = fp8.get_default_fp8_recipe()
+
+    def check_and_update_config_for_te_if_needed(self, config: Config) -> None:
+        if not self.enabled:
+            return
+
+        updates_info = []
+
+        def update_if_not_divisible(attr_name, divisor):
+            current_value = getattr(config, attr_name, 0)
+            if current_value % divisor != 0:
+                new_value = current_value + (divisor - current_value % divisor)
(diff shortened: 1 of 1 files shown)


thunder/benchmarks/benchmark_litgpt.py
```diff
@@ -69,75 +71,115 @@
...
-            f"Device compute capability {compute_capability} is insufficient. "
-            f"Compute capability {required_compute_capability} or higher is required for FP8 execution. "
-            "Please ensure you are using a compatible GPU and the correct driver version."
-        )
+    use_thunder_te: bool = True
+    use_legacy_thunder_te: bool = False
+
+    @property
+    def use_te_autocast(self) -> bool:
+        return self.enabled and not self.use_legacy_thunder_te
```
shino16, line 86: Beginner question: why do we exclude legacy TE here? Did we previously assume that `Benchmark_litGPT.use_te_fp8_autocast` is False when `transformerengine_v1` is specified?

thunder/benchmarks/benchmark_litgpt.py
```diff
@@ -69,75 +71,115 @@
...
+            raise ValueError(f"Unsupported low precision mode: {self.mode}. Supported: {SUPPORTED_LP_MODES}")
+
+        if not transformer_engine_available:
+            raise ImportError(
+                "Selected benchmark config is for TransformerEngine but could not import the TransformerEngine library!"
+            )
+
+        check_fp8_compute_capability()
 
+        self.fp8_recipe = fp8.get_default_fp8_recipe()
```
shino16: Maybe this should be run only when `self.fp8_recipe is None`? Is this meant for making the fp8 recipe customizable?

thunder/benchmarks/benchmark_litgpt.py
```diff
@@ -747,8 +783,8 @@ def train(self):
             input_ids, targets = next(self.train_data_iter)
             input_ids = input_ids.to(self.device)
             targets = targets.to(self.device)
-            if self.use_te_fp8_autocast:
-                with te.fp8_autocast():
+            if self.low_precision_handler.use_te_autocast:
+                with te.fp8_autocast(fp8_recipe=self.low_precision_handler.fp8_recipe):
```
shino16: FWIW, I thought the entire if-else could look like `with self.low_precision_handler.maybe_fp8_autocast()`.

thunder/benchmarks/benchmark_litgpt.py
riccardofelluga: transformerengine_v1 used to add it's own autocast to the thunder trace here 

https://github.com/Lightning-AI/lightning-thunder/blob/ab067fdfbe11fa83b779d7b09ad8c8c1ccf24dc0/thunder/core/trace.py#L418-L419

so no need to add another context manager with TEv1 executor

thunder/benchmarks/benchmark_litgpt.py
riccardofelluga: Great catch! Thanks I've missed that!

thunder/benchmarks/benchmark_litgpt.py
riccardofelluga: You're right, that way it's more elegant

thunder/benchmarks/benchmark_litgpt.py
shino16: Ah, thank you!

PR #2616, opened 2025-10-08
Propagated requires_grad to torch tensor

mattteochen:
## What does this PR do?

Propagates the `requires_grad` attribute during tensor conversion from numpy to torch.


thunder/tests/test_grad.py
```diff
@@ -155,9 +155,17 @@ def _from_numpy(x, like):
     """
     assert isinstance(like, torch.Tensor), f"_from_numpy: Unsupported type of the second argument {type(like)}"
     if isinstance(x, np.ndarray):
-        return torch.from_numpy(x).to(device=like.device)
+        t = torch.from_numpy(x).to(device=like.device, dtype=like.dtype)
+        # Preserve the requires_grad property to avoid cache mismatches
```
shino16, line 159: Maybe we should update the docstring too?

Issue #2549, opened 2025-09-29
CI slowness


t-vi:
We recently switched from 3090s through Azure to L4s on Lightning for the CI jobs that need GPUs but are CPU bound.
However it seems that L4 CPU power is really lacking.

Here are timings in comparison (no fancy stats, just the latest run on main).

Options
- Move grad to consistency tests with PyTorch rather than fdm based tests, this would be better also because we currently test different code paths (old grad vs. autodiff transform). https://github.com/Lightning-AI/lightning-thunder/issues/1874
- Parallelize slowest jobs to multiple L4,
- analyse if L40s would be a good alternative economically (needs to be OK in terms of price - so the speedup should justify the higher cost per minute - and availability). 

I could use a hand in doing these.

Azure:
```
lightning-thunder [dockers] (build_push cuda 12.6 | torch nightly | cudnn FE v1.10.0) Failing after 34m - build_push cuda 12.6 | torch nightly | cudnn FE v1.10.0 35 min

lightning-thunder (GPUs) Successful in 19m - Build #20250929.3 succeeded
[Details](https://github.com/Lightning-AI/lightning-thunder/runs/51460101923)

lightning-thunder (GPUs) (testing distributed w/ torch 2.8.0) Successful in 19m - testing distributed w/ torch 2.8.0 succeeded
[Details](https://github.com/Lightning-AI/lightning-thunder/runs/51460096109)

lightning-thunder (GPUs) (testing distributed w/ torch-nightly) Successful in 17m - testing distributed w/ torch-nightly succeeded
[Details](https://github.com/Lightning-AI/lightning-thunder/runs/51460008472)

lightning-thunder (GPUs) (testing grads w/ torch 2.8.0) Successful in 14m - testing grads w/ torch 2.8.0 succeeded
[Details](https://github.com/Lightning-AI/lightning-thunder/runs/51459872379)

lightning-thunder (GPUs) (testing grads w/ torch-nightly) Successful in 14m - testing grads w/ torch-nightly succeeded
[Details](https://github.com/Lightning-AI/lightning-thunder/runs/51459887005)

lightning-thunder (GPUs) (testing main w/ torch 2.8.0) Successful in 9m - testing main w/ torch 2.8.0 succeeded
[Details](https://github.com/Lightning-AI/lightning-thunder/runs/51459672271)

lightning-thunder (GPUs) (testing main w/ torch-nightly) Successful in 10m - testing main w/ torch-nightly succeeded
[...]

PR #2618, opened 2025-10-08
Update jvp computation in `test_grad.py`

### LINKED ISSUE #1874: Consider changing vjp tests
Copied from https://github.com/Lightning-AI/lightning-thunder/pull/1842#issuecomment-2718078365

It looks like the "consistency" vjp tests are much faster than the "correctness" vjp tests. With the new softplus tests we even hit a timeout on the "correctness" tests. Should we consider defaulting to "consistency," or maybe changing how the "correctness" tests work? Are the "consistency" tests really faster than the "correctness" tests, or just benefiting from caching?

fyi @beverlylytle @IvanYashchuk

mattteochen:
## What does this PR do?

Closes #1874, #2549.

Switching to torch jvp implementation gives a 3x [speedup](https://github.com/Lightning-AI/lightning-thunder/issues/1874#issuecomment-3370413738) on vjp correctness tests, which is essential to run CI faster. There are some future works to be investigated on top of this PR:
- complete types are currently differentiable, but they are not in Pytorch, currently the numerical jvp implementation just skips their computation: https://github.com/Lightning-AI/lightning-thunder/blob/cda32a46e19a214632191dc48371b72f5176afea/thunder/tests/test_grad.py#L202
- numerical mismatch in some tests as we accumulate errors in double precision when thunder executors are used (torch jvp runs in eager). Currently found with the nvfuser executor (see [this](https://github.com/Lightning-AI/lightning-thunder/issues/1874#issuecomment-3370413738) appendix).

I will create these issues after this work is merged.

### DIFF
--- a/thunder/tests/test_grad.py
+++ b/thunder/tests/test_grad.py
@@ -360,7 +360,7 @@ def check_vjp_torch(
       especially with float64, leading to small discrepancies leading to assertion failures.
     - `torch.func.jvp` has shown to have some issues on a few ops in float64 (see
       `_TORCH_JVP_UNSTABLE_OPS_FLOAT64`), producing a zero J*u.
-    
+
     In these unstable cases, we fall back to the finite-difference JVP used in `check_vjp`.
 
     Args:


thunder/tests/test_grad.py
```diff
@@ -268,6 +268,16 @@ def _dot(x, y):
     return sum([_tensor_dot(a, b) for a, b in zip(x, y)])
 
 
+def _thunder_vjp(f, *primals, v=None, executor="torch", set_compile_data=False):
+    jf = executor.make_callable(f, disable_torch_autograd=True)
```
shino16: Is this useful only when `set_compile_data` is set True?

thunder/tests/test_grad.py
```diff
@@ -423,20 +428,87 @@ def test_vjp_correctness(op, device, dtype, executor, comp):
...
+                torch_filtered_args,
+            )
+            if len(torch_filtered_args) == 0:
+                continue
+
+            make = partial(make_tensor_like, low=0, high=1)
+            u_torch = tree_map(make, torch_filtered_args)
+            # Convert to torch tensors on the requested device and dtype
+            u_torch = tree_map(
+                lambda t: t.clone().to(device=device, dtype=ltorch.to_torch_dtype(dtype))
```
shino16: If I'm not wrong, `make_tensor_like` should already make sure Tensors in `u_torch` share the same `device` and `dtype` with `torch_filtered_args`.

mattteochen: Yes correct, that main point was to call `clone`. I'll update the comment.

thunder/tests/test_grad.py
mattteochen: Yes, let's move inside the branch.

thunder/tests/test_grad.py
```diff
@@ -423,20 +428,87 @@ def test_vjp_correctness(op, device, dtype, executor, comp):
...
+                set_compile_data,
+                len(sample.kwargs) != 0,
+            )
+
+        # Fast path: if a pure torch reference exists and we're NOT using the plain Torch executor,
+        # compute J·u with torch.func.jvp on the torch reference, and J*·v with Thunder VJP.
+        if getattr(op, "torch_reference", None) is not None:
+            torch_args, torch_kwargs = sample.args, sample.kwargs
+            # Sanitize non-differentiable dtype arguments for torch fast path
+            sanitize_dtype = lambda x: ltorch.to_torch_dtype(x) if isinstance(x, dtypes.dtype) else x
```
beverlylytle: Is this necessary? I thought `sample_inputs` [took care of this](https://github.com/Lightning-AI/lightning-thunder/blob/ab067fdfbe11fa83b779d7b09ad8c8c1ccf24dc0/thunder/tests/opinfos.py#L362).  Or were you running into errors with the `kwargs`?  I'm wondering because maybe you could extract lines 451-457 to a method that takes the op.torch_reference/op.op as an argument and returns the filtered op and args.

thunder/tests/test_grad.py
beverlylytle: Why is the cloning needed?  What breaks if this reassignment of u_torch is deleted?

thunder/tests/test_grad.py
```diff
@@ -423,20 +428,87 @@ def test_vjp_correctness(op, device, dtype, executor, comp):
...
+            torch_args = tree_map(sanitize_dtype, torch_args)
+            torch_kwargs = tree_map(sanitize_dtype, torch_kwargs)
+            flat_torch_op, flat_torch_args, _ = flatten_func(op.torch_reference, torch_args, torch_kwargs)
+            torch_filtered_op, torch_filtered_args = _make_differentiable_wrapper(flat_torch_op, flat_torch_args)
+            # Convert to torch tensors on the requested device and dtype
+            torch_filtered_args = tree_map(
+                lambda t: t.clone().to(device=device, dtype=ltorch.to_torch_dtype(dtype))
```
beverlylytle: Again, why do we need to replace torch_filtered_args with clones of itself?

thunder/tests/test_grad.py
```diff
@@ -423,20 +428,87 @@ def test_vjp_correctness(op, device, dtype, executor, comp):
...
+            u_torch = tree_map(
+                lambda t: t.clone().to(device=device, dtype=ltorch.to_torch_dtype(dtype))
+                if isinstance(t, torch.Tensor)
+                else t,
+                u_torch,
+            )
+
+            try:
+                outs_p_torch, J_u_torch = torch.func.jvp(torch_filtered_op, torch_filtered_args, u_torch)
+            except Exception:
```
beverlylytle: This is pretty broad.  Can this be limited to a subclass of Exceptions? Or better, if we already know what will fail (int / bool) can we skip the call to torch.func.jvp and jump to the fallback immediately in those cases?

thunder/tests/test_grad.py
```diff
@@ -423,20 +428,87 @@ def test_vjp_correctness(op, device, dtype, executor, comp):
...
+            )
+
+            # Dot-product identity
+            Ju = (J_u_torch,) if not isinstance(J_u_torch, (tuple, list)) else J_u_torch
+            vv = (v_torch,) if not isinstance(v_torch, (tuple, list)) else v_torch
+            uu = (u_torch,) if not isinstance(u_torch, (tuple, list)) else u_torch
+            Ju_dot_v = _dot(Ju, vv)
+            u_dot_Jstarv = _dot(uu, J_star_v)
+            if Ju_dot_v.isnan().any():
+                continue
```
beverlylytle: Why do we ignore NaNs?  In the fdm version, the comment suggests that the NaNs are coming from bad behavior in fdm.  Do the NaNs in Thunder not correspond to the NaNs in PyTorch?

thunder/tests/test_grad.py
```diff
@@ -423,20 +428,87 @@ def test_vjp_correctness(op, device, dtype, executor, comp):
...
+            except AssertionError:
+                # Currently for some tests with nvfuser we expect numerical instability, so we fall back to numerical_jvp-based path
+                result = fallback_to_numerical_jvp()
+                if result is not None:
+                    return result
+                continue
+        else:
+            result = fallback_to_numerical_jvp()
+            if result is not None:
+                return result
```
beverlylytle: Could this be extracted to a snippet that's then passed to `run_snippet`?

thunder/tests/test_grad.py
```diff
@@ -423,20 +428,87 @@ def test_vjp_correctness(op, device, dtype, executor, comp):
...
+            vv = (v_torch,) if not isinstance(v_torch, (tuple, list)) else v_torch
+            uu = (u_torch,) if not isinstance(u_torch, (tuple, list)) else u_torch
+            Ju_dot_v = _dot(Ju, vv)
+            u_dot_Jstarv = _dot(uu, J_star_v)
+            if Ju_dot_v.isnan().any():
+                continue
+            try:
+                comp(Ju_dot_v, u_dot_Jstarv)
+            except AssertionError:
+                # Currently for some tests with nvfuser we expect numerical instability, so we fall back to numerical_jvp-based path
```
beverlylytle: If we really expect numerical instability only with certain ops and NvFuser, and our aim is to speed things up, maybe rather than trying vjp with PyTorch and failing and then trying the other path, we could go directly to the fdm version exactly in the cases we expect.  Moreover, the generic try-catch pattern would hide from us if suddenly a new op becomes unstable with NvFuser.

thunder/tests/test_grad.py
mattteochen: For `filtered_args`:

In some test cases:
```
unsupported operation: more than one element of the written-to tensor refers to a single memory location. Please clone() the tensor before performing the operation.
```

I assume that the torch API tries to access data that has not yet been [allocated](https://discuss.pytorch.org/t/about-error-more-than-one-element-of-the-written-to-tensor-refers-to-a-single-memory-location/85526/1).

For `u_torch`:
This is not needed.

thunder/tests/test_grad.py
mattteochen: Yes, kwargs introduce some incompatibility. For example, `cumsum`. To avoid specialization based on the input test I have applied this torch args filter. But exporting as a method is a good idea.
```
 cumsum() received an invalid combination of arguments - got (Tensor, int, dtype=floating), but expected one of:
 * (Tensor input, int dim, *, torch.dtype dtype = None, Tensor out = None)
 * (Tensor input, name dim, *, torch.dtype dtype = None, Tensor out = None)
 ```

File: thunder/tests/test_grad.py (hunk 4 above)
Comment (mattteochen) [reply thunder/tests/test_grad.py]: `torch_filtered_args ` must be cloned due to vjp exceptions (see above). `u_torch` does not need to be cloned; I'll update this last.

File: thunder/tests/test_grad.py (hunk 5 above)
Comment (mattteochen) [reply thunder/tests/test_grad.py]: Yes, your way is better.

File: thunder/tests/test_grad.py (hunk 8 above)
Comment (mattteochen) [reply thunder/tests/test_grad.py]: That's a good point. I'll add a whitelist of tests where we expect these numeric failures and fall back directly to the old path.

File: thunder/tests/test_grad.py (hunk 6 above)
Comment (mattteochen) [reply thunder/tests/test_grad.py]: There are some tests where both thunder and torch give NaN (also fdm I presume):
```
FAILED thunder/tests/test_grad.py::test_vjp_correctness_logsumexp_nvfuser_cuda_thunder.dtypes.float64 - AssertionError: Scalars are not close!
FAILED thunder/tests/test_grad.py::test_vjp_correctness_logsumexp_torch_cpu_thunder.dtypes.float64 - AssertionError: Scalars are not close!
FAILED thunder/tests/test_grad.py::test_vjp_correctness_logsumexp_torch_cuda_thunder.dtypes.float64 - AssertionError: Scalars are not close!
FAILED thunder/tests/test_grad.py::test_vjp_correctness_std_nvfuser_cuda_thunder.dtypes.float64 - AssertionError: Scalars are not close!
FAILED thunder/tests/test_grad.py::test_vjp_correctness_std_torch_cpu_thunder.dtypes.float64 - AssertionError: Scalars are not close!
FAILED thunder/tests/test_grad.py::test_vjp_correctness_std_torch_cuda_thunder.dtypes.float64 - AssertionError: Scalars are not close!
FAILED thunder/tests/test_grad.py::test_vjp_correctness_var_mean_nvfuser_cuda_thunder.dtypes.float64 - AssertionError: Scalars are not close!
FAILED thunder/tests/test_grad.py::test_vjp_correctness_var_mean_torch_cpu_thunder.dtypes.float64 - AssertionError: Scalars are not close!
FAILED thunder/tests/test_grad.py::test_vjp_correctness_var_mean_torch_cuda_thunder.dtypes.float64 - AssertionError: Scalars are not close!
FAILED thunder/tests/test_grad.py::test_vjp_correctness_var_nvfuser_cuda_thunder.dtypes.float64 - AssertionError: Scalars are not close!
FAILED thunder/tests/test_grad.py::test_vjp_correctness_var_torch_cpu_thunder.dtypes.float64 - AssertionError: Scalars are not close!
FAILED thunder/tests/test_grad.py::test_vjp_correctness_var_torch_cuda_thunder.dtypes.float64 - AssertionError: Scalars are not close!
```

The errors look like:
```
E               Expected nan but got nan.
E               Absolute difference: nan (up to 1e-07 allowed)
E               Relative difference: nan (up to 1e-07 allowed)
```

Comment (beverlylytle) [reply thunder/tests/test_grad.py]: Ah, so both are NaNs.  That's good.  Probably `comp` is based on [`torch.test.assert_close`](https://docs.pytorch.org/docs/stable/testing.html).  Would it be possible to modify `comp` with `equals_nan=True`?

File: thunder/tests/test_grad.py (hunk 2 above)
Comment (beverlylytle) [reply thunder/tests/test_grad.py]: That is interesting.  Which tests fail with this error?

Comment (mattteochen) [reply thunder/tests/test_grad.py]:
```
FAILED thunder/tests/t.py::test_vjp_correctness_cat_wrapper_torchcompile_cat_cuda_thunder.dtypes.float64 - RuntimeError: unsupported operation: more than one element of the written-t...
FAILED thunder/tests/t.py::test_vjp_correctness_cat_wrapper_torchcompile_xentropy_cpu_thunder.dtypes.float64 - RuntimeError: unsupported operation: more than one element of the written-t...
FAILED thunder/tests/t.py::test_vjp_correctness_cat_wrapper_torchcompile_xentropy_cuda_thunder.dtypes.float64 - RuntimeError: unsupported operation: more than one element of the written-t...
FAILED thunder/tests/t.py::test_vjp_correctness_isinf_nvfuser_cuda_thunder.dtypes.float64 - RuntimeError: unsupported operation: more than one element of the written-t...
FAILED thunder/tests/t.py::test_vjp_correctness_isinf_torch_cpu_thunder.dtypes.float64 - RuntimeError: unsupported operation: more than one element of the written-t...
FAILED thunder/tests/t.py::test_vjp_correctness_isinf_torch_cuda_thunder.dtypes.float64 - RuntimeError: unsupported operation: more than one element of the written-t...
FAILED thunder/tests/t.py::test_vjp_correctness_isnan_nvfuser_cuda_thunder.dtypes.float64 - RuntimeError: unsupported operation: more than one element of the written-t...
FAILED thunder/tests/t.py::test_vjp_correctness_isnan_torch_cpu_thunder.dtypes.float64 - RuntimeError: unsupported operation: more than one element of the written-t...
FAILED thunder/tests/t.py::test_vjp_correctness_isnan_torch_cuda_thunder.dtypes.float64 - RuntimeError: unsupported operation: more than one element of the written-t...
FAILED thunder/tests/t.py::test_vjp_correctness_logical_not_nvfuser_cuda_thunder.dtypes.float64 - RuntimeError: unsupported operation: more than one element of the written-t...
FAILED thunder/tests/t.py::test_vjp_correctness_logical_not_torch_cpu_thunder.dtypes.float64 - RuntimeError: unsupported operation: more than one element of the written-t...
FAILED thunder/tests/t.py::test_vjp_correctness_logical_not_torch_cuda_thunder.dtypes.float64 - RuntimeError: unsupported operation: more than one element of the written-t...
```

Comment (beverlylytle) [reply thunder/tests/test_grad.py]: What happens if you use `clone` [here](https://github.com/Lightning-AI/lightning-thunder/blob/90f2af016cd0350ba12febd4a4c2f7de67aea367/thunder/tests/opinfos.py#L477) instead?

Comment (mattteochen) [reply thunder/tests/test_grad.py]: These still fail:
```
FAILED thunder/tests/test_grad.py::test_vjp_correctness_cat_wrapper_nvfuser_cuda_thunder.dtypes.float64 - RuntimeError: unsupported operation: more than one element of the written-t...
FAILED thunder/tests/test_grad.py::test_vjp_correctness_cat_wrapper_torch_cpu_thunder.dtypes.float64 - RuntimeError: unsupported operation: more than one element of the written-t...
FAILED thunder/tests/test_grad.py::test_vjp_correctness_cat_wrapper_torch_cuda_thunder.dtypes.float64 - RuntimeError: unsupported operation: more than one element of the written-t...
FAILED thunder/tests/test_grad.py::test_vjp_correctness_cat_wrapper_torchcompile_cat_cpu_thunder.dtypes.float64 - RuntimeError: unsupported operation: more than one element of the written-t...
FAILED thunder/tests/test_grad.py::test_vjp_correctness_cat_wrapper_torchcompile_cat_cuda_thunder.dtypes.float64 - RuntimeError: unsupported operation: more than one element of the written-t...
FAILED thunder/tests/test_grad.py::test_vjp_correctness_cat_wrapper_torchcompile_xentropy_cpu_thunder.dtypes.float64 - RuntimeError: unsupported operation: more than one element of the written-t...
FAILED thunder/tests/test_grad.py::test_vjp_correctness_cat_wrapper_torchcompile_xentropy_cuda_thunder.dtypes.float64 - RuntimeError: unsupported operation: more than one element of the written-t...
```

Comment (beverlylytle) [reply thunder/tests/test_grad.py]: So then maybe a `clone` [here](https://github.com/Lightning-AI/lightning-thunder/blob/e809222579453676443849b25cbac27f001d1da2/thunder/tests/opinfos.py#L3685)?

Comment (mattteochen) [reply thunder/tests/test_grad.py]: Yeah, I am looking into the different generators right now. Thank you!

File: thunder/tests/test_grad.py [hunk 9]
```diff
@@ -268,6 +333,58 @@ def _dot(x, y):
...
+
+    # Dot-product identity
+    Ju = (J_u_torch,) if not isinstance(J_u_torch, (tuple, list)) else J_u_torch
+    vv = (v_torch,) if not isinstance(v_torch, (tuple, list)) else v_torch
+    uu = (u_torch,) if not isinstance(u_torch, (tuple, list)) else u_torch
+    Ju_dot_v = _dot(Ju, vv)
+    u_dot_Jstarv = _dot(uu, J_star_v)
+
+    # Currently, assert_close from torch.testing is used to check the equality of two tensors.
+    if getattr(comp, "__name__", "") == "assert_close" and getattr(comp, "__module__", "").startswith("torch.testing"):
```
beverlylytle: Some opinfos are given a custom_comparator, but after a brief glances through OpInfos.py, it looks like only one of them isn't based on `assert_close`.  Many of them are partials and I think they would fail this condition.  Since these `comp`s are relatively cheap operations, do you think it would be alright to do something like 
```
try:
    comp(Ju_dot_v, u_dot_Jstarv, equal_nan=True)
except TypeError:
    comp(Ju_dot_v, u_dot_Jstarv)
```

mattteochen: Yes, this makes sense as some of them will fail during the evaluation.

mattteochen:
> Looks good! Thank you for the iterations.
> 
> One last nit: The docstring for `check_vjp` is more or less the same as `check_vjp_torch`. Could you add some details to both explaining how they are different and why one is used over the other?

Yes, sure! Thank you!

t-vi:
Thank you @mattteochen @beverlylytle  for your diligence.

thunder/tests/opinfos.py
```diff
@@ -3682,7 +3679,7 @@ def cat_sample_generator(op, device, dtype, requires_grad, **kwargs):
 
     # Tests concatenating with a tensor broadcast along the concatenation dimension
     a = make((5,))
-    b = make((1,)).expand((5,))
+    b = make((1,)).expand((5,)).clone()
```
IvanYashchuk, line 3682: Cloning here changes strides of the tensor `b` turning a non-contiguous tensor into a contiguous tensor:
```py
import torch
print(torch.randn((1,)).expand((5,)).stride()) # (0,)
print(torch.randn((1,)).expand((5,)).clone().stride()) # (1,)
```

thunder/tests/opinfos.py
```diff
@@ -471,10 +471,7 @@ def elementwise_unary_generator(
     )
 
     for shape, strides, offset in strided_cases:
-        a = make_arg(
-            500,
-        ).as_strided(shape, strides, offset)
-        a = a.detach().requires_grad_(requires_grad)
+        a = make_arg(500).as_strided(shape, strides, offset).clone().detach().requires_grad_(requires_grad)
```
IvanYashchuk, line 474: Clone overrides the `as_strided` command producing contiguous tensors. We need to continue testing strided cases so we should revert this change and figure out a way to make the derivative tests pass.

thunder/tests/test_grad.py
```diff
@@ -268,6 +333,73 @@ def _dot(x, y):
...
+        initial_trace_vjp_f = thunder.trace()(vjp(f), primals, v)
+    return executor.make_callable(initial_trace_vjp_f.python_callable(), disable_torch_autograd=True)(primals, v)
+
+
+def check_vjp_torch(
+    f_torch, f_thunder, primals_torch, primals_thunder, comp, executor="torch", set_compile_data: bool = False
+):
+    """Check that the vector-Jacobian product of a function is correct.
+
+    This variant uses `torch.func.jvp` to compute J*u directly, which is
```
IvanYashchuk, line 351: The function is called "check_vjp_torch". You don't need to use `torch.func.jvp` when PyTorch provides "vjp" computation directly.

thunder/tests/test_grad.py
```diff
@@ -268,6 +333,73 @@ def _dot(x, y):
...
+    u_torch = tree_map(make, primals_torch)
+
+    outs_p_torch, J_u_torch = torch.func.jvp(f_torch, primals_torch, u_torch)
+
+    v_torch = tree_map(make, outs_p_torch)
+    _, J_star_v = _thunder_vjp(
+        f_thunder, *primals_thunder, v=v_torch, executor=executor, set_compile_data=set_compile_data
+    )
+
+    # Dot-product identity
```
IvanYashchuk, line 388: Instead of checking Dot-product identity you can simple check whether Thunder's vjp result matches PyTorch's vjp result.

thunder/tests/test_grad.py
```diff
@@ -423,18 +581,59 @@ def test_vjp_correctness(op, device, dtype, executor, comp):
...
+        flat_torch_op, flat_torch_args, _ = flatten_func(op.torch_reference, torch_args, torch_kwargs)
+        torch_filtered_op, torch_filtered_args = _make_differentiable_wrapper(flat_torch_op, flat_torch_args)
+
+        # If differentiable args include non-float/complex tensors, proactively fallback to numerical JVP
+        torch_not_supported = any(
+            isinstance(t, torch.Tensor) and not (torch.is_floating_point(t) or torch.is_complex(t))
+            for t in tree_flatten(torch_filtered_args)[0]
         )
+
+        # Fast path: if a pure torch reference exists,
```
IvanYashchuk, line 626: What are the opinfos that don't have PyTorch references?

thunder/tests/opinfos.py
beverlylytle: Oh, I did not realize that.  Interesting.

beverlylytle: So when this `clone` is removed, the callstack at the point of the error is:
```
thunder/tests/test_grad.py:633: in test_vjp_correctness
    result = _torch_jvp()
thunder/tests/test_grad.py:602: in _torch_jvp
    return run_snippet(
thunder/tests/framework.py:572: in run_snippet
    raise ex
thunder/tests/framework.py:564: in run_snippet
    snippet(*args, **kwargs)
thunder/tests/test_grad.py:539: in snippet_vjp_correctness_torch
    check_vjp_torch(
thunder/tests/test_grad.py:381: in check_vjp_torch
    outs_p_torch, J_u_torch = torch.func.jvp(f_torch, primals_torch, u_torch)
../../miniforge3/envs/tfes2/lib/python3.10/site-packages/torch/_functorch/eager_transforms.py:1043: in jvp
    return _jvp_with_argnums(
../../miniforge3/envs/tfes2/lib/python3.10/site-packages/torch/_functorch/eager_transforms.py:1095: in _jvp_with_argnums
    flat_duals = tuple(
../../miniforge3/envs/tfes2/lib/python3.10/site-packages/torch/_functorch/eager_transforms.py:1096: in <genexpr>
    fwAD.make_dual(p, t) for p, t in zip(flat_primals, flat_tangents)
```
It seems that torch's `make_dual` is doing some sort of in-place operation on the primals. The primals for a few of the tests are the result of either `expand` or `as_strided`.  Both of these operations have big warnings in pytorch ([here](https://docs.pytorch.org/docs/stable/generated/torch.as_strided.html) and [here](https://docs.pytorch.org/docs/stable/generated/torch.Tensor.expand.html)) about memory re-usage, and this is why the failure with the suggestion of cloning the input is occuring.  I don't think this is a bug for Thunder, but a limitation of torch's jvp.

mattteochen: It will be interesting to check if the torch vjp computation, as outlined by @IvanYashchuk, directly has the same issues or not. If not, we could change the test identity and remove the clone call, further improving the test latency #2645.

thunder/tests/test_grad.py
mattteochen: Only `OpInfo`.

thunder/tests/test_grad.py
mattteochen: This is even better, I've opened #2645.

PR #2620, opened 2025-10-09
[benchmark_inference] Fix replacing the MoE

kshitij12345:
Changes:
* Replacement of `Llama4TextMoe` didn't happen correctly as `module.named_children` returns an iterator over immediate children modules. As `Llama4TextMoe` is not an immediate child of the `model`, it wasn't replaced. We add `_replace_with_custom_fn_if_matches_filter_with_name` similar to torchao. Ref: https://docs.pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.named_children
* We also need some fixes to `from_transformers_llama4textmoe`.
* We materialize the model after replacement of the module to avoid materializing the `Llama4Moe` with the original module to save some memory.

Verified with - `python thunder/benchmarks/benchmark_inference.py --input-length 32 --output-length 32 --mode eager --num-iterations 10`

### DIFF
--- a/thunder/benchmarks/benchmark_inference.py
+++ b/thunder/benchmarks/benchmark_inference.py
@@ -20,6 +20,8 @@ import statistics
 import sys
 import time
 import warnings
+from typing import Any
+from collections.abc import Callable
 
 import torch
 import torch.nn as nn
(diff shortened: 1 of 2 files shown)


thunder/benchmarks/benchmark_inference.py
```diff
@@ -145,7 +183,20 @@ def __init__(self, config: InferenceBenchmarkConfig):
...
+
+        # NOTE: Replacement happens before model is materialized
+        #       otherwise, the memory usage will be increased due to
+        #       additional parameters materialized from the replacement module
+        if not self.config.disable_moe_replacement:
+            _replace_llama4_moe(model)
+        assert all(p.device == torch.device("meta") for p in model.parameters())
+
+        # Materialize the model on the device
+        model.to_empty(device=DEVICE)
```
kshitij12345, line 198: I think this should be ok as we don't load the model weights anyways so they are random already.

Should we initialize the model with `torch.nn.init` after this point?

crcrpar: Wouldn't tansformers automodel load weights when a model is on a non-meta device?

My thought was that we should load weights then replace our interesting layers with our fond definitions and reflect the trained weights to the custom def's

kshitij12345: > Wouldn't tansformers automodel load weights when a model is on a non-meta device?

I don't think that is true when using `AutoModelForCausalLM.from_config(config)`

Example:
```python
import torch

def _load_model() -> torch.nn.Module:
    model_id = "meta-llama/Llama-4-Maverick-17B-128E"
    config = AutoConfig.from_pretrained(model_id)

    if hasattr(config, "text_config"):
        config = config.text_config

    config.num_hidden_layers = 2

    with torch.device("cuda"):
        model = AutoModelForCausalLM.from_config(config, torch_dtype=torch.bfloat16)

    return model

if __name__ == "__main__":
    model = _load_model()
    print(model.lm_head.weight[0])

```

Output from script
```python
# First run
tensor([-0.0179,  0.0264, -0.0189,  ..., -0.0151,  0.0095, -0.0098],
       device='cuda:0', dtype=torch.bfloat16, grad_fn=<SelectBackward0>)

# Second run
tensor([ 0.0247,  0.0007, -0.0189,  ..., -0.0188,  0.0095,  0.0184],
       device='cuda:0', dtype=torch.bfloat16, grad_fn=<SelectBackward0>)

# Third run
tensor([-0.0352, -0.0142, -0.0036,  ..., -0.0427,  0.0167,  0.0050],
       device='cuda:0', dtype=torch.bfloat16, grad_fn=<SelectBackward0>)
```

EDIT: Also, found an issue on HF transformers discussing the behaviour of `AutoModel.from_config` - https://github.com/huggingface/transformers/issues/4685

kshitij12345: > My thought was that we should load weights then replace our interesting layers with our fond definitions and reflect the trained weights to the custom def's

Yes, ideally we should be doing that.

PR #2623, opened 2025-10-10
try getting version from `nvfuser_direct` first

crcrpar:
## What does this PR do?

This is to avoid the following warning
```
python/nvfuser_direct/__init__.py:11
  /opt/pytorch/nvfuser/python/nvfuser_direct/__init__.py:11: UserWarning: Be careful! You've imported nvfuser_direct when the nvfuser module is already imported.
    warnings.warn(
```

### DIFF
--- a/thunder/tests/conftest.py
+++ b/thunder/tests/conftest.py
@@ -1,4 +1,5 @@
 import importlib
+import importlib.util
 
 import pytest
 import pytest_benchmark


thunder/executors/nvfuserex.py
```diff
@@ -15,16 +15,18 @@ def nvfuser_version() -> LooseVersion | None:
         return None
 
     try:
+        import nvfuser_direct
+
+        if hasattr(nvfuser_direct, "version"):
+            return LooseVersion(nvfuser_direct.version())
+    except ImportError:
         import nvfuser
```
kshitij12345: I think this import should also be guarded with try-except for the scenario when `nvfuser` is not installed.

crcrpar: updated

thunder/executors/nvfuserex.py
```diff
@@ -14,17 +14,25 @@ def nvfuser_version() -> LooseVersion | None:
...
+        try:
+            import nvfuser
+        except ImportError:
+            pass
+        else:
+            if hasattr(nvfuser, "version"):
+                return LooseVersion(nvfuser.version())
+            else:
+                # NOTE: This import of nvFuser may or may not have version info
+                return LooseVersion("0.0.0")
```
kshitij12345, line 30:
```python
if hasattr(nvfuser, "version"):
  return LooseVersion(nvfuser.version())
else:
  # NOTE: This import of nvFuser may or may not have version info
  return LooseVersion("0.0.0")
```

can be moved into a helper function, `get_nvfuser_version(module) -> LooseVersion`. This would probably be simpler to read.

PR #2626, opened 2025-10-10
Add DIV_EXACT prim

### LINKED ISSUE #2599: HF ibm-granite/granite-3.1-3b-a800m-instruct: index_add_(): Expected dtype int32/int64 for index but got: Float
Image - pjnl-20251003 or above

Repro:
```python
import torch

dtype = torch.bfloat16
BS = 2
SEQ_LEN = 4096


HF_TOKEN = None

import thunder

model_name = "ibm-granite/granite-3.1-3b-a800m-instruct"
with torch.device("cuda"):
    config = AutoConfig.from_pretrained(model_name, trust_remote_code=False)
    config.num_hidden_layers = 2
    model = AutoModelForCausalLM.from_config(config, torch_dtype=dtype, trust_remote_code=False)


inp = torch.randint(0, 10, (BS, SEQ_LEN), device="cuda")


model(inp)
print("eager works")

# torch.compile(model)(inp)
# print("torch.compile works")

model = thunderfx(model)

#   File "<eval_with_key>.254", line 6, in forward
#     layer_output = zeros_2.index_add(0, batch_index, expert_outputs);  zeros_2 = batch_index = expert_outputs = None
#                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
# RuntimeError: index_add_(): Expected dtype int32/int64 for index but got: Float

# Call using an FX-traced Module, line 6 of the traced Module's generated forward function:
#     _set_grad_enabled = torch._C._set_grad_enabled(True);  _set_grad_enabled = None
#     layer_output = zeros_2.index_add(0, batch_index, expert_outputs);  zeros_2 = batch_index = expert_outputs = None

# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ <--- HERE
#     return layer_output
model(inp)
print("thunder default executors work")
print("DONE")

```

beverlylytle:
Fixes #2599

In the model HF ibm-granite/granite-3.1-3b-a800m-instruct a certain set of indices is computed by dividing two tensors of ints.  In general, `div` is differentiable, but not when both operands are of exact dtype and the rounding mode is "trunc".  Because there is no exception for this case, autodiff fetches the augmented forward part coming from div's grad transform, and this returns an inexact result.  This PR introduces a new prim DIV_EXACT which is the same as `prims.DIV` except without having a grad transform registered to it.  A check on dtypes forwards the call to DIV_EXACT instead of DIV.

### DIFF
--- a/thunder/core/prims.py
+++ b/thunder/core/prims.py
@@ -2624,7 +2624,7 @@ div_exact = _make_elementwise_binary_prim(
     PrimIDs.DIV_EXACT,
     "div_exact",
     number_fn=_div_numbers,
-    supported_input_dtypes=math_dtypes,
+    supported_input_dtypes=dtypes.exact_dtypes,
 )
 
 # We currently do not support floordiv on tensors.


thunder/core/prims.py
```diff
@@ -2618,6 +2619,14 @@ def _div_numbers(a: Number, b: Number) -> Number:
     supported_input_dtypes=math_dtypes,
 )
 
+# The non-differentiable version of div
+div_exact = _make_elementwise_binary_prim(
+    PrimIDs.DIV_EXACT,
+    "div_exact",
+    number_fn=_div_numbers,
+    supported_input_dtypes=math_dtypes,
```
kshitij12345: It would be better to use `exact_dtypes`.

https://github.com/Lightning-AI/lightning-thunder/blob/ab067fdfbe11fa83b779d7b09ad8c8c1ccf24dc0/thunder/core/dtypes.py#L247

thunder/tests/test_ops.py
```diff
@@ -525,3 +525,20 @@ def fn(a):
     jfn = thunder.jit(fn)
     a = torch.randn(5, 5, requires_grad=True)  # trigger grad transform
     assert_close(fn(a), jfn(a))
+
+
+def test_div_exact():
+    def fn(a, b, c):
+        indices = torch.div(a, b, rounding_mode="trunc")
+        # this would throw an error if indices are not ints
+        return c[indices]
```
kshitij12345, line 534: assert_close will also check `dtype`, so `c[indices]` shouldn't be required.

```python
>>> import torch
>>> torch.testing.assert_close(torch.tensor([0], dtype=torch.int32), torch.tensor([0], dtype=torch.int64))
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.12/dist-packages/torch/testing/_comparison.py", line 1587, in assert_close
    raise error_metas[0].to_error(msg)
AssertionError: The values for attribute 'dtype' do not match: torch.int32 != torch.int64.
```

beverlylytle: `a` and `b` are tensors of exact type, and so cannot have `requires_grad` True.  However, there has to be a differentiable tensor in the program for the function to pass through the relevant path of autodiff.  Thus some operation involving the result of `div` combined with a grad-requiring tensor is necessary to trigger corrected path.

thunder/core/prims.py
beverlylytle: Good call.

thunder/tests/test_ops.py
kshitij12345: Got it, thanks!

t-vi:
Is div_exact a good name? I thought floor_div was great to represent `//`

beverlylytle:
> Is div_exact a good name? I thought floor_div was great to represent //

I am open to changing the name, but since this prim is used for both floor_div and trunc_div I don't think it should be named floor_div.  Do you have another suggestion?

thunder/tests/test_ops.py
```diff
@@ -525,3 +525,20 @@ def fn(a):
     jfn = thunder.jit(fn)
     a = torch.randn(5, 5, requires_grad=True)  # trigger grad transform
     assert_close(fn(a), jfn(a))
+
+
+def test_div_exact():
```
IvanYashchuk, line 530: The test function works even without the change introduced in this PR:
```
git checkout 80461ba^
```py
In [1]: import torch, thunder

In [2]:     def fn(a, b, c):
   ...:         indices = torch.div(a, b, rounding_mode="trunc")
   ...:         # this would throw an error if indices are not ints
   ...:         return c[indices]
   ...: 

In [3]:     jfn = thunder.jit(fn)
   ...:     a = torch.randint(1, 5, (5,))
   ...:     b = torch.ones(5, dtype=torch.int32)
   ...:     c = torch.randn(5, 5)
   ...:     fn(a, b, c), jfn(a, b, c)
Out[3]: 
(tensor([[ 0.5290, -2.2824,  1.0693, -1.6769, -1.9725],
         [-1.0802, -0.4437, -0.7387, -1.1378,  0.6227],
         [-0.1425,  0.3009,  1.3933, -0.8111,  0.2195],
         [-0.1425,  0.3009,  1.3933, -0.8111,  0.2195],
         [-0.1425,  0.3009,  1.3933, -0.8111,  0.2195]]),
 tensor([[ 0.5290, -2.2824,  1.0693, -1.6769, -1.9725],
         [-1.0802, -0.4437, -0.7387, -1.1378,  0.6227],
         [-0.1425,  0.3009,  1.3933, -0.8111,  0.2195],
         [-0.1425,  0.3009,  1.3933, -0.8111,  0.2195],
         [-0.1425,  0.3009,  1.3933, -0.8111,  0.2195]]))
```
The `c` tensor must have requires_grad=True to trigger buggy code path.

beverlylytle: Oh, shoot.  `c` lost its require_grad in the various iterations.  Good catch.

beverlylytle: #2663

PR #2632, opened 2025-10-12
TE inference executor for 8 bit

t-vi:
## What does this PR do?

This implements inference (only) fp8 linears using transformer engine.
It is modelled after the bitsandbytes transform.

One question I'd have is what a good scaling would be. I'm currently using the max range (`fp8_max / tensor.absmax()`) on the weight and 1.0 on the input, but I have no idea what a good input scale would be (this would depend on the accumulation in fp8 matmuls).

## PR review

## Did you have fun?

### DIFF
--- a/thunder/transforms/te_inference.py
+++ b/thunder/transforms/te_inference.py
@@ -180,6 +180,9 @@ class TEInference8BitTransform(Transform):
             if isinstance(submodule, torch.nn.Linear) or submodule.__class__.__name__ == "GroupedLinear":
                 convert_linear_submodule(model, n)
 
+    def get_executor(self):
+        return get_te_inference_executor()
+
     def transform_state_dict_for_submodule(self, model: thunder.ThunderModule, submodule_name: str, state_dict: dict):
         # note that state dict entries do not include the submodule name as prefix
         if submodule_name not in self.quantized_submodule_names:


thunder/transforms/te_inference.py
```diff
@@ -0,0 +1,323 @@
...
+
+            out, *_ = transformer_engine.pytorch.ops.BasicLinear._functional_forward(
+                x,
+                w,
+                input_quantizer=xq,
+                with_quantized_compute=True,
+                weight_requires_grad=False,
+                input_requires_grad=False,
+            )
+
```
t-vi, line 65: This is the core of how I compute linears. Does this make sense? What is a good scale for the input? I used 1/absmax now such that `x[j, k] * w[i, k]` would be `< fp8_max`, not sure if that is needed.

thunder/transforms/te_inference.py
```diff
@@ -0,0 +1,323 @@
...
+                fp8_dtype=transformer_engine_torch.DType.kFloat8E4M3,
+                rowwise=True,
+                columnwise=False,
+            )
+
+            w = wq.create_tensor_from_data(qweight, fake_dtype=x.dtype, requires_grad=False)
+
+            minmax = x.aminmax()
+            xmax = torch.maximum(minmax.min.abs(), minmax.max.abs()).to(torch.float32)
+            xq = transformer_engine.pytorch.Float8Quantizer(
```
kshitij12345, line 49: Do we necessarily want to use `Float8Quantizer` which ties up with DelayedScaling recipe (but works for both Hopper and Blackwell archs).

Instead we can also use `MXFP8Quantizer` (note: this is only supported with Blackwell arch). For this recipe, the quantization of the input will not depend on scale from previous iteration.

https://github.com/NVIDIA/TransformerEngine/blob/7ad130efd52c3aa4a386d25f1d42b28d5aa20090/transformer_engine/pytorch/tensor/mxfp8_tensor.py#L29

t-vi: Sadly, I don't currently have easy access to Blackwell, so I'm keen to support hopper, too.
Of course, I'd 100% love to have something flexible enough for fp8 + fp4 on Hopper / Blackwell to the extend that it is supported.

PR #2635, opened 2025-10-13
Fix import of TE `Recipe`

ksivaman:
<details>

</details>

## What does this PR do?

Thunder uses internal TE API which is being refactored in https://github.com/NVIDIA/TransformerEngine/pull/2259. The TE PR ensures full backwards compatibility, however, the `Recipe` class was not defined in `fp8.py` to begin with and this only works because it is imported indirectly in it. This PR imports it directly in order to not break integration immediately.

## Future work

Use of internal TE API should be minimized, and imports from `fp8.py` should be changed as it will be eventually removed (currently renamed to `quantization.py` starting TE v2.9).

### DIFF
--- a/thunder/benchmarks/layers_for_inference_benchmark.py
+++ b/thunder/benchmarks/layers_for_inference_benchmark.py
@@ -300,7 +300,15 @@ class NVFP4InferenceLinear(nn.Module):
         )
 
     @staticmethod
-    def from_linear(linear: nn.Linear) -> NVFP4InferenceLinear:
+    def from_linear(linear: nn.Linear, fqn: str | None = None) -> NVFP4InferenceLinear:
+        """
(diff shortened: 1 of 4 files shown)


nvMelissa:
Adding @riccardofelluga as reviewer. Please take a look

thunder/executors/transformer_engineex_impl.py
```diff
@@ -30,10 +30,10 @@
 import transformer_engine.pytorch as te
 import transformer_engine.common.recipe as te_recipe
 from transformer_engine.pytorch.constants import MXFP8_BLOCK_SCALING_SIZE
+from transformer_engine.common.recipe import Recipe
 from transformer_engine.pytorch.fp8 import (
     _amax_and_scale_update,
     get_fp8_max,
-    Recipe,
```
riccardofelluga, line 36: Thanks for the update! LGTM

PR #2641, opened 2025-10-14
Enable E712

crcrpar:
## What does this PR do?

As per title.

Related #2218

cc @borda

### DIFF
--- a/thunder/executors/sdpaex.py
+++ b/thunder/executors/sdpaex.py
@@ -77,7 +77,7 @@ def _attention_mask_memory_efficient_helper(attn_mask: None | torch.Tensor, quer
     # When a boolean mask is used, it needs to be converted to an additive mask where zero'd elements are filled
     # with a very negative value that should become ~0 after softmax
     if attn_mask.dtype == torch.bool:
-        attn_mask = torch.masked_fill(torch.zeros_like(attn_mask, dtype=query.dtype), attn_mask == False, -math.inf)
+        attn_mask = torch.masked_fill(torch.zeros_like(attn_mask, dtype=query.dtype), not attn_mask, -math.inf)
 
     # Expand the number of heads in attention mask to match query, key, and value tensors.
     num_heads = query.shape[1]
--- a/thunder/tests/test_interpreter.py
+++ b/thunder/tests/test_interpreter.py
@@ -274,14 +274,14 @@ def test_dunder_bool_instance(jit):
     bx = bool(x)
     jitting = True
     jbx = jit(bool)(x)
-    assert bx == jbx == False
+    assert bx == jbx is False
 
     x.__bool__ = lambda: True  # dunder methods use class attribute, not instance attribute.
     jitting = False
     bx = bool(x)
     jitting = True
     jbx = jit(bool)(x)
-    assert bx == jbx == False
+    assert bx == jbx is False
 
 
 def test_function_call(jit):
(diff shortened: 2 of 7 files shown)


thunder/executors/sdpaex.py
```diff
@@ -77,7 +77,7 @@ def _attention_mask_memory_efficient_helper(attn_mask: None | torch.Tensor, quer
     # When a boolean mask is used, it needs to be converted to an additive mask where zero'd elements are filled
     # with a very negative value that should become ~0 after softmax
     if attn_mask.dtype == torch.bool:
-        attn_mask = torch.masked_fill(torch.zeros_like(attn_mask, dtype=query.dtype), attn_mask == False, -math.inf)
+        attn_mask = torch.masked_fill(torch.zeros_like(attn_mask, dtype=query.dtype), not attn_mask, -math.inf)
```
t-vi, line 80:
```suggestion
        attn_mask = torch.masked_fill(torch.zeros_like(attn_mask, dtype=query.dtype), ~attn_mask, -math.inf)
```

To my mind `== False` is less "line noise" than `~attn_mask` and the entire E712 is more than dubious.

t-vi: But of course, we can enable it, given that the linters love it.

crcrpar: I'm not sure about what/how you think so. https://docs.astral.sh/ruff/rules/true-false-comparison/ reads quite reasonable to me

t-vi: `not (torch.randn(5, 5) > 0)`

t-vi: It also misrepresents pep 8, which literally says

> Don’t compare boolean values to True or False using ==:

but the check does 

> Don’t compare *any* values to True or False using ==:

thunder/tests/test_interpreter.py
```diff
@@ -274,14 +274,14 @@ def __bool__(self):
     bx = bool(x)
     jitting = True
     jbx = jit(bool)(x)
-    assert bx == jbx == False
+    assert bx == jbx is False
```
shino16: [PEP 8](https://peps.python.org/pep-0008/#programming-recommendations) says comparing to truth values using `is` is even worse than using `==`. To my personal eyes, `is` looks totally fine though.

thunder/tests/test_interpreter.py
```diff
@@ -624,9 +624,9 @@ def bare_except():
             assert is_jitting_with_raise() == jitting
             return True
 
-    assert bare_except() == True
+    assert bare_except()
```
shino16, line 633: `assert X == True` is not equivalent to `assert X` if `X` may take non-boolean values. What do you think about keeping `== True` for tests on return values?

thunder/tests/test_interpreter.py
shino16, line 277: [PEP 8](https://peps.python.org/pep-0008/#programming-recommendations) says comparing to truth values using `is` is even worse than using `==`. But if we really intend it, it would make sense to use `is` for all those tests.

thunder/tests/test_interpreter.py
t-vi: I guess `is True` would not be objected to.
```suggestion
    assert bare_except() is True
```

shino16: I agree, and it makes sense too.

PR #2642, opened 2025-10-14
Enable E722

crcrpar:
## What does this PR do?

As per title, this enables [E722](https://docs.astral.sh/ruff/rules/bare-except) check.

Related: #2218


crcrpar:
@t-vi could you review this?

PR #2662, opened 2025-10-16
Clean up traces for symbolic values caching

beverlylytle:
When the symbolic values caching option is enabled, there are many duplicated calls to prims.eq and prims.shape that appear in any given bsym's subsymbols.  DCE is currently applied before the decent to the subsymbols happens.  When the descent to the subsymbols happens, it results in a very ugly and hard to read trace.  This PR applies dce to the bsym's subsymbols earlier on to tidy things up.

Fixes https://github.com/Lightning-AI/lightning-thunder/issues/2728


beverlylytle:
The initial draft in [e7c8bc9](https://github.com/Lightning-AI/lightning-thunder/pull/2662/commits/e7c8bc98234f7e0adf9da0d44cf07ce7c953c96a) applied dce to subsymbols within `remove_duplicate_number_proxies`, a function whose job it is to clean after symbolic values.  In [cb31f7f](https://github.com/Lightning-AI/lightning-thunder/pull/2662/commits/cb31f7fb6ea0c60ef65ca694ae9365b37addc008), dce is now applied in within `Symbol.__call__` when the subsymbols are created.  It unfortunately requires a local import of `dce` because of dependency hell.  I rather prefer the idea of all the symbolic values tidying up happening in one method.

thunder/core/symbol.py
```diff
@@ -323,6 +323,11 @@ def __call__(self, *args, **kwargs):
 
             result = tree_unflatten(flat_results, spec)
 
+            # When using symbolic values, there may be duplicate prims.eq and prims.shape subsymbols that can be removed.
+            from thunder.core.transform_common import dce
+
+            subsymbols = dce(subsymbols, output=result)
```
IvanYashchuk: Why are subsymbols dce'd at this point specifically? Later in the code there's some postprocessing of `results`. Can it affect how subsymbols are filtered?

thunder/core/transform_common.py
```diff
@@ -142,20 +142,29 @@ def keep_or_swap(p):
 #   that only produce non-proxy objects
 # NOTE needed_proxies is an in/out argument, it takes an initial set of Variables you want to keep, and return
 #   all the needed proxies of the input trace
-def dce(trace: Trace, needed_proxies: None | set[Variable] = None) -> Trace:
+def dce(
+    trace_or_bsyms: Trace | list[BoundSymbolInterface], needed_proxies: None | set[Variable] = None, output=None
+) -> Trace | list[BoundSymbolInterface]:
```
IvanYashchuk: The function `dce` now accepts either a Trace or a list of BoundSymbolInterface objects, but the parameter output is required only in the latter case and lacks type annotation or validation.

thunder/core/transform_common.py
```diff
@@ -142,20 +142,29 @@ def keep_or_swap(p):
...
-    producer_map: ProxyDict = producers(trace)
+    producer_map: ProxyDict = producers(trace_or_bsyms)
 
-    flat_trace_outputs, _ = tree_flatten(trace.output)
+    if isinstance(trace_or_bsyms, Trace):
+        bound_symbols = trace_or_bsyms.bound_symbols
+        output = trace_or_bsyms.output
+    else:
+        bound_symbols = trace_or_bsyms
+        output = output
```
IvanYashchuk:
```suggestion
```

thunder/core/transform_common.py
IvanYashchuk: Do we have a test for this utility function? Could you please update the test to include `list[BoundSymbolInterface]` type of inputs?

IvanYashchuk (approved):
Moving DCE to symbol creation results in much tidier outputs right where most contributors and users will inspect them. This is a targeted and elegant change, thank you for addressing technical debt so precisely!

beverlylytle:
I'm not convince that `Symbol.__call__` is the right place for this dce.  The dynamo tests are failing non-trivially, and they weren't with the other version.  

I don't have a minimal failing example yet, but I do see that in more complex tests that fusions are missing required bound symbols.
I think this has something to do with the fact that `results` is the output of a symbol's meta, and not the function itself.  I am seeing things like `check_len` with a meta that should return `None` having its subsymbol `prims.check_len` eliminated because of that.  Of course that doesn't explain why fusions are being gutted.

Things are still failing when I move the `dce(subsymbols)` line below to after the `results` are finalized.

thunder/core/symbol.py
beverlylytle: I had quickly moved this up because, in the other `is_prim` branch, there are no subsymbols and that was mucking things up.  When moving it down, I still encounter errors as mentioned in the comment below.

beverlylytle:
I don't have it fully figured out yet, but I'm narrowing in on the problem.  Consider
```
import torch

def foo(x):
    y = torch.cos(x) + torch.sin(x)
    return y

x = torch.randn(10, 10, requires_grad=True)

jfoo = thunderfx(foo)
jfoo(x)
```
No error is surfaced when running this code.  However, it does erroneously result in a split graph.  When the splitter is testing the node that has the add operation, a KeyError about 't1' is thrown and caught, resulting in the split  Looking more carefully at this node, one sees that the bound symbol that is attempting to be executed is `t1 = ltorch.add(t0, t1)`.  Note the name collision between the output and the args.  I can add some hack at the top of `Symbol.__call__` that adds unseen arg names to the trace:
```
        flat_args, _ = tree_flatten((args, kwargs))
        for arg in flat_args:
            if not hasattr(arg, 'name') or trace.has_name(arg.name):
                continue
            trace.add_name(arg.name)
```
and the graph is no longer split.  But obviously this isn't good.  I can't figure out how dce'ing the subsymbols results in this name collision.  The only time the dce'ing changes a list of subsymbols in the above example is for ltorch.add where it starts with `[i0 = prims.ne(alpha, 1), t1 = prims.add(a, b)]`, but dce removes the `prims.ne`.  

@IvanYashchuk What do you make of this?  Is it worth pulling more on this thread?  or can we go back to having the dce happen in `remove_duplicate_number_proxies`?

IvanYashchuk:
I think it's worth looking into further. What part of the codebase is generating `t1 = ltorch.add(t0, t1)`? It shouldn't happen. For normal non-view non-inplace operations, the output name should be new. This might be surfacing a hidden bug that is worth fixing.

thunder/dynamo/utils.py
```diff
@@ -374,7 +376,7 @@ def _run_with_cache_info():
             else thunder_symbol
         )
         # We need to be under trace context to generate proxies.
-        with thunder.core.trace.tracectx(TraceCtx()):
+        with thunder.core.trace.tracectx(tracectx):
             try:
                 function_to_run(*proxy_args, **proxy_kwargs)
```
beverlylytle, line 402: This was the source of all of my woes.  This pattern implicitly binds the provided `proxy_args` to the symbols being generated, as opposed to mapping the input args to the input args an established trace expects.  But the `proxy_args` were being created in a distinct TraceCtx, and this result in name collisions.  This was revealed in an application of DCE that exists way, way, deep down in the call stack when executing `function_to_run`.  DCE creates a producer map, and in accessing this map, a KeyError was raised, triggering a graph split.

thunder/core/transform_common.py
beverlylytle: Added a docstring and a test.

thunder/executors/nvfuserex_impl.py
```diff
@@ -706,14 +706,11 @@ def has_cuda_input_or_output(self, bsym: BoundSymbol) -> bool:
...
-        trace = TraceCtx(None)
-        trace.bound_symbols = bsyms
-        bsyms.append(prims.python_return.bind(output, output=None))
         needed_proxies: set[Variable] = set()
-        trace = dce(trace, needed_proxies)
+        bsyms = dce(bsyms, needed_proxies, output)
         # update the input_list by removing the unused inputs
         input_list[:] = [x for x in input_list if variableify(x) in needed_proxies]
-        return list(filter(lambda x: x.sym != prims.python_return, trace.bound_symbols))
+        return bsyms
```
beverlylytle, line 714: This isn't necessary for the PR, but since I changed the signature of `dce` I thought I might as well extra use of it.

thunder/tests/test_core.py
```diff
@@ -3295,22 +3325,29 @@ def clean(tr):
...
     with tracectx(trace):
-        x = prims.unpack_trivial(a, name="x")
-        y = prims.unpack_trivial(b, name="y")
+        x, y = return_bsym.flat_args
         packed_list = prims.pack_list(x, y)
         prims.python_return(packed_list)
 
     func = trace.python_callable()
-    actual = func()
+    print(trace)
```
beverlylytle: Oops, will remove.

thunder/core/symbol.py
```diff
@@ -340,7 +340,6 @@ def tag_tensorproxy_output_as_detached(proxy):
                 return proxy
 
             result = tree_map(tag_tensorproxy_output_as_detached, result)
-
```
shino16, line 343: If you're not particular, I'd prefer keeping this empty line.

thunder/core/transform_common.py
```diff
@@ -182,19 +203,24 @@ def dce(trace: Trace, needed_proxies: None | set[Variable] = None) -> Trace:
...
     # duplicate number proxies happen with the symbolic shapes and are
     # not covered by the above (due to being in tuples?).
     dced_bound_symbols = remove_duplicate_number_proxies(dced_bound_symbols)
-    dcetrace.bound_symbols = dced_bound_symbols
 
+    if isinstance(trace_or_bsyms, Trace):
+        result = from_trace(trace_or_bsyms)
+        result.bound_symbols = dced_bound_symbols
+    else:
+        result = dced_bound_symbols
```
shino16: This is nit-picky, but returning early for bsym lists will make the diff cleaner.
```suggestion
    if isinstance(trace_or_bsyms, list):
        return dced_bound_symbols

    dcetrace = from_trace(trace_or_bsyms)
```

thunder/tests/test_core.py
```diff
@@ -3295,22 +3325,28 @@ def clean(tr):
 
 
 def test_prims_pack_list():
```
shino16, line 3327: Wondering why this test is changing. Did this PR make this test fail?

thunder/core/transform_common.py
```diff
@@ -142,20 +142,41 @@ def keep_or_swap(p):
 #   that only produce non-proxy objects
 # NOTE needed_proxies is an in/out argument, it takes an initial set of Variables you want to keep, and return
 #   all the needed proxies of the input trace
-def dce(trace: Trace, needed_proxies: None | set[Variable] = None) -> Trace:
+def dce(
+    trace_or_bsyms: Trace | list[BoundSymbolInterface],
+    needed_proxies: None | set[Variable] = None,
+    output: Any = None,
+) -> Trace | list[BoundSymbolInterface]:
```
shino16: It would be nice if we could raise an error when a list of bsyms is given and `output` is not given, but it would just return an empty list. But `None` is a valid value for `bsym`'s output... I also find the difference in the meaning of `output=None` confusing when a trace is given and when a list of bsyms is given.

What do you think about making a separate function for a list of bsyms? It could be an intermediate function that calls the original `dce`, or the other way around.

thunder/tests/test_core.py
beverlylytle: Yes, the test was constructing a trace in an unnatural way, leading to Tensors, and not TensorProxies, being fed through the machinery.  dce'ing the subsymbols used tree_flatten, which doesn't support Tensors.  So I rewrote the test to use TensorProxies instead.

thunder/core/transform_common.py
beverlylytle: I had been following the pattern established by producers and consumers which ingests both traces and bsyms, but it seems that's too confusing when there's other input.

beverlylytle:
@KaelanDt This is ready for your review. Thanks!

beverlylytle:
Waiting on https://github.com/Lightning-AI/lightning-thunder/pull/2791 to be merged to fix the CI failures

PR #2666, opened 2025-10-16
Remove `--dtensor-single-gpu`

wujingyue:
Use dist.is_torchelastic_launched instead to tell whether the program
was launched with torchrun


thunder/benchmarks/benchmark_inference.py
kshitij12345, line 666: We should also remove this.

wujingyue: Which line? Or do you really mean the entire file?

kshitij12345: Sorry, the comment should have referred to this line - https://github.com/Lightning-AI/lightning-thunder/blob/c21bc52a0a7c2f4eb9258c29c6a0509fadf34bde/thunder/benchmarks/benchmark_inference.py#L661-L665

PR #2673, opened 2025-10-17
Remove stray print

kshitij12345:


t-vi (approved):
Thank you @kshitij12345

PR #2675, opened 2025-10-19
Record last_trace_cache_stop when miss the cache

wujingyue:
Currently, we record last_trace_cache_start **regardless of cache hit/miss** but record last_trace_cache_stop **only when hit**. This sounds like a bug.


wujingyue:
The CI error `FAILED thunder/tests/test_ops.py::test_core_vs_torch_consistency_instance_norm_torch_cuda_thunder.dtypes.float32 - AssertionError: Tensor-likes are not close!` is unrelated.

PR #2694, opened 2025-10-24
Fix nvfuser warning

kshitij12345:
```python
import thunder
import torch

def fn(x):
    return x.sin().cos()

x = torch.randn(32, 32, device="cuda")

tjit = thunder.jit(fn)
tjit(x)
```

Before PR, user will see `UserWarning: Be careful! You've imported nvfuser when the nvfuser_direct module is already imported.`


thunder/executors/nvfuserex_impl.py
```diff
@@ -483,8 +490,6 @@ def compute_contiguity(
     Returns:
         Tuple[Tuple[bool, ...], Tuple[int, ...]]: The contiguity and stride_order
     """
-    from nvfuser import compute_tensor_descriptor as nv_compute_td
```
jjsjann123, line 486: 👏

PR #2696, opened 2025-10-24
benchmark_inference: set requires_grad=False on params

kshitij12345:
While adding support for DTensor, settings `requires_grad=False` was moved and not applied on the single-GPU path.


wujingyue (approved):

KaelanDt (approved):
thank you @kshitij12345

PR #2697, opened 2025-10-27
benchmark_inference: Add CLI option to enable thunder CUDAGraph Transform

kshitij12345:
Command
```
python thunder/benchmarks/benchmark_inference.py --input-length 32 --output-length 3 --mode thunder --num-iterations 10 --enable-thunder-cudagraph
```

NOTE: Need to revert 13f7171784d6a953fd02879f325a7facfe124d0d

Running the above command leads to the following error (seems to fail during FusionDefinition execution)
```
  File "/opt/pytorch/lightning-thunder/thunder/transforms/cudagraph.py", line 130, in build_cuda_graph
    static_outputs = fn(*static_inputs)
                     ^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/torch/utils/_contextlib.py", line 121, in decorate_context
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/opt/pytorch/lightning-thunder/thunder/executors/torchex.py", line 169, in no_autocast_fn
    return fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^
  File "thunder.CUDAGraph5_39", line 116, in CUDAGraph5
  File "/opt/pytorch/lightning-thunder/thunder/executors/nvfuserex_impl.py", line 566, in __call__
    return fd.execute(
           ^^^^^^^^^^^
  File "/opt/pytorch/nvfuser/python/nvfuser_direct/__init__.py", line 318, in execute
    return self.fec.execute(
           ^^^^^^^^^^^^^^^^^
RuntimeError: 
Error from segmentation group 3: CUDA error: operation not permitted when stream is capturing
Search for `cudaErrorStreamCaptureUnsupported' in https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__TYPES.html for more information.
CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.
For debugging consider passing CUDA_LAUNCH_BLOCKING=1
Compile with `TORCH_USE_CUDA_DSA` to enable device-side assertions.

Exception raised from memcpy_and_sync at /opt/pytorch/pytorch/c10/cuda/CUDAFunctions.h:106 (most recent call first):
frame #0: c10::Error::Error(c10::SourceLocation, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >) + 0x88 (0x74e88646d008 in /usr/local/lib/python3.12/dist-packages/torch/lib/libc10.so)
frame #1: <unknown function> + 0x5cb0a (0x74e8bc9b2b0a in /usr/local/lib/python3.12/dist-packages/torch/lib/libc10_cuda.so)
frame #2: c10::cuda::c10_cuda_check_implementation(int, char const*, char const*, unsigned int, bool) + 0x1c8 (0x74e8bc9b27c8 in /usr/local/lib/python3.12/dist-packages/torch/lib/libc10_cuda.so)
```


wujingyue:
cc @mdavis36

thunder/benchmarks/benchmark_inference.py
```diff
@@ -291,6 +292,10 @@ def _thunder_jit_options(self) -> dict[str, Any]:
                 self._mask_transform = SDPAMaskTransform()
             res["transforms"] = [self._mask_transform]
             res["executors"] = [self._mask_transform.get_executor(), *thunder.get_default_executors()]
+        if self.config.enable_thunder_cudagraph:
+            from thunder.transforms.cudagraph import CUDAGraphTransform
```
wujingyue: https://peps.python.org/pep-0008/#imports recommends putting `import`s at the top of a file. I'd do that by default unless you see a clear reason otherwise. 

Advantages:
- Clear dependencies (easy to see what modules are used)
- Avoids redundant imports
- Modules are loaded once, at startup, improving performance for repeated calls
- Makes circular import issues easier to detect

thunder/benchmarks/benchmark_inference.py
```diff
@@ -676,6 +681,7 @@ def parse_args() -> argparse.Namespace:
 
     parser.add_argument("--save-results", action="store_true", help="Save results to JSON file")
     parser.add_argument("--output-dir", type=str, default="./results", help="Directory to save results")
+    parser.add_argument("--enable-thunder-cudagraph", action="store_true", help="Pass CUDAGraphTransform to Thunder")
```
wujingyue, line 688: Non-blocking question: should this also affect torch.compile? I understand your code doesn't and how you name the flag indicates you didn't intend to. But it might be better so we compare apples to apples.

thunder/benchmarks/benchmark_inference.py
kshitij12345: Good point, updated.

thunder/benchmarks/benchmark_inference.py
kshitij12345: We already use `mode="reduce-overhead"` (i.e. using CUDA grpahs) for inductor in the benchmark script.

https://github.com/Lightning-AI/lightning-thunder/blob/b3a94d1440d72b47d8ade85ad306c57ab1d1549b/thunder/benchmarks/benchmark_inference.py#L300-L301

These are all the options inductor currently supports

```
Can be either “default”, “reduce-overhead”, “max-autotune” or “max-autotune-no-cudagraphs”

* ”default” is the default mode, which is a good balance between performance and overhead

* ”reduce-overhead” is a mode that reduces the overhead of python with CUDA graphs, useful for small batches. Reduction of overhead can come at the cost of more memory usage, as we will cache the workspace memory required for the invocation so that we do not have to reallocate it on subsequent runs. Reduction of overhead is not guaranteed to work; today, we only reduce overhead for CUDA only graphs which do not mutate inputs. There are other circumstances where CUDA graphs are not applicable; use TORCH_LOG=perf_hints to debug.

* ”max-autotune” is a mode that leverages Triton or template based matrix multiplications on supported devices and Triton based convolutions on GPU. It enables CUDA graphs by default on GPU.

* ”max-autotune-no-cudagraphs” is a mode similar to “max-autotune” but without CUDA graphs

To see the exact configs that each mode sets you can call torch._inductor.list_mode_options()
```

In a follow-up, we should enable specifying other modes.

Also, I am curious to check the perf for both thunder and torch.compile with default modes. Wdyt?

Docs: https://docs.pytorch.org/docs/stable/generated/torch.compile.html

mattteochen: From some investigations, I have found the comparisons between [thunder_no_cuda_graphs, torch.compile_default] vs [thunder_cuda_graphs, torch.compile_reduce_overhead] to be very different (It was on llama3 though). Tuning the inductor path is necessary to make a fair comparison imo (in a follow-up).

tbqh:
Seems to be broken with different parameters:
`python thunder/benchmarks/benchmark_inference.py --input-length 4096 --output-length 4 --mode thunder --enable-nv-linear --warmup-iterations 2 --num-iterations 2 --enable-thunder-cudagraph`

Causes:
```Traceback (most recent call last):
  File "/opt/pytorch/nvfuser/lightning-thunder/thunder/benchmarks/benchmark_inference.py", line 737, in <module>
    main()
  File "/opt/pytorch/nvfuser/lightning-thunder/thunder/benchmarks/benchmark_inference.py", line 719, in main
    benchmark = InferenceBenchmark(config)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/pytorch/nvfuser/lightning-thunder/thunder/benchmarks/benchmark_inference.py", line 279, in __init__
    self.model = self._compile_model(model)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/pytorch/nvfuser/lightning-thunder/thunder/benchmarks/benchmark_inference.py", line 308, in _compile_model
    return thunderfx(model, **self._thunder_jit_options)
                              ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/pytorch/nvfuser/lightning-thunder/thunder/benchmarks/benchmark_inference.py", line 298, in _thunder_jit_options
    res["transforms"].append(CUDAGraphTransform())
    ~~~^^^^^^^^^^^^^^
KeyError: 'transforms'
```


Edit:
Looks like it's just the `--enable-nv-linear` flag. Is this expected to not be compatible with `--enable-thunder-cudagraph`?

kshitij12345:
@tbqh Thanks for reporting, have pushed a patch to fix the issue.

mattteochen:
Just a FYI for anyone testing this, the transform + https://github.com/NVIDIA/Fuser/issues/5434#issuecomment-3454322149 are expected to work on blackwell

kshitij12345:
Ping @KaelanDt for review

tbqh:
Thanks for the `--enable-nv-linear` fix, this PR is working well.

Before:
```
Time Between Output Tokens (TBOT): 7.25 ms
Prefill Time: 15.68 ms
Decode Time: 7.25 ms
```

After:
```
Time Between Output Tokens (TBOT): 4.47 ms
Prefill Time: 15.27 ms
Decode Time: 4.47 ms
```

PR #2698, opened 2025-10-27
tests: Re-enable zero-size shapes in elementwise unary opinfo

aobolensk:
Enable back zero-size test cases that are actually passing

### DIFF
--- a/.github/workflows/docs-build.yml
+++ b/.github/workflows/docs-build.yml
@@ -26,6 +26,7 @@ jobs:
     env:
       ARTIFACT_DAYS: 0
       PYPI_LOCAL_DIR: "pypi_pkgs/"
+      TORCH_URL: "https://download.pytorch.org/whl/cpu"
     steps:
       - uses: actions/checkout@v5
       - uses: actions/setup-python@v6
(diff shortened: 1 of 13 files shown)


IvanYashchuk:
@KaelanDt, I don't have the Lightning account to view CI logs. Could you please review the changes and CI failures if they are related?

aobolensk:
> Thank you for enabling this again!
> 
> There seems to be some related failures in the logs:
> 
> ```shell
> thunder/tests/distributed/test_dtensor.py::DTensorTest::test_dtensor_opinfo_neg_nvfuser
> thunder/tests/distributed/test_dtensor.py::DTensorTest::test_dtensor_opinfo_silu_nvfuser
> thunder/tests/distributed/test_dtensor.py::DTensorTest::test_dtensor_opinfo_reciprocal_nvfuser
> ```
> 
> Would you be able to investigate? If that is not possible, could you add a skip to those nvFuser tests and open a new issue?

@riccardofelluga  Ok, that explains better what is going on, thanks for sharing! So it seems 0-shape tensors are not really working well on nvFuser backend, but as for other backends it is fine. Unfortunately, I'm not ready to fix nvFuser backend specific stuff (no HW available). I was actually observing that everything is good on CPU.

Does it make sense for you to enable them globally, but skip again only for nvFuser?

In this case, we will increase code coverage on some targets at least.

riccardofelluga:
> @aobolensk  Ok, that explains better what is going on, thanks for sharing! So it seems 0-shape tensors are not really working well on nvFuser backend, but as for other backends it is fine. Unfortunately, I'm not ready to fix nvFuser backend specific stuff (no HW available). I was actually observing that everything is good on CPU.
> 
> Does it make sense for you to enable them globally, but skip again only for nvFuser?
> 
> In this case, we will increase code coverage on some targets at least.

Thanks for checking! Yess it does totally make sense to skip nvFuser ones but let's also open an issue for it so that we can track it in the future

PR #2709, opened 2025-11-01
tests: Fix variable reference in mul_to_multimul_checker

aobolensk:
Incorrect value is mistakenly being checked in `mul_to_multimul_checker` specifically for CPU case


IvanYashchuk:
@KaelanDt, could you please take a look and help merge?

aobolensk:
> Do you think it would be useful to add an assert to check that multimul is indeed not used when one or more tensors are not on CPU?

That's possible, but isn't that a purpose of this specific checker (`mul_to_multimul_checker`) to guardrail that? :)

PR #2719, opened 2025-11-05
Add cutlass python dsl executor for `quack-kernels`

crcrpar:
## What does this PR do?

As per title, this adds cutlass python dsl executor.
In this PR, the kernels defined in https://github.com/Dao-AILab/quack, except matmul, are registered. Also, backward is not integrated.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -6,12 +6,12 @@
 # Thank you, our previous code owners for their service:
 #   @carmocca @borda
 
-* @mruberry @lantiga @t-vi @KaelanDt
+* @mruberry @lantiga @KaelanDt
 
 # CI/CD and configs
-/.azure/        @KaelanDt @lantiga @t-vi
-/.github/       @KaelanDt @lantiga @t-vi
-/.lightning/    @KaelanDt @lantiga @t-vi
-/dockers/       @KaelanDt @lantiga @t-vi
-Makefile        @KaelanDt @lantiga @t-vi
-*.yml           @KaelanDt @lantiga @t-vi
+/.azure/        @KaelanDt @lantiga
+/.github/       @KaelanDt @lantiga
+/.lightning/    @KaelanDt @lantiga
+/dockers/       @KaelanDt @lantiga
+Makefile        @KaelanDt @lantiga
+*.yml           @KaelanDt @lantiga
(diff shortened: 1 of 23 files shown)


thunder/executors/cutlass_dsl_ex.py
```diff
@@ -0,0 +1,559 @@
...
+        normalized_shape: Sequence[int],
+        weight: TensorProxy | None = None,
+        bias: TensorProxy | None = None,
+        eps: Number = 1e-5,
+    ) -> bool:
+        if (
+            a.dtype not in {dtypes.float16, dtypes.bfloat16, dtypes.float32}
+            or weight.ndim != 1
+            or a.shape[-1] != weight.shape[0]
+            or weight.dtype not in {dtypes.float32}
```
riccardofelluga, line 378: Can weight be None? In that case this would need to check before trying to access `.ndim`

thunder/executors/cutlass_dsl_ex.py
```diff
@@ -0,0 +1,559 @@
...
+
+
+# NOTE: This constraint comes from https://github.com/Dao-AILab/quack/blob/59631e98/quack/reduction_base.py#L35-L38
+def is_last_dim_divisible(dtype: dtypes.dtype, last_dim_size: int) -> bool:
+    return last_dim_size % (128 // 8 // dtype.bytes) == 0
+
+
+quack_version: LooseVersion
+try:
+    import quack
```
kiya00: do we need to add this into requirements to install it?

thunder/tests/test_cutlass_dsl_ex.py
```diff
@@ -0,0 +1,103 @@
...
+    ref_x = x.clone().detach()
+    if dtype != torch.float32:
+        ref_x = ref_x.to(torch.float32)
+    targets = torch.randint(0, 128, (128,), dtype=torch.int64)
+
+    jitted = jit_with_cutlass_dsl_ex(F.cross_entropy)
+
+    expected = F.cross_entropy(ref_x, targets, reduction="none")
+    actual = jitted(x, targets, reduction="none")
+    torch.testing.assert_close(expected, actual)
```
kiya00, line 59: It seems the backward is not tested

thunder/executors/cutlass_dsl_ex.py
crcrpar: I'd not think we should do so. Because `pip install quack-kernels` seems to install cuda python packages such as `nvidia-cutlass-dsl` and I don't know how to having requirements.txt install cuda python packages that respect users local environments

thunder/executors/cutlass_dsl_ex.py
crcrpar: good catch. will check it

thunder/tests/test_cutlass_dsl_ex.py
crcrpar: I've not managed to have backward work

PR #2724, opened 2025-11-11
DTensor: skip nvfuser test if NVFUSER_DISABLE=multidevice is set

kshitij12345:
In some environment, the nvfuser multidevice could be explicitly disabled.

### DIFF
--- a/thunder/tests/distributed/test_moe.py
+++ b/thunder/tests/distributed/test_moe.py
@@ -146,7 +146,7 @@ def parallelize_moe_model(model: llama4_moe.Llama4MoE, device_mesh: torch.distri
 class TestLlama4MoEDistributed(DistributedParallelTestCase):
     def test_llama4_moe_distributed(self):
         if "multidevice" in os.environ.get("NVFUSER_DISABLE", ""):
-            raise unittest.SkipTest("test_dtensor_grouped_mm: nvfuser multidevice is disabled")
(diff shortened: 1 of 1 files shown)


thunder/tests/distributed/test_moe.py
```diff
@@ -143,6 +145,9 @@ def parallelize_moe_model(model: llama4_moe.Llama4MoE, device_mesh: torch.distri
 
 class TestLlama4MoEDistributed(DistributedParallelTestCase):
     def test_llama4_moe_distributed(self):
+        if "multidevice" in os.environ.get("NVFUSER_DISABLE", ""):
+            raise unittest.SkipTest("test_dtensor_grouped_mm: nvfuser multidevice is disabled")
```
beverlylytle:
```suggestion
            raise unittest.SkipTest("test_llama4_moe_distributed: nvfuser multidevice is disabled")
```

kshitij12345:
> can you explain to me why not all of the distributed tests require this escape route?

Other nvFuser tests do not involve communication (eg. for pointwise ops) in which case they still pass (skipping them should also be ok) . These 2 are the tests which fail internally where we have `NVFUSER_DISABLE=multidevice` set.

kshitij12345:
Ping @KaelanDt for review

KaelanDt (approved):
thank you @kshitij12345

PR #2734, opened 2025-11-13
Fix AttributeError when cache position logic queries device of integer start_pos in benchmark_inference.py

IvanYashchuk:
This PR fixes an error that occurs when querying the `device` attribute from `start_pos`, which is a Python integer rather than a tensor. The issue was discovered while running the inference benchmark script:
```
python thunder/benchmarks/benchmark_inference.py --model-name meta-llama/Llama-4-Maverick-17B-128E --mode eager --input-length 1024 --output-length 32 --batch-size 1 --num-iterations 20 --num-layers 2
```
The following traceback highlights the error:
```py
  File "/home/iyashchuk/dev/lightning-thunder/thunder/benchmarks/benchmark_inference.py", line 413, in generate
    first_token = self.prefill(input_ids, past_key_values)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/iyashchuk/dev/lightning-thunder/thunder/benchmarks/benchmark_inference.py", line 386, in prefill
    return self.get_next_token(input_ids, past_key_values)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/iyashchuk/dev/lightning-thunder/thunder/benchmarks/benchmark_inference.py", line 369, in get_next_token
    cache_position = start_pos + torch.arange(0, input_ids.shape[1], device=start_pos.device, dtype=start_pos.dtype)
                                                                            ^^^^^^^^^^^^^^^^
AttributeError: 'int' object has no attribute 'device'
```

The problematic cache position construction was introduced in https://github.com/Lightning-AI/lightning-thunder/commit/72a9d866b38e8ec0c3a29b1e6544c4787754236c.

Explicit construction of cache_position in this way is not necessary, because the model code already correctly generates it as needed (see [llama4 transformers implementation](https://github.com/huggingface/transformers/blob/8cb5963cc22174954e7dca2c0a3320b7dc2f4edc/src/transformers/models/llama4/modeling_llama4.py#L516-L520)).

cc @crcrpar

### DIFF
--- a/thunder/__init__.py
+++ b/thunder/__init__.py
@@ -729,6 +729,7 @@ def jit(
 
                     # Updates cache statistics
                     cs.cache_hits += 1
+                    cs.last_trace_cache_stop = time.perf_counter_ns()
                     cs.last_traces = comp_traces
                     cs.last_interpreted_instructions = None
                     cs.last_interpreter_log = None
@@ -759,6 +760,7 @@ def jit(
 
                 # Updates cache statistics
                 cs.cache_hits += 1
+                cs.last_trace_cache_stop = time.perf_counter_ns()
                 cs.last_traces = comp_traces
                 cs.last_interpreted_instructions = None
                 cs.last_interpreter_log = None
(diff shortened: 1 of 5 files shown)


IvanYashchuk:
@kshitij12345, @riccardofelluga could you please review the change?

kshitij12345 (approved):
LGTM, thanks!

KaelanDt (approved):
thank you @IvanYashchuk

PR #2744, opened 2025-11-17
replace 'cache_mode' option in tests with 'cache'

beverlylytle:
I noticed that a few tests were using the compile option "cache_mode" instead of the expected "cache".  This unexpected compile option was silently ignored.  Luckily, the tests were mostly wanting to use the default caching strategy, so nothing lost.  However, one test was attempting to use the caching strategy "same values", which doesn't exist.  This PR switches it to "NO_CACHING".

### DIFF
--- a/thunder/tests/test_grad.py
+++ b/thunder/tests/test_grad.py
@@ -1336,11 +1336,11 @@ def test_torch_autograd_module(executor, device, _):
     a = make_tensor((2, 3), device=device, dtype=torch.float32, requires_grad=True)
     g = make_tensor((2, 4), device=device, dtype=torch.float32)
 
-    for cache_mode in ("constant values", "same input"):
+    for cache_mode in ("constant values", "no caching"):
         lc = executor.make_callable(
             linear,
             disable_torch_autograd=False,
-            cache_mode=cache_mode,
+            cache=cache_mode,
         )
         lc.zero_grad()
         a.grad = None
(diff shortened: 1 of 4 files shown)


thunder/tests/test_grad.py
```diff
@@ -1336,11 +1336,11 @@ def test_torch_autograd_module(executor, device, _):
     a = make_tensor((2, 3), device=device, dtype=torch.float32, requires_grad=True)
     g = make_tensor((2, 4), device=device, dtype=torch.float32)
 
-    for cache_mode in ("constant values", "same input"):
+    for cache_mode in ("constant values", "no caching"):
```
kiya00, line 1339: seems there's a "same input" option?
https://github.com/Lightning-AI/lightning-thunder/blob/93aa254082162605f724c3be7451b641b6669eba/thunder/core/options.py#L55-L60

KaelanDt (dismissed):
thank you @beverlylytle. let's keep `same input` in `test_grad.py` though

thunder/tests/test_grad.py
beverlylytle: According to [this line](https://github.com/Lightning-AI/lightning-thunder/blob/93aa254082162605f724c3be7451b641b6669eba/thunder/core/jit_ext.py#L2138), it's not supported....

beverlylytle:
No.  Would you expect something to change?

PR #2716, opened 2025-11-04
Support for `polar`, `view_as_complex`, `view_as_real`

mattteochen:
This PR implements the operatios in the title; needed, for example in [`llama4`](https://github.com/huggingface/transformers/blob/f4fc42216cd56ab6b68270bf80d811614d8d59e4/src/transformers/models/llama4/modeling_llama4.py) RoPE stages. 

Closes https://github.com/Lightning-AI/lightning-thunder/issues/2681

Issue #2677, opened 2025-10-20
Symbolic Values support for benchmark_inference.py


kshitij12345:
While running `python thunder/benchmarks/benchmark_inference.py --input-length 32 --output-length 3 --mode thunder --num-iterations 10` , one of the FXGraph we receive is the following.

FXGraph

<details>

```python
class GraphModule(torch.nn.Module):
    def forward(self, view_as_real_1: "f32[1, 1, 8, 64, 2]", xq_out: "f32[1, 1, 40, 128]", query_states: "bf16[1, 1, 40, 128]", key_states: "bf16[1, 1, 8, 128]", value_states: "bf16[1, 8, 1, 128]", l_past_key_values_cumulative_length_0_: "Sym(s78)", k_out: "bf16[1, 8, 35, 128]", l_cache_position_: "i64[1]", v_out: "bf16[1, 8, 35, 128]", l_stack0_0_: "bf16[1, 1, 1, 8192]", l_self_modules_layers_modules_0_modules_self_attn_modules_o_proj_parameters_weight_: "bf16[5120, 5120]", l_inputs_embeds_: "bf16[1, 1, 5120]", l_self_modules_layers_modules_0_modules_post_attention_layernorm_parameters_weight_: "bf16[5120]", l_self_modules_layers_modules_0_modules_feed_forward_modules_gate_proj_parameters_weight_: "bf16[16384, 5120]", l_self_modules_layers_modules_0_modules_feed_forward_modules_up_proj_parameters_weight_: "bf16[16384, 5120]", l_self_modules_layers_modules_0_modules_feed_forward_modules_down_proj_parameters_weight_: "bf16[5120, 16384]", l_self_modules_layers_modules_1_modules_input_layernorm_parameters_weight_: "bf16[5120]", l_self_modules_layers_modules_1_modules_self_attn_modules_q_proj_parameters_weight_: "bf16[5120, 5120]", l_self_modules_layers_modules_1_modules_self_attn_modules_k_proj_parameters_weight_: "bf16[1024, 5120]", l_self_modules_layers_modules_1_modules_self_attn_modules_v_proj_parameters_weight_: "bf16[1024, 5120]", freqs_cis_1: "c64[1, 1, 64]"):
[...]

PR #2745, opened 2025-11-17
Fix `benchmark_inference.py` segfault and superficial errors with multiple devices

shino16:
- Makes `_call_dispatch` use `__get__(slf, type(slf))` instead of `__get__(slf)`, working around a CPython segfault ([python/cpython#132747](https://github.com/python/cpython/issues/132747)). From Python 3.13.3 this workaround will no longer be necessary, but I think we can keep it anyway.
- Creates `TensorProxy.shape`'s provenance only when the tensor's history exists, so that incomplete provenance record will not be created. This fixes the `AssertionError` in mentioned in [#2677 (comment)](https://github.com/Lightning-AI/lightning-thunder/issues/2677#issuecomment-3542984406).
- Makes `thunder/core/update_aliases.py` skip numel checks (introduced in [#2716](https://github.com/Lightning-AI/lightning-thunder/pull/2716)) when relevant shapes are dynamic.

Multi-device support of `benchmark_inference.py` is still incomplete. Refer to #2677 for progress.

### DIFF
--- a/thunder/tests/distributed/test_ddp.py
+++ b/thunder/tests/distributed/test_ddp.py
@@ -165,7 +165,7 @@ class DDPTest(DistributedParallelTestCase):
             m = ToyModel().to(device)
             jitted_m = thunder.jit(
                 m,
-                cache_mode=CACHE_OPTIONS.CONSTANT_VALUES,
+                cache=CACHE_OPTIONS.CONSTANT_VALUES,
                 executors=executors_map[executor].executors_list(),
             )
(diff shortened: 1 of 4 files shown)


beverlylytle:
Solving #2677 completely is going to blow this PR up into a monster, touching lots of files.  I think it would be better to fix each issue individually. It will make it easier to choose targeted reviewers for each change.  This will also potentially save time later, if an issue pops up and bisection is used to find the culprit commit.  So, let's just start with the seg fault fix.

shino16:
Your suggestion makes sense, I'll submit separate small PRs. And thank you for your reviews in #2746! @beverlylytle

PR #2646, opened 2025-10-14
Relax Test Tolerances for TE executor tests

riccardofelluga:
## Summary

This PR introduces dtype-specific numerical tolerances for Transformer Engine (TE) tests to improve test stability and reduce false positives due to numerical precision differences across different floating-point data types.

## Key Changes

- **Added `te_assert_close` function**: A new assertion wrapper that automatically applies appropriate tolerances based on tensor dtype and scalar inputs
- **Dtype-specific tolerances**: Implements tolerances aligned with [Transformer Engine's numerical testing specification](https://github.com/NVIDIA/TransformerEngine/blob/7ad130efd52c3aa4a386d25f1d42b28d5aa20090/tests/pytorch/test_numerics.py#L155-L167):
  - `float32`: `rtol=1.3e-6, atol=1e-5`  
  - `float16`: `rtol=1e-3, atol=1e-5`
  - `bfloat16`: `rtol=1.6e-2, atol=1e-5`
  - `float` scalars: `rtol=1.3e-6, atol=1e-5`

Issue #2444, opened 2025-08-20
[TE] Flaky test test_te_activation_checkpointing_correctness[jit-delayed_scaling]


kshitij12345:
CI - https://github.com/Lightning-AI/lightning-thunder/actions/runs/17099686020/job/48492598499#step:5:377

kshitij12345:
The failure can be repro-ed by running the test multiple times  (requires `pip install pytest-repeat`).

Also, we need to reset the TE FP8GlobalState if there is an exception, so need to apply the following patch

Patch

<details>

```patch
diff --git a/thunder/tests/test_transformer_engine_v2_executor.py b/thunder/tests/test_transformer_engine_v2_executor.py
index 3ff4045e..f3b0ed4c 100644
--- a/thunder/tests/test_transformer_engine_v2_executor.py
+++ b/thunder/tests/test_transformer_engine_v2_executor.py
@@ -501,10 +501,27 @@ def test_te_activation_checkpointing_trace(fp8_recipe: recipe.Recipe, compile_pa
         assert len(saved_tensors - {"l_x_", "l_w_"}) == 2
 
 
+def reset_te_state(fn):
+    # Reset the state if there is an Exception while executing the test
+    # otherwise the following runs will also fail.
+    from functools import wraps
+
+    @wraps(fn)
+    def wrapper(*args, **kwargs):
+        try:
+            fn(*args, **kwargs)
+        finally:
+            from transformer_engine.pytorch.fp8 import FP8GlobalStateManager
+            FP8GlobalStateManager.reset()
+
+    return wrapper
+
+
 @requiresCUDA
 @pytest.mark.parametrize("fp8_recipe", recipes, ids=recipe_ids)
 @pytest.mark.parametrize("compile_path", ["jit", "ThunderFX"])
 @pytest.mark.filterwarnings("ignore::FutureWarning")  # Coming from TE v2.3
+@reset_te_state
 def test_te_activation_checkpointing_correctness(fp8_recipe: recipe.Recipe, compile_path: str):
     if not fp8_recipe:
         pytest.skip(
@@ -642,6 +659,3 @@ def test_te_activation_checkpointing_correctness(fp8_recipe: recipe.Recipe, comp
         # check that amax history is the same as TE
         for te_amax, th_amax in zip(te_amax_hist, th_amax_hist):
             assert_close(te_amax[:, :-1], th_amax)
-
-        # Reset state manager before next run.
-        FP8GlobalStateManager.reset()
[...]

Issue #2252, opened 2025-06-18
TE: Flaky Test - test_te_linear_forward_backward_multiple_iteration[delayed_scaling : AssertionError: Tensor-likes are not close! 


kshitij12345:
Using pjnl-20250610 on RTX6000

```bash
# Install - pip install pytest-repeat
pytest thunder/tests/test_transformer_engine_v2_executor.py -k test_te_linear_forward_backward_multiple_iteration[delayed --count 50
```

Output
```
================================================================================================================== short test summary info ===================================================================================================================
FAILED thunder/tests/test_transformer_engine_v2_executor.py::test_te_linear_forward_backward_multiple_iteration[delayed_scaling-12-50] - AssertionError: Tensor-likes are not close!
FAILED thunder/tests/test_transformer_engine_v2_executor.py::test_te_linear_forward_backward_multiple_iteration[delayed_scaling-15-50] - AssertionError: Tensor-likes are not close!
FAILED thunder/tests/test_transformer_engine_v2_executor.py::test_te_linear_forward_backward_multiple_iteration[delayed_scaling-21-50] - AssertionError: Tensor-likes are not close!
================================================================================================= 3 failed, 47 passed, 700 deselected, 301 warnings in 4.63s =================================================================================================
```

Sample error message
```
_________________________________________________________________________________________ test_te_linear_forward_backward_multiple_iteration[delayed_scaling-21-50] __________________________________________________________________________________________

fp8_recipe = recipe_type=DelayedScaling, margin=0, format=HYBRID, amax_history_len=1024, fp8_dpa=False, fp8_mha=False

    @requiresCUDA
    @pytest.mark.parametrize("fp8_recipe", recipes, ids=recipe_ids)
    def test_te_linear_forward_backward_multiple_iteration(fp8_recipe: recipe.Recipe):
        if not fp8_recipe:
            pytest.skip(
                "When recipe is None a new recipe is created for each iteration. This makes the results not numerically comparable."
            )
    
        if fp8_recipe and not (fp8_recipe.delayed() or is_mxfp8_supported):
            pytest.skip(msg_mxfp8)
    
        # Test Description:
        # In this test, we verify whether a model using TransformerEngine Linear
        # and transformer_engine executor converge to same state.
        # Since, the FP8 operations are stateful, we want to verify that
        # our output matches over multiple iterations (where state handling comes into picture)
        dtype = torch.bfloat16
        device = "cuda"
[...]

PR #2748, opened 2025-11-18
Fix for flaky TE tests

riccardofelluga:
## What does this PR do?

After better investigating TE test failures from #2444 and #2252 I've concluded that there are a couple of ways to resolve this. When using delayed scaling recipe, the numerical difference stems from Thunder not using fused amax update, therefore the solution would be to implement that, however, the flakiness does not only limit to that recipe so the best course of action here in my mind is to use the same tolerances found in TE test suite here:

https://github.com/NVIDIA/TransformerEngine/blob/7e593c3be96b3eebc384da1a2ab307727065c9ab/tests/pytorch/utils.py#L68-L118

This PR fixes both #2444 and #2252.

I've made a previous PR relaxing the tolerances but I think I made the mistake of taking the wrong TE specs earlier here #2646

cc. @kshitij12345

### DIFF
--- a/thunder/core/proxies.py
+++ b/thunder/core/proxies.py
@@ -2009,7 +2009,8 @@ def tensorproxy(t: torch.Tensor, /, *, name: None | str, history: None | tuple =
     # See Note [DistributedDataParallel and distparallel_type]
     distparallel_type = getattr(t, "distparallel_type", None)
     _thunder_fsdp_padding_size = getattr(t, "_thunder_fsdp_padding_size", None)
-    if using_symbolic_values():
+    # For parameters, shapes should be static.
+    if using_symbolic_values() and not isinstance(t, torch.nn.Parameter):
         shape_attr = ProvenanceRecord(PseudoInst.LOAD_ATTR, inputs=[copy.copy(history), wrap_const("shape").provenance])
         shape = tuple(
             IntegerProxy(
--- a/thunder/tests/test_jit_general.py
+++ b/thunder/tests/test_jit_general.py
@@ -1581,6 +1581,24 @@ def test_cache_symbolic_values_dict():
     assert_close(actual, expected)
 
 
+def test_cache_symbolic_values_nn_parameter_static_shape():
+    linear = torch.nn.Linear(2, 2)
+    x = torch.randn(2, 2)
+
+    jlinear = thunder_jit(linear, cache="symbolic values")
+
+    jlinear(x)
+    exec_trc = thunder.last_traces(jlinear)[-1]
+    for bsym in exec_trc.bound_symbols:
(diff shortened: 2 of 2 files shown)


thunder/tests/test_transformer_engine_executor.py
```diff
@@ -53,21 +52,16 @@
...
-        tolerances = dict(rtol=1.6e-2, atol=1e-5)
-
-    kwargs.update(tolerances)
+# Returns the estimated numerical error for a given TE recipe based on:
+# https://github.com/NVIDIA/TransformerEngine/blob/7e593c3be96b3eebc384da1a2ab307727065c9ab/tests/pytorch/utils.py#L68-L118
+def te_assert_close(actual, expected, recipe=None, **kwargs):
+    if recipe is not None:
+        if is_nvfp4_available and recipe.nvfp4():
+            # Tolerances for NVFP4
+            return torch.testing.assert_close(actual, expected, rtol=0.25, atol=0.125)
```
beverlylytle: Maybe it doesn't matter, but do we want to ignore the input kwargs here?  Would it be worthwhile to mimic the pattern set in the deleted code, something like:
```
if recipe is not None:
    if ... and ...:
        tolerances = {'rtol':0.25, 'atol':0.125}
    else:
        tolerances = {...}
    kwargs.update(tolerances)
```

thunder/tests/test_transformer_engine_executor.py
```diff
@@ -53,21 +52,16 @@
...
+        else:
+            # Tolerances for all FP8 recipes
+            return torch.testing.assert_close(actual, expected, rtol=0.125, atol=0.0675)
```
beverlylytle: What about [e5m2](https://github.com/NVIDIA/TransformerEngine/blob/7e593c3be96b3eebc384da1a2ab307727065c9ab/tests/pytorch/utils.py#L100)?

thunder/tests/test_transformer_engine_executor.py
riccardofelluga: You're right, that is a better pattern. Updated it.

thunder/tests/test_transformer_engine_executor.py
riccardofelluga: Thanks for the catch, I've really missed that. I've added the appropriate guarding

riccardofelluga:
> Thanks @riccardofelluga
> 
> Should a similar patch be applied to DDP and FSDP transformer_engine tests (as we have observed some failures for them)?
> 
> NOTE: We don't test the TE DDP and FSDP tests in Github CI.

@kshitij12345  You're right, I've updated distributed tests too 👍

PR #2749, opened 2025-11-18
Symbolic values: nn.Parameter to have static shapes even with symbolic values

kshitij12345:
Related: #2677 

It should be ok to treat parameter shape as static as they usually don't change. torch.compile(dynamic=True) also does the same.

Reference: 
1. https://github.com/Lightning-AI/lightning-thunder/issues/2677#issuecomment-3543030803
2. https://github.com/Lightning-AI/lightning-thunder/issues/2677#issuecomment-3543228690

### DIFF
--- a/thunder/tests/test_jit_general.py
+++ b/thunder/tests/test_jit_general.py
@@ -1592,6 +1592,11 @@ def test_cache_symbolic_values_nn_parameter_static_shape():
     for bsym in exec_trc.bound_symbols:
         if bsym.sym.name == prims.PrimIDs.UNPACK_TRIVIAL and "weight" in bsym.output.name:
             assert bsym.output.shape == (2, 2)
+        elif bsym.sym.name == prims.PrimIDs.UNPACK_TRIVIAL and "bias" in bsym.output.name:
+            assert bsym.output.shape == (2,)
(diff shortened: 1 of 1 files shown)


thunder/tests/test_jit_general.py
```diff
@@ -1581,6 +1581,19 @@ def foo(a, v):
...
+    x = torch.randn(2, 2)
+
+    jlinear = thunder_jit(linear, cache="symbolic values")
+
+    jlinear(x)
+    exec_trc = thunder.last_traces(jlinear)[-1]
+    for bsym in exec_trc.bound_symbols:
+        if bsym.sym.name == prims.PrimIDs.UNPACK_TRIVIAL and "weight" in bsym.output.name:
+            assert bsym.output.shape == (2, 2)
+
```
beverlylytle, line 1600: Maybe also add a check that the other input has a shape modeled on IntegerProxies?

beverlylytle (approved):
Thank you!

kshitij12345:
Ping @KaelanDt for review, thanks!

riccardofelluga (commented):
Nice idea! 

Would it make sense to check the prologue too in the test or is it overkilled?

PR #2754, opened 2025-11-18
Add type argument to descriptor __get__

shino16:
This PR makes `_call_dispatch` use `__get__(slf, type(slf))` instead of `__get__(slf)`, working around a CPython segfault ([python/cpython#132747](https://github.com/python/cpython/issues/132747)).

This will eliminate the segmentation fault mentioned in https://github.com/Lightning-AI/lightning-thunder/issues/2677#issuecomment-3422790012. From Python 3.13.3 this workaround will no longer be necessary, but I think we can keep it anyway.

### DIFF
--- a/thunder/benchmarks/benchmark_inference.py
+++ b/thunder/benchmarks/benchmark_inference.py
@@ -17,7 +17,6 @@ import argparse
 import json
 import os
 import statistics
-import sys
 import time
 import warnings
 from typing import Any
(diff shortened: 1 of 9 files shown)


riccardofelluga (approved):
Way to go! Great catch! 

For info, what changes under the hood when we give the type to the __get__ method?

shino16:
Thanks! I believe there is no change. [The official doc](https://docs.python.org/3/howto/descriptor.html#overview-of-descriptor-invocation) says
> If a descriptor is found for a.x, then it is invoked with: desc.\_\_get\_\_(a, type(a)).

> A dotted lookup such as super(A, obj).m searches obj.\_\_class\_\_._\_mro\_\_ for the base class B immediately following A and then returns B.\_\_dict\_\_['m'].\_\_get\_\_(obj, A). If not a descriptor, m is returned unchanged.

(copied from https://github.com/Lightning-AI/lightning-thunder/issues/2677#issuecomment-3542984406)

shino16:
@KaelanDt This is ready for your review. Thanks!

PR #2746, opened 2025-11-18
[Experimental] Insert `prims.shape` to ensure NumberProxy is materialized for all dynamic shapes

shino16:
Builds upon #2745. Closes #2677.

The following command runs successfully.

```sh
torchrun --nproc-per-node 2 thunder/benchmarks/benchmark_inference.py \
  --num-iterations 10 --mode thunder --thunder-cache "symbolic values"
```

PR #2755, opened 2025-11-19
Set TensorProxy shape's history to None if tensor's history is unavailable

shino16:
This is to fix the AssertionError reported in https://github.com/Lightning-AI/lightning-thunder/issues/2677#issuecomment-3542984406.
See https://github.com/Lightning-AI/lightning-thunder/pull/2746#discussion_r2537639544 for the discussion on correctness.

I added a test for this, which was the first test related to DTensors with dynamic shapes.


thunder/tests/distributed/test_dtensor.py
```diff
@@ -6,6 +6,7 @@
 
 import pytest
 import torch
+from torch.distributed.tensor import DTensor
```
kshitij12345: This import should be moved after 

```python
if not torch.distributed.is_available():
    pytest.skip(allow_module_level=True)
```

PR #2762, opened 2025-11-20
Remove TE v1 executor 

riccardofelluga:
## What does this PR do?

After the latest update to transformer_engine library(2.10) this executor has stopped working and, since the TEv2 is now bringing the same functionality and more, it is time to move on :D

### DIFF
--- a/thunder/executors/utils.py
+++ b/thunder/executors/utils.py
@@ -96,8 +96,33 @@ class Region:
         return s
 
 
-# Helper to use torch.autograd.Function as an implementation for a symbol.
 class Context:
+    """Helper to use torch.autograd.Function as an implementation for a symbol.
+
+    This class provides a minimal interface for saving and retrieving tensors between
+    forward and backward passes, compatible with torch.autograd.Function implementations.
+
+    Usage Pattern:
+        Forward Pass:
+            1. Create Context() instance
+            2. Pass to autograd function which calls ctx.save_for_backward(...)
+            3. Extract tensors with ctx.pop_saved_tensors()
+
+        Backward Pass:
+            1. Restore tensors using set_saved_tensors(ctx, saved_tensors) context manager
+            2. Call backward function which accesses ctx.saved_tensors
+
+    Example:
+        # Forward
+        ctx = Context()
+        out = CustomFunction.forward(ctx, x, weight)
+        saved = ctx.pop_saved_tensors()
+
+        # Backward
+        with set_saved_tensors(ctx, saved):
+            grads = CustomFunction.backward(ctx, grad_out)
+    """
+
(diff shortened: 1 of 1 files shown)


thunder/executors/utils.py
```diff
@@ -97,7 +97,6 @@ def __repr__(self) -> str:
 
 
 # Helper to use torch.autograd.Function as an implementation for a symbol.
-# See `transformer_engine_v1ex.py` for example.
```
crcrpar, line 100: what would be an example of this class, after this v1 removal?

thunder/plugins/fp8.py
crcrpar, line 1: would there be some script and/or test to make sure there isn't unexpected regressions?

riccardofelluga: For convergence, we do compare outputs between Thunder TE executor and vanilla TE in the tests already, but regarding performance there aren't any specific tests but the benchmarks. If we were to compare performance we can do so by using `benchmark_litgpt.py` in something like this:

```shell
# Using TE executor
python thunder/benchmarks/benchmark_litgpt.py --model_name Llama-2-7b-hf --compile thunder --checkpoint_activations False --low_precision_mode fp8-default-te --use_sdpa False
# Using TE without Thunder
python thunder/benchmarks/benchmark_litgpt.py  --model_name Llama-2-7b-hf --compile eager --checkpoint_activations False --low_precision_mode fp8-default-te --use_sdpa False
```

thunder/executors/utils.py
riccardofelluga: Added the doc string with the example usage

thunder/executors/utils.py
```diff
@@ -96,9 +96,33 @@ def __repr__(self) -> str:
         return s
 
 
-# Helper to use torch.autograd.Function as an implementation for a symbol.
-# See `transformer_engine_v1ex.py` for example.
 class Context:
+    """Helper to use torch.autograd.Function as an implementation for a symbol.
+
+    This class provides a minimal interface for saving and retrieving tensors between
```
kshitij12345, line 102: `for saving tensors and other metadata`

kshitij12345:
I still see CI job `test_transformer_engine_v1_executor.py` waiting and is tagged required. Is there something that needs to be changed from Github settings (as this PR removes the CI job from corresponding .yml)?

image

riccardofelluga:
> I still see CI job `test_transformer_engine_v1_executor.py` waiting and is tagged required. Is there something that needs to be changed from Github settings (as this PR removes the CI job from corresponding .yml)?
> 
> image

I've changed the CI file but maybe the CI will record that only after this PR 🤔

PR #2763, opened 2025-11-20
Translate shape for fd.ops.full

shino16:
Part of fix for #2677. As explained in https://github.com/Lightning-AI/lightning-thunder/issues/2677#issuecomment-3422790012, this patch fixes an error which is caused by feeding a `NumberProxy` for shape directly into `fd.ops.full`.


thunder/executors/nvfuserex_impl.py
```diff
@@ -1043,12 +1043,13 @@ def full(
     fd: FusionDefinition,
     lc_to_nv_map: dict,
 ) -> Any:
+    nv_shape = getnv(shape, fd, lc_to_nv_map, inline_number=True)
```
shino16, line 1046: In my interpretation, we put `inline_number=True` simply because we don't want to create `fd.define_scalar` for each dimension every time.

PR #2764, opened 2025-11-20
Fix `KeyError 'i23'` with symbolic shapes and `reshape` bsyms

shino16:
Partial fix for #2677. `update_aliases.py` manually inserts `reshape` bsyms for aliases with different shapes. In this PR, we additionally insert `prims.shape` to acquire the new symbolic shape in the trace.

The tests will fail due to this line.
https://github.com/Lightning-AI/lightning-thunder/blob/1a2c7d972197ea0a59e1016af56a9d8c78d8d549/thunder/core/update_aliases.py#L75
It will disappear once #2760 is merged.


thunder/tests/test_update_aliases.py
```diff
@@ -361,7 +363,7 @@ def fn(x):
         return y
 
     a = make_tensor((2, 3), dtype=torch.float32, device=device)
-    jfn = executor.make_callable(fn, skip_inplace_alias_updates=True)
```
shino16, line 364: I removed `skip_inplace_alias_updates=True` assuming it was meant to be deleted in 3df3d26201b8abfc0e3c01a200e86e6fd199057a. Maybe this test didn't work back then without this option.

beverlylytle:
While this PR makes sense for what I imagine to be most cases, I think we are still not gracefully handling the following situation:
```
def f(x, y) : ...
jf = thunder.jit(f, cache='symbolic values')

x = torch.randn(4)

jf(x, x)

y = torch.randn(4)
z = y[0:2]

jf(y,z)
```
In the second application of `jf`, we won't go through the checks about numel and we will instead jump directly to a compiled function that will attempt to reshape `z` into `y`, which will produce an error.  I'd really rather skip this reshaping and replacing hubbub, but I was running into trouble with numerical errors with nvfuser when I tried.

shino16:
Hmm, that's a good point. I will wait until seeing how https://github.com/Lightning-AI/lightning-thunder/pull/2760#issuecomment-3562836815 will end up.

PR #2769, opened 2025-11-25
Fix comparison to view_groups

beverlylytle:
In the insertion of `update_aliases`, we need to keep track of which tensors share memory.  The variable `view_groups` contains that information in the form of the original variable names assigned to those tensors (and no shallow copies thereof).  In the process of determining whether or not to insert an `update_aliases` bsym, we were mistakenly checking whether aliased tensors were contained in `view_groups`.  This PR amends that by checking, instead, whether an alias's pre-image under `swap_map` is in `view_groups`.

### DIFF
--- a/thunder/tests/test_jit_general.py
+++ b/thunder/tests/test_jit_general.py
@@ -1237,18 +1237,28 @@ def test_complex_backward_custom_autograd():
 def test_autograd_function_apply():
     # see https://github.com/Lightning-AI/lightning-thunder/issues/1248#issuecomment-2388655917
     # for why `torch.foo` instead of `torch.Tensor.foo`
-    def forward(ctx, x):
-        saved_for_backward = (x,)
-        return torch.sin(x), saved_for_backward
 
-    def backward(ctx, grad_output, *saved_tensors):
-        (x,) = saved_tensors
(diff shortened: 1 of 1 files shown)


thunder/core/update_aliases.py
```diff
@@ -166,15 +175,21 @@ def insert_alias_updates(computation_trace: Trace, alias_tensor_indices: list[li
...
+        ):
             if _is_inplace_op(bsym) and in_tensors:
                 in_tensors = {in_tensors[0]}
+                unswapped_in_tensors = {unswapped_in_tensors[0]}
             else:
                 in_tensors = set(in_tensors)
             out_tensors = set(map(variableify, filter(lambda p: isinstance(p, TensorProxy), bsym.flat_proxy_outs)))
             encountered.update(in_tensors)
-            group = set(reduce(set.union, filter(lambda g: any(g.intersection(in_tensors)), view_groups), set()))
+            group = set().union(*filter(lambda g: any(g.intersection(unswapped_in_tensors)), view_groups))
```
shino16: Not so related, but do we need `any` here?

beverlylytle: No, we don't.  Removed.

PR #2771, opened 2025-11-26
Updated cudnn executor checker

### LINKED ISSUE #2752: Remove embedding dimension check from `cudnn` executor
[This](https://github.com/Lightning-AI/lightning-thunder/blob/93aa254082162605f724c3be7451b641b6669eba/thunder/executors/cudnn_sdpa.py#L337) guard is linked to an old cuDNN version, which shouldn't affect runs with a newer cuDNN version.

cc @mruberry

mattteochen:
Closes #2752 

For reference: https://github.com/NVIDIA/cudnn-frontend/blob/v1.16.0/include/cudnn_frontend/node/scaled_dot_product_flash_attention.h#L816

### DIFF
--- a/thunder/tests/test_cudnn_executor.py
+++ b/thunder/tests/test_cudnn_executor.py
@@ -90,10 +90,10 @@ def grad_scaled_dot_product_attention_reference_generator(op, device, dtype, req
 
     # Standard dimensions - should work on all GPUs (d_q <= 128, d_kv <= 128)
     standard_dims = [
-        (64, 64),    # standard small
-        (32, 32),    # divisible by 8 small
-        (96, 96),    # divisible by 8 mid
-        (120, 88),   # divisible by 8 asymmetric
+        (64, 64),  # standard small
+        (32, 32),  # divisible by 8 small
+        (96, 96),  # divisible by 8 mid
+        (120, 88),  # divisible by 8 asymmetric
     ]
     for d_q, d_kv in standard_dims:
         q = make(b, h, s_q, d_q)


thunder/executors/cudnn_sdpa.py
```diff
@@ -334,9 +334,36 @@ def _cudnn_sdpa_checker(
...
+                        return False
+
+            is_supported_dim = d_q <= 256 and d_kv <= 256
+
+        elif cc_major == 10 and cudnn_backend_version >= 91100:  # Blackwell with cuDNN 9.11+
+            if d_q == 192:
+                is_supported_dim = d_kv == 128
+            else:
+                is_supported_dim = d_q <= 128 and d_kv <= 128
+    if not is_supported_dim:
```
riccardofelluga, line 364: nit. instead of checking on this variable, could have it been possible to check on the cudnn version instead? (requires change of the previous if/else structure

riccardofelluga (approved):
Thanks! Just a nit but it looks good to me. 

Do we need to add tests on the new shapes?

as a last comment make sure to check in with @vedaanta

mattteochen:
> Thanks! Just a nit but it looks good to me.
> 
> Do we need to add tests on the new shapes?
> 
> as a last comment make sure to check in with @vedaanta

Yes, good point. I've added some specific input tests to check the new supported dimensions.

mattteochen:
@vedaanta could you please take a look? Thanks!

vedaanta (approved):
thanks @mattteochen for these updated head_dim checks.
The support surface LGTM.

mattteochen:
@KaelanDt could you please review this? Thanks!

PR #2774, opened 2025-11-26
Clone tensor before mutation for in-place tests

### LINKED ISSUE #2773: `test_update_aliases` does not correctly test function outputs
https://github.com/Lightning-AI/lightning-thunder/blob/9e12768086aa6ef904927f6cc4760b8c8156fd32/thunder/tests/test_update_aliases.py#L122-L135

This is a test for in-place ops such as `Tensor.square_`, `Tensor.abs_`, etc.

https://github.com/Lightning-AI/lightning-thunder/blob/9e12768086aa6ef904927f6cc4760b8c8156fd32/thunder/tests/test_update_aliases.py#L132-L135

`op.torch_reference` is precisely those ops like `Tensor.square_`, so it returns the same tensor as its first argument. It ends up being mutated by both jitted op and eager op, and `torch.testing.assert_close(actual, expected)` is trivially True.

You can verify this with this patch:

```diff
diff --git a/thunder/tests/test_update_aliases.py b/thunder/tests/test_update_aliases.py
index 7cafcbec..e5fe7b8c 100644
--- a/thunder/tests/test_update_aliases.py
+++ b/thunder/tests/test_update_aliases.py
@@ -132,6 +132,7 @@ def test_update_aliases(op, device, dtype, executor, _):
     j_op = executor.make_callable(op.torch_reference)
     actual = j_op(*args, **sample.kwargs)
     expected = op.torch_reference(*args, **sample.kwargs)
+    assert id(args[0]) == id(actual) == id(expected), f"{id(args[0]) = }, {id(actual) = }, {id(expected) = }"
     torch.testing.assert_close(actual, expected, equal_nan=True)
```

This assertion passes on all ops except for the case `op = tanhshrink` (which is also strange because `tanhshrink` is not even in-place).

shino16:
Fixes #2773. It clones the first argument into `op` so that jitted op and eager-mode op will mutate different tensors.

Alongside, I removed `_inplace_to_out_of_place[tanhshrink] = tanhshrink, -1`, because `tanhshrink` is not in-place (e.g. compare with [`relu_`](https://github.com/Lightning-AI/lightning-thunder/blob/9e12768086aa6ef904927f6cc4760b8c8156fd32/thunder/torch/__init__.py#L2121)). And I couldn't find an in-place version of `tanhshrink`.

### DIFF
--- a/thunder/benchmarks/benchmark_litgpt.py
+++ b/thunder/benchmarks/benchmark_litgpt.py
@@ -691,7 +691,10 @@ class Benchmark_litGPT:
                 executors.insert(0, transformer_engine_ex)
                 transforms.insert(0, TransformerEngineTransform())
 
-            if "dynamo" in self.compile:
+            if "jit" in self.compile:
+                model = thunder.jit(model, executors=executors, transforms=transforms, **jit_options)
+
+            else:
                 if self.distributed_mode == "fsdp2":
                     print("Resetting cache size for when fsdp2 and using thunder as backend torch.compile")
                     import torch._dynamo.config as dynamo_config
@@ -704,10 +707,6 @@ class Benchmark_litGPT:
                 # using __wrapped__ to access the original torch.compile function did not work
                 # so we are using the lower level torch._dynamo.optimize function
                 model = torch._dynamo.optimize(backend=self.backend)(model)
-            else:
-                jit_options = {}
-                jit_options["fp8_shard_intermediate_activation"] = self.fp8_shard_intermediate_activation
-                model = thunder.jit(model, executors=executors, transforms=transforms, **jit_options)
         elif self.compile != "eager":
             raise ValueError(f"Invalid compile option: {self.compile}")
 


thunder/torch/__init__.py
```diff
@@ -2313,9 +2313,6 @@ def tanhshrink(a: TensorLike, /) -> TensorLike:
     return a - tanh(a)
 
 
-_inplace_to_out_of_place[tanhshrink] = tanhshrink, -1
-
```
kshitij12345, line 2317: Nice catch!

kshitij12345 (approved):
LGTM, thanks @shino16

riccardofelluga (approved):
Nice! 

Is there a way to test this? Should we add a test to check that this is always the case?

thunder/tests/test_update_aliases.py
```diff
@@ -129,9 +129,11 @@ def test_update_aliases(op, device, dtype, executor, _):
     if op.name == "polygamma_":
         args[0], args[1] = args[1], args[0]
 
+    args_ref = [args[0].clone().detach()] + args[1:]
     j_op = executor.make_callable(op.torch_reference)
     actual = j_op(*args, **sample.kwargs)
-    expected = op.torch_reference(*args, **sample.kwargs)
+    expected = op.torch_reference(*args_ref, **sample.kwargs)
+    assert id(actual) != id(expected)
```
beverlylytle: This assert checks that our testing infrastructure is correct, not that the source code is correct.  It's an inexpensive check, but we need to always keep in mind what these inexpensive checks add up to when run for every in-place op. Isn't the more interesting thing to test that the id of `actual` matches the id of args[0]?

beverlylytle (approved):
Nice!

thunder/tests/test_update_aliases.py
shino16: Interesting! That makes a lot of sense.

shino16:
I added `assert id(actual) == id(args[0])` to the test. I think this resolves @riccardofelluga's question.

> Is there a way to test this? Should we add a test to check that this is always the case?

PR #2778, opened 2025-11-29
Support min/max for symbolic values

### LINKED ISSUE #2737: Support torch.sym_max/sym_min
## 🚀 Feature

Thunder needs to support `torch.sym_max` and `torch.sym_min` operations, which are used by TorchDynamo to represent max/min operations on symbolic values, particularly for sliding window attention patterns.

## Current Behavior
Likely unsupported or falls back to eager evaluation.

## Expected Behavior
Support symbolic max/min that preserves symbolic expressions:
- `max(Sym(s0 - 1055), 0)` → `Sym(Max(0, s0 - 1055))`
- Generate appropriate runtime checks

## Example from torch.compile

```py
sub: "Sym(s67 - 1056)" = l_kwargs_past_key_values_layers_0_cumulative_length - 1056
add_1: "Sym(s67 - 1055)" = sub + 1
sym_max: "Sym(Max(0, s67 - 1055))" = torch.sym_max(add_1, 0)
```

This pattern is used for computing KV-cache offsets with sliding windows.

## Minimal Reproduction Case
```py
import torch
import thunder

@thunder.jit
def compute_kv_offset(cumulative_length: int, sliding_window: int = 1056) -> int:
    """Compute offset for sliding window attention."""
    offset_raw = cumulative_length - sliding_window + 1
    # Should support: torch.sym_max for symbolic values
    kv_offset = max(offset_raw, 0)  # or torch.sym_max(offset_raw, 0)
    return kv_offset

assert compute_kv_offset(100) == 0    # max(100 - 1056 + 1, 0) = 0
assert compute_kv_offset(2000) == 945  # max(2000 - 1056 + 1, 0) = 945

print(compute_kv_offset._lc_cs.last_epilogue_traces[-1])
# def epilogue():
#   return 945
```

## Use Case
Critical for HF transformer models with sliding window attention (e.g., Mistral, Llama with sliding window).

## Success Criteria
- [ ] Support `torch.sym_max(symbolic, constant)`
- [ ] Support `torch.sym_min(symbolic, constant)`
- [ ] Support `max(symbolic, constant)` as alias
- [ ] Generated code produces correct runtime values
- [ ] Symbolic expression preserved for further optimization

shino16:
Closes #2737. This adds new primitives for builtin `min`/`max`. This enables keeping `min`/`max` in the computation trace instead of embedding concrete values. We also support `torch.sym_min`/`sym_max` because those are what Dynamo produces.

### DIFF
--- a/thunder/benchmarks/benchmark_litgpt.py
+++ b/thunder/benchmarks/benchmark_litgpt.py
@@ -691,7 +691,10 @@ class Benchmark_litGPT:
                 executors.insert(0, transformer_engine_ex)
                 transforms.insert(0, TransformerEngineTransform())
 
-            if "dynamo" in self.compile:
+            if "jit" in self.compile:
+                model = thunder.jit(model, executors=executors, transforms=transforms, **jit_options)
+
+            else:
                 if self.distributed_mode == "fsdp2":
                     print("Resetting cache size for when fsdp2 and using thunder as backend torch.compile")
                     import torch._dynamo.config as dynamo_config
@@ -704,10 +707,6 @@ class Benchmark_litGPT:
                 # using __wrapped__ to access the original torch.compile function did not work
                 # so we are using the lower level torch._dynamo.optimize function
                 model = torch._dynamo.optimize(backend=self.backend)(model)
-            else:
-                jit_options = {}
-                jit_options["fp8_shard_intermediate_activation"] = self.fp8_shard_intermediate_activation
-                model = thunder.jit(model, executors=executors, transforms=transforms, **jit_options)
         elif self.compile != "eager":
             raise ValueError(f"Invalid compile option: {self.compile}")
 


thunder/clang/__init__.py
```diff
@@ -1743,6 +1743,24 @@ def minimum(a, b):
     return _elementwise_binary_wrapper(a, b, prim=prims.minimum)
 
 
+@clangop()
+def py_max(a: NumberLike | Number, b: NumberLike | Number):
```
crcrpar: could you tell me what "py" suffix means?

shino16: I thought I should avoid name collision. I've just realized there are so many name collisions already.

shino16: But I named the new primitive `prims.py_max` to clarify the difference from `prims.maximum`, and the same could apply here. What do you think?

riccardofelluga: Why does this need to be differentiated with the naming?

shino16: I had a quick off-GitHub conversation and decided to go with the existing `clang.maximum`. Thank you for your comments!

crcrpar: I was purely curious what py meant, was it pytorch or python, or something different?

shino16: Ah haha, I meant Python because `max` here represents Python's builtin `max`, not `torch.max`.

thunder/tests/test_jit_general.py
```diff
@@ -864,6 +864,26 @@ def foo(a, scalar):
...
 
+@pytest.mark.parametrize("min_op,max_op", [(torch.sym_min, torch.sym_max), (min, max)], ids=("torch", "builtin"))
+def test_symbolic_values_min_max(min_op, max_op):
+    def foo(seq_len, cumulative_len, max_len, sliding_window) -> int:
+        max_cache_len = min_op(max_len, seq_len + 128)
+        offset_raw = cumulative_len - sliding_window + 1
+        kv_offset = max_op(offset_raw, 0)
+        return max_cache_len, kv_offset
+
+    samples = torch.randint(1, 100, (4, 20))
```
beverlylytle: Can this be deleted?

beverlylytle (approved):
Nice!

thunder/tests/test_jit_general.py
shino16: Ahh good catch, thanks!

Issue #2738, opened 2025-11-13
Support symbolic values in torch.arange and other tensor operations


IvanYashchuk:
## 🚀 Feature

Thunder needs to support using symbolic scalar values as arguments to tensor creation operations, particularly `torch.arange` with symbolic start/end bounds.

## Current Behavior
Likely requires concrete values for tensor operations.

## Expected Behavior
Support symbolic values as arguments to:
- `torch.arange(symbolic_start, symbolic_end, ...)`
- `torch.zeros(symbolic_size, ...)`
- `.view(symbolic_shape, ...)` and other reshaping ops
- `.index_copy_(dim, symbolic_indices, ...)`

## Example from torch.compile

```py
# cache_position: "i64[1]" = torch.arange(cumulative_length, add, device=...)
cache_position: "i64[s50]" = torch.arange(
    l_kwargs_past_key_values_layers_0_cumulative_length, 
    add,  # add is Sym(s50 + s67)
    device=device(type='cuda', index=0)
)
```
The output tensor has symbolic shape `"i64[s50]"`.

## Minimal Reproduction Case

```py
import torch
import thunder

@thunder.jit
def create_cache_position(cumulative_length: int, seq_len: int) -> torch.Tensor:
    """Create cache position indices using symbolic bounds."""
    new_cumulative_length = cumulative_length + seq_len
    
    # torch.arange with symbolic start and end
    cache_position = torch.arange(
        cumulative_length,
        new_cumulative_length,
        device='cuda'
    )
    
    return cache_position

result = create_cache_position(1024, 32)
assert result.shape == (32,)
assert result[0].item() == 1024
assert result[-1].item() == 1055
```

## Related Operations
- `torch.arange(symbolic, symbolic)`
- `torch.zeros(symbolic, *shape)`
- `tensor.view(-1, symbolic)`
- `tensor.scatter(1, symbolic_indices, value)`

## Success Criteria
- [ ] `torch.arange` accepts symbolic start/end
- [ ] Output tensor shape is correctly inferred as symbolic
- [ ] Other tensor ops support symbolic size arguments
- [ ] Generated code handles runtime dimensions correctly

shino16:
With #2779 and the following patch on nvFuser, `torch.arange` will work under nvFuser. I'm exploring what other ops that we need to add to nvFuser's concretization process.

```diff
diff --git a/csrc/evaluator_common.cpp b/csrc/evaluator_common.cpp
[...]

PR #2779, opened 2025-11-30
Enhance symbolic arithmetics

### LINKED ISSUE #2736: Support symbolic arithmetic operations
## 🚀 Feature

Thunder JIT needs to support arithmetic operations on symbolic values, producing new symbolic expressions. This includes addition, subtraction, and composition of symbolic values.

## Current Behavior
Arithmetic on dynamic values likely forces eager evaluation and concretizing the tracing time results.

## Expected Behavior
Arithmetic operations on symbolic values should produce symbolic results:
- `Sym(s0) + 1` → `Sym(s0 + 1)`
- `Sym(s0) - 1056` → `Sym(s0 - 1056)`
- `Sym(s0) + Sym(s1)` → `Sym(s0 + s1)`

## Example from torch.compile

```py
add: "Sym(s67 + 1)" = l_kwargs_past_key_values_layers_0_cumulative_length + 1
add: "Sym(s50 + s67)" = l_kwargs_past_key_values_layers_0_cumulative_length + s50
add_10: "Sym(s50 + s59)" = l_kwargs_past_key_values_layers_1_cumulative_length + s50
```

## Minimal Reproduction Case
```py
import torch
import thunder

@thunder.jit
def symbolic_arithmetic(cumulative_length: int, seq_len: int) -> tuple[int, int, int]:
    """Both inputs should be symbolic."""
    # Pattern 1: Symbolic + constant
    new_length_1 = cumulative_length + 1
    
    # Pattern 2: Symbolic - constant
    offset = cumulative_length - 1056
    
    # Pattern 3: Symbolic + symbolic
    combined = cumulative_length + seq_len
    
    return new_length_1, offset, combined

result = symbolic_arithmetic(1024, 32)
assert result == (1025, -32, 1056)

print(symbolic_arithmetic._lc_cs.last_epilogue_traces[-1])
# This shows that the values are concretized and not treated symbolically
# def epilogue():
#   return (1025, -32, 1056)
```

## Operations to Support
- [ ] Addition: `symbolic + constant`, `symbolic + symbolic`
- [ ] Subtraction: `symbolic - constant`, `symbolic - symbolic`
- [ ] Multiplication: `symbolic * constant`
- [ ] Floor division: `symbolic // constant`

## Success Criteria
- [ ] All arithmetic operations preserve symbolic types
- [ ] Symbolic expressions can be composed
- [ ] Generated code correctly handles runtime values
- [ ] Works with both scalar-scalar and scalar-tensor operations

shino16:
Closes #2736. The main purpose is to implement `math.ceil`, which is needed for supporting `clang.arange` ([ref](https://github.com/Lightning-AI/lightning-thunder/blob/9e12768086aa6ef904927f6cc4760b8c8156fd32/thunder/clang/__init__.py#L198)).

To test it against both `int` and `float`, I added a logic to test the type of NumberProxy's underlying values in prologue.

I took a quick side step to implement all ops tested in `test_elementwise.py`. This involved the following patches.

#### Rounding ops: `ceil`, `floor`, `trunc`, `round`

- Registered their NumberProxy versions to `pythonex`, backed with `math.ceil/floor/trunc`, `builtins.round`.
- Patched them to compute in the promoted type (not in int).

#### Bitwise ops: `bitwise_not`, `bitwise_left_shift`, `bitwise_right_shift`

- Registered their NumberProxy version to `pythonex`, backed with `operator.inv/lshift/rshift`
- Routed e.g. `NumberProxy.__lshift__` to the corresponding `clangop`.

#### `floor_divide`

- Patched `where`, which `floor_divide` depends on, to support all-NumberProxy inputs

#### `pos` (as opposed to `neg`)

- Thunder has no implementation of `pos`. I changed `NumberProxy.__pos__` and `TensorProxy.__pos__` to just `return self`, which aligns with Python and PyTorch (try `for x in (1000, torch.randn(3)): assert +x is x`).

### DIFF
--- a/thunder/benchmarks/benchmark_litgpt.py
+++ b/thunder/benchmarks/benchmark_litgpt.py
@@ -691,7 +691,10 @@ class Benchmark_litGPT:
                 executors.insert(0, transformer_engine_ex)
                 transforms.insert(0, TransformerEngineTransform())
 
-            if "dynamo" in self.compile:
+            if "jit" in self.compile:
+                model = thunder.jit(model, executors=executors, transforms=transforms, **jit_options)
+
+            else:
                 if self.distributed_mode == "fsdp2":
                     print("Resetting cache size for when fsdp2 and using thunder as backend torch.compile")
                     import torch._dynamo.config as dynamo_config
@@ -704,10 +707,6 @@ class Benchmark_litGPT:
                 # using __wrapped__ to access the original torch.compile function did not work
                 # so we are using the lower level torch._dynamo.optimize function
                 model = torch._dynamo.optimize(backend=self.backend)(model)
-            else:
-                jit_options = {}
-                jit_options["fp8_shard_intermediate_activation"] = self.fp8_shard_intermediate_activation
-                model = thunder.jit(model, executors=executors, transforms=transforms, **jit_options)
         elif self.compile != "eager":
             raise ValueError(f"Invalid compile option: {self.compile}")
 


thunder/core/interpreter.py
```diff
@@ -398,7 +398,7 @@ def __init__(
         self._with_provenance_tracking = with_provenance_tracking
         if with_provenance_tracking:
             assert isinstance(uncacheable_classes, (list, tuple))
-            uncacheable_classes = tuple(set(uncacheable_classes) | {NoneType, int, str, float, bool})
+            uncacheable_classes = tuple(set(uncacheable_classes) | {NoneType, int, str, float, bool, complex})
```
shino16: Not directly related, but we do need this for complex inputs.

On `main` (f6f0496cc5f0c9071b62ef79a72ffed1a1bcb296),

```py
import thunder

def f(x, y, z):
    return x + y + z

jf = thunder.jit(f)
print(jf(1j, 1j, 1j), thunder.cache_misses(jf)) # 3j, 1
print(jf(1j, 2j, 1j), thunder.cache_misses(jf)) # 3j, 1
print(jf(1j, 1j, 2j), thunder.cache_misses(jf)) # 3j, 1

print(thunder.last_prologue_traces(jf)[-1])
# check_number_type_and_value applied only to args[0]
```

This caused a new test `test_where_on_numbers` to fail, so I patched it in this PR.

thunder/executors/nvfuserex_impl.py
```diff
@@ -1472,6 +1488,7 @@ def bitwise_not(a: TensorProxy | Number, *, fd: FusionDefinition, lc_to_nv_map:
 register_supported(PrimIDs.BITWISE_NOT, bitwise_not, _elementwise_unary_check)
 
 
+@_returns_int_for_numbers
```
kiya00, line 1491: This appears to introduce a new number-only input type for nvfuser ceil, I noticed the test cases use CPU only, has this path also been tested?

thunder/core/interpreter.py
kshitij12345: Would it be better to add `complex` here?

https://github.com/Lightning-AI/lightning-thunder/blob/f6f0496cc5f0c9071b62ef79a72ffed1a1bcb296/thunder/core/jit_ext.py#L2167

thunder/core/prims.py
```diff
@@ -1945,9 +1945,12 @@ def _numpy_array_to_torch_tensor_meta(a: TensorProxy, /) -> TensorProxy:
 #   usually produce an output with that same datatype (SAME).
 #   Sometimes, however, elementwise operations can produce an output with a different
 #   datatype than the inputs. For example, comparison operations like eq and lt always
-#   produce boolean results (ALWAYS_BOOL), math.ceil/math.floor produces integer outputs for number inputs while preserves datatype for tensor inputs, and other operations, like abs, map
-#   complex numbers to floats (COMPLEX_TO_FLOAT).
+#   produce boolean results (ALWAYS_BOOL), ceil/floor produces integer outputs for
+#   number inputs while preserves datatype for tensor inputs (INT_FOR_NUMBER), and
+#   other operations, like abs, map complex numbers to floats (COMPLEX_TO_FLOAT).
 #   The ELEMENTWISE_PRIM_OUTPUT_DTYPE_KIND enum describes these three behaviors so that
+#   operations, like abs, map complex numbers to floats (COMPLEX_TO_FLOAT).
```
kshitij12345: This seems like stray copy.

thunder/core/prims.py
```diff
@@ -1945,9 +1945,12 @@ def _numpy_array_to_torch_tensor_meta(a: TensorProxy, /) -> TensorProxy:
 #   usually produce an output with that same datatype (SAME).
 #   Sometimes, however, elementwise operations can produce an output with a different
 #   datatype than the inputs. For example, comparison operations like eq and lt always
-#   produce boolean results (ALWAYS_BOOL), math.ceil/math.floor produces integer outputs for number inputs while preserves datatype for tensor inputs, and other operations, like abs, map
-#   complex numbers to floats (COMPLEX_TO_FLOAT).
+#   produce boolean results (ALWAYS_BOOL), ceil/floor produces integer outputs for
+#   number inputs while preserves datatype for tensor inputs (INT_FOR_NUMBER), and
+#   other operations, like abs, map complex numbers to floats (COMPLEX_TO_FLOAT).
 #   The ELEMENTWISE_PRIM_OUTPUT_DTYPE_KIND enum describes these three behaviors so that
```
kshitij12345: This line is redundant with the `The ELEMENTWISE_PRIM_OUTPUT_DTYPE_KIND enum describes these four behaviors so that` below.

thunder/executors/nvfuserex_impl.py
kshitij12345: Also, I am not sure if nvfuser will pick up a prim with Number only input due to the `has_cuda_input_or_output` in the check below.

https://github.com/Lightning-AI/lightning-thunder/blob/f6f0496cc5f0c9071b62ef79a72ffed1a1bcb296/thunder/executors/nvfuserex_impl.py#L852-L860

I tried a sample program

```python
import thunder
import torch
import math

def fn(x_t, x, y):
    return math.ceil(x) + math.ceil(y)

tfn = thunder.jit(fn, cache="symbolic values")

print(tfn(torch.randn(3, 3, device="cuda") ,1., 2.))

# print(thunder.last_traces(tfn)[0])
print(thunder.last_traces(tfn)[-1])

```

And the trace looks like
```python
import math
import operator
import torch

@torch.no_grad()
@no_autocast
def computation(x, y):
  # x: "float 1.0"
  # y: "float 2.0"
  i7 = math.ceil(x)  # i7: "int 1"
    # i7 = prims.ceil(x)  # i7: "int 1"
  i8 = math.ceil(y)  # i8: "int 2"
    # i8 = prims.ceil(y)  # i8: "int 2"

  # /opt/pytorch/lightning-thunder/test.py:6:       return math.ceil(x) + math.ceil(y)
  i9 = operator.add(i7, i8)  # i9: "int 3"
    # i9 = prims.add(i7, i8)  # i9: "int 3"
  del i7, i8
  return (i9,)
```

If nvfuser will not pick-up the prim, I think nvfuser related changes should be made in a follow-up PR.

thunder/tests/test_elementwise.py
```diff
@@ -61,36 +78,58 @@ def filter_b_values(vals, special):
...
+
+    if cache_option == "symbolic values":
+        assert thunder.cache_misses(cfoo) == len(a_types) * len(b_types)
 
-        for a in a_vals:
-            for b in b_vals:
-                actual = cfoo(a, b)
-                expected = foo(a, b)
-                assert_close(actual, expected)
+        fixed_a = random.choice(a_vals)
```
kshitij12345: What are we trying to test here? I think product(fixed_a, b_vals) is already tested above.

thunder/tests/test_elementwise.py
```diff
@@ -13,23 +16,37 @@
...
+    operator.xor: ((bool, int), (bool, int), None),
+    operator.lshift: ((bool, int), (int,), "shift_count"),
+    operator.rshift: ((bool, int), (int,), "shift_count"),
+}
+
+
+@instantiate(
+    dtypes=NOTHING,
+    devicetypes=(devices.DeviceType.CPU,),
+    decorators=(
```
kshitij12345, line 38: Nice!

thunder/tests/test_elementwise.py
shino16: I thought it may be worth testing when one argument is a constant Number and the other is a NumberProxy.

The test takes 0.21s including this and 0.05s excluding this. I thought it was worth the time, but I'd have no argument against excluding it.

thunder/core/interpreter.py
shino16: Hmm, maybe both? I notice `bool` is listed only here too.

thunder/core/prims.py
shino16: Thanks for the catch!

thunder/executors/nvfuserex_impl.py
shino16: > Also, I am not sure if nvfuser will pick up a prim with Number only input due to the has_cuda_input_or_output in the check below.

Whether or not nvfuser supports scalar ops matters when such ops appear as a subsymbol. Once #2662 is merged, nvFuser will be able to fuse `torch.arange`. Without `@_returns_int_for_numbers`, it will give this error:

```py
import torch, thunder

def fn(a):
    return torch.arange(a, device="cuda")
  # [t9] = nvFusion0(a)
  #   # i3 = prims.sub(a, 0)  # i3: "int 10"
  #   # f4 = prims.convert_element_type(i3, float)  # f4: "float 10.0"
  #   # f5 = prims.div(f4, 1.0)  # f5: "float 10.0"
  #   # i6 = prims.ceil(f5)  # i6: "int 10"
  #   # t9 = prims.iota(i6, start=0, step=1, device=devices.Device("cuda:0"), dtype=dtypes.int64)  # t9: "cuda:0 i64[[IntegerProxy name=i6, value=10, static=CONSTRAINT.CONSTRAINABLE]]"


jfn = thunder.jit(fn, cache="symbolic values")
jfn(10)
```
```
Exception occurred: RuntimeError: Expected isIntegralType(*length->getDataType()) . length must be integer, but get dtype double
Exception raised from iota at /opt/pytorch/nvfuser/csrc/ops/arith.cpp:398 (most recent call first):
frame #0: nvfuser::nvfCheckFail(char const*, char const*, long, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) + 0xfb (0x731564672ba9 in /opt/pytorch/nvfuser/python/nvfuser_direct/../build/libnvfuser_codegen.so)
frame #1: nvfuser::iota(nvfuser::Val*, nvfuser::Val*, nvfuser::Val*, nvfuser::DataType) + 0xb3b (0x731564d78e3b in /opt/pytorch/nvfuser/python/nvfuser_direct/../build/libnvfuser_codegen.so)
```

I added a new test for this in `test_nvfuser.py`, but this is xfailed because we need a patch inside nvFuser. https://github.com/Lightning-AI/lightning-thunder/issues/2738#issuecomment-3604494766

kshitij12345: Got it, thanks for the explanation.

thunder/tests/test_elementwise.py
kshitij12345: I think thunder treats all Numbers as dynamic with symbolic values. So, I don't think we test anything different here from above.

```python
import thunder
import torch
import math

def fn(x, y):
    return x + y

tfn = thunder.jit(fn, cache="symbolic values")

tfn(1, 2)

print(thunder.last_traces(tfn)[-1])

```

Trace
```python
import operator
import torch

@torch.no_grad()
@no_autocast
def computation(x, y):
  # x: "int 1"
  # y: "int 2"

  # /opt/pytorch/lightning-thunder/test.py:6:       return x + y
  i3 = operator.add(x, y)  # i3: "int 3"
    # i3 = prims.add(x, y)  # i3: "int 3"
  return (i3,)
```

thunder/tests/test_nvfuser.py
```diff
@@ -1061,3 +1061,18 @@ def test_scatter(executor, device: str, dtype: dtypes.dtype):
...
+@instantiate(
+    executors=(nvFuserExecutor,),
+    dtypes=NOTHING,
+)
+def test_arange_symbolic_values(executor, device: str, dtype: dtypes.dtype):
+    from thunder.tests.opinfos import arange_opinfo
+
+    for sample in arange_opinfo.sample_inputs(device, thunder.float32):
+        compiled_func = executor.make_callable(torch.arange, cache="symbolic values")
+        out = compiled_func(*sample.args, **sample.kwargs)
```
kshitij12345, line 1076: I think we should also add a a check to verify that there is only one bsym in trace and it is a nvfuser symbol.

thunder/core/proxies.py
```diff
@@ -455,6 +455,9 @@ def __eq__(self, other):
             return False
         return str(self) == str(other)
 
+    def __bool__(self) -> bool:
+        return bool(self.value)
```
kshitij12345, line 459: Can you comment on why this is needed?

IvanYashchuk: `.value` should only be accessed by trace passes. In the current implementation this conversion would be baked into the trace as a constant with no relation to the Proxy object from which it was converted from. Ideally, it would be a symbol `prims.bool` as done for other operations.

shino16: It's part of 08bf07e, which enables calling `torch.arange` with `device="cuda"` option. It calls

https://github.com/Lightning-AI/lightning-thunder/blob/f6f0496cc5f0c9071b62ef79a72ffed1a1bcb296/thunder/torch/__init__.py#L666

which calls

https://github.com/Lightning-AI/lightning-thunder/blob/f6f0496cc5f0c9071b62ef79a72ffed1a1bcb296/thunder/torch/__init__.py#L145

which calls the lookaside of `bool` for `device or ...`. The commit 08bf07e ensures `bool(<StringProxy>)` is handled in the same way as IntegerProxy

shino16: EDIT: I realized I was wrong. StringProxy does not have `constraint` attribute.

<details>

> Ideally, it would be a symbol prims.bool as done for other operations.

I only imagined cases where `StringProxy.__bool__` was called by `bool`'s lookaside.

https://github.com/Lightning-AI/lightning-thunder/blob/ec69a756a6294b33377c961f96b468238c3da72d/thunder/core/jit_ext.py#L646-L651

This makes StringProxy statically constrained, forcing recompile when it changes.

What I didn't imagine is cases where `bool(<StringProxy>)` is called outside of the JIT context, in which case it would break correctness. I'm out now, so I'll check this later.

</details>

shino16: `StringProxy.__bool__` is just something that makes the new test in `test_nvfuser.py` jit-able with symbolic values enabled, and this simple implementation aligns with collection proxies.

I am wondering if `StringProxy` is something we should treat symbolically. When input strings differ e.g. device="cuda:0" vs "cuda:1" vs "cpu", we should recompile rather than making it a symbolic StringProxy. And if we take that way, `StringProxy.value` and `bool(<StringProxy>.value)` will be constant, not symbolic. Please correct me if I'm off the point, because I am not so sure about the roles of non-number proxies.

I suggest making a minimal change here just to make the test jit-able, and think about symbolic StringProxies separately, as we have done little about it yet.

thunder/core/interpreter.py
shino16: `torch.Tensor` is listed only at the call site... It's confusing that the same thing is specified in two different places with small differences.

I removed the line `uncacheable_classes=(torch.Tensor, int, float, str, NoneType)` from `jit_ext.py` and added `torch.Tensor` inside the function `interpret`. What do you think?

thunder/core/proxies.py
kshitij12345: Thanks for explaining, I agree that making string value symbolic especially for device string doesn't make sense as we can't expect backend to generate efficient code for different devices.

Also, an alternative fix which wouldn't require supporting `__bool__` for str would be rewrite 
```python
return device or get_default_device()
```

to 
```python
if device is not None:
    return device
return get_default_device()
```

thunder/core/interpreter.py
```diff
@@ -397,8 +399,11 @@ def __init__(
         self._callbacks: dict[INTERPRETER_CALLBACKS, Callable] = callbacks
         self._with_provenance_tracking = with_provenance_tracking
         if with_provenance_tracking:
-            assert isinstance(uncacheable_classes, (list, tuple))
-            uncacheable_classes = tuple(set(uncacheable_classes) | {NoneType, int, str, float, bool})
+            if uncacheable_classes is None:
+                uncacheable_classes = ()
+            uncacheable_classes = tuple(
+                set(uncacheable_classes) | {NoneType, int, str, float, bool, complex, torch.Tensor}
+            )
```
kshitij12345, line 406: On looking at this a bit further, I think your previous fix to add `complex` in `interpreter.py` was correct.

My understanding is that `interpreter.py` only understand basic Python and nothing about external libraries like PyTorch. `jit_ext.py` extends the capability to teach it about PyTorch support. So, I don't think `torch.Tensor` should be present here but instead be passed from `jit_ext.py`.

Sorry for the confusion.

thunder/core/proxies.py
shino16: Nice fix! I didn't come up with it. I think `StringProxy.__bool__` is a nice thing to have anyway.

thunder/core/interpreter.py
shino16: Never mind, thank you for the explanation! That explains why `torch.Tensor` appears only outside of `interpreter.py`. `jit_ext.py` has little to do with complex inputs, so I think the original PR was fine? I reverted back the last change.

KaelanDt (approved):
looks great, thanks @shino16

PR #2786, opened 2025-12-03
Default thunder to the dynamo path and make jit explicit for benchmark_litgpt.py

riccardofelluga:
To make it easier for benchmarking, this changes the compile options by removing the dynamo keyword and requiring `jit` to explicitly use the jit path

### DIFF
--- a/thunder/benchmarks/benchmark_litgpt.py
+++ b/thunder/benchmarks/benchmark_litgpt.py
@@ -79,20 +79,18 @@ class LowPrecisionHandler:
     enabled: bool = False
 
     use_thunder_te: bool = True
-    use_legacy_thunder_te: bool = False
 
     @property
     def use_te_autocast(self) -> bool:
-        return self.enabled and not self.use_legacy_thunder_te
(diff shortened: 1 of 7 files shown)


thunder/benchmarks/benchmark_litgpt.py
```diff
@@ -695,7 +695,12 @@ def setup_compile(self, model):
                 executors.insert(0, transformer_engine_ex)
                 transforms.insert(0, TransformerEngineTransform())
 
-            if "dynamo" in self.compile:
+            if "jit" in self.compile:
+                jit_options = {}
+                jit_options["fp8_shard_intermediate_activation"] = self.fp8_shard_intermediate_activation
```
kshitij12345: This option was part of TE v1 executor and shouldn't have any effect now that we have removed it.

riccardofelluga: Thanks! Updated it

shino16 (approved):
Nice!

Issue #2790, opened 2025-12-03
Investigate `torch.Tensor.__setitem__` as unsupported in `thunderx`


mattteochen:
Setting this operator in https://github.com/Lightning-AI/lightning-thunder/blob/f6f0496cc5f0c9071b62ef79a72ffed1a1bcb296/thunder/dynamo/utils.py#L48 gives 10x decode throughput improvement on `openai/gpt-oss-20b` and `Qwen/Qwen3-32B` inside SGLang `bench_one_batch.py`.

Repro (this only works on the NV internal [thunder-sglang-integration](https://gitlab-master.nvidia.com/kaiximatteoc/sglang/-/tree/thunder-dynamo-recompilation) codebase):
```
SGLANG_USE_THUNDER_GRAPH_RUNNER=1 python3 -m sglang.bench_one_batch   --model-path openai/gpt-oss-20b --trust-remote-code   --model-impl transformers --dtype bfloat16   --json-model-override-args '{"quantization_config": null}' --cuda-graph-bs 1 --tp-size 4 --tp-strategy dtensor --load-format-dummy
```

kiya00:
It also fits my observation when running the benchmark on each graph:
```
  Graph Name: graph7_thunder_1
    Operators used in the graph:
      <built-in function setitem>
```
the benchmark script can be a minimal repro:
```
from math import inf
from math import nan
NoneType = type(None)
import torch
from torch import device
import torch.fx._pytree as fx_pytree
import torch.utils._pytree as pytree
import torch
from thunder.dynamo.benchmark_utils import TorchInductorSpecification
from thunder.dynamo.benchmark_utils import KernelTime

from thunder.dev_utils import nvtx_profile_transform


class DynamoModule(torch.nn.Module):
    def forward(self, l_forward_batch_token_to_kv_pool_k_buffer_0_: "bf16[2615232, 8, 64][512, 64, 1]cuda:0", l_forward_batch_out_cache_loc: "i64[1][1]cuda:0", l_k_: "bf16[1, 8, 64][512, 64, 1]cuda:0"):
         # File: /sgl-workspace/sglang/python/sglang/srt/mem_cache/memory_pool.py:803 in set_kv_buffer, code: self.k_buffer[layer_id - self.start_layer][loc] = cache_k
[...]

PR #2792, opened 2025-12-09
Implement `setitem` without copy_

shino16:
Fixes https://github.com/Lightning-AI/lightning-thunder/issues/2790. Previously, `inp[idx] = value` was translated into `_copy_(inp, clang.copy_with_setitem(inp, idx, val))`, where `copy_with_setitem` was an opaque primitive that only torchex could execute. This PR makes the entire `setitem` opaque.

### Demo

```py
import torch, thunder

def fn(x, idx, val):
    x[idx] = val

jf = thunder.jit(fn)
x = torch.zeros(3, 4, 5, device="cpu")
x_ref = x.clone()
idx = torch.tensor([1], device="cpu")
val = torch.ones(1, 4, 5, device="cpu")
fn(x_ref, idx, val)
jf(x, idx, val)

torch.testing.assert_close(x, x_ref)
```

```py
import torch

@torch.no_grad()
@no_autocast
def computation(x, idx, val):
  # x: "cpu f32[3, 4, 5]"
  # idx: "cpu i64[1]"
  # val: "cpu f32[1, 4, 5]"
  (t4,) = update_aliases((x,))
  del x

  # /opt/pytorch/lightning-thunder/tmp/main.py:4:           x[idx] = val
  t5 = setitem(t4, idx, val)  # t5: "cpu f32[3, 4, 5]"
    # t5 = ltorch.setitem_(t4, idx, val)  # t5: "cpu f32[3, 4, 5]"
      # t5 = prims.setitem(t4, idx, val)  # t5: "cpu f32[3, 4, 5]"
  del t4
  return ()
```

### DIFF
--- a/thunder/benchmarks/__init__.py
+++ b/thunder/benchmarks/__init__.py
@@ -2081,7 +2081,7 @@ class LitGPTCausalSelfAttentionBenchmark(Benchmark, metaclass=UserFacingBenchmar
         from litgpt.model import CausalSelfAttention
 
         module = (
-            CausalSelfAttention(self.config)
+            CausalSelfAttention(self.config, 0)
             .to(device=self.device, dtype=self.tdtype)
             .requires_grad_(self.requires_grad)
         )
--- a/thunder/benchmarks/benchmark_litgpt.py
+++ b/thunder/benchmarks/benchmark_litgpt.py
@@ -691,7 +691,10 @@ class Benchmark_litGPT:
                 executors.insert(0, transformer_engine_ex)
                 transforms.insert(0, TransformerEngineTransform())
 
-            if "dynamo" in self.compile:
+            if "jit" in self.compile:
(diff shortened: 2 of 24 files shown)


kiya00:
The failures in CI seems relevant

thunder/core/transforms.py
```diff
@@ -2594,6 +2574,34 @@ def index_put_aug_fwd(
     return VJPDual(primal, residuals)
 
 
+@register_augmented_forward(prims.PrimIDs.SETITEM)
```
shino16, line 2578: We have to define forward and backward separately. If not, when we extract the forward trace from the combined fw&bw function, we hit into `prims.setitem(clone(g), index, 0)` in backward, which survives through DCE because `setitem` is tagged as `DONT_DCE`.

PR #2795, opened 2025-12-10
[WIP] See if we can make symbolic values default

shino16:
Let's see how many tests fail when we make `cache="symbolic values"` the default option.

I don't intend to merge this PR. Unmerged PRs and dirty commits are contained.


beverlylytle:
FYI: the intention for Q4 is to enable symbolic values for thunderfx, not for thunder.jit.

PR #2797, opened 2025-12-11
Handled DTensor `placements` as positional argument for `redistribute `

### LINKED ISSUE #2794: DTensor `redistribuite` is not supported from `thunderfx` if `placements` is provided as positional arg
Currently thunder supports:
```python
z = z.redistribute(placements=[Replicate()], async_op=True)
```

This will mark the redistribution node as unsupported:
```python
z = z.redistribute(mesh, [Replicate()], async_op=True)
```

The issue should be [this](https://github.com/Lightning-AI/lightning-thunder/blob/f6f0496cc5f0c9071b62ef79a72ffed1a1bcb296/thunder/dynamo/utils.py#L1218) dictionary access, while `placements` could be passed as a positional argument.

mattteochen:
The PyTorch DTensor redistribute API can accept `placements` as a positional argument. `thunderfx` lacked this support, making the node unsupported and causing additional graph breaks.

Closes #2794.

### DIFF
--- a/thunder/core/jit_ext.py
+++ b/thunder/core/jit_ext.py
@@ -940,27 +940,40 @@ def _general_jit_torch_ops_higher_order_autograd_function_apply(fwd, bwd, *fwd_a
         length = 5
         return "".join(secrets.choice(string.ascii_lowercase) for _ in range(length))
 
-    args_tensor_mask = unwrap(fwd_kwargs["args_tensor_mask"])
+    # Support both stable PyTorch (with args_tensor_mask) and nightly (without it)
+    if "args_tensor_mask" in fwd_kwargs:
+        args_tensor_mask = unwrap(fwd_kwargs["args_tensor_mask"])
+    else:
+        args_tensor_mask = None
+
     # TODO(crcrpar): Think about making use of `non_differentiable_idx`
     # note that this key is quite new: https://github.com/pytorch/pytorch/pull/134087
     # non_differentiable_idx = fwd_kwargs.get("non_differentiable_idx")
-    length_of_tensor_args = sum(args_tensor_mask)
 
-    # N.B.(crcrpar) When `torch.compile(..., dynamic=True)`,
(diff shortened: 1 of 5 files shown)


thunder/dynamo/utils.py
```diff
@@ -1216,8 +1216,15 @@ def dtensor_from_local_prim_wrapper(x, mesh=mesh, placements=placements):
...
-                placements = kwargs["placements"]
+                args = closure_vars.nonlocals.get("args_as_value", ())
+                kwargs = closure_vars.nonlocals.get("kwargs_as_value", {})
+
+                # Handle positional args: redistribute(device_mesh, placements)
+                # and keyword args: redistribute(placements=...)
+                if len(args) >= 2:
+                    placements = args[1]
+                else:
+                    placements = kwargs["placements"]
```
crcrpar: The linked issue seems to say placements is a positional arg. When we pass placements as a keyword arg, would we see it in kwargs?

mattteochen: The current status is that we support this node only if `placements` is a kwarg. We didn't parse the positional argument.

The answer to your question is yes.

shino16: Is there a chance to see `redistribute` with `placements` unspecified?

mattteochen: From the [doc](https://docs.pytorch.org/docs/stable/distributed.tensor.html#torch.distributed.tensor.DTensor.redistribute), it could be None or unspecified, but it will trigger a Dynamo assertion.

kshitij12345: Looks like the docs don't match the implementation as `redistribute` raises an error if `placements` is None

https://github.com/pytorch/pytorch/blob/d38164a545b4a4e4e0cf73ce67173f70574890b6/torch/distributed/tensor/_api.py#L540-L541

mattteochen:
@KaelanDt for review

mattteochen:
Hi @KaelanDt, could you please review this? Thank you.

Issue #2803, opened 2025-12-15
`args_tensor_mask` has been removed from `torch.ops.higher_order.autograd_function_apply`


mattteochen:
A recent Pytorch [PR](https://github.com/pytorch/pytorch/pull/166788/changes#diff-4ad321e59efa9c5873050471d6855a1e328a0ba05ec85f963a06958b5d862880L746) changed the `autograd_function_apply` function API by dropping the usage of `args_tensor_mask` keyword argument.

Currently, thunder [reads](https://github.com/Lightning-AI/lightning-thunder/blob/fb989d485cf50fcd7645d500fb229f28e318ce52/thunder/core/jit_ext.py#L943) this key in the `kwarg` dict leading to a `KeyError`.

PR #2802, opened 2025-12-15
[WIP] Refactor autograd function handling to utilize `non_differentiable_idx`

mattteochen:

PR #2804, opened 2025-12-15
Fix KeyError for args_tensor_mask removed in PyTorch nightly

IvanYashchuk:
A recent PyTorch [PR #166788](https://github.com/pytorch/pytorch/pull/166788) changed the autograd_function_apply higher-order op API by removing the args_tensor_mask keyword argument.

Currently, Thunder reads this key in the fwd_kwargs dict leading to a KeyError:
https://github.com/Lightning-AI/lightning-thunder/blob/fb989d485cf50fcd7645d500fb229f28e318ce52/thunder/core/jit_ext.py#L943

This PR adds a compatibility layer for both older PyTorch releases (which still provide args_tensor_mask) and newer PyTorch nightly versions (which removed it). When the key is not present, we infer the tensor mask by checking which arguments are TensorProxy instances.

Let's see in CI if further changes needed anywhere else.

Fixes https://github.com/Lightning-AI/lightning-thunder/issues/2803.

### DIFF
--- a/thunder/core/jit_ext.py
+++ b/thunder/core/jit_ext.py
@@ -940,7 +940,15 @@ def _general_jit_torch_ops_higher_order_autograd_function_apply(fwd, bwd, *fwd_a
         length = 5
         return "".join(secrets.choice(string.ascii_lowercase) for _ in range(length))
 
-    args_tensor_mask = unwrap(fwd_kwargs["args_tensor_mask"])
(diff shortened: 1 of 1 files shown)


IvanYashchuk:
Closing in favor of https://github.com/Lightning-AI/lightning-thunder/pull/2802.

Issue #2700, opened 2025-10-29
Double first iteration time for `Llama-2-7b-hf` after nvFuser direct bindings


riccardofelluga:
## 🐛 Bug

As per title, the following command takes double the time for the first iteration at commit 8b542cf16e12cf04f6375691621bb3adb0c4acea :

```
python thunder/benchmarks/benchmark_litgpt.py --model_name Llama-2-7b-hf --compile thunder_inductor_cat --checkpoint_activations False --low_precision_mode none --micro_batch_size 1 --global_batch_size 64 --use_sdpa False --block_size 4096 --max_iters 10 --warmup_iters 5
```

compared to before #2502 : 

```python
# Before
iter 0: loss 0.1650, iter time: 105627.25ms, t: 4096
# After 
iter 0: loss 0.1650, iter time: 215099.65ms, t: 4096
```

cc @rdspring1 @kshitij12345 @crcrpar

rdspring1:
Direct bindings doesn't use an LRU cache to cache `FusionDefinition` creation. It was proposed in [[RFC] Create LRU cache for direct bindings. #4893](https://github.com/NVIDIA/Fuser/pull/4893) but shelved until it is needed. From Thunder's perspective, you simply wrap `create_fd` in `thunder/executors/nvfuserex_impl.py` with `LruFusionCache`.

### Codediff to add `LruFusionCache`

<details>

```diff
diff --git a/thunder/executors/nvfuserex_impl.py b/thunder/executors/nvfuserex_impl.py
index b65d6944..5531cd9d 100644
--- a/thunder/executors/nvfuserex_impl.py
+++ b/thunder/executors/nvfuserex_impl.py
@@ -77,6 +77,7 @@ if nvfuser_version() >= DIRECT_BINDINGS_SUPPORTED_VERSION:
     from nvfuser_direct import (
         DataType,
         FusionDefinition,
+        LruFusionCache,
         multidevice,
         ParallelType,
         execute_with_dtensors,
@@ -297,6 +298,15 @@ def multidevice_schedule(fd: FusionDefinition, in_dtensors: list[Proxy]) -> None
             in_tv.set_allocation_domain(in_tv.get_loop_domain(), new_contiguity=True)
 
 
+# This function wraps nvfuser_direct's LruFusionCache with a version check.
+def FusionCacheDecorator(func: callable):
+    # For legacy bindings, the decorator does nothing.
[...]

PR #2806, opened 2025-12-15
Add LruFusionCache when direct bindings is used

rdspring1:
This PR enables an LRU Fusion Cache decorator for direct bindings to address issue #2700.

```
Legacy Fusion Cache
iter 0: loss 0.1641, iter time: 102852.81ms, t: 4096
iter 1: loss 0.1196, iter time: 20309.14ms, t: 4096
iter 2: loss 0.0771, iter time: 19994.30ms, t: 4096

ToT Direct w/o LRU Cache
iter 0: loss 0.1641, iter time: 204575.74ms, t: 4096
iter 1: loss 0.1196, iter time: 22914.21ms, t: 4096
iter 2: loss 0.0771, iter time: 20006.35ms, t: 4096

Direct with LRU Cache
iter 0: loss 0.1641, iter time: 115281.74ms, t: 4096
iter 1: loss 0.1196, iter time: 20228.57ms, t: 4096
iter 2: loss 0.0771, iter time: 19998.50ms, t: 4096
```

* LRU Cache is 12% slower because equality check was updated to cover block
scale operations, which create non-trivial allocation domains.

### DIFF
--- a/thunder/executors/nvfuserex_impl.py
+++ b/thunder/executors/nvfuserex_impl.py
@@ -43,7 +43,7 @@ from thunder.core.utils import check
 from thunder.core.trace import TraceCtx, from_trace, TraceProvenance
 from thunder.core.symbol import BoundSymbol, BoundSymbolRHS, Symbol, has_tags
 from thunder.core.devices import Device, DeviceType, cpu
-from thunder.core.transform_common import dce, cse_single_bsym, replace_redundant_inputs
+from thunder.core.transform_common import dce, dce_bsyms, cse_single_bsym, replace_redundant_inputs
 from thunder.core.profile import annotate_for_profile
 from thunder.core.compile_data import get_compile_option
 from thunder.torch.experimental.dtensor_torch_and_prims import DTensorPrimIDs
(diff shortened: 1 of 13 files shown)


thunder/executors/nvfuserex_impl.py
```diff
@@ -71,7 +71,7 @@
 # NOTE This impl file is here because nvFuser may not be available, so it's imported conditionally
 #   by nvfuserex.py when nvFuser is available.
 
-DIRECT_BINDINGS_SUPPORTED_VERSION = LooseVersion("0.2.34")
+DIRECT_BINDINGS_SUPPORTED_VERSION = LooseVersion("0.2.35")
```
kshitij12345, line 74: Is this the minimum version which ships with LruFusionCache?

thunder/executors/nvfuserex_impl.py
```diff
@@ -298,6 +298,17 @@ def multidevice_schedule(fd: FusionDefinition, in_dtensors: list[Proxy]) -> None
...
 
 
+# This function wraps nvfuser_direct's LruFusionCache with a version check.
+def FusionCacheDecorator(func: callable):
+    # For legacy bindings, the decorator does nothin.
+    if nvfuser_version() < DIRECT_BINDINGS_SUPPORTED_VERSION:
+        return func
+    from nvfuser_direct import LruFusionCache
+
+    return LruFusionCache(max_fusions=16384)(func)
```
kshitij12345, line 308: Default value for `max_fusions` is 16384, we can remove this here.

crcrpar: qq: will there be an option of setting cache size? Or it would not be that useful

thunder/executors/nvfuserex_impl.py
rdspring1: Yes

thunder/executors/nvfuserex_impl.py
rdspring1: Usually we pick a reasonable number to avoid out-of-memory issues. We never thought to change it at runtime.

PR #2828, opened 2026-08-12
Remove F405 from Ruff ignore list

arubittu:
Removes `F405` from Ruff’s global ignore list by importing the names used in `thunder.core.utils` explicitly. The intentional wildcard imports stay in place so the module namespace does not change. The metadata wildcard import is replaced with explicit re-exports.

Addresses part of #2218.

Testing:
- `ruff check . --select F405`
- `ruff check thunder/core/utils.py thunder/__init__.py`
- `pytest thunder/tests/test_core.py -q` (161 passed)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -156,7 +156,6 @@ ignore = [
     # TODO(crcrpar): Resolves the following ignores as these are added while enabling ruff check in pre-commit
     "F821",  # https://docs.astral.sh/ruff/rules/undefined-name/
     "E402",  # https://docs.astral.sh/ruff/rules/module-import-not-at-top-of-file/
-    "F405",  # https://docs.astral.sh/ruff/rules/undefined-local-with-import-star-usage/
     "E712",  # https://docs.astral.sh/ruff/rules/true-false-comparison/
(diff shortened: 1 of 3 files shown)


arubittu:
The pre-commit.ci failure comes from the `pyupgrade` hook crashing on Python 3.14 with `TypeError: cannot use a bytes pattern on a string-like object`. All other pre-commit hooks passed. This is the same issue addressed by #2824.

arubittu:
I checked the failed CI logs:

- the stable PyTorch main and gradient suites pass
- the PyTorch nightly jobs treat the new `inline_inbuilt_nn_modules` deprecation warning as an error during test collection
- the notebook jobs stop during setup because pip cannot uninstall the Debian `cryptography` package without a `RECORD` file
- Transformer Engine cannot find a matching wheel and its source build expects the missing `nccl_dev_cap.hpp` PyTorch header
- Read the Docs imports a PyTorch build without `Tensor.align_as`
- pre-commit still hits the Python 3.14 `pyupgrade` crash addressed by #2824

These failures appear to come from CI images or dependency changes rather than this Ruff cleanup.

PR #2832, opened 2026-09-15
[DO NOT MERGE] Resolve CI: torchaudio, dynamo FutureWarning, torch 2.13 named tensors

bhimrazy:
## What does this PR do?

Combines three separate fixes into one branch so CI can be evaluated together, ahead of splitting back into #2829, #2830, #2831 individually:

- Drop unused `torchaudio` from `requirements/devel.txt` — the pin (`adjust-torch-versions.py` tracking torch) is unsatisfiable at torch 2.14 (torchaudio stopped releasing in lockstep after 2.11), and nothing in the repo imports it.
- Ignore the deprecated `torch._dynamo.config.inline_inbuilt_nn_modules` `FutureWarning`, which fires at import time from `torch/testing/_internal/common_utils.py` and breaks collection of all distributed tests.
- Fix `import thunder` on PyTorch 2.13: named tensors were removed upstream, so `Tensor.align_as`, `align_to`, `has_names`, `refine_names` and `rename` are gone. Gates those five behind a new `_TORCH_GREATER_EQUAL_2_13` constant.

<details>
<summary>Why combined</summary>

Each fix individually only exposes the next failure in the chain, making it hard to tell from CI alone whether a given fix works. This branch merges all three so we can see the full picture, then will be split back into the standalone PRs once confirmed.

</details>

> Changes in this PR are currently in raw form. Glad that the CI is finally passing now.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -74,6 +74,10 @@ markers = [
 ]
 filterwarnings = [
     "error::FutureWarning",
+    # torch reads this deprecated dynamo config at import time, as a default argument in
+    # torch/testing/_internal/common_utils.py, which fails collection of the distributed tests
+    # todo: drop once torch stops reading it
+    "ignore:torch._dynamo.config.inline_inbuilt_nn_modules is deprecated:FutureWarning",
 ]
 timeout = 900
 # xfail_strict = true  # todo
--- a/thunder/constants.py
+++ b/thunder/constants.py
@@ -1,0 +1,7 @@
+import operator
+
+from lightning_utilities.core.imports import compare_version
+
+# PyTorch version constants
+# `use_base_version` so that dev builds like "2.13.0a0+gitabc123" compare as "2.13.0"
+_TORCH_GREATER_EQUAL_2_13 = compare_version("torch", operator.ge, "2.13.0", use_base_version=True)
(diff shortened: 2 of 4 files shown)


.lightning/workflows/all-tests.yaml
```diff
@@ -72,13 +72,16 @@ run: |
          --random-order-seed=42 \
          --numprocesses=3
   elif [ "${testing}" == "ops" ]; then
+    # cross_entropy under nvfuser is the slowest test here and landed right on the old 240s cap
+    # (240.24s, vs 185s for its next-slowest dtype) with 9 workers sharing the one GPU. 360s
+    # matches every other job in this file.
```
lianakoleva, line 77: maybe we can put this in the PR description, not needed here

```suggestion
```

thunder/core/interpreter.py
```diff
@@ -2449,6 +2483,9 @@ def mapping(self):
     def isdisjoint(self, other):
         return all((k not in self.mapping) for k in other)
 
+    def __len__(self):
+        return len(self.mapping)
+
```
lianakoleva, line 2488: is this code used anywhere? i don't see

